PR #1788, opened 2021-11-15
Execution spec

yvan-sraka:


docs/execution.md
```diff
@@ -0,0 +1,208 @@
...
+```
+
+Where:
+* `bytecode` is smart contract bytecode that is understood by the SCE but not the CSS
+* `max_gas` is the maximum amount of gas that the execution of `bytecode` is allowed to cost
+* `gas_price` is the price per unit of gas that the caller is willing to pay for the execution
+* `coins` are extra coins that are spent by tge CSS and are available in the execution context of `bytecode`
+
+Blocks that include ExecuteSC operations must respect the following criteria, or else they are invalid:
+* ExecuteSC spends `max_gas * gas_price + coins` from the sender address in the CSS ledger, which must therefore be available
```
AureliaDolo, line 37: is max_gas * gas_price considered as a reward and should it be split among the block creator, the endorsers and the endorsed block ?

docs/execution.md
```diff
@@ -0,0 +1,208 @@
...
+OperationType::ExecuteSC {
+    max_gas: u32,
+    gas_price: Amount,
+    coins: Amount,
+    bytecode: Vec<u8>,
+}
+```
+
+Where:
+* `bytecode` is smart contract bytecode that is understood by the SCE but not the CSS
```
AureliaDolo, line 31: Do we enforce it to be wasm byte code ? If not maybe add a language field that could be an enum representing all supported languages (that could have just one variant for now)

docs/execution.md
```diff
@@ -0,0 +1,208 @@
...
+* `max_gas` is the maximum amount of gas that the execution of `bytecode` is allowed to cost
```
AureliaDolo, line 32: No more min gas here ?

docs/execution.md
AureliaDolo: This is spend immediately in CSS coins ?

docs/execution.md
```diff
@@ -0,0 +1,208 @@
...
+* `gas_price` is the price per unit of gas that the caller is willing to pay for the execution
+* `coins` are extra coins that are spent by tge CSS and are available in the execution context of `bytecode`
+
+Blocks that include ExecuteSC operations must respect the following criteria, or else they are invalid:
+* ExecuteSC spends `max_gas * gas_price + coins` from the sender address in the CSS ledger, which must therefore be available
+* all ExecuteSC operations included in the block must be ordered by decreasing `gas_price`
```
AureliaDolo, line 38: Why not by gas_price * max_gas ?

docs/execution.md
```diff
@@ -0,0 +1,208 @@
...
+* the sum of the max_gas values of all ExecuteSC operations in the block must not be higher than a protocol constant `max_block_gas`
+
+When included in a block, from the CSS point of view, an ExecuteSC operation will only spend `max_gas * gas_price + coins` coins from the sender address and do nothing else.
+
+When observing an ExecuteSC operation, the SCE will:
+* if any of the following steps fail, the bytecode is not executed and all its changes to the SCE ledger are rollbacked. The sender loses `max_gas * gas_price + coins`.
+* parse `bytecode` as smart contract bytecode
+* execute the bytecode in `bytecode`:
+  * if hard config-defined engine resource limits are hit (e.g. too much RAM used), execution fails
```
AureliaDolo, line 47: Link #1267 and #1266 here

docs/execution.md
```diff
@@ -0,0 +1,208 @@
...
+Smart-contract specific addresses are also part of the SCE ledger (but not CSS ledgers).
+
+When an SCE block execution requests a coin transfer form the SCE ledger to the CSS ledger, a `TransferToConsensus{target_addr: Address, origin_slot: Slot, amount: Amount}` message is sent from the SCE to the CSS.
+`origin_slot` is the slot of the block causing the `TransferToConsensus`.
+
+When the SCE receives a `BlockCliqueChanged` message:
+* the SCE reset its Candidate SCE ledger to the Final SCE ledger
+* the SCE sorts all blockclique blocks by increasing slot
+* for each block B in that order, the SCE processes the ExecuteSC operations in the block in the order they appear in the block. For each such operation `op`:
+  * credit the block creator in the SCE ledger with `op.max_gas * op.gas_price`
```
AureliaDolo, line 86: Why do we credit the block creator in the sce ledger ? And as it's considered as fee why only credit the block creator and not the other fee recipients ?

docs/execution.md
```diff
@@ -0,0 +1,208 @@
...
+```mermaid
+flowchart TB
+    Address-->Bytecode
+    subgraph SCE Ledger Data
+      Address-->Balance
+      Address-->DB
+    end
+    subgraph Program
+      Bytecode
+    end
```
AureliaDolo, line 119: I don't understand why that part of the graph is outside of the ledger data ?

docs/execution.md
```diff
@@ -0,0 +1,208 @@
...
+```
+
+* Holds separate Final and Candidate SCE ledgers, that match each address to a balance, database, and program area
+  * the `database` is a key-value map where the key is a sha256 hash, and the value contains arbitrary bytes
+  * the `program` area contains bytecode with decorated public/private functions
+* Contains an execution engine that will be fed the sorted and ExecuteSC operations with the block they belong to:
+  * Bytecode is interpreted as WASM and executed
+  * massa-specific "systcalls" are made available to the execution engine
+  * the execution engine is fully deterministic
```
AureliaDolo, line 128: How do we ensure that ?

docs/execution.md
```diff
@@ -0,0 +1,208 @@
...
+  * program_data: `Vec<u8>` // to be better defined in followup issues
+
+### SCE WASM execution engine syscalls
+
+A smart-contract will be able to call into Massa-specific API's, providing by the "Massa runtime".
+We could use "host functions" in wasmer to provide these API's to a running contract. See https://gitlab.com/massalabs/massa/-/issues/360
+
+Available "syscalls":
+
+- `Call(SC_addr, function_name, parameters)`: synchronously calls a function in the same, or another smart contract
```
AureliaDolo, line 143: Why do someone would want to use call on itself? Does that lead to potential security issues ?

docs/execution.md
```diff
@@ -0,0 +1,208 @@
...
+- `TransferToConsensus(amount)`: transfers massa coins from SCE ledger to the consensus ledger. On the SCE side, this just burns the coins.
+- `WriteDatabase(key, value)`: writes in the database of the current smart contract
```
AureliaDolo, line 145: Rights are checked inside that write function ?

docs/execution.md
damip: > is max_gas * gas_price considered as a reward and should it be split among the block creator, the endorsers and the endorsed block ?

that's an open question... it could be credited as a reward on the CSS side, or as a reward on the SCE side... any opinions ?


> This is spend immediately in CSS coins ?

yes (see my modification of the skeleton code)

AureliaDolo: > that's an open question... it could be credited as a reward on the CSS side, or as a reward on the SCE side... any opinions ?

That's more an economics question than a dev one :)

docs/execution.md
damip: The idea is that the CSS side does not interpret this field at all. It is only interpreted in the SCE.

For now we went for having only one type of bytecode: WASM, but for sure we could add a field describing which bytecode (and version!) it is

docs/execution.md
damip: yeah for simplicity, let's not add min_gas for now

docs/execution.md
damip: > This is spend immediately in CSS coins ?

yes, `max_gas * gas_price + coins` is spent in CSS



> Why not by gas_price * max_gas ?


Actually by decreasing `(gas_price * max_gas) / op_size`

docs/execution.md
damip: see one of the previous comments. We could consider it a fee, OR we could credit it only on the SCE side... both are options, any preference ?

docs/execution.md
damip: looks like github does not render mermaid ?

https://github.com/massalabs/massa/blob/ebeca6cf9c515ed62b88a143f769c883d8ebfdb8/docs/execution.md

docs/execution.md
damip: no that's ok, you might want to call one of your own already-defined public methods instead of writing a new one. No security issues I can see, do you see one ?

docs/execution.md
damip: this is not a final list, just a preliminary idea.

But if we go the way `WriteDatabase(key, value)`, there is no rights check to be done because it is always targeted at the current contect, which always has write rights to itself

docs/execution.md
damip: I guess we should credit it on the SCE side

docs/execution.md
damip: can you :) ?

docs/execution.md
damip: OK let's credit in SCE :)

docs/execution.md
AureliaDolo: It's not supported but you can install a plug in https://addons.mozilla.org/fr/firefox/addon/github-mermaid/

docs/execution.md
AureliaDolo: Have you a reason to prefer that option ?

damip: it's a bit arbitrary but the parallel SCE ledger can be used with more flexibility than the CSS one

docs/execution.md
damip: that's unfortunate...

adrien-zinger (approved):

docs/execution.md
adrien-zinger: > It's not supported but you can install a plug in https://addons.mozilla.org/fr/firefox/addon/github-mermaid/

This quote resolved the conversation

docs/execution.md
adrien-zinger: Writing these issues has automatically linked/quote this PR in the issues.

AureliaDolo:
Seems like we don't really keep that up to date, maybe we should merge it now and update it when #1837 is merged @adrien-zinger @gterzian @yvan-sraka @damip

yvan-sraka:
I suggest merging it, and to open follow-up issue :)

AureliaDolo:
Maybe we can merge that into add_cautious bassoon ?

yvan-sraka:
> Maybe we can merge that into add_cautious bassoon ?

Why?!

AureliaDolo:
> > Maybe we can merge that into add_cautious bassoon ?
> 
> Why?!

Because everything related to smart contracts will be added in that PR so why not the spec too ?

AureliaDolo:
What's the difference with the spec.md file in massa-execution?

yvan-sraka:
> Because everything related to smart contracts will be added in that PR so why not the spec too?

Okay right, but in the future, we should not work like that!

AureliaDolo:
> > Because everything related to smart contracts will be added in that PR so why not the spec too?
> 
> Okay right, but in the future, we should not work like that!

You're right ! :crab:

PR #1795, opened 2021-11-16
warn on stale

damip:


consensus/src/block_graph.rs
```diff
@@ -856,7 +856,7 @@ pub struct BlockGraph {
     to_propagate: BlockHashMap<(Block, OperationHashSet, Vec<EndorsementId>)>,
     attack_attempts: Vec<BlockId>,
     new_final_blocks: BlockHashSet,
-    new_stale_blocks: BlockHashMap<Slot>,
+    new_stale_blocks: BlockHashMap<(PublicKey, Slot)>, // creator, slot
```
AureliaDolo, line 859: It may be clearer to use a named struct

consensus/src/consensus_worker.rs
```diff
@@ -1345,10 +1345,15 @@ impl ConsensusWorker {
...
 
         // add stale blocks to stats
-        let new_stale_block_ids_slots = self.block_db.get_new_stale_blocks();
+        let new_stale_block_ids_creators_slots = self.block_db.get_new_stale_blocks();
         let timestamp = UTime::now(self.clock_compensation)?;
-        for (_b_id, _b_slot) in new_stale_block_ids_slots.into_iter() {
+        for (b_id, (b_creator, b_slot)) in new_stale_block_ids_creators_slots.into_iter() {
             self.stale_block_stats.push_back(timestamp);
+
+            let creator_addr = Address::from_public_key(&b_creator)?;
```
AureliaDolo, line 1353: Maybe we can add a staking public keys hashset to avoid recomputing an address just for logging

consensus/src/block_graph.rs
damip: since it's just 2 elements, I prefer not creating a structure just for that. The comment is quite clear :)

consensus/src/consensus_worker.rs
damip: it's an option indeed... but it adds a lot of complexity (we need to maintain another index)

if the performance hit is low (which it is), this is OK

consensus/src/block_graph.rs
adrien-zinger: An alternative could be defining a type for this context or use import .. as ..

damip: true, but let's leave it like this for now as the tuple is small and heavily explicit :)

damip:
bors merge

PR #1806, opened 2021-11-17
Add current timestamp in graph block intervals

yvan-sraka:
Close https://github.com/massalabs/massa/issues/1236


AureliaDolo:
This PR updates the old api and client

PR #1808, opened 2021-11-17
Completely split crypto and signature

adrien-zinger:
The purpose is to start a refactoring of the signature implementation.

The signature part can be a crate itself so I split everywhere. Next step is split PublicKey, PrivateKey, etc, in multiple files to increase code readability.

There is a lot of code duplication in this file and a weak test coverage. This branch can be the beginning of big journey.


AureliaDolo:
I'm not sure what does that bring ?

adrien-zinger:
> I'm not sure what does that bring ?

This is the first step of reducing code duplcation in signature.rs. It should increase the readability and produce a better to maintain and to refactor (if we want too use traits later for example).

Making a separated crate make this part more clear I think.

This commit also remove some clippy warnings and too verbose code while using the signature module.

yvan-sraka (approved):

yvan-sraka:
`[REF]` standing for "Refactoring" is a bit non-obvious for me!

AureliaDolo (approved):

consensus/src/tests/scenarios_reward_split.rs
```diff
@@ -220,7 +220,7 @@ async fn test_reward_split() {
                         .unwrap()
                         .checked_div_u64((1 + cfg.endorsement_count).into())
                         .unwrap()
-                        .saturating_sub(third.checked_mul_u64(2).unwrap())
```
AureliaDolo: That should not change just moving signatures around

adrien-zinger: > That should not change just moving signatures around

You're right, it should be done in another PR. NB: Before it was "2 * 0", I don't understand why we can't see this version of the main branch

adrien-zinger:
bors merge

PR #1819, opened 2021-11-18
Add a simple check of balance before transaction

adrien-zinger:
Doesn't affect transaction anyway, just print a warning.

### DIFF
--- a/massa-client/src/cmds.rs
+++ b/massa-client/src/cmds.rs
@@ -651,12 +651,24 @@ impl Command {
                         Err(e) => return repl_err!(e),
                     };
 
+                    let mut res = String::new();
+                    if let Ok(addresses_info) = client.public.get_addresses(vec![addr]).await {
+                        for info in addresses_info {
+                            if info.ledger_info.locked_balance > amount {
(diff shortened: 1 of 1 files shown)


massa-client/src/cmds.rs
```diff
@@ -651,12 +651,24 @@ impl Command {
                         Err(e) => return repl_err!(e),
                     };
 
+                    let mut res = String::new();
+                    if let Ok(addresses_info) = client.public.get_addresses(vec![addr]).await {
+                        for info in addresses_info {
+                            if info.ledger_info.locked_balance > amount {
```
AureliaDolo, line 657: Candidate balance is the most relevant here. It's the balance at the most recent blocks that we consider valid. Final balance is the balance at the latest final blocks and locked balance is the balance that is locked after a roll sell.

Moreover, you must check that amount + fee is enough

AureliaDolo, line 657: Candidate balance is the most relevant here. It's the balance at the most recent blocks that we consider valid. Final balance is the balance at the latest final blocks and locked balance is the balance that is locked after a roll sell.

Moreover, you must check that amount + fee is enough

AureliaDolo (changes requested):

PR #1827, opened 2021-11-18
Add a simple check of balance before transaction

yvan-sraka:
Doesn't affect transaction anyway, just print a warning

### DIFF
--- a/massa-client/src/cmds.rs
+++ b/massa-client/src/cmds.rs
@@ -472,11 +472,13 @@ impl Command {
                 match amount.checked_add(fee) {
                     Some(total) => {
                         if let Ok(addresses_info) = client.public.get_addresses(vec![addr]).await {
-                            if !addresses_info.is_empty()
-                                && addresses_info[0].ledger_info.candidate_ledger_info.balance
-                                    < total
-                            {
-                                println!("Warning: this operation may be rejected due to insuffisant balance");
+                            match addresses_info.get(0) {
+                                Some(info) => {
+                                    if info.ledger_info.candidate_ledger_info.balance < total {
+                                        println!("Warning: this operation may be rejected due to insuffisant balance");
+                                    }
+                                }
+                                None => println!("Warning: address {} not found", addr),
                             }
                         }
                     }
--- a/massa-client/src/cmds.rs
+++ b/massa-client/src/cmds.rs
@@ -469,6 +469,14 @@ impl Command {
                 let amount = parameters[2].parse::<Amount>()?;
                 let fee = parameters[3].parse::<Amount>()?;
 
+                let mut res = String::new();
+                if let Ok(addresses_info) = client.public.get_addresses(vec![addr]).await {
+                    for info in addresses_info {
+                        if info.ledger_info.candidate_ledger_info.balance < amount + fee {
+                            res.push_str("Warning: this operation may be rejected due to insuffisant balance\n");
+                        }
+                    }
+                }
                 send_operation(
                     client,
                     wallet,
--- a/massa-client/src/cmds.rs
+++ b/massa-client/src/cmds.rs
@@ -469,14 +469,22 @@ impl Command {
                 let amount = parameters[2].parse::<Amount>()?;
                 let fee = parameters[3].parse::<Amount>()?;
 
-                let mut res = String::new();
-                if let Ok(addresses_info) = client.public.get_addresses(vec![addr]).await {
-                    for info in addresses_info {
-                        if info.ledger_info.candidate_ledger_info.balance < amount + fee {
-                            res.push_str("Warning: this operation may be rejected due to insuffisant balance\n");
+                match amount.checked_add(fee) {
+                    Some(total) => {
+                        if let Ok(addresses_info) = client.public.get_addresses(vec![addr]).await {
+                            if !addresses_info.is_empty()
+                                && addresses_info[0].ledger_info.candidate_ledger_info.balance
+                                    < total
+                            {
+                                println!("Warning: this operation may be rejected due to insuffisant balance");
+                            }
                         }
                     }
+                    None => {
+                        bail!("The total amount hit the limit overflow, operation rejected")
+                    }
                 }
+
                 send_operation(
                     client,
                     wallet,
--- a/massa-client/src/cmds.rs
+++ b/massa-client/src/cmds.rs
@@ -473,6 +473,19 @@ impl Command {
                 let roll_count = parameters[1].parse::<u64>()?;
                 let fee = parameters[2].parse::<Amount>()?;
 
+                if let Ok(addresses_info) = client.public.get_addresses(vec![addr]).await {
+                    match addresses_info.get(0) {
+                        Some(info) => {
+                            if info.ledger_info.candidate_ledger_info.balance < fee
+                                || roll_count > info.rolls.candidate_rolls
+                            {
+                                println!("WARNING: this operation may be rejected due to insuffisant balance or roll count");
+                            }
+                        }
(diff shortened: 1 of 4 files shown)


yvan-sraka:
Previously here https://github.com/massalabs/massa/pull/1819

massa-client/src/cmds.rs
```diff
@@ -651,12 +651,24 @@ impl Command {
                         Err(e) => return repl_err!(e),
                     };
 
+                    let mut res = String::new();
+                    if let Ok(addresses_info) = client.public.get_addresses(vec![addr]).await {
+                        for info in addresses_info {
+                            if info.ledger_info.locked_balance > amount {
```
AureliaDolo, line 657: Candidate balance is the most relevant here. It's the balance at the most recent blocks that we consider valid. Final balance is the balance at the latest final blocks and locked balance is the balance that is locked after a roll sell.

Moreover, you must check that amount + fee is enough

AureliaDolo (changes requested):

models/src/amount.rs
```diff
@@ -11,6 +11,13 @@ pub const AMOUNT_DECIMAL_FACTOR: u64 = 1_000_000_000;
 #[derive(Clone, Copy, Debug, PartialEq, Eq, Ord, PartialOrd, Default)]
 pub struct Amount(u64);
 
+impl std::ops::Add for Amount {
```
AureliaDolo, line 14: We choose to use saturating add or checked add and not just add

@damip I don't think we documented why and when to use one instead of the other ?

adrien-zinger: We could have an overflow that make the node panic :thinking:
Alright I'll use checked_add in this case

massa-client/src/cmds.rs
```diff
@@ -469,6 +469,22 @@ impl Command {
                 let amount = parameters[2].parse::<Amount>()?;
                 let fee = parameters[3].parse::<Amount>()?;
 
+                match amount.checked_add(fee) {
+                    Some(total) => {
+                        if let Ok(addresses_info) = client.public.get_addresses(vec![addr]).await {
+                            if !addresses_info.is_empty()
```
AureliaDolo, line 475: I think if there is no address info we want to display something like "no balance information found for this address"

(can you use `get` here ? the `.is_empty` quite confused me at first)

massa-client/src/cmds.rs
```diff
@@ -469,6 +469,22 @@ impl Command {
...
+                                && addresses_info[0].ledger_info.candidate_ledger_info.balance
+                                    < total
+                            {
+                                println!("Warning: this operation may be rejected due to insuffisant balance");
+                            }
+                        }
+                    }
+                    None => {
+                        bail!("The total amount hit the limit overflow, operation rejected")
```
AureliaDolo, line 484: I'm not sure we want to fail here, maybe just display the warning and send the operation.

AureliaDolo (changes requested):

massa-client/src/cmds.rs
adrien-zinger: Really? I thought that the node may try to process the same operation and panic anyway. So here or later it doesn't change a lot

models/src/amount.rs
adrien-zinger: Sadly, this trait is needed by checked_add

massa-client/src/cmds.rs
adrien-zinger: yep!

massa-client/src/cmds.rs
AureliaDolo: You point is valid and I do not have a clear answer here :/

I would not assume that the client and the node will fail on the same limits. Nor that the client will always be directly connected to the node. But it make sense to reduce as much as possible the load on the node.

models/src/amount.rs
damip: We always use saturating_XX and checked_XX on amounts because they are critical. Let's not use std::ops::Add

massa-client/src/cmds.rs
damip: We should NEVER panic for things that come from the outside (API or network) ! Otherwise the system can be attacked very easily ! If there is an overflow, just consider the operation invalid.

AureliaDolo: In that case it's only the client that panics

damip: still not good :) it should just tell us that the operation is invalid

adrien-zinger: There is a lot of cases where the client can panic. For example if a key is not found in the wallet or at every rpc error. But I can also just print a error/warning.

Whatever, the main question is: should I return here without calling the node? Or should I call the node even if I know that it will fail? I think we can just say `Sorry it's impossible ...` and return directly.

massa-client/src/cmds.rs
adrien-zinger: I changed that :wink:

models/src/amount.rs
adrien-zinger: done

massa-client/src/cmds.rs
damip: >There is a lot of cases where the client can panic. For example if a key is not found in the wallet or at every rpc error. But I can also just print a error/warning.

The client should absolutely NOT panic in any of those cases, just print an error explaining what the user typed wrong.
This needs to be corrected. Can someone create the issues ?

In general, untrusted input validation should be there to prevent crashes.

> should I return here without calling the node? Or should I call the node even if I know that it will fail? I think we can just say Sorry it's impossible ... and return directly.

The node should be called anyways.
The warning message is optional: try to get the info, print the warning if necessary, but if gathering info fails, just ignore it.
In any case, try sending the operation, and if there is an error, print the error explaining what happened.

adrien-zinger: Thank you @damip. I'll rewrite this part and request you a new verification. Check this issue #1850, we'll look at this later, anyway we wanted to verify if we can test scenarios with assert_cmd and it was already in the backlog

adrien-zinger: @AureliaDolo feel free to close this thread if you're agree!

AureliaDolo (approved):

massa-client/src/cmds.rs
```diff
@@ -469,6 +469,24 @@ impl Command {
                 let amount = parameters[2].parse::<Amount>()?;
                 let fee = parameters[3].parse::<Amount>()?;
 
+                match amount.checked_add(fee) {
+                    Some(total) => {
+                        if let Ok(addresses_info) = client.public.get_addresses(vec![addr]).await {
+                            match addresses_info.get(0) {
+                                Some(info) => {
+                                    if info.ledger_info.candidate_ledger_info.balance < total {
+                                        println!("WARNING: this operation may be rejected due to insuffisant balance");
```
yvan-sraka, line 478: Warning's prints could be painted on yellow using `console` crate!

massa-client/src/cmds.rs
```diff
@@ -2,7 +2,7 @@
 
 use crate::repl::Output;
 use crate::rpc::Client;
-use anyhow::{bail, Result};
+use anyhow::{anyhow, bail, Result};
```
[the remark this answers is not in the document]
yvan-sraka: And `clippy` is right!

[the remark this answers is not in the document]
AureliaDolo: Clippy is always right :christmas_tree:

adrien-zinger (approved):

yvan-sraka:
bors merge

PR #1832, opened 2021-11-19
Add mock of execution module in consensus

adrien-zinger:
Add mocks in consensus for the execution controller

### DIFF
--- a/consensus/src/tests/mock_execution_controller.rs
+++ b/consensus/src/tests/mock_execution_controller.rs
@@ -1,0 +1,76 @@
+// Copyright (c) 2021 MASSA LABS <info@massa.net>
+
+use execution::{ExecutionCommand, ExecutionCommandSender};
+use models::{Block, BlockHashMap};
+use time::UTime;
+use tokio::{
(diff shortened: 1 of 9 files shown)


consensus/src/tests/mock_execution_controller.rs
```diff
@@ -0,0 +1,76 @@
+// Copyright (c) 2021 MASSA LABS <info@massa.net>
+
+use execution::{ExecutionCommand, ExecutionCommandSender};
+use models::{Block, BlockHashMap};
+use time::UTime;
+use tokio::{
+    sync::mpsc::{channel, Receiver, Sender},
+    time::sleep,
+};
+
```
gterzian, line 10: I think you want to add something similar to the current `PoolCommandSink`, to prevent deadlock if the channel were to fill-up, and to properly shut it down at the end of the test.

gterzian: Looks good otherwise!

yvan-sraka (approved):

PR #1838, opened 2021-11-22
store more final blocks

damip:


consensus/src/block_graph.rs
```diff
@@ -3727,6 +3727,22 @@ impl BlockGraph {
             }
         }
 
+        // retain extra history according to the config
+        // this is useful to avoid desync on temporary connection loss
+        for a_block in self.active_index.iter() {
```
AureliaDolo, line 3732: is this ok to go through every active block ?

consensus/src/block_graph.rs
```diff
@@ -3727,6 +3727,22 @@ impl BlockGraph {
...
+            if let Some(BlockStatus::Active(ActiveBlock { block, .. })) =
+                self.block_statuses.get(a_block)
+            {
+                let (_b_id, latest_final_period) =
+                    self.latest_final_blocks_periods[block.header.content.slot.thread as usize];
+                if block.header.content.slot.period
+                    >= latest_final_period.saturating_sub(self.cfg.force_keep_final_periods)
```
AureliaDolo, line 3739: You can compute that just once for each thread and re use the value

consensus/src/block_graph.rs
damip: there is no index by slot, and we already do it a few times elsewhere in the prune_active method. So it doesn't change the overall complexity of the method, it just adds one to the complexity pre-factor

damip:
bors merge

adrien-zinger:
Just a comment about the `prune_active` function logic.

It's a little bit strange that we need to create a temporary variable that flag what we want to keep and then remove the difference.

And then in the loop that remove unused active blocks, with the `block_statuses` variable, we remove the discarded_active just before reinserting almost the same thing inthere.

damip:
@adrien-zinger can you propose an optimization in a followup issue ?

consensus/src/block_graph.rs
AureliaDolo: We do it only once before, why don't you put everything you need in that one loop ?

damip: The force-keep part needs to be done after "grow with parents & fill thread holes twice", otherwise we will also "grow" the force-kept set and try to fill holes that cannot exist there

damip: If you have something special in mind, feel free to propose it

AureliaDolo: You could fill `to retain` with this criterion and a `to_grow` variable that would only have the blocks we want to grow dependencies from. And once `to_grow` has grow you can extend `to_retain` with it.

consensus/src/block_graph.rs
AureliaDolo: It would be a way to make the code easier to read.

AureliaDolo (commented):
Overall, I think it would be a great opportunity to review the clean up process, especially after having removed storage.

adrien-zinger:
@damip I'll check better this part later and if I see an optimisation I'll create an issue or a PR ;)

consensus/src/block_graph.rs
damip: would that improve performance and readability ? looks like one more memory allocation

damip:
someone wants to take over this PR for further improvements ?

To me it is already OK. A followup can be made to try to improve performance.

One requirement: this needs to be merged this week to be ready for 5.0.

consensus/src/block_graph.rs
AureliaDolo: But one less loop on everything

AureliaDolo (approved):
Opened up #1843 as a follow up

AureliaDolo:
bors merge

PR #1839, opened 2021-11-22
Use a `& 'static` ref of settings rather than cloning configuration at runtime across modules

yvan-sraka:
There are a lot of changes here move around! To get what this PR does you can start by looking at changes made in  `massa-node`!


yvan-sraka:
A follow-up could be to also read configuration settings this way in `consensus` module :)

massa-node/src/main.rs
```diff
@@ -182,8 +181,66 @@ async fn launch(
...
+        .with_max_level(match SETTINGS.logging.level {
+            4 => Level::TRACE,
+            3 => Level::DEBUG,
+            2 => Level::INFO,
+            1 => Level::WARN,
+            _ => Level::ERROR,
+        })
+        .init();
+
+    // run
```
AureliaDolo, line 243: Why did you move the run part outside of its function ?

AureliaDolo (commented):
Why did you renamed some xxxConfig with xxxSettings ?

massa-node/src/main.rs
yvan-sraka: because having it in a separate function call have no purpose here

adrien-zinger:
> Why did you renamed some xxxConfig with xxxSettings ?

The reason is to keep in mind what is new and what is older.

adrien-zinger:
I have tested to a node and it works!

adrien-zinger (approved):

yvan-sraka:
bors merge

PR #1847, opened 2021-11-23
[OPTI] Fix warnings in consensus/block_graph

adrien-zinger:
Takes the file block_graph and works on clippy warnings

### DIFF
--- a/consensus/src/block_graph.rs
+++ b/consensus/src/block_graph.rs
@@ -110,6 +110,21 @@ impl From<ActiveBlock> for ExportActiveBlock {
     }
 }
 
+impl From<Box<ActiveBlock>> for ExportActiveBlock {
+    fn from(block: Box<ActiveBlock>) -> Self {
+        ExportActiveBlock {
+            block: block.block,
+            parents: block.parents,
+            children: block.children,
+            dependencies: block.dependencies,
+            is_final: block.is_final,
+            block_ledger_changes: block.block_ledger_changes,
+            roll_updates: block.roll_updates,
+            production_events: block.production_events,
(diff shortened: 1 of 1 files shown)


consensus/src/block_graph.rs
```diff
@@ -110,6 +110,21 @@ impl From<ActiveBlock> for ExportActiveBlock {
     }
 }
 
+impl From<Box<ActiveBlock>> for ExportActiveBlock {
```
AureliaDolo, line 113: Why did you boxed active block ?

adrien-zinger: For almost the same error as #1265.

In the enum bellow, the variant Active<ActiveBlock> is a lot larger in memory than the other variants.
So in memory, because the data has to be aligned, all other variants takes the same size than Active<ActiveBlock> and keep useless space (an offset) to be aligned.

```rust
enum BlockStatus {
    Incoming(HeaderOrBlock),
    WaitingForSlot(HeaderOrBlock),
    WaitingForDependencies {
        header_or_block: HeaderOrBlock,
        unsatisfied_dependencies: BlockHashSet, // includes self if it's only a header
        sequence_number: u64,
    },
    Active(ActiveBlock), // I replaced it by Active<Box<ActiveBlock>>
    Discarded {
        header: BlockHeader,
        reason: DiscardReason,
        sequence_number: u64,
    },
}
```

Note: There is also some macros used to tell specifically "don't align that!" but it will produce a slower execution.

AureliaDolo (approved):

yvan-sraka:
bors merge

PR #1851, opened 2021-11-24
Feature use get endorsement by xxx in api and client

AureliaDolo:


AureliaDolo:
@qdrn @Noxhaa this will change the api

qdrn:
Ok thanks for notifying. As long as it's retro-compatible it's not a big deal, we need to take special care when things change and there is a risk to break the explorer.

AureliaDolo:
@yvan-sraka I'm not sure if there is still something to do in the client

yvan-sraka:
I will check that! Maybe I can do that in follow-up?

yvan-sraka (approved):

yvan-sraka:
I checked the client :heavy_check_mark:!

yvan-sraka:
bors merge

PR #1853, opened 2021-11-25
Added logs to fix #1817

### LINKED ISSUE #1817: Add block receptions to higher lvl log
I think it's important to log when blocks (and maybe headers) are received in order to make statistics about block reception timings. Currently it's only available in log lvl 4 but lvl 4 log is way too heavy to run a node.

It could be only the hash and timing. No need to log the whole block as is currently done in lvl 4.

It could be in lvl 2 or 3. To be discussed.

AureliaDolo:


qdrn:
It log blocks several time (each time it is received). As discussed in #1817 it's better to log it only the first time it is processed.
I think changing the log lvl of this line is enough https://github.com/massalabs/massa/blob/main/consensus/src/block_graph.rs#L1891
It would be nice to log the same thing for block header as well.

It would be nice to log the corresponding slot as well. Something like that (to be adapted for blocks):
`info!("consensus.block_graph.incoming_header block_id {} header {}", block_id, header.content.slot);`

AureliaDolo:
> It would be nice to log the same thing for block header as well.

The header is quite huge so I wouldn't display at a high log level.

qdrn (approved):

AureliaDolo:
bors merge

PR #1855, opened 2021-11-25
Fix #1146 added when episode ends client command.

### LINKED ISSUE #1146: Add get end client command
In GitLab by @AureliaDolo

Add a get_end command to the client to get the end timestamp of that episode.

And maybe in a countdown form like : 3 days and 7h17 remaining in episode X

AureliaDolo:


models/src/version.rs
```diff
@@ -8,7 +8,7 @@ use std::{convert::TryInto, fmt, str::FromStr};
 #[derive(Debug, Clone, Copy, PartialEq, Eq)]
 pub struct Version {
     instance: [char; 4], // ascii uppercase alpha
-    major: u32,
+    pub major: u32,
```
yvan-sraka, line 11: Why not make all fields of this struct `pub`?

time/src/lib.rs
```diff
@@ -300,4 +300,14 @@ impl UTime {
         let datetime: DateTime<Utc> = DateTime::from_utc(naive, Utc);
         format!("{}", datetime)
     }
+
+    pub fn days_hours_mins_secs(&self) -> (u64, u64, u64, u64) {
```
yvan-sraka, line 304: I'm no sure about the naming, and maybe it would be more obvious to return a custom named struct rather than a tuple struct?

yvan-sraka (dismissed):

time/src/lib.rs
```diff
@@ -300,4 +300,14 @@ impl UTime {
...
+        let days = self.0 / (1000 * 24 * 60 * 60);
+        let hours = (self.0 - days * 1000 * 24 * 60 * 60) / (1000 * 60 * 60);
+        let mins = (self.0 - days * 1000 * 24 * 60 * 60 - hours * 1000 * 60 * 60) / (1000 * 60);
+        let secs =
+            (self.0 - days * 1000 * 24 * 60 * 60 - hours * 1000 * 60 * 60 - mins * 1000 * 60)
```
damip, line 309: Rule 1 : never reinvent cryptographic primitives, use libs, otherwise you WILL introduce security problems
Rule 2 : never reinvent calendar primitives, use libs, otherwise you WILL miscalculate (ex: leap seconds, precession corrections, etc...)

time/src/lib.rs
AureliaDolo: Any ideas on the name of the struct ?

time/src/lib.rs
```diff
@@ -300,4 +300,19 @@ impl UTime {
         let datetime: DateTime<Utc> = DateTime::from_utc(naive, Utc);
         format!("{}", datetime)
     }
+
```
adrien-zinger, line 303: Add a small documentation here

massa-client/src/cmds.rs
```diff
@@ -482,6 +488,21 @@ impl Command {
                 )
                 .await
             }
+            Command::when_episode_ends => {
+                if let (Some(end), version) = match client.public.get_status().await {
+                    Ok(node_status) => (node_status.algo_config.end_timestamp, node_status.version),
+                    Err(e) => bail!("RpcError: {}", e),
+                } {
+                    let (days, hours, mins, secs) =
+                        end.saturating_sub(UTime::now(0)?).days_hours_mins_secs()?; // TODO set compensation millis
```
adrien-zinger, line 497: Can we find easily the compensation in the client? Or do we need to update the node API?

adrien-zinger (dismissed):

models/src/version.rs
AureliaDolo: Good question

massa-client/src/cmds.rs
AureliaDolo: I added the compensation millis to the api.

So @qdrn @Noxhaa get_status has one more field.

time/src/lib.rs
AureliaDolo: It's a duration not a date so a minute is always 60 secs and so on :)

yvan-sraka (approved):

yvan-sraka (approved):

models/src/version.rs
damip: in what case do we need to directly access any of those fields ?

time/src/lib.rs
damip: a day is not always 24 hours ;)

massa-client/src/cmds.rs
damip: We should NOT send the compensation_millis from the node to the client !
In many cases, the client won't even be running on the same computer as the node.
Use compensation_millis = 0 in the client

time/src/lib.rs
AureliaDolo: To do that we must use [datetime](https://docs.rs/chrono/0.4.19/chrono/struct.DateTime.html), but from what I recall we didn't want to go down that path.

massa-client/src/cmds.rs
AureliaDolo: So the now in the client that is used to process data won't be the same now as in the node that produced that data ?

models/src/version.rs
AureliaDolo: I used major to know which testnet will end

massa-client/src/cmds.rs
damip: yup, that's ok, especially if the client is running on a different computer

time/src/lib.rs
damip: ok then let's keep it like that...

models/src/version.rs
damip: Makes me wonder if this PR is useful enough to justify making such things public... especially since this whole PR will be reverted at the end of the testnet, so the time spent working on it should be extremely limited

AureliaDolo: What's the issue in making that public ? From what I understand after @yvan-sraka 's work on the config it will be available everywhere anyway

time/src/lib.rs
AureliaDolo: Done !

yvan-sraka (approved):

models/src/version.rs
damip: the less public things we expose, the better. Here we're talking about version.major, which is not only used in config, but also in inter-node communication, and will be used for versioning in the future. Remember when we did a whole bunch of "pub" removals ?

time/src/lib.rs
damip: why not just return a formatted string ?

pub fn format_human_readable(&self) -> String

you can also use: https://crates.io/crates/hhmmss

AureliaDolo: This crate does not give days.

format human readable is ambiguous : is it human readable as a date or as a duration ?

models/src/version.rs
AureliaDolo: Done

massa-client/src/cmds.rs
AureliaDolo: Done !

models/src/version.rs
AureliaDolo: Removed the pub

massa-client/src/cmds.rs
AureliaDolo: Done

yvan-sraka (approved):

AureliaDolo:
bors merge

PR #1861, opened 2021-11-26
Fix #1824: error should be also handled in JSON format

### LINKED ISSUE #1824: Error should be also handled in JSON format

yvan-sraka:


AureliaDolo (approved):

adrien-zinger (commented):
You should add in your commit that you also add a macro. Or do it in another commit.

PR #1865, opened 2021-11-27
correct pos deserialization ram usage

damip:
The deserialization of the random bits used to take all the remainder of the buffer. Now it takes only the slice it needs.


sebastien-forestier (approved):
Heaptrack now says 40MB RAM for bootstrap instead of 4GB.

damip:
Bors merge

PR #1866, opened 2021-11-27
send only required blocks to bootstrap

damip:


damip:
I used the opportunity to clean up a whole bunch of useless and performance-hindering mess that was introduced recently in the block graph bootstrap graph/block import/export system

adrien-zinger (approved):
@damip Look good to me, I asked myself about how many blocks we were speaking about while bootstrapping before? VS the number of blocks required.

AureliaDolo (approved):

damip:
> @damip Look good to me, I asked myself about how many blocks we were speaking about while bootstrapping before? VS the number of blocks required.

initialy, about 100, after the optims about 1500

damip:
bors merge

PR #1868, opened 2021-11-27
demote node absent warn to debug

damip:


yvan-sraka:
It's a debug level message? But it start with error? Do we just want to silent it? Could we document it in a comment?

yvan-sraka (approved):

yvan-sraka:
bors merge

PR #1880, opened 2021-11-30
docmuent compensation millis

damip:


AureliaDolo (commented):
Maybe add that it should not be used in the client

damip:
> Maybe add that it should not be used in the client

No, I think it is self-explanatory now.

It's like if you have a screen that is too blue, and you setup a color correction on your program A.
If another program B remotely connects to your program A, it does not make any sense to apply the same correction in B as it may not run on the same computer at all.
B can decide either to not do any correction (compensation_millis=0 in the client) or to compute its own local correction.

AureliaDolo (approved):

AureliaDolo:
bors merge

PR #1918, opened 2021-12-03
Fix warning in massa-hash

adrien-zinger:
Rename hash::hash to hash::from

### DIFF
--- a/bootstrap/src/client_binder.rs
+++ b/bootstrap/src/client_binder.rs
@@ -45,7 +45,7 @@ impl BootstrapClientBinder {
             let mut random_bytes = [0u8; BOOTSTRAP_RANDOMNES_SIZE_BYTES];
             StdRng::from_entropy().fill_bytes(&mut random_bytes);
             self.duplex.write_all(&random_bytes).await?;
-            let rand_hash = Hash::hash(&random_bytes);
+            let rand_hash = Hash::from(&random_bytes);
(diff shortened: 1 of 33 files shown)


gterzian (approved):
LGTM. What was the warning?

AureliaDolo (commented):
These warnings can be fixed, if you want
pos.rs L81 L87 L93 L202 L208 use of `ok_or` followed by a function call
pos.r2 l 187 struct `RollCounts` has a public `len` method, but no `is_empty` method
operation_pool.rs L244 avoid using `collect()` when not needed
operation_pool.rs L330 use of `ok_or` followed by a function call

Why did you rename Hash::hash into Hash::from ?

yvan-sraka:
bors merge

adrien-zinger:
> These warnings can be fixed, if you want pos.rs L81 L87 L93 L202 L208 use of `ok_or` followed by a function call pos.r2 l 187 struct `RollCounts` has a public `len` method, but no `is_empty` method operation_pool.rs L244 avoid using `collect()` when not needed operation_pool.rs L330 use of `ok_or` followed by a function call
> 
> Why did you rename Hash::hash into Hash::from ?

Hash::hash had a redondant warning. I think "from" is more idiomatic. It's my side quest to resolve all warnings in massa world :superhero: I resolve one file at a time!

PR #1966, opened 2021-12-08
Add is_staking to AddressInfo and AddressState.

### LINKED ISSUE #1963: add if an address is staking (!= has rolls) in wallet info
previously in the client in wallet info and address info we had [staking] next to addresses that were also in staking addresses

AureliaDolo:
Fix #1963 

I will update the wiki once it's merged.


AureliaDolo:
By the way if you have a better tag than [staking] it would be nice

AureliaDolo:
And staking keys should remain only in the private api.

@yvan-sraka what happens if a command is implemented both in the private and in the public api, given that it will be slightly different (in this case move that updated implementation to the private api one)

PR #1968, opened 2021-12-08
Add hints in wallet add/generate private key.

### LINKED ISSUE #1796: Automatically add newly created/added addresses to node's staking keys, if connected
When creating or adding an address to the wallet, if the node is connected sent it given addresses to use as staking keys.

Or maybe just add an option to do that 

That raises the question: should we check if addresses in the wallet are the same as staking addresses and if not add them where they are missing

AureliaDolo:
Fix #1796

### DIFF
--- a/massa-client/src/cmds.rs
+++ b/massa-client/src/cmds.rs
@@ -12,6 +12,7 @@ use models::{
 };
 use serde::Serialize;
 use signature::{generate_random_private_key, PrivateKey, PublicKey};
+use std::collections::HashMap;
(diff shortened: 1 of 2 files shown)


yvan-sraka (approved):

AureliaDolo:
By the way @damip do you have an opinion on what should return the json output of wallet_add_private_keys and wallet_generate_private_key ? Should it returns only the addresses, or also the private/public keys ?

damip:
I think displaying the added addresses is enough

PR #1978, opened 2021-12-10
Improve stop execution worker.

AureliaDolo:
I tried to fix the shutdown issue but was not able to


AureliaDolo:
Cherry picked here #1979

PR #1837, opened 2021-11-22
Adding VM to skeleton

gterzian:
Part of https://github.com/massalabs/massa/issues/1830

PR #1979, opened 2021-12-10
Start the integration of execution

adrien-zinger:
### Integrate the execution module developed in cautious bassoon. Just being *fully agnostic* with a common and reusable interface.

:stars: Remove the logic of VM wasmer and start a full agnostic solution

- I first rebased the add_vm_skeleton branch into master (then I merged main while working)
- Add the agnostic solution
- Add bootstrap sce ledger
- send smart contract command line
- Improve the stop


damip:
@adrien-zinger  can you take the following changes into account ?

https://github.com/massalabs/massa/pull/1837/commits/8398d61b98d54df0486dae3af1f3841d9a3f2b98

* I massively simplified the messaging
* * I repaired the STOP mechanism by moving the condition variable wait at the bottom of the loop

damip:
In the add_vm_to_skeleton version, we used to call reset_vm at the beginning of every blockclique change. It was removed in this PR, thus making the execution algos incorrect

adrien-zinger:
> In the add_vm_to_skeleton version, we used to call reset_vm at the beginning of every blockclique change. It was removed in this PR, thus making the execution algos incorrect

I've reintegrated the thread system and so set back the self.reset_to_final();

gterzian:
@adrien-zinger Ok looks great, can we replace the `static` with members on the `VM`?

gterzian:
I will open a PR on this branch to integrate `sc-runtime` with traits...

damip:
`let mut remaining_operation_count = settings::MAX_GAS_PER_BLOCK as usize;`  was put in Consensus block creation. This is wrong. Corrected it.

massa-execution/src/worker.rs
```diff
@@ -240,121 +239,194 @@ impl ExecutionWorker {
...
-            .ordered_pending_css_final_blocks
-            .iter()
-            .cloned()
+        // 3 - process CSS-active blocks
+
+        // define `sce_active_blocks = blockclique_blocks UNION pending_css_final_blocks`
+        let new_blockclique_blocks = blockclique.iter().filter_map(|(b_id, b)| {
+            if b.header.content.slot <= self.last_final_slot {
+                // eliminate blocks that are not from a stricly later slot than the current latest SCE-final one
+                // (this is an optimizeation)
```
AureliaDolo, line 390: typo

massa-execution/src/lib.rs
```diff
@@ -1,3 +1,5 @@
+#![feature(map_first_last)]
```
AureliaDolo, line 1: What does that feature ?

massa-execution/src/worker.rs
```diff
@@ -240,121 +239,194 @@ impl ExecutionWorker {
...
+        // 1 - reset the SCE state back to its latest final state
 
-        // stop the current VM execution and reset state to final
+        // revert the VM to its latest SCE-final state by clearing its active slot history.
         // TODO make something more iterative/conservative in the future to reuse unaffected executions
         self.reset_to_final();
 
-        // stop the current VM execution and reset state to final
-        // TODO reset VM
+        // set `last_active_slot = last_final_slot
```
AureliaDolo, line 310: Not sure about that comment

massa-execution/src/worker.rs
```diff
@@ -240,121 +239,194 @@ impl ExecutionWorker {
...
+                    .map(|(s, _v)| *s)
+                {
+                    // there is a block B at slot S in `pending_css_final_blocks`:
+                    Some(b_slot) if b_slot == s => {
+                        // remove B from `pending_css_final_blocks`
+                        // cannot panic, checked above
+                        let (_s, (b_id, b)) = self
+                            .pending_css_final_blocks
+                            .pop_first()
+                            .expect("pending_css_final_blocks wa unexpectedly empty");
```
AureliaDolo, line 346: typo: was

massa-execution/src/worker.rs
```diff
@@ -240,121 +239,194 @@ impl ExecutionWorker {
...
+                        // call the VM to execute the SCE-final block B at slot S
+                        self.push_request(ExecutionRequest::RunFinalStep(ExecutionStep {
+                            slot: s,
+                            block: Some((b_id, b)),
+                        }));
+                        // set `last_active_slot = last_final_slot = S`
```
AureliaDolo, line 352: why that comment ?

AureliaDolo (commented):
I feel like I read the same thing multiple times but slightly different, Would be nice if at some point we take some time to clean up the code before merging it

massa-execution/src/lib.rs
damip: > I feel like I read the same thing multiple times but slightly different, Would be nice if at some point we take some time to clean up the code before merging it

Can you give more details ?

massa-execution/src/worker.rs
damip: it's just there so that every comment matches word-to-word one line in the spec, and that both can be read linearly

massa-execution/src/worker.rs
damip: corrected

massa-execution/src/worker.rs
damip: corrected

massa-execution/src/worker.rs
damip: see Greg's comments on https://github.com/massalabs/massa/pull/2019

=> match comments 1-to-1 with the spec for linear reading

massa-execution/src/worker.rs
AureliaDolo: maybe here it's a little bit overkill to just repeat the code in a comment.

massa-execution/src/worker.rs
AureliaDolo: not sure it helps with linear reading in this case

AureliaDolo:
@damip in vs code there is an option to format on save, that can help with the sanity if you forget to run `cargo fmt`

AureliaDolo (commented):
There are quite a few clippy warnings that could easily be tackled.

For what I recall, there may be quite a few methods that return results only because they were computing an address from a public key, So it may be worth having a look on that at some point.

massa-execution/src/lib.rs
damip: > What does that feature ?

allows accessing the first and last elments of a btreemap/btreeset

massa-execution/src/worker.rs
damip: removed

massa-execution/src/worker.rs
damip: removed

damip:
corrected all comments here https://github.com/massalabs/massa/pull/2042

AureliaDolo (approved):

AureliaDolo (changes requested):
There is an issue with cargo check

AureliaDolo (dismissed):

damip (approved):

damip:
bors merge

PR #1986, opened 2021-12-11
Add slot filler

adrien-zinger:
> *"In the execution module, run misses between the last active executed slot and the current latest slot"* - @damip 

@damip sorry I reopen this because I accidentally merged some non compiling code! :baby_chick:


damip:
was this merged anywhere ?

PR #1989, opened 2021-12-13
Execution: re-add test for start/stop without commmands

gterzian:
Re-adding tests, since the initial problem only became apparent with a test where no commands where sent.


gterzian:
@damip this one is ready for review.

damip (approved):

PR #1995, opened 2021-12-14
Put context and interface on vm

gterzian:


gterzian:
@adrien-zinger I'll work more on this tomorrow, and perhaps you can already take a look and let me know what you think?

I moved some stuff from the `sc-runtime` repo to make it easier to make this change, so the idea is that an `Interface` trait would instead be provided by `sc-runtime`, and pass to calls like `call`.

adrien-zinger:
> @adrien-zinger I'll work more on this tomorrow, and perhaps you can already take a look and let me know what you think?
> 
> I moved some stuff from the `sc-runtime` repo to make it easier to make this change, so the idea is that an `Interface` trait would instead be provided by `sc-runtime`, and pass to calls like `call`.

Hey! Looking at the code, I'm still not sure that the "context" should be in the interface because Context is for only the massa part and not for the both --> That the purpose of the interface, to be *context* :free:.

May be I still don't understand what you're trying to do so if I'm wrong don't take the following into account. :pray: 
As well as the interface is defined in the runtime-sc we don't want to import massa structures. The execution context contain too much non agnostic informations

```rust
#[derive(Clone)]
/// Stateful context, providing an execution context to massa between module executions
pub(crate) struct ExecutionContext {
    pub ledger_step: SCELedgerStep,
    pub max_gas: u64,
    pub coins: Amount,
    pub gas_price: Amount,
    pub slot: Slot,
    pub opt_block_id: Option<BlockId>,
    pub opt_block_creator_addr: Option<Address>,
    pub call_stack: VecDeque<Address>,
}
```

For the traits, be aware, use also the runtime-sc, if you prefer, import all the module to work on this improvement. Or an alternative would be to start from *runtime-sc* and trying to redefine the interface structure and use traits instead. Doing this you'll be sure that your work won't be incompatible later. The reality is that we need to embed the interface in an Env that implement WasmerEnv and that, that caused lot of lifetime issues when we tried with @yvan-sraka.

gterzian:
> Hey! Looking at the code, I'm still not sure that the "context" should be in the interface because Context is for only the massa part and not for the both --> That the purpose of the interface, to be context free.

Yes I agree, which is why I was trying out making `Interface` a trait, which would be imported by massa and implemented by a struct which would own a reference to the context. 

In other words, the interface would remain unaware of the context, and other massa things, only the massa implementation of the interface would be. 

I will open a pr in the runtime repo to propose the change, because from the code here I can see that it can compile without problem.

gterzian:
Closing this in favor of a new PR for https://github.com/massalabs/massa/issues/2021

PR #2003, opened 2021-12-15
Fix issue in `test_if_node_stop`

yvan-sraka:
https://github.com/massalabs/massa/runs/4532141114?check_suite_focus=true#step:6:82


massa-client/src/tests/scenarios.rs
```diff
@@ -1,40 +1,51 @@
...
+#[serial]
+async fn client_exit_gracefully() {
+    run_client_cmd("exit").await.unwrap();
+}
+
+#[tokio::test]
+#[serial]
+async fn node_stop_gracefully() {
+    run_client_cmd("node_stop").await.unwrap();
+    // TODO: @adrien-zinger do we have a better way than `.join().unwrap()` to
```
AureliaDolo, line 49: there is no join anymore

massa-client/src/tests/scenarios.rs
```diff
@@ -1,40 +1,51 @@
...
+        let output = std::str::from_utf8(
+            &Command::new("cargo")
+                .args(&["run", "--", cmd, "--json"])
+                .assert()
+                .get_output()
+                .stdout,
+        )?
+        .to_string();
+        // TODO: compare with tcp connect error as a JSON object?
+        if !output.contains("tcp connect error") {
```
AureliaDolo, line 19: This and the multiple tentative are only relevant for the first command after a node is launched, not for every command

massa-client/src/tests/scenarios.rs
adrien-zinger: Can we join something?

massa-client/src/tests/scenarios.rs
```diff
@@ -1,40 +1,51 @@
...
-async fn test_if_node_stop() {
-    let massa_node_thread_handle = thread::spawn(|| {
-        Command::cargo_bin("massa-node")
+async fn run_node() {
+    // TODO: @adrien-zinger do we have a better way to spawn `massa_node`?
+    std::thread::spawn(|| {
+        Command::new("cargo")
+            .arg("run")
+            .current_dir("../massa-node")
             .unwrap()
```
adrien-zinger, line 35: I'm afraid that this could run since the end of time. How do you replace the timeout?

yvan-sraka: What good value did you advise for a timeout?

massa-client/src/tests/scenarios.rs
yvan-sraka: Humm, my concern was more: make the test robust to network issue/wait for the node to be started, if the node is already launched, this issue just not occur, so it not try several tentative/wait and if it fails for other reason it just propagates the error, this is for me the correct behavior.

massa-client/src/tests/scenarios.rs
yvan-sraka: I guess this is just an old comment I left and there is nothing to join here. My concern was more to have a review on: did I handle everything right?

yvan-sraka:
> @yvan-sraka I don't get it :thinking:

@adrien-zinger did you don't get what I try to do? (having a node around the corner to test any client command (to automate what described here #1347) or did you don't get what is the bug that currently occurs? (https://github.com/massalabs/massa/runs/4532141114?check_suite_focus=true#step:6:82) or did you don't get the solution I suggest?

massa-client/src/tests/scenarios.rs
AureliaDolo: I would rather have a assert node is running function that retries a command until it's ok and then another one for just running a command and retrieving the output.

massa-client/src/tests/scenarios.rs
AureliaDolo: So you should remove it

AureliaDolo (commented):
Overall seems ok

yvan-sraka:
> And you committed your peers.json (for the stats right ? :p)

Fixed!

massa-client/src/tests/scenarios.rs
AureliaDolo: If you go for the full scenario, I would tell 10 minutes, given that genesis timestamp is between 3 and 5 minutes in the future

AureliaDolo:
> > And you committed your peers.json (for the stats right ? :p)
> 
> Fixed!

No !

adrien-zinger:
> > @yvan-sraka I don't get it thinking
> 
> @adrien-zinger did you don't get what I try to do? (having a node around the corner to test any client command (to automate what described here #1347) or did you don't get what is the bug that currently occurs? (https://github.com/massalabs/massa/runs/4532141114?check_suite_focus=true#step:6:82) or did you don't get the solution I suggest?

I was tired probably, no I get it.

massa-client/src/tests/scenarios.rs
adrien-zinger: this test looks ok for me

massa-client/src/tests/scenarios.rs
adrien-zinger: I you just test if the node start and run, a timeout will run in error whatever you do. I think that the test for stoping the node is better.

AureliaDolo: the full scenario here #1347 finish by stopping the node, so that should not be an issue :)

yvan-sraka:
> > > And you committed your peers.json (for the stats right ? :p)
> > 
> > Fixed!
> 
> No!

Fixed my fix!

massa-client/src/tests/scenarios.rs
adrien-zinger: what if you failed stoping the node?

adrien-zinger (changes requested):
can we remove my name from comments that we will probably never delete after that? :rofl:

AureliaDolo:
I'm trying something out there #2014

yvan-sraka:
Closing in favor of https://github.com/massalabs/massa/pull/2014

PR #2006, opened 2021-12-16
Implement send_smart_contract client command.

### LINKED ISSUE #1998: [Client] Create Execute SC operation
I was thinking about adding a command like that 
```
send_smart_contract <sender_addresses> <path_to_byte_code> <max_gas> <gas_price> <available_coins> <fee>
```

Feel free to comment if you have any suggestions to make that more user friendly. (There will be no tooling around the massa app for now)

AureliaDolo:
Fix #1998


AureliaDolo:
I'm not sure  that it's the easier way to use that.

massa-client/src/cmds.rs
```diff
@@ -151,6 +154,17 @@ pub enum Command {
...
     )]
     send_transaction,
+
+    #[strum(
+        ascii_case_insensitive,
+        props(
+            args = "SenderAddress PathToBytecode MaxGas GasPrice Coins Fee",
+            todo = "[unstable] "
+        ),
+        message = "create and sends an operation containing byte code"
```
Noxhaa, line 164: Missing a "s" and an explanation

Creates and sends an operation containing bytecode to execute on-chain

massa-client/src/cmds.rs
```diff
@@ -613,6 +627,59 @@ impl Command {
...
+                                        if info.ledger_info.candidate_ledger_info.balance < total {
+                                            println!("WARNING: this operation may be rejected due to insuffisant balance");
+                                        }
+                                    }
+                                    None => println!("WARNING: address {} not found", addr),
+                                }
+                            }
+                        }
+                        None => {
+                            println!("WARNING: The total amount hit the limit overflow, operation will certainly be rejected");
```
Noxhaa, line 662: WARNING: the total [...]

Noxhaa (changes requested):

yvan-sraka (approved):

massa-client/src/cmds.rs
AureliaDolo: For consistency 'create and send'

massa-client/src/cmds.rs
AureliaDolo: Thanks !

yvan-sraka:
This is not the purpose of this issue, but I suggest coloring all the :warning: **WARNING** `println!` in yellow using `console` crate https://docs.rs/console/latest/console/#colors-and-styles :rainbow:!

massa-client/src/cmds.rs
```diff
@@ -613,6 +627,59 @@ impl Command {
                 }
                 Ok(Box::new(()))
             }
+            Command::send_smart_contract => {
+                if parameters.len() != 6 {
+                    bail!("wrong param numbers");
```
damip, line 632: wrong number of parameters

massa-client/src/cmds.rs
```diff
@@ -667,3 +734,12 @@ async fn send_operation(
 pub fn parse_vec<T: std::str::FromStr>(args: &[String]) -> anyhow::Result<Vec<T>, T::Err> {
     args.iter().map(|x| x.parse::<T>()).collect()
 }
+
+fn get_file_as_byte_vec(filename: &PathBuf) -> Result<Vec<u8>> {
```
damip, line 738: since we are in an async context, we could use asynchronous io that wouldn't block a whole thread. Also no need to touch the metadata and risk problems if the file changes in-between or something.

```rust 
async fn get_file_as_byte_vec(filename: &PathBuf) -> Result<Vec<u8>> {
  tokio::fs::read(filename).await
}
```

damip (changes requested):

massa-client/src/cmds.rs
AureliaDolo: I count 6 params :confused:

damip: I was just improving the error message :)

AureliaDolo: ah ok x)

massa-client/src/cmds.rs
damip: okay

damip (approved):

yvan-sraka:
bors merge

massa-client/src/cmds.rs
```diff
@@ -613,6 +625,59 @@ impl Command {
...
+                            if let Ok(addresses_info) =
+                                client.public.get_addresses(vec![addr]).await
+                            {
+                                match addresses_info.get(0) {
+                                    Some(info) => {
+                                        if info.ledger_info.candidate_ledger_info.balance < total {
+                                            println!("WARNING: this operation may be rejected due to insuffisant balance");
+                                        }
+                                    }
+                                    None => println!("WARNING: address {} not found", addr),
```
gterzian, line 655: should we `bail` in the cases where it's sure to fail?

gterzian (approved):
Looks good overall, just some nits...

massa-client/src/cmds.rs
AureliaDolo: We are not sure it will fail here. It may succeed if at the time we create the operation the address is not known of the network, but if it receives enough coins within its range of validity, it may be integrated.

Anyway the client does not guarantee that the operation is valid, only structurally valid. These warnings were added so people understand more easily why their transactions are never included.

gterzian: ok, thanks for the info.

Noxhaa (approved):

AureliaDolo:
bors merge

damip:
@AureliaDolo it will mostly be used through the API or called through the client CLI from a deployment tool

AureliaDolo:
> > I'm not sure that it's the easier way to use that.
> 
> @AureliaDolo it will mostly be used through the API or called through the client CLI from a deployment tool, but this will allow easy manual testing

Yeah but here I choose that the user is able to provide a path to a file containing bytes, instead of just copy pasting the bytes

damip:
> > > I'm not sure that it's the easier way to use that.
> > 
> > 
> > @AureliaDolo it will mostly be used through the API or called through the client CLI from a deployment tool, but this will allow easy manual testing
> 
> Yeah but here I choose that the user is able to provide a path to a file containing bytes, instead of just copy pasting the bytes

Yes that's a good choice

AureliaDolo:
bors merge

yvan-sraka:
bors merge

yvan-sraka:
bors merge

yvan-sraka:
bors merge

AureliaDolo:
bors merge

damip:
still broken...

AureliaDolo:
bors merge

yvan-sraka:
bors merge

AureliaDolo:
bors merge

damip:
Don't give up :D

PR #2019, opened 2021-12-19
Update vm feed algo

damip:


gterzian (changes requested):
Could we get more comments with links to a spec?

massa-execution/src/worker.rs
```diff
@@ -243,7 +241,7 @@ impl ExecutionWorker {
             while self.last_active_slot < end_step {
                 self.last_active_slot =
```
gterzian, line 242: I would prefer to assign to a variable like `next_slot`, with a comment `slot S`, followed by the call to `push_request` commented as `call the VM to execute an SCE-active miss at slot S`, and then assign `next_slot` to `self.last_active_slot` with a comment `set last_active_slot = S`, so that the code matches the spec more clearly. Ideally we should do this kind of commenting/matching of the spec everywhere.

massa-execution/src/worker.rs
```diff
@@ -255,106 +253,137 @@ impl ExecutionWorker {
...
-        mut finalized_blocks: BlockHashMap<Block>,
+        blockclique: BlockHashMap<Block>,
+        finalized_blocks: BlockHashMap<Block>,
     ) -> Result<(), ExecutionError> {
-        // filter out any already-finalized blocks
-        // this can happen for example on bootstrap because the consensus vs execution bootstrap snapshots may be out of sync
-        blockclique.retain(|_b_id, b| b.header.content.slot > self.last_final_slot);
-        finalized_blocks.retain(|_b_id, b| b.header.content.slot > self.last_final_slot);
-
         // stop the current VM execution and reset state to final
```
gterzian, line 259: `/// Revert the VM to its latest SCE-final state by clearing its active slot history.`

It would be even better to have a `Step 1` in the spec, link to the paragraph at the top of the function, and add `Step 1` to the comment, so that both the code and spec follow a clear number of steps which can be easily reviewed for correctness.

massa-execution/src/worker.rs
```diff
@@ -255,106 +253,137 @@ impl ExecutionWorker {
...
         // TODO make something more iterative/conservative in the future to reuse unaffected executions
         self.reset_to_final();
+        self.last_active_slot = self.last_final_slot;
```
gterzian, line 262: `/// Step 2: set last_active_slot = last_final_slot` (for example).

massa-execution/src/worker.rs
```diff
@@ -255,106 +253,137 @@ impl ExecutionWorker {
...
-        // TODO reset VM
-
-        // gather pending finalized CSS
-        let mut css_final_blocks: Vec<(BlockId, Block)> = self
-            .ordered_pending_css_final_blocks
-            .drain(..)
-            .chain(finalized_blocks.into_iter())
-            .collect();
-        css_final_blocks.sort_unstable_by_key(|(_, b)| b.header.content.slot);
+        // process SCE-final slots
```
gterzian, line 264: The spec I believe reads `Then, process CSS-final blocks:`, so either one should be updated.

massa-execution/src/worker.rs
```diff
@@ -255,106 +253,137 @@ impl ExecutionWorker {
...
+        self.pending_css_final_blocks
+            .extend(finalized_blocks.into_iter().filter_map(|(b_id, b)| {
```
gterzian, line 266: here I would assign the result of the `filter_map` to a variable named `new_css_final_blocks`, to match the wording of the spec.

massa-execution/src/worker.rs
```diff
@@ -255,106 +253,137 @@ impl ExecutionWorker {
...
+            let mut cur_slot = self.last_final_slot.get_next_slot(self.cfg.thread_count)?;
+            while cur_slot <= max_css_final_slot {
+                match self
+                    .pending_css_final_blocks
+                    .first_key_value()
+                    .map(|(s, _v)| *s)
+                {
+                    // there is a CSS-final block at cur_slot
+                    Some(b_slot) if b_slot == cur_slot => {
+                        // remove the entry from pending_css_final_blocks (cannot panic, checked above)
```
gterzian, line 287: Can we switch the order of those two steps, so that is matches the spec?

massa-execution/src/worker.rs
```diff
@@ -255,106 +253,137 @@ impl ExecutionWorker {
...
-        // list maximum thread slots
-        let mut max_thread_slot = vec![self.last_final_slot; self.cfg.thread_count as usize];
-        for (_b_id, block) in css_final_blocks.iter() {
-            max_thread_slot[block.header.content.slot.thread as usize] = std::cmp::max(
-                max_thread_slot[block.header.content.slot.thread as usize],
-                block.header.content.slot,
-            );
-        }
+                    // there is no CSS-final block at cur_slot, but there are CSS-final blocks later
+                    Some(_b_slot) => {
```
gterzian, line 301: should we not add a check here that the slot is indeed later?

massa-execution/src/worker.rs
```diff
@@ -255,106 +253,137 @@ impl ExecutionWorker {
...
+                        // check whether there is a CSS-final block later in the same thread
+                        let mut check_slot = Slot::new(cur_slot.period + 1, cur_slot.thread);
+                        while check_slot <= max_css_final_slot {
+                            if self.pending_css_final_blocks.contains_key(&check_slot) {
+                                break;
+                            }
+                            check_slot.period += 1;
+                        }
+                        if check_slot <= max_css_final_slot {
+                            // subsequent CSS-final block found in the same thread as cur_slot
```
gterzian, line 311: `cur_slot` should be renamed, or clearly commented, as `slot_s`, or something similar, to match the spec(or the spec should use different wording).

massa-execution/src/worker.rs
```diff
@@ -255,106 +253,137 @@ impl ExecutionWorker {
...
-        let mut max_thread_slot = vec![self.last_final_slot; self.cfg.thread_count as usize];
-        for (_b_id, block) in css_final_blocks.iter() {
-            max_thread_slot[block.header.content.slot.thread as usize] = std::cmp::max(
-                max_thread_slot[block.header.content.slot.thread as usize],
-                block.header.content.slot,
-            );
-        }
+                    // there is no CSS-final block at cur_slot, but there are CSS-final blocks later
+                    Some(_b_slot) => {
+                        // check whether there is a CSS-final block later in the same thread
```
gterzian, line 302: Can we encapsulate this logic in a function/method that would return an option, and break on none?

massa-execution/src/worker.rs
```diff
@@ -255,106 +253,137 @@ impl ExecutionWorker {
...
                 }
+
+                cur_slot = cur_slot.get_next_slot(self.cfg.thread_count)?;
             }
         }
 
-        // list remaining CSS finals + new blockclique
-        self.ordered_active_blocks = self
-            .ordered_pending_css_final_blocks
+        // process SCE-active blocks
```
gterzian, line 334: This, or the spec, should be updated so that they match, I believe the spec currently reads `blocks that remain to be processed are in blockclique_blocks UNION pending_css_final_blocks`

massa-execution/src/worker.rs
```diff
@@ -255,106 +253,137 @@ impl ExecutionWorker {
...
             )
             .collect();
+        if let Some(max_css_active_slot) = sce_active_blocks.last_key_value().map(|(s, _v)| *s) {
+            let mut cur_slot = self.last_final_slot.get_next_slot(self.cfg.thread_count)?;
+            while cur_slot <= max_css_active_slot {
+                match sce_active_blocks.first_key_value().map(|(s, _v)| *s) {
+                    // there is a CSS-active block at cur_slot
+                    Some(b_slot) if b_slot == cur_slot => {
+                        // remove the entry from sce_active_blocks (cannot panic, checked above)
+                        let (_b_slot, (b_id, block)) = sce_active_blocks.pop_first().unwrap();
```
gterzian, line 357: please use `expect`, and write something there instead of the comment above.

massa-execution/src/worker.rs
```diff
@@ -255,106 +253,137 @@ impl ExecutionWorker {
...
-                    .filter(|(_b_id, b)| b.header.content.slot > self.last_final_slot),
+                self.pending_css_final_blocks
+                    .iter()
+                    .map(|(k, (b_id, b))| (*k, (b_id, b))),
             )
             .collect();
+        if let Some(max_css_active_slot) = sce_active_blocks.last_key_value().map(|(s, _v)| *s) {
+            let mut cur_slot = self.last_final_slot.get_next_slot(self.cfg.thread_count)?;
+            while cur_slot <= max_css_active_slot {
+                match sce_active_blocks.first_key_value().map(|(s, _v)| *s) {
```
gterzian, line 353: Please assign `sce_active_blocks.first_key_value().map(|(s, _v)| *s)` to a named variable for clarity.

massa-execution/src/worker.rs
gterzian: Also, the spec seems to use `blockclique_blocks UNION pending_css_final_blocks` as an immutable concept, may need to be reworded to match the code.

massa-execution/src/worker.rs
```diff
@@ -255,106 +253,137 @@ impl ExecutionWorker {
...
 
-        // sort active blocks
-        self.ordered_active_blocks
-            .sort_unstable_by_key(|(_b_id, b)| b.header.content.slot);
+                    // there is no CSS-active block at cur_slot
+                    Some(_b_slot) => {
+                        // execute a miss as an SCE-active VM step
+                        self.push_request(ExecutionRequest::RunActiveStep(ExecutionStep {
+                            slot: cur_slot,
+                            block: None,
```
gterzian, line 372: could you not simply do this with one call to `push_request`, where `block` would be an option acquired earlier?

massa-execution/src/worker.rs
```diff
@@ -255,106 +253,137 @@ impl ExecutionWorker {
...
+                        }));
+                        // update cursor
+                        self.last_active_slot = cur_slot;
```
gterzian, line 375: It seems that you could do this once above.

gterzian: The spec could also be reworded to do this step first, since it happens in both cases.

massa-execution/src/worker.rs
```diff
@@ -255,106 +253,137 @@ impl ExecutionWorker {
...
             }
-            self.push_request(ExecutionRequest::RunActiveStep(ExecutionStep {
-                slot: self.last_active_slot,
-                block: Some((b_id, block)),
-            }));
         }
 
-        // apply misses until now()
+        // fill with misses until current slot
         self.fill_misses_until_now()?;
```
gterzian, line 387: I think this function could be specified independently, and referred to in the spec at this point, as well. Also the current description of `fill up the remaining slots until now with misses` does not seem to match all the details that go on inside `fill_misses_until_now`.

massa-execution/src/worker.rs
damip: done

massa-execution/src/worker.rs
damip: done

massa-execution/src/worker.rs
damip: done

massa-execution/src/worker.rs
damip: done

massa-execution/src/worker.rs
damip: done

massa-execution/src/worker.rs
damip: done

massa-execution/src/worker.rs
damip: done

massa-execution/src/worker.rs
damip: done

massa-execution/src/worker.rs
damip: done

massa-execution/src/worker.rs
damip: done

massa-execution/src/worker.rs
damip: done, done

massa-execution/src/worker.rs
damip: done

massa-execution/src/worker.rs
damip: I didn't do it for 3 reasons:
* it makes the spec clearer
* it better matches the spec
* there is a third "None => break," case that would make the split heavier

Basically it has no impact on performance and makes the code more explicit. Duplication is very minimal.

massa-execution/src/worker.rs
damip: I updated the spec to make things match, but same comment as above: I prefer making things explicit than factoring them here

massa-execution/src/worker.rs
damip: * I made the method description more explicit and linked it to the spec
* I made the function itself look more like the spec and added comments
* in the spec, this function now has its own "step" that can be referred to

massa-execution/src/worker.rs
gterzian: Ok

massa-execution/src/worker.rs
gterzian: Ok

massa-execution/src/worker.rs
gterzian: Great, thanks

PR #2032, opened 2021-12-21
Automatically update submodule on build

adrien-zinger:
Because we need to be ultra transparent with submodules!
Add a build.rs that automatically update submodules on build


yvan-sraka (approved):

yvan-sraka:
Bors merge

PR #2025, opened 2021-12-21
Integrate trait-based runtime interface

gterzian:
Close #2021

PR #2033, opened 2021-12-21
[Execution] Output event system

### LINKED ISSUE #2005: Output event system
## Rationale

Allow smart contracts to emit arbitrary messages during execution, that can be seen by outside observers.

A bit like this: http://www2.phys.canterbury.ac.nz/dept/docs/manuals/unix/DEC_5.0a_Docs/HTML/ARH9VBTE/VMCHPXXX.HTM

## output_event syscall to output an event from within a smart contract

```
output_event("json string");
```

* fails if the string is longer than a config-defined max
* costs constant gas
* does not return anything
* is also allowed in readonly execution contexts

When this syscall is called, the following element is added to an `call_output_events: Vec<SCOutputEvent>` that lists all the events generated by the execution of the smart contract, in the order in which they were emitted:

```rust
struct SCOutputEvent {
    "context": Execution context when the call happened: slot, address, block, call stack etc...,
    "data": the emitted JSON string
}
```

## API access

The API must allow listing all the events that happened on the execution of a given ExecuteSC operation in a block, or during a given slot range (up to a config-defined maximum history length), with an optional filter according to the emitter smart contract address.

When running a readonly call, output events are listed as well and sent among the API output

AureliaDolo:
Resolve #2005


AureliaDolo:
If anyone has better name ideas

AureliaDolo:
Add generate event to sc runtime interface

AureliaDolo:
I did mostly the api part for now. I was thinking of merging this and finishing the work on execution and interface part once #2025 is merged into #1979

adrien-zinger (approved):
[this issue](https://github.com/massalabs/massa/issues/2035) in massa and [this issue](https://github.com/massalabs/sc-runtime/issues/40) in the runtime are the next steps.

We can merge this part in the feature branch and continue forward the development with [the new interface](https://github.com/massalabs/massa/pull/2025).

massa-models/src/output_event.rs
```diff
@@ -0,0 +1,44 @@
...
+    data: String,
+}
+
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct ExecutionContext {
+    slot: Slot,
+    bytes_address: Address,
+    caller_address: Address,
+    block: Option<BlockId>,
+    // todo I don't know how to represent the call stack
```
gterzian, line 19: I think this has already been done [here](https://github.com/massalabs/massa/blob/399fb6e02a977903d3e085084978c671b19403ba/massa-execution/src/types.rs#L31)? cc @damip

damip: yes indeed

AureliaDolo:
@yvan-sraka please don't merge anything in my branch

yvan-sraka:
Right! Do not hesitate to dismiss my commit :)

massa-execution/src/interface_impl.rs
```diff
@@ -155,4 +158,22 @@ impl Interface for InterfaceImpl {
             .map(|addr| addr.to_bs58_check())
             .collect())
     }
+
+    fn generate_event(&self, data: String) -> Result<()> {
+        let slot = context_guard!(self).slot;
```
damip, line 163: let's not use context_guard! at every line. This causes CPU-wide sync every time and is very heavy. Better lock once at the beginning of the function (see how it was implemented in other methods of InterfaceImpl)

massa-execution/src/interface_impl.rs
```diff
@@ -155,4 +158,22 @@ impl Interface for InterfaceImpl {
...
+        let bytes_address = *context_guard!(self).call_stack.back().unwrap(); // TODO what's an empty stack ?
```
damip, line 164: for now the stack should always be non-empty, but to be future-proof and robust, we did a clean fail in the other methods of InterfaceImpl. Let's do the same here

massa-execution/src/interface_impl.rs
```diff
@@ -155,4 +158,22 @@ impl Interface for InterfaceImpl {
...
+        let caller_address = *context_guard!(self).call_stack.front().unwrap();
+        let block = context_guard!(self).opt_block_id;
+        let call_stack = context_guard!(self).call_stack.clone();
+        let context = EventExecutionContext {
+            slot,
+            bytes_address,
```
damip, line 170: Let's remove the entries "bytes_address" and "caller_address" from EventExecutionContext. It's redundant with call_stack

damip (changes requested):

yvan-sraka (approved):
LGTM

damip (approved):

AureliaDolo:
bords merge

AureliaDolo:
bors merge

damip:
bors merge

damip:
bors merge

damip:
bors merge

PR #2036, opened 2021-12-22
Improve vm execution handling

damip:
Implement the VM execution handling so that it follows https://github.com/massalabs/massa/wiki/vm_ledger_interaction

Note that I had to remove the OperationSC because it didn't readily allow implementing some failure cases (for example still paying the coins to the sender on bytrcode parse failure) and workarounds made OperationSC not very useful.

Note that I also refactored address::from_public_key so that it does not return an error (it doesn't generate any errors anyway, it was a useless complication) because we don't want execution to fail in such a way that we don't give the op.coins to the sender

### DIFF
--- a/massa-execution/src/vm.rs
+++ b/massa-execution/src/vm.rs
@@ -139,6 +139,12 @@ impl VM {
     ) -> SCELedgerChanges {
         let mut context = self.execution_context.lock().unwrap();
 
+        // credit the sender with "coins"
+        let _result =
+            context
+                .ledger_step
+                .set_balance_delta(operation.sender, operation.coins, true);
+
         // credit the block creator with max_gas*gas_price
         let _result = context.ledger_step.set_balance_delta(
             block_creator_addr,
@@ -148,12 +154,6 @@ impl VM {
                 .unwrap(),
             true,
         );
-
-        // credit the sender with "coins"
-        let _result =
-            context
-                .ledger_step
-                .set_balance_delta(operation.sender, operation.coins, true);
 
         // fill context for execution
         context.gas_price = operation.gas_price;
--- a/massa-execution/src/vm.rs
+++ b/massa-execution/src/vm.rs
@@ -127,6 +127,7 @@ impl VM {
     }
 
     /// Prepare (update) the shared context before the new operation
+    /// returns a snapshot copy of the current caused ledger changes
     /// TODO: do not ignore the results
     /// TODO consider dispatching with edorsers/endorsed as well
     fn prepare_context(
@@ -137,11 +138,7 @@ impl VM {
         slot: Slot,
     ) -> SCELedgerChanges {
         let mut context = self.execution_context.lock().unwrap();
-        // credit the sender with "coins"
-        let _result =
-            context
-                .ledger_step
-                .set_balance_delta(operation.sender, operation.coins, true);
+
         // credit the block creator with max_gas*gas_price
         let _result = context.ledger_step.set_balance_delta(
             block_creator_addr,
@@ -151,8 +148,12 @@ impl VM {
                 .unwrap(),
             true,
         );
-        // Save the Initial ledger changes before execution
-        // It contains a copy of the initial coin credits that will be popped back if bytecode execution fails in order to cancel its effects
+
+        // credit the sender with "coins"
+        let _result =
+            context
+                .ledger_step
+                .set_balance_delta(operation.sender, operation.coins, true);
 
         // fill context for execution
         context.gas_price = operation.gas_price;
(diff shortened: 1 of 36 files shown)


massa-execution/src/vm.rs
```diff
@@ -162,7 +163,7 @@ impl VM {
         context.opt_block_id = Some(block_id);
         context.opt_block_creator_addr = Some(block_creator_addr);
         context.call_stack = vec![operation.sender].into();
-        context.ledger_step.caused_changes.clone()
+        context.ledger_step.caused_changes.clone();
```
adrien-zinger, line 166: it's not a return value anymore?

damip: my bad
corrected

damip:
I will finish specifying syscalls in a followup

massa-execution/src/vm.rs
```diff
@@ -127,6 +127,7 @@ impl VM {
     }
 
     /// Prepare (update) the shared context before the new operation
+    /// returns a snapshot copy of the current caused ledger changes
```
gterzian, line 138: nit: If this is a separate sentence, it would be best to use an uppercase, and have a dot at the end of the line above.

massa-execution/src/vm.rs
```diff
@@ -151,8 +148,12 @@ impl VM {
                 .unwrap(),
             true,
         );
-        // Save the Initial ledger changes before execution
-        // It contains a copy of the initial coin credits that will be popped back if bytecode execution fails in order to cancel its effects
+
+        // credit the sender with "coins"
```
gterzian: nit: I'd rather have upper case at the start of each sentence.

massa-execution/src/vm.rs
```diff
@@ -194,10 +195,14 @@ impl VM {
             for (op_idx, operation) in block.operations.clone().into_iter().enumerate() {
                 let operation_sc = OperationSC::try_from(operation.content);
                 if operation_sc.is_err() {
-                    // only fail if the operation cannot parse the sender address
+                    // fails if the operation cannot parse the sender address or if the bytecode is invalid
```
gterzian: same

massa-execution/src/vm.rs
```diff
@@ -194,10 +195,14 @@ impl VM {
...
                     continue;
                 }
                 let operation = &operation_sc.unwrap();
+
+                // Prepare context and save the Initial ledger changes before execution
```
gterzian, line 216: no need for the upper-case `Initial`, unless you're referring to some construct in the code...

massa-execution/src/vm.rs
```diff
@@ -194,10 +195,14 @@ impl VM {
...
+                // The returned snapshot contains a copy of the initial coin credits
+                // that will be popped back if bytecode execution fails in order to cancel its effects
```
gterzian: I don't think you need to add `in order to cancel its effects`. That makes it read like popping the coins back will have some other side-effect(does it?).

massa-execution/src/vm.rs
```diff
@@ -127,6 +127,7 @@ impl VM {
     }
 
     /// Prepare (update) the shared context before the new operation
+    /// returns a snapshot copy of the current caused ledger changes
     /// TODO: do not ignore the results
     /// TODO consider dispatching with edorsers/endorsed as well
```
gterzian, line 140: typo

massa-execution/src/vm.rs
```diff
@@ -139,6 +139,12 @@ impl VM {
     ) -> SCELedgerChanges {
         let mut context = self.execution_context.lock().unwrap();
 
+        // credit the sender with "coins"
```
gterzian, line 153: It would be great if we could refer/link to a spec.

gterzian (changes requested):
LGMT except some nits

massa-execution/src/vm.rs
damip: done

massa-execution/src/vm.rs
damip: done

massa-execution/src/vm.rs
damip: this is gone

massa-execution/src/vm.rs
damip: done

massa-execution/src/vm.rs
damip: done

massa-execution/src/vm.rs
damip: done

massa-execution/src/vm.rs
damip: done

damip:
Greg said LGTM. Merging

PR #2040, opened 2021-12-24
Read only execution 

### LINKED ISSUE #1977: Readonly Smart Contract Calls
### Discussed in https://github.com/massalabs/massa/discussions/1970

<div type='discussions-op-text'>

<sup>Originally posted by **damip** December  9, 2021</sup>
# Rationale

When using Massa in a Web3 context for example, one should be able to perform read-only Smart Contract calls.

See: https://ethereum.stackexchange.com/questions/765/what-is-the-difference-between-a-transaction-and-a-call/770

# Massa implementation

## API

Add a "sc_readonly_call" API endpoint 

Parameters:
* max_gas: u64  // max gas allowed for the readonly run
* bytecode: `Vec<u8>`  // bytecode to execute

Return value:
* executed_at: Slot  // slot at which the execution happened
* result:
  * (optional) error: Error 
  * (optional) return_value : Vec<u8>  // return value of the bytecode's main() function (TODO this needs to be discussed)
  * (optional) output_events: `Vec<SCOutputEvent>`  // output events generated during execution

 ## Operation

* when the sc_readonly_call is called, the bytecode's main() function will be called with the following execution context:
  * the execution will be done from the point of view of the current slot (even if there is currently no block there)
  * empty initial call stack (anonymous call)

* during the call, everything is allowed as with usual SC calls, except that any syscall attempting to WRITE anything to the SCE ledger will fail. Async message sending also fails when called from within a readonly context.

</div>

gterzian:
Close #1977


damip:
@gterzian  updated spec: https://github.com/massalabs/massa/discussions/1970

gterzian:
@adrien-zinger @damip Ready for review, I'll do the API part in a separate PR.

massa-execution/src/vm.rs
```diff
@@ -72,15 +75,15 @@ impl VM {
...
         // check if that step was already executed as the earliest active step
-        let history_item = if let Some(cached) = self.pop_cached_step(step) {
+        let history_item = if let Some(cached) = self.pop_cached_step(&step) {
             // if so, pop it
             cached
         } else {
             // otherwise, clear step history an run it again explicitly
             self.step_history.clear();
-            self.run_step_internal(step)
+            self.run_step_internal(&step)
```
adrien-zinger, line 86: I guess that you can reset these lines and just borrow the step

massa-execution/src/vm.rs
```diff
@@ -174,6 +177,63 @@ impl VM {
...
+
+            // Set the max gas.
+            context.max_gas = max_gas;
+
+            // Set the simulated gas price.
+            context.gas_price = simulated_gas_price;
+
+            // Return a clone of the current state, to revert to it after execution.
+            context.ledger_step.caused_changes.clone()
+        };
```
adrien-zinger, line 209: It looks like something more generic, can we find a way to use the prepare_context or extract a function?

massa-models/src/execution.rs
```diff
@@ -0,0 +1,22 @@
+use crate::Slot;
```
adrien-zinger, line 1: execution_*types*.rs?

massa-execution/src/vm.rs
```diff
@@ -174,6 +177,63 @@ impl VM {
...
+            .caused_changes = saved_changes;
+
+        // Send result back.
+        let execution_response = ExecuteReadOnlyResponse {
+            executed_at: slot,
+            // TODO: specify result.
+            result: run_result.map_or_else(
+                |_| ReadOnlyResult::Error("Failed to run in read-only mode".to_string()),
+                |_| ReadOnlyResult::Ok,
+            ),
```
adrien-zinger, line 228: We can return all the response or at least the `run_result.res`, what do you think?

adrien-zinger (commented):
It looks good enough for me. I just make some comments that can be fixed, ignored or turned into issues!

massa-execution/src/vm.rs
gterzian: I'm not sure, do we actually want to return anything other than the output events? @damip Looking at the latest version at https://github.com/massalabs/massa/discussions/1970#discussioncomment-1879779 it seems we don't.

massa-execution/src/vm.rs
gterzian: I removed the borrow because it required borrowing in the run loop, which is weird because you actually want to take ownership of the values(for example in the new request to execute read-only stuff).

massa-models/src/execution.rs
gterzian: I don't think so, since it's arleady in `models` it means it will be some sort of "types".

massa-execution/src/vm.rs
gterzian: I don't think its' worth it in this PR. If you see a clear alternative feels free to open an issue/pr and propose the change.

gterzian:
I will still integrate the spec in this PR(or do this in a follow-up if you want to merge this already today).

gterzian:
Actually let's merge this first(after approval from reviewers), since it makes it easier to link to the md file after it's been already checked in...

massa-execution/src/vm.rs
```diff
@@ -174,6 +177,63 @@ impl VM {
...
+        simulated_gas_price: Amount,
+        bytecode: Vec<u8>,
+        address: Option<Address>,
+        result_sender: oneshot::Sender<ExecuteReadOnlyResponse>,
+    ) {
+        let saved_changes = {
+            let mut context = self.execution_context.lock().unwrap();
+
+            // Set the call stack, using the provided address, or a random one.
+            let address = address.unwrap_or_else(|| {
```
damip, line 194: the specification should be updated to reflect this random-address-by-default behavior

massa-execution/src/vm.rs
```diff
@@ -174,6 +177,63 @@ impl VM {
...
+
+            // Return a clone of the current state, to revert to it after execution.
+            context.ledger_step.caused_changes.clone()
+        };
+
+        // run in the intepreter
+        let run_result = assembly_simulator::run(&bytecode, max_gas, &*self.execution_interface);
+
+        // cancel the effects of execution only, pop back init_changes
+        self.execution_context
```
damip, line 215: is not necessary to save/pop the context at all, since we  don't apply its results in the ledger.

Just reset it at the beginning, run the stuff, and forget it.

massa-execution/src/vm.rs
```diff
@@ -174,6 +177,63 @@ impl VM {
...
+        &self,
+        slot: Slot,
+        max_gas: u64,
+        simulated_gas_price: Amount,
+        bytecode: Vec<u8>,
+        address: Option<Address>,
+        result_sender: oneshot::Sender<ExecuteReadOnlyResponse>,
+    ) {
+        let saved_changes = {
+            let mut context = self.execution_context.lock().unwrap();
```
damip, line 191: use self.clear_and_update_context(); (see the run_step_internal function) to initialize the context, otherwise it will simply be inherited from the last call. This way the rest of the context will also be reset to a default state here, to ensure that if we add fields to the context, there is no persistence between calls in case we forget to port them here

damip (changes requested):

massa-execution/src/vm.rs
damip: we don't need to save/ pop changes at all actually :) just reset the context, run and that's it (just don't apply local_changes to the ledger at the end). The save/pop is there in run_step_internal because we need to keep some changes but not others, and we always apply at least some changes to the final ledger. Here it's not the case, so we can simplify :)

gterzian:
@damip @adrien-zinger Ok, thanks for the review and I've addressed the comments, please take another look and merge if ok. I'm adding the API in another branch.

massa-execution/src/vm.rs
gterzian: done

massa-execution/src/vm.rs
gterzian: Ok thanks, done

massa-execution/src/vm.rs
gterzian: Ok thanks, done.

damip (approved):
OK for me

PR #2048, opened 2021-12-29
For cautious bassoon: update runtime version and impl interface

adrien-zinger:
In this PR I update the latest runtime version and fix some clippy warns in execution module.

I'll manage in a next commit to implement the interface. Nevertheless it's ready for a first review.

:notes: Note for reviewers:
To fix a clippy warning I've created a new structure in types.rs, feel free to suggest a different name more appropriated!
https://github.com/massalabs/massa/blob/5a008ce67adcf54c8807c82a967c44f5a9c84972/massa-execution/src/types.rs#L95-L128

### DIFF
--- a/massa-execution/src/config.rs
+++ b/massa-execution/src/config.rs
@@ -1,3 +1,4 @@
+use massa_time::MassaTime;
 use std::path::PathBuf;
 
 use serde::{Deserialize, Serialize};
@@ -10,4 +11,12 @@ pub const CHANNEL_SIZE: usize = 256;
 pub struct ExecutionSettings {
     /// Initial SCE ledger file
     pub initial_sce_ledger_path: PathBuf,
+    /// Thread count
+    pub thread_count: u8,
+    /// Genesis timestmap
+    pub genesis_timestamp: MassaTime,
+    /// period duration
+    pub t0: MassaTime,
+    /// clock compensation in milliseconds
+    pub clock_compensation: i64,
 }
--- a/massa-execution/src/config.rs
+++ b/massa-execution/src/config.rs
@@ -6,11 +6,17 @@ use serde::{Deserialize, Serialize};
 /// Max size for channels used with communication with other components.
 pub const CHANNEL_SIZE: usize = 256;
 
-/// Execution configuration
+/// Execution setting parsed with .toml in `massa-node/src/settings.rs`
 #[derive(Debug, Deserialize, Serialize, Clone)]
 pub struct ExecutionSettings {
     /// Initial SCE ledger file
     pub initial_sce_ledger_path: PathBuf,
+}
+
+#[derive(Debug, Deserialize, Serialize, Clone)]
+pub struct ExecutionConfigs {
+    /// Execution settings
+    pub settings: ExecutionSettings,
     /// Thread count
     pub thread_count: u8,
     /// Genesis timestmap
--- a/massa-execution/src/interface_impl.rs
+++ b/massa-execution/src/interface_impl.rs
@@ -4,9 +4,18 @@ use std::str::FromStr;
 
 use crate::types::ExecutionContext;
 use anyhow::{bail, Result};
-use assembly_simulator::{Interface, InterfaceClone};
+use assembly_simulator::{Bytecode, Interface, InterfaceClone};
 use massa_models::Address;
 use std::sync::{Arc, Mutex};
+
+macro_rules! context_guard {
+    ($self:ident) => {
+        $self
+            .context
+            .lock()
+            .expect("Failed to acquire lock on context.")
+    };
+}
 
 #[derive(Clone)]
 pub(crate) struct InterfaceImpl {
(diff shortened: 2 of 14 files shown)


massa-execution/src/config.rs
```diff
@@ -10,4 +11,12 @@ pub const CHANNEL_SIZE: usize = 256;
 pub struct ExecutionSettings {
     /// Initial SCE ledger file
     pub initial_sce_ledger_path: PathBuf,
+    /// Thread count
+    pub thread_count: u8,
```
damip, line 15: does this mean that the `execution` section of the config file needs to contain a duplicate of parameters like thread_count ?

damip: how is clock_compensation defined ? it should be dynamically computed at every bootstrap and passed

damip (approved):
some config nits, the rest is a welcome improvement !

InterfaceImpl did not change tho ?

massa-execution/src/interface_impl.rs
```diff
@@ -29,15 +38,73 @@ impl Interface for InterfaceImpl {
...
+        let (block_id, index) = {
+            let context = context_guard!(self);
+            (context.opt_block_id, context.call_stack.len())
+        };
+        let (block_id, index) = match (block_id, u8::try_from(index)) {
+            (Some(block_id), Ok(index)) => (block_id, index),
+            _ => bail!("Failed to read current context"),
+        };
+        let mut data = block_id.to_bytes().to_vec();
+        data.push(index);
```
damip, line 66: index is an u8 ? that may overflow. It should be u64 (or at least u32), use u64::to_be_bytes() for the byte conversion to remain deterministic

massa-execution/src/interface_impl.rs
```diff
@@ -29,15 +38,73 @@ impl Interface for InterfaceImpl {
...
+        let address = Address(massa_hash::hash::Hash::from(&data));
```
damip, line 67: I find hash::from disturbingly misleading in what it does :)

massa-execution/src/interface_impl.rs
```diff
@@ -29,15 +38,73 @@ impl Interface for InterfaceImpl {
...
+    fn set_data_for(
+        &self,
+        address: &assembly_simulator::Address,
+        key: &str,
+        value: &Bytecode,
+    ) -> Result<()> {
+        let mut context = context_guard!(self);
+        let addr = Address::from_str(address)?;
+        let key = massa_hash::hash::Hash::from_bs58_check(key)?;
+        if context.call_stack.contains(&addr) {
```
damip, line 97: access should be granted only if `addr == context.call_stack.back()` (you don't want something you call -possibly very indirectly- to write-access your own data) or if `context.created_addresses.contains(&addr)` (you can write-access SCs that you created during the same execution, useful for initialization)

massa-execution/src/interface_impl.rs
```diff
@@ -29,15 +38,73 @@ impl Interface for InterfaceImpl {
...
+        context.ledger_step.set_module(address, module.clone());
+        Ok(address.to_string())
+    }
+
+    /// Requires the data at the address
+    fn get_data_for(&self, address: &assembly_simulator::Address, key: &str) -> Result<Bytecode> {
+        let context = context_guard!(self);
+        let addr = &Address::from_str(address)?;
+        let key = massa_hash::hash::Hash::from_bs58_check(key)?;
+        return match context.ledger_step.get_data_entry(addr, &key) {
```
damip, line 78: "return" keyword is not necesary

damip (changes requested):

massa-execution/src/interface_impl.rs
damip: but that's another issue

massa-execution/src/interface_impl.rs
```diff
@@ -29,15 +38,73 @@ impl Interface for InterfaceImpl {
...
+    /// Requires a new address that contains the sent bytecode.
+    ///
+    /// Generate a new address with a concatenation of the firsts N bytes of the blockId and the firsts N bytes of the block creator address.
+    /// Where N = ADDRESS_SIZE_BYTES / 2
+    ///
+    /// Insert in the ledger the given bytecode in the generated address
+    fn create_module(&self, module: &Bytecode) -> Result<assembly_simulator::Address> {
+        let (block_id, index) = {
+            let context = context_guard!(self);
+            (context.opt_block_id, context.call_stack.len())
```
damip, line 59: index should not be the length of the call stack, but the index of the ledger entry creation within the block

massa-execution/src/config.rs
damip: solved

massa-execution/src/interface_impl.rs
adrien-zinger: modified

massa-execution/src/interface_impl.rs
adrien-zinger: modified

massa-execution/src/interface_impl.rs
adrien-zinger: done!

massa-execution/src/interface_impl.rs
adrien-zinger: You can chck the new version. The next index is the number of owned addresses.
Implemented like in #2051

massa-execution/src/interface_impl.rs
```diff
@@ -27,17 +36,117 @@ impl InterfaceClone for InterfaceImpl {
...
-                .lock()
-                .expect("Failed to acquire lock on context.");
-            context.ledger_step.get_module(&address)
-        };
+        let address = Address::from_str(address)?;
+        let bytecode = { context_guard!(self).ledger_step.get_module(&address) };
         match bytecode {
-            Some(bytecode) => Ok(bytecode),
+            Some(bytecode) => {
+                context_guard!(self).call_stack.push_back(address);
```
damip, line 43: why are we pushing the address on the call stack here ? we just want to read the module from an address

massa-execution/src/interface_impl.rs
```diff
@@ -27,17 +36,117 @@ impl InterfaceClone for InterfaceImpl {
...
+        let (block_id, index) = {
+            let context = context_guard!(self);
+            (context.opt_block_id, context.owned_addresses.len())
+        };
+        let block_id = match block_id {
+            Some(block_id) => block_id,
+            _ => bail!("Failed to read current context"),
+        };
+        let mut data = block_id.to_bytes().to_vec();
+        data.append(&mut index.to_be_bytes().to_vec());
```
damip, line 72: non-deterministic !

owned_addresses.len() returns an usize which is u64 on 64bit systems, 32 on 32bit systems etc... let's force conversion to u64 beforehand

massa-execution/src/interface_impl.rs
adrien-zinger: This is a naming issue. When we get a module with the interface it's necessary to execute it

massa-execution/src/interface_impl.rs
```diff
@@ -27,17 +36,117 @@ impl InterfaceClone for InterfaceImpl {
...
+        let address = Address(massa_hash::hash::Hash::from(&data));
+        let res = address.to_bs58_check();
+        let mut context = context_guard!(self);
```
damip, line 75: locking & unlocking is heavy (CPU sync instructions), maybe use only one context guard for the whole function

massa-execution/src/interface_impl.rs
```diff
@@ -27,17 +36,117 @@ impl InterfaceClone for InterfaceImpl {
...
+        context.ledger_step.set_module(address, module.clone());
+        context.owned_addresses.push_back(address);
```
damip, line 77: I think owned_addresses should be an AddressHashSet for fast lookup on write access rights check

massa-execution/src/interface_impl.rs
```diff
@@ -27,17 +36,117 @@ impl InterfaceClone for InterfaceImpl {
...
+    fn set_data_for(
+        &self,
+        address: &assembly_simulator::Address,
+        key: &str,
+        value: &Bytecode,
+    ) -> Result<()> {
+        let addr = Address::from_str(address)?;
+        let key = massa_hash::hash::Hash::from_bs58_check(key)?;
+        let mut context = context_guard!(self);
+        if context.owned_addresses.contains(&addr) {
```
damip, line 106: `if(context.owned_addresses.contains(&addr) || context.call_stack.back() == Some(addr)) `

massa-execution/src/interface_impl.rs
```diff
@@ -27,17 +36,117 @@ impl InterfaceClone for InterfaceImpl {
...
+        let key = massa_hash::hash::Hash::from_bs58_check(key)?;
+        match context.ledger_step.get_data_entry(addr, &key) {
+            Some(bytecode) => Ok(bytecode),
+            _ => bail!("Data entry not found"),
+        }
+    }
+
+    fn set_data(&self, key: &str, value: &Bytecode) -> Result<()> {
+        let mut context = context_guard!(self);
+        let addr = match context.call_stack.front() {
```
damip, line 127: ontext.call_stack.back()

damip:
context.owned_addresses should be cleared at the beginning of an exec (also at the beginning of readonly execs), otherwise it will only accumulate forever

massa-execution/src/interface_impl.rs
damip: good idea !

massa-execution/src/interface_impl.rs
damip: uuh

adrien-zinger:
> context.owned_addresses should be cleared at the beginning of an exec (also at the beginning of readonly execs), otherwise it will only accumulate forever

Oups! I forgot this obvious little details 🤣

massa-execution/src/interface_impl.rs
```diff
@@ -27,17 +36,117 @@ impl InterfaceClone for InterfaceImpl {
...
+    ///
+    /// Generate a new address with a concatenation of the block_id hash and the index of address owned in context.
+    ///
+    /// Insert in the ledger the given bytecode in the generated address
+    fn create_module(&self, module: &Bytecode) -> Result<assembly_simulator::Address> {
+        let (block_id, index) = {
+            let context = context_guard!(self);
+            (context.opt_block_id, context.owned_addresses.len())
+        };
+        let block_id = match block_id {
```
damip, line 67: instead of using `[block_id, index]`, let's use `['B', context.slot.to_bytes_key(), index]` in a block context and `['A', context.slot.to_bytes_key(), index]` in an async context.

So here it would be `['B', context.slot.to_bytes_key(), index]`

That way:
* no ID collisions with async executions
* no crash in readonly contexts

massa-execution/src/interface_impl.rs
adrien-zinger: get_module should be get_module_for_execution

damip (changes requested):
Fundamental problem:

ownled_addresses is cleared at each execution.

So if there are multiple ExecuteSC in a single block, here will be ID collisions in generated addresses between them 

See comment https://github.com/massalabs/massa/pull/2048#pullrequestreview-842006673

damip:
I think the "call" abi is missing (?)

massa-execution/src/interface_impl.rs
damip: hmmm... let's discuss that :)

adrien-zinger:
> I think the "call" abi is missing (?)

There is no implementation of ABI in the massa side. The execution-module is fully agnostic and the ABIs are in sc-runtime. Maybe we can also talk on this discussion: https://github.com/massalabs/massa/discussions/1984 after an update of the current implementation in `sc-runtime(v0.2.3)`

massa-execution/src/interface_impl.rs
adrien-zinger: @damip, I'm not sure to understand what do you mean with the slot's ID.  Maybe the new version will be correct enough for you, I tried to do my best to avoid collisions! Feel free to modify this part of the code if you think it's still incorrect.

massa-execution/src/interface_impl.rs
adrien-zinger: fixed

massa-execution/src/interface_impl.rs
adrien-zinger: fixed

massa-execution/src/interface_impl.rs
adrien-zinger: changed!

massa-execution/src/interface_impl.rs
adrien-zinger: changed!

massa-execution/src/interface_impl.rs
adrien-zinger: FIxed, but linked with another thread ;) https://github.com/massalabs/massa/pull/2048#discussion_r776802683

massa-execution/src/interface_impl.rs
adrien-zinger: Can we close the thread?

damip: ok

damip:
Various problems corrected

damip (approved):

damip:
bors merge

damip:
Bors merge

PR #2057, opened 2022-01-01
resolve readony addres creation ambiguity

damip:
If there are address creations at slot B's execution, and then a readonly call at that same slot that also creates addresses, the new creations will overwrite the ones created at the slot. This PR removes this ambiguity.


yvan-sraka (approved):

yvan-sraka:
bors merge

damip:
bors merge

PR #2061, opened 2022-01-03
Renamed `Hash::from` into `Hash::compute_from`

### LINKED ISSUE #2002: MassaHash::from() is ambiguous
Hash::from in massa_hash does not create a MassaHash from bytes as one would expect (like from_bytes) but builds the MassaHash by computing the sha256 of the provided bytes.

This behavior is counter-intuitive, or at least ambiguous and can cause silent errors that are hard to detect .

We need to force a more explicit usage by forbidding "from" and adding a "hash_bytes" method for example

AureliaDolo:
Fix #2002


yvan-sraka (approved):
LGTM

yvan-sraka (approved):

yvan-sraka:
bors merge

AureliaDolo:
bors merge

PR #2082, opened 2022-01-06
Repair exec loop

damip:
On multiple condvar wakeup signals in the execution loop, only one was processed. Now all are processed. This repairs clean shutdown and improves the processing of batches of blocks.


AureliaDolo (approved):

AureliaDolo:
bors merge

PR #2083, opened 2022-01-06
log generated execution events

damip:
log generated execution events for debugging purposes


AureliaDolo (commented):
Won't that be lost in the other logs ?

damip:
> Won't that be lost in the other logs ?

there are very few "info" logs, only startup, block creation, slot selection

AureliaDolo (approved):

AureliaDolo:
I unlinked the issues as merging this would have solved them and this is just a quick fix

damip:
bors merge

PR #2086, opened 2022-01-06
Add get sce ledger entry command.

### LINKED ISSUE #2072: Missing sce legder data in api

AureliaDolo:
Fix #2072

### DIFF
--- a/massa-execution/src/vm.rs
+++ b/massa-execution/src/vm.rs
@@ -3,6 +3,7 @@ use crate::interface_impl::InterfaceImpl;
 use crate::sce_ledger::{FinalLedger, SCELedger, SCELedgerChanges};
 use crate::types::{ExecutionContext, ExecutionData, ExecutionStep, StepHistory, StepHistoryItem};
 use crate::{config::ExecutionConfigs, ExecutionError};
+use massa_models::api::SCELedgerInfo;
 use massa_models::prehash::Map;
 use massa_models::AMOUNT_ZERO;
 use massa_sc_runtime::Interface;
@@ -78,6 +79,34 @@ impl VM {
             .ledger_step
             .final_ledger_slot
             .clone()
+    }
+
+    // clone bootstrap state (final ledger and slot)
+    pub fn get_sce_ledger_entry_for_addresses(
+        &self,
+        addresses: Vec<Address>,
+    ) -> Map<Address, SCELedgerInfo> {
+        let ledger = &self
+            .execution_context
+            .lock()
+            .unwrap()
+            .ledger_step
+            .final_ledger_slot
+            .ledger;
+        addresses
+            .into_iter()
+            .map(|ad| {
+                let entry = ledger.0.get(&ad).cloned().unwrap_or_default();
+                (
+                    ad,
+                    SCELedgerInfo {
+                        balance: entry.balance,
+                        module: entry.opt_module,
+                        datastore: entry.data,
+                    },
+                )
+            })
+            .collect()
     }
 
     /// runs an SCE-final execution step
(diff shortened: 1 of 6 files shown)


AureliaDolo:
There is an issue with the get address api command: if an address is not in the consensus ledger, it will be displayed as not found, even if it exist in the sce ledger. Should we make every consensus related field optional ? Or make a sce specific command ?

damip:
@AureliaDolo it should be an OR:
* both consensus and SCE are optional, and let's call them "sequencial balance" and "parallel balance" respectively, to be future-proof and not break the API too much when we change to a unique ledger with those 2 columns.
* the address is not found if it has 0 rolls and is absent from both consensus and SCE

AureliaDolo:
> @AureliaDolo it should be an OR:
> 
>     * both consensus and SCE are optional, and let's call them "sequencial balance" and "parallel balance" respectively, to be future-proof and not break the API too much when we change to a unique ledger with those 2 columns.
> 
>     * the address is not found if it has 0 rolls and is absent from both consensus and SCE

Ok so this means that this command need to be fully updated

AureliaDolo:
@adrien-zinger @gterzian I'm kinda lost on how to access the current ledger from  the execution worker

AureliaDolo:
see #2126 for the wiki update

massa-execution/src/worker.rs
```diff
@@ -287,9 +304,52 @@ impl ExecutionWorker {
...
+            } => todo!(),
+            ExecutionCommand::GetSCOutputEventBySCAddress {
+                sc_address,
+                response_tx,
+            } => todo!(),
+            ExecutionCommand::GetSCELedgerForAddresses {
+                response_tx,
+                addresses,
+            } => {
+                let ledger = &self
```
gterzian, line 324: Instead of sharing the execution context between the worker and context, you can use a `ExecutionRequest`, like was done in https://github.com/massalabs/massa/pull/2056/

gterzian (changes requested):
Looks good so far, and I think it can be done without sharing the execution context.

massa-execution/src/worker.rs
AureliaDolo: Thanks !

AureliaDolo: It's much cleaner that way :)

massa-execution/src/worker.rs
```diff
@@ -116,6 +122,7 @@ impl ExecutionWorker {
 
         // Init VM
         let mut vm = VM::new(cfg.clone(), bootstrap_ledger)?;
+        let execution_context = vm.execution_context.clone();
```
gterzian, line 125: this can now be removed.

gterzian (changes requested):
Looks good, with just one nit. Also there are some failing tests which would need looking into.

massa-execution/src/vm.rs
```diff
@@ -21,7 +22,7 @@ pub(crate) struct VM {
     thread_count: u8,
     step_history: StepHistory,
     execution_interface: Box<dyn Interface>,
-    execution_context: Arc<Mutex<ExecutionContext>>,
+    pub execution_context: Arc<Mutex<ExecutionContext>>,
```
gterzian, line 25: this also doesn't need to be public anymore.

gterzian (approved):
great, thanks!

gterzian:
@AureliaDolo maybe squash into a single commit as part of the necessary rebase?

AureliaDolo:
> @AureliaDolo maybe squash into a single commit as part of the necessary rebase?

Done ! :D

adrien-zinger (approved):

adrien-zinger (approved):

AureliaDolo:
bors merge

PR #2105, opened 2022-01-09
repair config override

damip:
config/config.toml was not loaded anymore to override the default config. This PR restores the documented behavior


AureliaDolo:
That was not how it's supposed to work !

PR #2109, opened 2022-01-09
Fix cd

qdrn:
This commit should fix the CD.

This a temporary solution for Mac OS until we find something more robust.


damip (approved):

qdrn:
Glad I could make Yvan happy :D

PR #2137, opened 2022-01-12
repair SCE ledger updates

damip:


massa-execution/src/sce_ledger.rs
```diff
@@ -32,7 +32,9 @@ impl SCELedgerEntry {
         }
 
         // module
-        self.opt_module = update.update_opt_module.clone();
+        if let Some(opt_module) = &update.update_opt_module {
+            self.opt_module = opt_module.clone();
+        }
```
adrien-zinger, line 37: Can just be 
```rust
        if let Some(module) = update.update_opt_module {
            self.opt_module = Some(module);
        }
```

without manage all the process an Option right? You never make a real use of the super Option

damip:
The compiler in the CI crashes  :/

adrien-zinger:
> The compiler in the CI crashes :/

Do you think that we should define a version of rust for this episode?

adrien-zinger (approved):

damip:
> > The compiler in the CI crashes :/
> 
> Do you think that we should define a version of rust for this episode?

good idea. Can you make a PR ?

AureliaDolo:
> > > The compiler in the CI crashes :/
> > 
> > 
> > Do you think that we should define a version of rust for this episode?
> 
> good idea. Can you make a PR ?

Isn't it a risk that we forget to update it ?

AureliaDolo:
bors merge

damip:
> > > > The compiler in the CI crashes :/
> > > 
> > > 
> > > Do you think that we should define a version of rust for this episode?
> > 
> > 
> > good idea. Can you make a PR ?
> 
> Isn't it a risk that we forget to update it ?

yes, that's why we need to make an issue to roll it back

PR #2139, opened 2022-01-12
Add initial peers separated from the peer file.

### LINKED ISSUE #1997: Add initial_peers.json and gitignore peers.json
Like it is done every where else

AureliaDolo:
Fix #1997


AureliaDolo:
I would really appreciate an extra look at the gitignore file :)

yvan-sraka (dismissed):
That's neat :sparkles:!

AureliaDolo:
bors merge

massa-network/src/peer_info_database.rs
```diff
@@ -230,7 +230,7 @@ impl PeerInfoDatabase {
 
         // load from file
         let mut peers = serde_json::from_str::<Vec<PeerInfo>>(
```
damip, line 232: there is a problem here.

What it does:
* load initial peers
* add bootstrap peers if any

What it should do:
* load initial peers
* try to append the storage peers file if it exists
* add bootstrap peers if any

damip (changes requested):

massa-network/src/peer_info_database.rs
AureliaDolo: I overwrite the content of peers.json on purpose. If the peers are ok they will be sent back anyway, and if they aren't there is no point keeping them.

damip: keeping them is extremely important. We cannot trust peers sent from others. We should prioritize peers that we have tested ourselves in the past (also maintain which peers we have banned).

Otherwise there is no point in having a non-initial peers file at all

yvan-sraka:
bors merge

AureliaDolo:
Note that I deleted peers.json and added it to the gitignore

massa-network/src/peer_info_database.rs
```diff
@@ -228,13 +228,21 @@ impl PeerInfoDatabase {
...
-            &tokio::fs::read_to_string(&cfg.peers_file).await?,
+            &tokio::fs::read_to_string(&cfg.initial_peers_file).await?,
         )?
         .into_iter()
         .map(|p| (p.ip, p))
         .collect::<HashMap<IpAddr, PeerInfo>>();
+        peers.extend(
+            // previously known peers
+            serde_json::from_str::<Vec<PeerInfo>>(
+                &tokio::fs::read_to_string(&cfg.peers_file).await?,
```
damip, line 241: The file might not exist (actually by default it doesn't !). Its existence needs to be checked beforehand (see my previous explanation)

damip (changes requested):

yvan-sraka (dismissed):

massa-network/src/peer_info_database.rs
AureliaDolo: Done !

yvan-sraka (approved):

damip (approved):
LGTM

damip:
bors merge

PR #2146, opened 2022-01-13
Check static config Hash on bootstrap.

### LINKED ISSUE #1992: Check hash of protocol related config at bootstrap
Instead of just checking the version we could hash all protocol related config and make sure the hashes match. Anyway, if this is not ok that will lead to desynchronization.

Since #1848 is merged that should be quite easy

AureliaDolo:
Fix #1992 

I still have to figure out how to compute that hash for real

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1397,6 +1397,7 @@ dependencies = [
  "massa_bootstrap",
  "massa_consensus",
  "massa_execution",
+ "massa_hash",
  "massa_logging",
  "massa_models",
  "massa_network",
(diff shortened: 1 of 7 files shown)


damip:
This is not necessary anymore given the new config system.
All the critical settings are now hardcoded in the source code, including the version number.
So checking the version number is enough.
Also if someone wants to cheat, they can always do it by sending a fake hash.

damip (changes requested):
to be closed

AureliaDolo:
This is useful in case someone altered their settings.rs and built the binaries by themselves. Preventing them from joining the main network would limit the number of nodes that will produce only invalid blocks

damip:
If people already altered the code, they will probably also comment out the verification system. Too much work and future maintenance (adding every new parameter to the verification process) for very marginal benefit. Closing

PR #2149, opened 2022-01-14
Execution: use assembly-simulator bytecode type

### LINKED ISSUE #2138: The value in SCE datastores is wrongly typed as Bytecode
The value in the datastore key->value mapping is not bytecode and should not be typed as such. It's just bytes (`vec<u8>`). It compiles because for now Bytecode happens to be mapped to `Vec<u8>` but is incorrect in many places in the code.

gterzian:
Close #2138


adrien-zinger (approved):

adrien-zinger:
bors merge

PR #2141, opened 2022-01-12
Temporary set a version of rust

adrien-zinger:
Set a version temporary, waiting for a nightly release compatible.
Version rustc: 1.59.0

> If you see this commit in a month and it's not already revert, try to revert it quick!

PR #2155, opened 2022-01-15
Revert the version of Rust fixed

### LINKED ISSUE #2142: Revert #2141 after bugfix (and update wiki)
Revert #2141 after https://github.com/rust-lang/rust/pull/92781 is merged and the compiler bug fixed

And don't forget to update the default nightly locally too. 
And maybe we should add a line about that in the doc, as currently testnet users pinned a specific nightly as default.

AurelienFT:
Revert #2141  because the bug on the nightly compiler has been fixed : rust-lang/rust#92781

This will partially resolve  #2142.

I don't know which version should be documented for the episode so I didn't add it to this PR.

### DIFF
--- a/.github/workflows/cd.yml
+++ b/.github/workflows/cd.yml
@@ -21,7 +21,7 @@ jobs:
       - uses: actions-rs/toolchain@v1
         with:
           profile: minimal
-          toolchain: nightly-2021-12-03
+          toolchain: nightly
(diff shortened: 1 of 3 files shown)


AureliaDolo (dismissed):
Thanks ! We use the latest nightly :)

damip (approved):

damip:
bors merge

PR #2169, opened 2022-01-17
Fix #1899: rename `hhasher` related structs

### LINKED ISSUE #1899: Rename hhasher related structs
* [x] HHasher
* [x] EndorsementHashMap/Set
* [x] AddressHashMap/Set
* [x] BlockHashMap/Set
* [x] OperationHashMap/Set

yvan-sraka:
The core purpose of this PR is to simplify a lot codebase by using transparent and explicit types rather than alias defined in models: an `AddressHashSet` became an `Set<Address>` and so on!


massa-protocol-worker/src/protocol_controller.rs
```diff
@@ -92,189 +92,17 @@ pub async fn start_protocol_controller(
     ))
 }
 
-//put in a module to block private access from Protocol_worker.
-mod nodeinfo {
```
yvan-sraka, line 96: Working on this PR, I spotted that this module is just dead code duplicated in `massa-protocol-exports` ... :smiling_face_with_tear:

damip: Feel free to remove dead code :)
Why doesn't the compiler warn about it tho ?

AureliaDolo (approved):

massa-models/src/hash.rs
```diff
@@ -1,41 +1,48 @@
-// Copyright (c) 2021 MASSA LABS <info@massa.net>
```
damip, line 1: we need to keep this for now

damip: also hash.rs is a confusing name. It's not a file dealing with hashes

damip: maybe prehashed.rs or something

massa-models/src/hash.rs
```diff
@@ -1,41 +1,48 @@
...
-
+use massa_signature::{PublicKey, Signature};
+use serde::{Deserialize, Serialize};
 use std::collections::{HashMap, HashSet};
-use std::convert::TryInto;
 use std::hash::{BuildHasherDefault, Hasher};
 use std::marker::PhantomData;
 
+#[derive(Debug, Serialize, Deserialize)]
+pub struct PubkeySig {
```
damip, line 10: I would not have put PubkeySig in this file...

massa-models/src/hash.rs
yvan-sraka: Fixed :heavy_check_mark:

massa-protocol-worker/src/protocol_controller.rs
adrien-zinger: :fearful:

yvan-sraka: My bad, it was the whole file that was duplicated and not included in the project ...

adrien-zinger (dismissed):

damip (changes requested):

massa-models/src/hash.rs
yvan-sraka: In a `pubkey.rs` file in `massa-models`?

damip: yes,  or in crypto, or in composite, as you prefer. All seem better-fit than prehashed.rs

yvan-sraka: Fixed :heavy_check_mark:

damip (approved):

damip:
Personnally I would have named them "PreHashSet" and "PreHashMap" instead of "Set" and "Map" which are quite misleading

yvan-sraka:
bors merge

yvan-sraka:
> @yvan-sraka Personally I would have named them "PreHashSet" and "PreHashMap" instead of "Set" and "Map" which are quite misleading

I see it as an abstraction: I split the underlying mechanism (that could be seen as an optimization, hidden in the library) and the user wanted behavior (how the data structure should behave)!

This separation of concern should allow us in future to implement, change or test optimizations, always with respects to the contract made with user (having data structure that still behaves like a `Set` or a `Map`), that how I see all this!

PR #2175, opened 2022-01-18
Fix: `wallet_remove_addresses` does not update wallet file

### LINKED ISSUE #2172: `wallet_remove_addresses` does not update wallet file
1. `wallet_generate_private_key`
2. `wallet_remove_addresses xxxx`
3. `exit`
4. Launch client again
5. `wallet_info` address is still there!

AurelienFT:
I have fixed by calling `save()`. `save()` can fail with a Result so I changed the return type to be able to propagate the error.

Fix #2172.

### DIFF
--- a/massa-client/src/cmds.rs
+++ b/massa-client/src/cmds.rs
@@ -471,11 +471,11 @@ impl Command {
                 let mut res = "".to_string();
                 for key in parse_vec::<Address>(parameters)?.into_iter() {
                     match wallet.remove_address(key) {
-                        Some(_) => {
+                        Ok(_) => {
                             res.push_str(&format!("Removed address {} from the wallet\n", key));
                         }
-                        None => {
-                            res.push_str(&format!("Address {} wasn't in the wallet\n", key));
+                        Err(err) => {
+                            res.push_str(&err.to_string());
                         }
                     }
                 }
--- a/massa-client/src/cmds.rs
+++ b/massa-client/src/cmds.rs
@@ -12,7 +12,7 @@ use massa_models::{
 };
 use massa_signature::{generate_random_private_key, PrivateKey, PublicKey};
 use massa_time::MassaTime;
-use massa_wallet::Wallet;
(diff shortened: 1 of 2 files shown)


massa-client/src/cmds.rs
```diff
@@ -471,11 +471,11 @@ impl Command {
                 let mut res = "".to_string();
                 for key in parse_vec::<Address>(parameters)?.into_iter() {
                     match wallet.remove_address(key) {
-                        Some(_) => {
+                        Ok(_) => {
                             res.push_str(&format!("Removed address {} from the wallet\n", key));
                         }
-                        None => {
```
damip, line 477: let's keep this branch with `Err(MissingKeyError(_)) => res.push_str(&format!("Address {} wasn't in the wallet\n", key))` because it's a common error that needs proper formatting

massa-wallet/src/lib.rs
```diff
@@ -77,8 +77,9 @@ impl Wallet {
         }
     }
 
-    pub fn remove_address(&mut self, address: Address) -> Option<(PublicKey, PrivateKey)> {
-        self.keys.remove(&address)
+    pub fn remove_address(&mut self, address: Address) -> Result<(), WalletError> {
+        self.keys.remove(&address).ok_or(WalletError::MissingKeyError(address))?;
+        self.save()
     }
```
damip, line 84: also need to run "cargo fmt" before pushing for sanity

damip (changes requested):

massa-wallet/src/lib.rs
damip: otheerwise, all good !

AureliaDolo: You can set up you code editor to run cargo fmt on save

AurelienFT: Yes I have made it but only on the massa_client subfolder. Done.

damip (approved):

damip:
bors merge

PR #2185, opened 2022-01-18
Remove some code duplication in bootstrap

adrien-zinger:
Actually, it's more a _"Beautify"_ than a deletion of code duplication. If you have some suggestions that improve the PR, you can assign it to you and continue the work!!


yvan-sraka (approved):
That's neat ✨, let's merge this and continue in a follow-up!

AureliaDolo (approved):
That's great, I guess there is plenty of stuff like that all around,  I wonder if there is a solution to detect that automatically ?

yvan-sraka:
bors merge

PR #2189, opened 2022-01-18
Finish output event system

### LINKED ISSUE #2080: Finish output event system

AureliaDolo:
Fix #2080


massa-execution/src/interface_impl.rs
```diff
@@ -230,16 +230,35 @@ impl Interface for InterfaceImpl {
     }
 
     fn generate_event(&self, data: String) -> Result<()> {
-        let context = context_guard!(self);
+        let mut context = context_guard!(self);
         let slot = context.slot;
```
damip, line 234: no need to copy all of those, just use context.slot when you need it

massa-execution/src/interface_impl.rs
```diff
@@ -230,16 +230,35 @@ impl Interface for InterfaceImpl {
...
         let block = context.opt_block_id;
         let call_stack = context.call_stack.clone();
+        let read_only = context.read_only;
+        if !read_only {
```
damip, line 238: always increment created_event_index AFTER generating it, even if we are in readonly

massa-execution/src/interface_impl.rs
```diff
@@ -230,16 +230,35 @@ impl Interface for InterfaceImpl {
...
+        }
+
+        let mut to_hash: Vec<u8> = slot.to_bytes_key().to_vec();
+        to_hash.append(&mut context.created_event_index.to_be_bytes().to_vec());
+        if read_only {
+            to_hash.push(0u8);
+        } else {
+            to_hash.push(1u8);
+        }
+
```
damip, line 249: do the created_event_index incrementation here, always (even if read_only)

massa-execution/src/interface_impl.rs
```diff
@@ -230,16 +230,35 @@ impl Interface for InterfaceImpl {
...
         let context = EventExecutionContext {
```
damip, line 250: keep the context lock to be able to push the event into it, don't override it with this variable

massa-execution/src/interface_impl.rs
```diff
@@ -230,16 +230,35 @@ impl Interface for InterfaceImpl {
...
             slot,
             block,
             call_stack,
```
damip, line 253: here you can just context.call_stack.clone()

massa-execution/src/interface_impl.rs
```diff
@@ -230,16 +230,35 @@ impl Interface for InterfaceImpl {
...
         };
-        let event = SCOutputEvent { context, data };
+        let id = Hash::compute_from(&to_hash); // FIXME two read only smart contracts at the same slot will have the same id if no not read only event was produced in between
+        let event = SCOutputEvent {
+            id,
+            context,
+            data,
+            read_only,
+        };
         // store the event somewhere
```
damip, line 262: send the event into the dedicated field in context

massa-execution/src/types.rs
```diff
@@ -33,6 +33,7 @@ pub(crate) struct ExecutionContext {
     pub gas_price: Amount,
     pub slot: Slot,
     pub created_addr_index: u64,
+    pub created_event_index: u64,
     pub opt_block_id: Option<BlockId>,
```
damip, line 37: also add a container listing all the generated events

massa-models/src/output_event.rs
```diff
@@ -1,12 +1,13 @@
...
-
 use crate::{Address, BlockId, DeserializeCompact, SerializeCompact, Slot};
+use massa_hash::hash::Hash;
+use serde::{Deserialize, Serialize};
+use std::collections::VecDeque;
 
 #[derive(Debug, Clone, Serialize, Deserialize)]
 /// By product of a byte code execution
 pub struct SCOutputEvent {
+    pub id: Hash,
```
damip, line 9: maybe let's define an EventId wrapper like we did for BlockId etc...

damip (changes requested):

massa-execution/src/types.rs
```diff
@@ -22,6 +25,8 @@ pub(crate) struct StepHistoryItem {
 
     // list of SCE ledger changes caused by this execution step
     pub ledger_changes: SCELedgerChanges,
+
+    pub events: HashMap<Hash, SCOutputEvent>,
```
damip, line 29: it might be more practical for the user to keep the order of the emitted messages. Maybe ` Vec<(Hash, SCOutputEvent)> `would be nice. What do you think ?

massa-execution/src/vm.rs
```diff
@@ -234,8 +234,8 @@ impl VM {
                 |_| ReadOnlyResult::Error("Failed to run in read-only mode".to_string()),
                 |_| ReadOnlyResult::Ok,
             ),
-            // TODO: integrate with output events.
-            output_events: None,
+            // integrate with output events.
+            output_events: context.events.clone(),
```
damip, line 238: use `mem::take` to avoid cloning

massa-execution/src/vm.rs
```diff
@@ -311,6 +311,7 @@ impl VM {
             slot: step.slot,
             opt_block_id,
             ledger_changes: mem::take(&mut context.ledger_step.caused_changes),
+            events: context.events.clone(),
```
damip, line 314: use mem::take as above

massa-execution/src/vm.rs
```diff
@@ -311,6 +311,7 @@ impl VM {
             slot: step.slot,
```
damip, line 311: I comment here because Github doesn't let me comment at the line I want.

* clear_and_update_context should include a reset to zero of the event counter as well as a clear() of the list of generated events

* prepare_context should also return a cloned snapshot of the events that can be rollbacked in case of failure, just like `context.ledger_step.caused_changes = ledger_changes_backup;` for example

damip (changes requested):

massa-execution/src/types.rs
AureliaDolo: This is a place holder for a struct with multiple indexes as we want to search by caller, smart contract and slot. I'd rather add an index field in the SCOutputEvent

damip: let's do that

damip:
Next step will be adding the events in limited-size database when the corresponding history step becomes final, and making this database searchable together with the active history itself

massa-execution/src/vm.rs
```diff
@@ -267,8 +290,8 @@ impl VM {
                 |_| ReadOnlyResult::Error("Failed to run in read-only mode".to_string()),
                 |_| ReadOnlyResult::Ok,
             ),
-            // TODO: integrate with output events.
-            output_events: None,
+            // integrate with output events.
+            output_events: mem::take(&mut context.events).export(),
```
damip, line 294: why not have output_events be also an EventStore so that you can just take the whole event store ?

damip: and EventStore could have a method `EventStore::extend(&mut self, other: &EventStore)` to merge another event store into it while keeping a config-defined max number of elements.

This extend can then be used to extend the final EventStore with a history element that finalized

massa-execution/src/types.rs
```diff
@@ -24,6 +29,59 @@ pub(crate) struct StepHistoryItem {
...
+                Entry::Vacant(e) => {
+                    e.insert(vec![id].into_iter().collect());
+                }
+            }
+            if let Some(caller) = event.context.call_stack.front() {
+                self.caller_to_id
+                    .get_mut(caller)
+                    .unwrap_or(&mut vec![].into_iter().collect()) // TODO maybe implement default for Set
+                    .insert(id);
+            }
```
AureliaDolo, line 61: In a nutshell here I do the same thing twice (but for a different map) and I'm not sure which way is the better any opinions @yvan-sraka @adrien-zinger

massa-execution/src/vm.rs
AureliaDolo: I'm quite not sure about where to store final events: 
- the final ledger is in a ledger step in an execution context in the vm 
- for now I store the events in a step history item in the step history in the vm 
- when I query events, for now, I iter on step history items in vm methods
- from what I understood, I don't want/need to have access to past events from the execution context.

So I guess I can have a final EventStore in the VM, prune it every now and then, and updating it during run_final_step ? So upon query I'll have to iter on the step history items and on the final ones. (That means that I should return a Vec<(SCOutputEvent, bool)> instead of a Vec<SCOutputEvent> to indicate whether the event is final or not)

AureliaDolo: > why not have output_events be also an EventStore so that you can just take the whole event store ?

That more a user interface issue, I don't know how people expect this to behave. I don't feel like returning a bunch of indexes on top of everything else through the api is really usefull, but I  may be wrong. @z80dev what do you expect from that ?

massa-execution/src/types.rs
AureliaDolo: Thanks for your help guys ! :)

AureliaDolo:
I did  [this](https://mermaid.live/edit/#pako:eNqNlU2P4jAMhv9K1BMjMfcVh5UYYDVIizgwOyekypuYNqJNqnywIMp_3zRtoXTSgVtkP3Ycx3lzjqhkGE2iREGRko_5VhBCPlfk9fUn-UgVAptJK0zXvDFYvHNtpDp1zUthUO2AYte4OCK1hksxk859rPN0ElRQeUBado1Lg3kfrGz17pk0g871aF1Uu5G3TNL9kr0Mp5ktfiNLUM1SEAnqQXBxQGE2zlSfqx_nD7DHU0mmjCnUepg6QGax7PtCvN9YqjMKm188IJW3zTFDg13LBptuoLk_10KY5n4a8k_BoImtlwG-w9w7PDvNr6MQ8K7frs0_GazG6mUIXY1WUJB30Cn5RErsjxDZKXI9bVPXJTzCb6U8UVE3cNXGBeu7jYKnucFR7jiszISzcb0Kstpwz2o3vWOi3WW1UUGcNjiFLEP1RIBuIzQN0r1neH_11dgPYzkc4wT0MEAlFzrWBuh-GHIZ4kLxRh2CiG5fdngXp0UGWQzumcVcMDw-Zn0THsGyMPHfSi5izp6hfHKpfCHflOBu7lFP5D_RHMjpBj5sYSXGsRTZaRjpiVWQsULDDmMlkvvHUE2BB35xAVltCxMltbnNwPADxmktlzGtha4cENavKcBqd_jvwzqV1BNyHZG-5xp-v91NnSGkzkFZ9qIQjaMcVQ6cub_xXEVtI5Nijtto4pYMd2Azs4224uJQ6yVkwbhrRTTZQaZxHIE1cnMSNJoYZbGF5hzcV5s31OU_pFeKTQ)  to see a little bit more clearly

damip (changes requested):
some changes

massa-execution/src/interface_impl.rs
```diff
@@ -390,19 +390,25 @@ impl Interface for InterfaceImpl {
...
 
+    /// generate an execution event and stores it
     fn generate_event(&self, data: String) -> Result<()> {
-        let context = context_guard!(self);
-        let slot = context.slot;
-        let block = context.opt_block_id;
-        let call_stack = context.call_stack.clone();
+        let mut execution_context = context_guard!(self);
+
+        let mut to_hash: Vec<u8> = execution_context.slot.to_bytes_key().to_vec();
```
gterzian, line 397: Could you please add a comment explaining what this is?

massa-execution/src/types.rs
```diff
@@ -26,6 +31,217 @@ pub(crate) struct StepHistoryItem {
...
+
+    /// maps direct event producer to a set of event ids
+    smart_contract_to_id: Map<Address, Set<SCOutputEventId>>,
+}
+
+impl EventStore {
+    /// add event to the store and all its indexes
+    pub fn insert(&mut self, id: SCOutputEventId, event: SCOutputEvent) {
+        if let Entry::Vacant(entry) = self.id_to_event.entry(id) {
+            self.slot_to_id
```
gterzian, line 59: I think you want to use [`entry`](https://doc.rust-lang.org/std/collections/hash_map/struct.HashMap.html#method.entry), and in any case use something like [`or_insert_with`](https://doc.rust-lang.org/std/collections/hash_map/enum.Entry.html#method.or_insert_with) or [`unwrap_or_else`](https://doc.rust-lang.org/std/option/enum.Option.html#method.unwrap_or_else) to only allocate the default when necessary(altough I guess if `default()` doesn't allocate anything these days it doesn't matter).

massa-execution/src/types.rs
```diff
@@ -26,6 +31,217 @@ pub(crate) struct StepHistoryItem {
...
+    pub fn clear(&mut self) {
+        self.id_to_event.clear();
+        self.slot_to_id.clear();
+        self.caller_to_id.clear();
+        self.smart_contract_to_id.clear();
+    }
+
+    /// Remove exess events considering a config defined max
+    pub fn prune(&mut self, max_final_events: usize) {
+        // todo make setting static
```
gterzian, line 97: Please open an issue for the todo.

massa-execution/src/types.rs
```diff
@@ -26,6 +31,217 @@ pub(crate) struct StepHistoryItem {
...
+        if self.id_to_event.len() > max_final_events {
+            let diff = self.id_to_event.len() - max_final_events;
```
gterzian, line 99: You could store `len` in a variable to avoid calling it twice.

massa-execution/src/vm.rs
```diff
@@ -200,14 +260,14 @@ impl VM {
     /// See https://github.com/massalabs/massa/wiki/vm_ledger_interaction
     /// TODO: do not ignore the results
     /// TODO: consider dispatching gas fees with edorsers/endorsees as well
-    /// Returns (backup of local ledger changes, backup of created_addr_index)
+    /// Returns (backup of local ledger changes, backup of created_addr_index, backup of events, backup of created_events_index, backup of unsafe rng)
```
gterzian, line 263: I think it's best to use multiple lines for this.

gterzian (approved):
LGTM with comments

massa-execution/src/interface_impl.rs
damip: changed

massa-execution/src/types.rs
damip: changed

massa-execution/src/types.rs
damip: changed

damip:
I changed a lot of things. Should be more simple, consistent and user-friendly. Ready for reviews

massa-execution/src/vm.rs
```diff
@@ -409,6 +451,12 @@ impl VM {
             let block_creator_addr = Address::from_public_key(&block.header.content.creator);
             // run all operations
             for (op_idx, operation) in block.operations.iter().enumerate() {
+                // get operation ID
+                // TODO for the refactoring, have operations include their operationID (without necessarily serializing it)
```
gterzian, line 455: Please open an issue or add it to an existing issue.

massa-execution/src/types.rs
```diff
@@ -374,20 +166,13 @@ pub(crate) enum ExecutionRequest {
...
-    GetSCOutputEventBySCAddress {
-        sc_address: Address,
+    /// Get events with filter criteria
+    GetSCOutputEvents {
+        start_slot: Option<Slot>,
+        end_slot: Option<Slot>,
+        emitter_address: Option<Address>,
+        original_caller_address: Option<Address>,
+        origin_operation_id: Option<OperationId>,
         response_tx: oneshot::Sender<Vec<SCOutputEvent>>,
```
gterzian, line 176: Can we get some docs on each member please?

gterzian (commented):
LGTM in general, with some comments. I will let @AureliaDolo review in detail.

adrien-zinger (dismissed):

AureliaDolo (commented):
From what I understand you've done three things here 
- removing the event store concept
- merging all the api endpoints
- adding origin operation id as a criterion

The two first points should have been discussed here https://github.com/massalabs/massa/pull/2033 when they were defined.

For the event store point I don't see why your solution is better. If we do that we'll have to recompute every thing at each query. While using an event store we just have to merge everything we get from the stores (in the future it could be easily parallelized)

massa-execution/src/types.rs
```diff
@@ -79,6 +80,76 @@ impl EventStore {
         self.smart_contract_to_id.clear();
     }
 
+    pub fn prune(&mut self, max_final_events: usize) {
```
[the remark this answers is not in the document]
damip: changed

massa-execution/src/types.rs
damip: changed

damip:
> From what I understand you've done three things here
> 
>     * removing the event store concept
> 
>     * merging all the api endpoints
> 
>     * adding origin operation id as a criterion
> 
> 
> The two first points should have been discussed here #2033 when they were defined.
> 
> For the event store point I don't see why your solution is better. If we do that we'll have to recompute every thing at each query. While using an event store we just have to merge everything we get from the stores (in the future it could be easily parallelized)

@AureliaDolo 

You're right, this is mostly a proposal you can get some inspiration from. And after trying it I see that the EventStore approach was better. Feel free to rollback. That being said, some ideas from my contribution seem interesting:
* EventId is not necessary and its hash computation is costly (nobody will search an event by its ID), events can be uniquely and more efficiently indexed by the pair `(slot, index_in_slot)`
* indexing done by slot, by SC address that emitted the event (back of the call stack), original caller (front of the call stack) and optional origin OperationID
* allowing a flexible multi-criteria search with a single endpoint

massa-execution/src/types.rs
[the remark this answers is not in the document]
AureliaDolo: @adrien-zinger @yvan-sraka any ideas on that ?

massa-execution/src/vm.rs
AureliaDolo: Done

massa-execution/src/interface_impl.rs
AureliaDolo: Done

massa-execution/src/types.rs
AureliaDolo: I hopefully copy pasted the doc everywhere, but I'm not sure how clear it is.

massa-execution/src/vm.rs
AureliaDolo: see #2234

AureliaDolo:
> EventId is not necessary and its hash computation is costly (nobody will search an event by its ID), events can be uniquely and more efficiently indexed by the pair (slot, index_in_slot)

I'm not so sure about that: using a real Id that is a hash allow us to used the prehashed trait, and avoid hashing too much

massa-execution/src/types.rs
AureliaDolo: Done

yvan-sraka (approved):
LGTM

massa-execution/src/types.rs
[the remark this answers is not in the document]
adrien-zinger: What about :point_down: ?

```rust
#[inline]
/// Remove a given event_id from a `Set<SCOutputEventId>`
/// The Set is stored into a Hashmap `ctnr` at a `key` address. If
/// the Set resulted from the operation is empty, remove the entry
/// from the `ctnr`
/// 
/// Used in `alternatve_prune ()`
fn remove<T>(ctnr: &mut HashMap<T, Set<SCOutputEventId>>, key: &T, evt_id: &SCOutputEventId) {
    match ctnr.get_mut(key) {
        Some(ele) => {
            ele.remove(evt_id);
            if ele.is_empty() {
                ctnr.remove(key);
            }
        }
        _ => ele.remove(key)
    }
}

/// Prune the exess of events from the event store,
/// While there is a slot found, pop slots and get the `event_ids`
/// inside, remove the event from divers containers.
/// 
/// Return directly if the event_size < max_final_events
pub fn alternative_prune(&mut self, max_final_events: usize) {
    let mut events_size = self.id_to_event.len();
    if (events_size <= max_final_events) {
        return
    }
    let mut slots = self.slot_to_id.keys().collect::<Vec<_>>();
    slots.sort_unstable_by_key(|s| cmp::Reverse(*s));
    loop {
        let event_id = match slots.pop() {
            Some(slot) => slot,
            _ => return
        };
        let event_ids = match self.slot_to_id.get(slot) {
            Some(event_ids) => event_ids,
            _ => continue
        };
        for event_id in event_ids.iter() {
            let event = match self.id_to_event.remove(&id) {
                Some(event) => event,
                _ => continue,
            };
            remove(&mut self.slot_to_id, &event.context.slot);
            if let Some(caller) = event.context.call_stack.front() {
                remove(&mut self.caller_to_id, caller);
            }
            if let Some(sc) = event.context.call_stack.back() {
                remove(&mut self.smart_contract_to_id, sc);
            }
            if let Some(op) = event.context.origin_operation_id {
                remove(&mut self.operation_id_to_event_id, op);
            }
            events_size -= 1;
            if events_size <= max_final_events {
                return
            }
        }
    }
}
```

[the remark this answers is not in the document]
AureliaDolo: Thanks !

yvan-sraka (approved):

adrien-zinger (approved):

AureliaDolo:
bors merge

PR #2190, opened 2022-01-18
Quick fix `massa_trace` (simplify it a bit)

yvan-sraka:


AureliaDolo:
Is the output the same ?

yvan-sraka:
> Is the output the same ?

It slightly changes (for the better, I think)!

yvan-sraka:
I'm wondering about remove `massa_trace:` prefix?

yvan-sraka:
(and what's neat with `tokio::tracing` it that it use https://doc.rust-lang.org/std/macro.module_path.html macro to indicate where the log comme form in codebase)

AurelienFT (commented):
Be careful, there is a whole peers.json file with thousands of peers in the changes of this PR. I think it's a bad push because there is nothing related to this in the PR.

AureliaDolo:
> Be careful, there is a whole peers.json file with thousands of peers in the changes of this PR. I think it's a bad push because there is nothing related to this in the PR.

Soon it will be in the gitignore :) #2139

AureliaDolo (commented):
I'm not sure about that since we're changing the logs output. Maybe if you wrote a quick note intend to people that wrote scripts using theses logs explaining how it has changed.

damip (approved):

yvan-sraka:
@damip posted on Discord:

> technically, tha tracing crate should already not be evaluating JSON for the following branch at all: https://docs.rs/tracing/latest/src/tracing/macros.rs.html#605-618
> 
> It means that the JSON evaluation is likely coming from the following branch: https://docs.rs/tracing/latest/src/tracing/macros.rs.html#620-623
> 
> but this second branch also has an extra level filter: https://docs.rs/tracing/latest/src/tracing/macros.rs.html#620-623
> 
> So all of this filtering must be failing. We should investigate why. Maybe we have a trace-level subscriber somewhere

AureliaDolo:
So @yvan-sraka what do you want to do with this pr ?

yvan-sraka:
Merge it! But first maybe understand more what happens by investigating with `tokio-tracing`?

What can I do, to not stale the PR for too long, is trying to hide the changes required by tokio-tracing behind an `instrument` feature?

AureliaDolo (approved):

PR #2199, opened 2022-01-20
Improve warning message.

### LINKED ISSUE #1810: Improve "Node xxx sent us critically wrong xxxxxx" warning
That warning may imply that that node is purposefully sending us invalid stuff, but it only means that this stuff is invalid *from our point of view* 

I suggest adding "This may be caused by desynchronization" at the end of that warning to be explicit

AureliaDolo:
Fix #1810 

Actually retrieving the exact reason why the object in incorrect requires a bit of refactoring. I'm not sure if it's worth it

### DIFF
--- a/massa-protocol-worker/src/protocol_worker.rs
+++ b/massa-protocol-worker/src/protocol_worker.rs
@@ -1351,7 +1351,7 @@ impl ProtocolWorker {
                     .await;
                     self.update_ask_block(block_ask_timer).await?;
                 } else {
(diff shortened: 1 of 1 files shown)


yvan-sraka (approved):
It could be nice to have all these redundant strings into some `const` value :) otherwise LGTM

AureliaDolo:
> It could be nice to have all these redundant strings into some `const` value :) otherwise LGTM

Not sure about that, if we hide these at the start of the file I'm not sure we're gonna use it

AureliaDolo:
bors merge

PR #2205, opened 2022-01-24
Split block graph tests + Structurize channels in consensus

adrien-zinger:
In this PR I managed to extract the tests from block_graph in order to create a mind map, separing concepts, before a big refacto and a benchmark development. Maybe not as usefull as expected but the work is done.

Eventually, this PR also has a structuration of channels used by the consensus in two main `struct`, one private and one public. It solved the signature warnings of some main functions. It's supposed to increase the readability.


yvan-sraka (approved):

adrien-zinger:
bors merge

Issue #2197, opened 2022-01-20
Use thread count from the static config instead of passing it as arg


AureliaDolo:
Thread count is now in the static config, shouldn't it be available here, without the need to duplicate it here ? Or am I missing on something @yvan-sraka

_Originally posted by @AureliaDolo in https://github.com/massalabs/massa/pull/2194#discussion_r787508042_

I think these function could be updated ? (searching for thread_count: u8) (I could be worth having a look at other commonly passed as args config parameters)
* [ ] pub fn get_thread(&self, thread_count: u8) -> u8
* [ ]  pub fn get_next_slot(&self, thread_count: u8) -> Result<Slot, ModelsError>
* [ ] pub fn get_block_slot_timestamp(
    thread_count: u8,
    t0: MassaTime,
    genesis_timestamp: MassaTime,
    slot: Slot,
) -> Result<MassaTime, ModelsError> // should take only slot as a param
* [ ]  pub fn get_latest_block_slot_at_timestamp(
    thread_count: u8,
    t0: MassaTime,
    genesis_timestamp: MassaTime,
    timestamp: MassaTime,
) -> Result<Option<Slot>, ModelsError>
* [ ] pub fn get_current_latest_block_slot(
    thread_count: u8,
    t0: MassaTime,
    genesis_timestamp: MassaTime,
    clock_compensation: i64,
) -> Result<Option<Slot>, ModelsError>
* [ ] pub fn time_range_to_slot_range(
    thread_count: u8,
    t0: MassaTime,
    genesis_timestamp: MassaTime,
    start_time: Option<MassaTime>,
    end_time: Option<MassaTime>,
) -> Result<(Option<Slot>, Option<Slot>), ModelsError>
* [ ] pub fn new(pool_settings: &'static PoolSettings, thread_count: u8) -> EndorsementPool
* [ ] fn new(op: Operation, thread_count: u8) -> Result<Self, PoolError> 
* [ ] pub fn new(
        pool_settings: &'static PoolSettings,
        thread_count: u8,
        operation_validity_periods: u64,
    ) -> OperationPool 
* [ ] pub async fn start_pool_controller(
    pool_settings: &'static PoolSettings,
    thread_count: u8,
    operation_validity_periods: u64,
    protocol_command_sender: ProtocolCommandSender,
    protocol_pool_event_receiver: ProtocolPoolEventReceiver,
) -> Result<(PoolCommandSender, PoolManager), PoolError>
* [ ] pub fn new(
[...]

PR #2211, opened 2022-01-25
Split massa-consensus

adrien-zinger:
Split the consensus into multiple crates
worker, exports and also massa-graph and proof-of-stake-exports

This PR contains also the development of a split of configs and errors between POS, graph, ledger, and consensus.

I didn't manage the split of the existing unit tests to simplify the review. I would prefer to move the unit tests from massa-consensus-worker with another PR. This choice has an impact on the visibility of some structures and functions, and on some _test feature_ flags.


adrien-zinger:
Currently, I've an issue with the tempfiles, I'm investigating

yvan-sraka (dismissed):

massa-graph/src/active_block.rs
```diff
@@ -0,0 +1,58 @@
...
+}
+
+impl ActiveBlock {
+    /// Computes the fitness of the block
+    pub fn fitness(&self) -> u64 {
+        1 + self.block.header.content.endorsements.len() as u64
+    }
+}
+
+impl From<&ActiveBlock> for POSBlock {
```
gterzian, line 50: Good example of something we can address with the refactoring, to avoid cloning...

gterzian (approved):
Great, thanks! Agree on doing tests as follow-up...

massa-consensus-exports/src/tools.rs
```diff
@@ -0,0 +1,62 @@
+use std::collections::HashMap;
```
AureliaDolo, line 1: These are tools meant to be only used in tests, I'd rename/move the file to make that clear

massa-graph/src/settings.rs
```diff
@@ -0,0 +1,72 @@
...
+use serde::{Deserialize, Serialize};
+use std::{path::PathBuf, usize};
+
+/// Consensus full configuration (static + user defined)
+///
+/// Assert that `THREAD_COUNT >= 1 || T0.to_millis() >= 1 || T0.to_millis() % THREAD_COUNT == 0`
+#[derive(Debug, Deserialize, Serialize, Clone)]
+pub struct LedgerConfig {
+    /// Number of threads
+    pub thread_count: u8,
```
AureliaDolo, line 17: So you duplicate thread count in all the configs that need it right ?

massa-consensus-exports/src/settings.rs
```diff
@@ -312,84 +323,152 @@ impl ConsensusConfig {
...
+/// Execution & Protocol Sender/Receiver
+pub struct ConsensusChannels {
+    pub execution_command_sender: ExecutionCommandSender,
+    pub execution_event_receiver: ExecutionEventReceiver,
+    pub protocol_command_sender: ProtocolCommandSender,
+    pub protocol_event_receiver: ProtocolEventReceiver,
+    pub pool_command_sender: PoolCommandSender,
+}
+
+const SERIALIZATION_CONTEXT: SerializationContext = SerializationContext {
```
AureliaDolo, line 414: Probable a discussion for another issue : shouldn't we get rid of the serialization context ?

massa-consensus-exports/src/settings.rs
```diff
@@ -312,84 +323,152 @@ impl ConsensusConfig {
...
+    max_ask_blocks_per_message: 10,
+    max_operations_per_message: 1024,
+    max_endorsements_per_message: 1024,
+    max_bootstrap_message_size: 100000000,
+    max_bootstrap_pos_entries: 1000,
+    max_bootstrap_pos_cycles: 5,
+    max_block_endorsements: 8,
+};
+
+use std::path::Path;
```
AureliaDolo, line 433: why this import is not as the top of the file ?

AureliaDolo (commented):
Why are pos.md and endorsement.md in the worker ? (mAybe in more of a hacking or internal section of the wiki)

yvan-sraka:
> Why are pos.md and endorsement.md in the worker ? (maybe in more of a hacking or internal section of the wiki)

It's both something we like to have included in `cargo doc` output and linked in a hacking/internal section of wiki :)

damip (approved):
Didn't read all the details, but the general scheme seems good !

yvan-sraka:
bors merge

massa-consensus-exports/src/tools.rs
gterzian: I think that's a good point, best to move those to the worker where the tests are.

gterzian:
Since there are still unresolved discussions and suggested changes, we should wait for these to be resolved before trying to merge.

Also as a general note, after having addressed a review, it's best to ask for a second review, as opposed to assuming that the changes made were as the reviewer intended and to immediately try to merge. It's possible that the reviewer still has a comment, or that the changes do not completely address the earlier review.

cc @massalabs/core-team

AureliaDolo:
> Since there are still unresolved discussions and suggested changes, we should wait for these to be resolved before trying to merge.
> 
> Also as a general note, after having addressed a review, in most cases it's a good idea to ask for a second review, as opposed to assuming that the changes made were as the reviewer intended and to immediately try to merge. It's possible that the reviewer still has a comment after the changes, for example if they do not completely address the earlier review comments.
> 
> The exception would be for really minor or even optional changes, which the reviewer can make clear by marking the comments with "nit", or perhaps by using the ["LGTM with comments"](https://google.github.io/eng-practices/review/reviewer/speed.html#lgtm-with-comments) convention.
> 
> We could consider eventually writing down clearly in our own doc what standards we want to apply to our own review process, I've opened an issue for this at #2220.
> 
> cc @massalabs/core-team

Related to #1069 and #1927

massa-graph/src/settings.rs
adrien-zinger: The thread count management will be the object of another issue/PR

massa-graph/src/active_block.rs
adrien-zinger: Can you clarify your thoughts? :) We could see if we can move the ActiveBlock in models or use a kind of code as @yvan-sraka showed up in the meet today:

```rust
struct Grounded;
struct Launched;

struct Rocket<T = Grounded> {
    stage: std::marker::PhantomData<T>,
}

impl Rocket<Grounded> {
    pub fn launch(self) -> Rocket<Launched> {}
}

impl Rocket<Launched> {
    pub fn accelerate(&mut self) {}
    pub fn decelerate(&mut self) {}
}

impl<T> Rocket<T> {
    pub fn color(&self) -> Color {}
    pub fn weight(&self) -> Kilograms {}
}
```

massa-graph/src/settings.rs
AureliaDolo: #2197

massa-consensus-exports/src/settings.rs
AureliaDolo: Probably linked to #2197 too

AureliaDolo (dismissed):

massa-consensus-exports/src/tools.rs
adrien-zinger: Actually, one of these functions is also used in an implementation of ConsensusConfig (the `impl<&Path>`) that cannot be moved easily. We should first think about _"Where we should put the common configs tools and implementations"_. I'm open to any proposals!!

massa-graph/src/active_block.rs
adrien-zinger: Ok, I just pushed one version with the atoms of the code in models, ⁣:wink:  what do you think?

gterzian (dismissed):
LGTM with one comment

AureliaDolo:
It looks like you also tackled #237 here ?

yvan-sraka (dismissed):
LGTM

massa-graph/src/active_block.rs
adrien-zinger: @gterzian Do you think you can close this discussion? :)

massa-consensus-exports/src/settings.rs
adrien-zinger: I fixed that point forward with the config refactoring

massa-graph/src/active_block.rs
adrien-zinger: I resolve it because the code has changed. Nevertheless, we'll keep the discussion open in future PR.

massa-models/src/error.rs
```diff
@@ -32,4 +34,8 @@ pub enum ModelsError {
     TimeOverflowError,
     /// Time error {0}
     TimeError(#[from] massa_time::TimeError),
+    /// invalid roll update: {0}
+    InvalidRollUpdate(String),
+    /// Ledger changes, Amount overflow
```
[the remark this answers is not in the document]
adrien-zinger: I close that and open the #2257

AureliaDolo (approved):

adrien-zinger:
bors merge

PR #2221, opened 2022-01-28
thread count constant

adrien-zinger:
I'm not really happy of this version. The main reason is that we cannot share for the current version with cargo the need of a _"test configuration"_ (or _"test feature"_ call it how you prefer).

The thread_count... **isn't a constant, it has never been a constant**. Yes, we had a constant value, but quickly the usage of the thread_count in the project is not the usage of a constant anymore. We require an injection of dependencies in order to test where we want to test. And that's why I don't understand why there is an issue.

Maybe I didn't understand the problem:
1. We want to avoid cloning the config?
  -> We just have to get a massive configuration and continue to pass a reference. But we cannot check with this technic if we accidentally changed a value (for example, it happened to me with the `ledger_path`)

2. We want to control and simplify the development, providing the configuration when required?
  -> We can centralize the all configuration in a RwLock _(if mutable in test)_ and an Arc _(if not)_, then we can control the development and avoid accidental modifications. Stop passing `cfg` argument with information known at compilation and values read from the config file. But, we'll have to deal with a static value that might be seen as a bad practice.

3. We want to know what is the config used in test and the config in release?
  -> Both solutions are usable, we don't have to recreate the all configuration management for that, Just be clear on what is in the configuration. We can centralize the definition and some factories tools. If we choose the dependency injection method, we could also think about passing an Arc reference instead of borrowing/cloning/dispatching. We could have _ONE config_. Everyone will be happy.

Eventually, in this PR you'll see how it can be with RwLock for the thread count.


damip:
I personally don't know what the best solution is, I tried only a couple. That being said, here is my opinion:

* I vote against using RwLocks because every time we read the config (and we do it very often !), it will cause a CPU-wide sync event affecting all cores. We don't want this overhead, especially in a multithreaded application.

* I vote for defining a big NodeConfig containing all parameters needed by all modules. NodeConfig is loaded from file and/or constants, then its fields are used in massa-node to initialize the fields of ConsensusConfig (containing all parameters that the consensus crate needs), NetworkConfig (containing all parameters that the network crate needs) etc... Note that some parameters like thread_count are used by multiple modules. That way, config passing is more verbose but always explicit and simple to understand (I still struggle with the config since it was changed, I find it overly complicated, and a high entry barrier in code readability and dev onboarding). This method has no performance overhead and a very low risk of error.
I feel like the only argument against this method is that it requires explicitly passing some config values (eg. thread_count) to serialization functions, slot computation functions etc...

adrien-zinger:
> I personally don't know what the best solution is, I tried only a couple. That being said, here is my opinion:
> 
> * I vote against using RwLocks because every time we read a value from the config (and we do it very often !), it will cause a CPU-wide sync event affecting all cores. We don't want this overhead, especially in a multithreaded application.
> 
> * I vote for defining a big NodeConfig containing all parameters needed by all modules. NodeConfig is loaded from file and/or constants, then its fields are used in massa-node to initialize the fields of ConsensusConfig (containing all parameters that the consensus crate needs), NetworkConfig (containing all parameters that the network crate needs) etc... Note that some parameters like thread_count are used by multiple modules. That way, config passing is more verbose but always explicit and simple to understand (I still struggle with the config since it was changed, I find it overly complicated, and a high entry barrier in code readability and dev onboarding). This method has no performance overhead and a very low risk of error.
> I feel like the only argument against this method is that it requires explicitly passing some config values (eg. thread_count) to serialization functions, slot computation functions etc... and I think we can afford this.
> 

After this development I totally agree with you

Issue #2222, opened 2022-01-30
Send peer list when refusing incoming connection


damip:
When refusing an incoming connection (because we have too many), try to send the peer list (with short timeout) before closing.

There should also be a maximum amount of simultaneous "try to peer send on refusal" attempts, beyond which the incoming connections are directly refused.

PR #2231, opened 2022-01-31
Send a list of peer if max connection reached

adrien-zinger:
Manage a special operation if the handshake return an error with
a peer list.

Link to the issue : #2222 

In this PR I did some modification in handshake errors and network errors in order to get a list of new candidates nodes when a candidate has too many connections. The main logic is done in the handshake.

I also planned to remove some warnings and simplify the `HandshakeWorker` object. I think it would be great to move the implementation of all _connecting's_ functions into a dedicated file with a cool flowchart. Furthermore, I think I can now do that.

### DIFF
--- a/massa-network/src/handshake_worker.rs
+++ b/massa-network/src/handshake_worker.rs
@@ -44,7 +44,6 @@ impl HandshakeWorker {
     /// Manage a new connection and perform a normal handshake
     ///
     /// Used for incomming and outgoing connections.
-    /// Add the node with
     /// It will spawn a new future with an HandshakeWorker from the given `reader`
     /// and `writer` from your current node to the distant `connectionId`
     ///
--- a/massa-network/src/handshake_worker.rs
+++ b/massa-network/src/handshake_worker.rs
@@ -115,9 +115,6 @@ impl HandshakeWorker {
         .await
         {
             Err(_) => throw!(HandshakeTimeout),
-            Ok(Err(NetworkError::HandshakeError(HandshakeErrorType::PeerListReceived(list)))) => {
-                throw!(PeerListReceived, list)
-            }
             Ok(Err(e)) => return Err(e),
             Ok(Ok((_, None))) => throw!(HandshakeInterruption, "init".into()),
             Ok(Ok((_, Some((_, msg))))) => match msg {
@@ -126,6 +123,7 @@ impl HandshakeWorker {
                     random_bytes: rb,
                     version,
                 } => (NodeId(pk), rb, version),
+                Message::PeerList(list) => throw!(PeerListReceived, list),
                 _ => throw!(HandshakeWrongMessage),
             },
         };
--- a/massa-network/src/handshake_worker.rs
+++ b/massa-network/src/handshake_worker.rs
@@ -54,6 +54,7 @@ impl HandshakeWorker {
     /// * private_key : our private key.
     /// * timeout_duration: after timeout_duration millis, the handshake attempt is dropped.
     /// * connection_id : Node we are trying to connect for debuging
+    /// * version : Node version used in handshake initialization (check peers compatibility)
     pub fn spawn(
         socket_reader: ReadHalf,
         socket_writer: WriteHalf,
--- a/massa-network/src/handshake_worker.rs
+++ b/massa-network/src/handshake_worker.rs
@@ -6,17 +6,19 @@ use super::{
     messages::Message,
 };
 use crate::{
-    error::{HandshakeErrorType, NetworkError},
-    ReadHalf, WriteHalf,
+    error::{throw_handshake_error as throw, HandshakeErrorType, NetworkError},
+    ConnectionId, ReadHalf, WriteHalf,
 };
 use futures::future::try_join;
 use massa_hash::hash::Hash;
+use massa_logging::massa_trace;
 use massa_models::node::NodeId;
 use massa_models::Version;
 use massa_signature::{sign, verify_signature, PrivateKey};
 use massa_time::MassaTime;
 use rand::{rngs::StdRng, RngCore, SeedableRng};
-use tokio::time::timeout;
+use tokio::{task::JoinHandle, time::timeout};
+use tracing::debug;
 
 /// Type alias for more readability
 pub type HandshakeReturnType = Result<(NodeId, ReadBinder, WriteBinder), NetworkError>;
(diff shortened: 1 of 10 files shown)


massa-network/src/handshake_worker.rs
```diff
@@ -39,35 +41,57 @@ pub struct HandshakeWorker {
 impl HandshakeWorker {
     /// Creates a new handshake worker.
     ///
+    /// Manage a new connection and perform a normal handshake
+    ///
+    /// Used for incomming and outgoing connections.
+    /// Add the node with
```
AureliaDolo, line 47: It looks like this sentences does not have a end ?

massa-network/src/network_worker.rs
```diff
@@ -995,36 +975,54 @@ impl NetworkWorker {
...
+                    Err(NetworkError::PeerConnectionError(
+                        NetworkConnectionErrorType::MaxPeersConnectionReached(_),
+                    )) => {
+                        debug!("Maximum connection reached");
+                        debug!("inbound connection from addr={} refused", remote_addr);
+                        let peer_list = self.peer_info_db.get_advertisable_peer_ips();
+                        WriteBinder::new(writer)
+                            .send(&Message::PeerList(peer_list))
+                            .await?;
+                        ReadBinder::new(reader).next().await?;
```
AureliaDolo, line 1020: Why are you reading here?

massa-network/src/network_worker.rs
```diff
@@ -995,36 +975,54 @@ impl NetworkWorker {
...
+                            .insert(connection_id, (remote_addr.ip(), false));
+                        self.manage_successfull_connection(connection_id, reader, writer)?;
+                    }
+                    Err(NetworkError::PeerConnectionError(
+                        NetworkConnectionErrorType::MaxPeersConnectionReached(_),
+                    )) => {
+                        debug!("Maximum connection reached");
+                        debug!("inbound connection from addr={} refused", remote_addr);
+                        let peer_list = self.peer_info_db.get_advertisable_peer_ips();
+                        WriteBinder::new(writer)
```
AureliaDolo, line 1017: Maybe here add a timeout, to avoid writing for too long if there is an issue

massa-network/src/network_worker.rs
adrien-zinger: I'm not sure, it's because I saw that the handshake can go in timeout error if the distant node never read. I will test that and if it's not necessary I'll remove it

massa-network/src/network_worker.rs
adrien-zinger: Good point

massa-network/src/handshake_worker.rs
adrien-zinger: Oups

massa-network/src/network_worker.rs
AureliaDolo: Until now the handshake was symmetric, so when you were sending anything it was expected that the distant node was also sending something. Maybe with the timeout on the write it should be ok.

Maybe it would be worth having a look at the read side of that to know where we can expect this kind of message.

massa-network/src/handshake_worker.rs
adrien-zinger: Fixed

massa-network/src/network_worker.rs
damip: I think you should add a comment to explain what this read does.

Also let's wait simultaneously on that read and that send with a config-defined timeout peer_propagation_timeout_on_connection_refused: MassaTime

damip (changes requested):

massa-network/src/network_worker.rs
adrien-zinger: Done :)

massa-network/src/network_worker.rs
adrien-zinger: Done!

AureliaDolo (approved):

massa-network/src/network_worker.rs
```diff
@@ -1045,6 +1027,83 @@ impl NetworkWorker {
...
+    /// The function is used while `manage_in_connections()` if the current
+    /// node reach the number of maximum in connection.
+    ///
+    /// Timeout if the connection as not been done really quickly resolved
+    /// but doesn't throw the timeout error.
+    ///
+    /// Return an error ONLY if we get a `NetworkError`.
+    ///
+    /// ```bash
+    /// In connection node            Curr node
```
yvan-sraka, line 1040: :heart:

yvan-sraka (approved):
LGTM

massa-network/src/network_worker.rs
```diff
@@ -1005,36 +974,49 @@ impl NetworkWorker {
...
+                        cur_connection_id.0 += 1;
+                        self.active_connections
+                            .insert(connection_id, (remote_addr.ip(), false));
+                        self.manage_successfull_connection(connection_id, reader, writer)?;
+                    }
+                    Err(NetworkError::PeerConnectionError(
+                        NetworkConnectionErrorType::MaxPeersConnectionReached(_),
+                    )) => {
+                        self.send_peer_list_in_handshake(reader, writer, remote_addr)
+                            .await?
```
damip, line 1014: this will make the whole program crash if the peer list fails ton send, which will happen very often

yvan-sraka (dismissed):

massa-network/src/network_worker.rs
```diff
@@ -1045,6 +1027,83 @@ impl NetworkWorker {
...
+    ///
+    /// In the `symetric read & write` the current node simulate a handshake managed
+    /// by the *connection node* in `HandshakeWorker::run()`, the current node
+    /// send a ListPeer as a message.
+    async fn send_peer_list_in_handshake(
+        &self,
+        reader: ReadHalf,
+        writer: WriteHalf,
+        remote_addr: SocketAddr,
+    ) -> Result<(), NetworkError> {
```
damip, line 1059: either:
* return an error in both cases (timeout AND read/write failure) but ignore it (or just print a debug message) in the caller (to not make the whole program crash !!!) 
* or do not return anything here and ignore errors (maybe with a debug message) within the send_peer_list_in_handshake function, and rename the function `try_send_peer_list_in_handshake` to make it clear that it might fail silently


I vote for solution 1

damip (changes requested):

massa-network/src/network_worker.rs
```diff
@@ -1005,36 +974,49 @@ impl NetworkWorker {
...
+                        });
+                        cur_connection_id.0 += 1;
+                        self.active_connections
+                            .insert(connection_id, (remote_addr.ip(), false));
+                        self.manage_successfull_connection(connection_id, reader, writer)?;
+                    }
+                    Err(NetworkError::PeerConnectionError(
+                        NetworkConnectionErrorType::MaxPeersConnectionReached(_),
+                    )) => {
+                        self.send_peer_list_in_handshake(reader, writer, remote_addr)
```
damip, line 1013: this function will block the main loop during the await ! let's spawn a separate thread

damip (changes requested):

massa-network/src/network_worker.rs
adrien-zinger: fixed!

massa-network/src/network_worker.rs
adrien-zinger: I made a proposal

massa-network/src/network_worker.rs
adrien-zinger: fixed

damip (approved):

damip:
@adrien-zinger  LGTM when conflicts are resolved

massa-network/src/handshake_worker.rs
```diff
@@ -39,35 +41,56 @@ pub struct HandshakeWorker {
...
+    /// Used for incomming and outgoing connections.
+    /// It will spawn a new future with an HandshakeWorker from the given `reader`
+    /// and `writer` from your current node to the distant `connectionId`
+    ///
     /// # Arguments
     /// * socket_reader: receives data.
     /// * socket_writer: sends data.
     /// * self_node_id: our node id.
     /// * private_key : our private key.
     /// * timeout_duration: after timeout_duration millis, the handshake attempt is dropped.
```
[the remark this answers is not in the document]
damip: @adrien-zinger  can you add it in this PR ?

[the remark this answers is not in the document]
adrien-zinger: yep done!

AureliaDolo (approved):

damip (approved):

damip:
On last comment: if the main loop quits, wait for all queued tasks to finish

damip (changes requested):
On last comment: if the main loop quits, wait for all queued tasks to finish after the loop

damip (approved):

yvan-sraka (approved):

damip:
bors merge

yvan-sraka:
bors merge

PR #2233, opened 2022-02-01
Add shared storage

### LINKED ISSUE #2178: Shared "storage" library to avoid block copies
Part of https://github.com/massalabs/massa/issues/2162

Can we start with getting an overview of current "block copies"?

Also, how much do we need to copy eventually anyway, for example due to outgoing network messages?

gterzian:
Close #2178


gterzian:
@massalabs/core-team I've started trying to use a "shared storage" within consensus, and looking at how it could be used in other components(like API), and I'm thinking perhaps we need to share more information than just a block, such as for example the map of `BlockStatus` currently owned by `consensus`. 

The reason being that something like a `Block` in and of itself doesn't contain much information. If we only share that, other components would still have to query `consensus` in order to use the block, I think.

Let me know what you think, I'll give it a try tomorrow...

AureliaDolo:
@gterzian In my opinion in storage we should store every information we may reuse, especially the block status. Maybe we could do something like an [ECS](https://docs.rs/ecs/latest/ecs/) ?

AureliaDolo (commented):
Do you plan to use the shared storage with protocol and network in this pr ?

gterzian:
@adrien-zinger If you want to help today, you could continue the work of my last commit(I think it explains itself), there are still a lot of similar errors to fix:
```
error[E0609]: no field `header` on type `massa_models::BlockId`
    --> massa-graph/src/block_graph.rs:3001:44
     |
3001 |                         [final_block.block.header.content.slot.thread as usize] = (
     |                                            ^^^^^^ unknown field
     |
     = note: available fields are: `0`

error[E0609]: no field `header` on type `massa_models::BlockId`
    --> massa-graph/src/block_graph.rs:3003:43
     |
3003 |                         final_block.block.header.content.slot.period,
     |                                           ^^^^^^ unknown field
     |
     = note: available fields are: `0`

Some errors have detailed explanations: E0063, E0308, E0609.
For more information about an error, try `rustc --explain E0063`.
error: could not compile `massa_graph` due to 47 previous errors
```

Basically, if you need to use the `header`(or something else on the block), it should use the block from storage. For simple things like `Slot` we can actually add them directly on `ActiveBlock` I think. 

I'll pick things up tomorrow...

massa-graph/src/export_active_block.rs
```diff
@@ -58,7 +58,7 @@ impl TryFrom<ExportActiveBlock> for ActiveBlock {
             a_block.block.addresses_to_endorsements(&endorsement_ids)?;
         Ok(ActiveBlock {
             creator_address: Address::from_public_key(&a_block.block.header.content.creator),
-            block: a_block.block,
+            block: a_block.block.header.compute_block_id()?,
```
damip, line 61: that's a very heavy computation (hashing the whole block)

gterzian: We can remove it, by instead of using `try_into`, simply creating the block in block graph when bootstrapping, where the id is already available. This is more of a quick fix for the compilation.

gterzian:
Next step: removing the block from `to_propagate`, and instead trying to use shared storage in network/protocol.

gterzian:
Added some basic intergration into `protocol` and `network`, as discussed with @damip will next look into storing the serialized format and reusing it in the `NodeWorker`.

gterzian:
@damip Could you please take a look at the last three commits, and let me know what you think?

massa-network/src/node_worker.rs
```diff
@@ -179,17 +179,47 @@ impl NodeWorker {
...
+        let storage = self.storage.clone();
         let node_writer_handle = tokio::spawn(async move {
             loop {
                 match writer_command_rx.recv().await {
                     Some(to_send) => {
                         let bytes_vec: Vec<u8> = match to_send {
                             ToSend::Msg(msg) => msg.to_bytes_compact().unwrap(),
                             ToSend::Block(block_id) => {
-                                break;
+                                // TODO: precisely allocate?
```
AureliaDolo, line 190: we know a block is at most max_block_size. And when all blocks will be full it will be exactly that.

massa-network/src/node_worker.rs
```diff
@@ -179,17 +179,47 @@ impl NodeWorker {
...
+                                    let serialized = stored_block.block.to_bytes_compact()?;
+                                    res.extend(&serialized);
+                                    stored_block.serialized = Some(serialized);
+                                }
+
+                                res
                             }
                             ToSend::Header(block_id) => {
-                                break;
+                                // TODO: precisely allocate?
```
AureliaDolo, line 207: We don't know the size of the header, but it's computed each time we create a block, maybe event if we don't know its size it may be worth to have at least some heuristics.

gterzian:
I was thinking of looking into caching of objects as they are de-serialized when coming in from over the network, and I actually think it might not be a good idea since at that point we don't know much about their validity.

gterzian:
The last commit shows how we could store the serialized form of a block as it comes in over the network, and reuse it when sending it out later. Does not include the various other processing steps mentioned at https://github.com/massalabs/massa/issues/2264#issuecomment-1033679531.

@massalabs/core-team

massa-models/src/storage.rs
```diff
@@ -0,0 +1,70 @@
...
+}
+
+#[derive(Clone, Default)]
+pub struct Storage {
+    blocks: Arc<RwLock<Map<BlockId, Arc<RwLock<StoredBlock>>>>>,
+    operations: Arc<RwLock<Map<OperationId, Arc<RwLock<Operation>>>>>,
+    endorsements: Arc<RwLock<Map<EndorsementId, Arc<RwLock<Endorsement>>>>>,
+}
+
+impl Storage {
```
yvan-sraka, line 19: I would love to rather have a something like a:
```rust
impl Storage for Block {
    pub fn store(&self, id: BlockId, item: Block, serialized: Vec<u8>) { todo!() }
    pub fn retrieve_block(&self, id: &BlockId) -> Option<Arc<RwLock<StoredBlock>>> { todo!() }
}
```
But this completely be part of a follow-up :slightly_smiling_face:!

massa-models/src/storage.rs
```diff
@@ -0,0 +1,70 @@
+use crate::prehash::Map;
+use crate::{Block, BlockId, Endorsement, EndorsementId, Operation, OperationId};
+use parking_lot::RwLock;
+use std::sync::Arc;
+
+pub struct StoredBlock {
```
yvan-sraka, line 6: Can we, rather than having a `StoredBlock` struct that wrap a block with his serialized representation, have something more like a serialization cache in a separate structure, that would be updated when a block is updated!

gterzian: Could you please elaborate on "separate structure" and "updated when a block is updated"?

Note that the point of having one struct wrap both the block and the serialized representation is to have them both covered by the same lock.

yvan-sraka: Yes I was thinking about having an `HashMap<BlockId, SerializedBlock>` but yes a unique structure make more sense behind a lock mechanism you're right!

massa-graph/src/export_active_block.rs
damip: I really think objects should eventually carry their own ID

massa-models/src/storage.rs
damip: Having a single map allows for better consistency without having to use more locks. I vote for having either a StoredBlock wrapper like was done here or having the Block structure itself carry its serialized representation so we don't bother with multiple structures. As you prefer

gterzian:
For clarity, this is the "main" branch of the shared storage project. Also includes the integration into consensus/graph.

massa-graph/src/export_active_block.rs
```diff
@@ -32,67 +32,6 @@ pub struct ExportActiveBlock {
...
-            dependencies: a_block.dependencies.clone(),
-            is_final: a_block.is_final,
-            block_ledger_changes: a_block.block_ledger_changes.clone(),
-            roll_updates: a_block.roll_updates.clone(),
-            production_events: a_block.production_events.clone(),
-        }
-    }
-}
-
-impl TryFrom<ExportActiveBlock> for ActiveBlock {
```
gterzian, line 50: Note: removed those because shared storage is not easily available from here, and also because there weren't used in more than one place anyway, so this is now simply code in `block_graph`, where shared storage is available.

massa-consensus-worker/src/consensus_worker.rs
```diff
@@ -144,18 +144,8 @@ impl ConsensusWorker {
...
-                .into_iter()
-                .filter_map(|block_id| {
-                    block_db
-                        .get_active_block(&block_id)
-                        .map(|a_block| (block_id, a_block.block.clone()))
-                })
-                .collect(),
-        );
+
+        // TODO: use shared-storage.
```
AurelienFT, line 148: If we want to merge this is that a problem that this is still a todo and the past behaviour has been removed ? Or it's ok because you store here : https://github.com/massalabs/massa/pull/2233/files#diff-990c58de6c3e1194da0ee5db7bcac14cdd0816ff8682805753c594f83e304511R589 ?

AurelienFT (commented):
I'm new to this part of the codebase so I'm not able to make a great review but the changes are readable

massa-consensus-worker/src/consensus_worker.rs
gterzian: Good observation. For now we'll keep this branch open, and makes changes to it, as well as try to merge the other PRs into it. 

This particular TODO is addressed in https://github.com/massalabs/massa/pull/2322

massa-models/src/active_block.rs
```diff
@@ -2,16 +2,16 @@ use crate::{
...
 
 /// Block that was checked as final, with some useful precomputed data
 #[derive(Debug, Clone, serde::Serialize, serde::Deserialize)]
 pub struct ActiveBlock {
     /// The creator's address
     pub creator_address: Address,
-    /// The block itself, as it was created
-    pub block: Block,
+    /// The id of the block
+    pub block: BlockId,
```
AurelienFT, line 14: I think this field should be renamed to block_id even if it implies a lot of changes in the codebase.

massa-models/src/storage.rs
```diff
@@ -0,0 +1,32 @@
...
+impl Storage {
+    pub fn store_block(&self, block_id: BlockId, block: Block, serialized: Vec<u8>) {
+        // TODO: first check, and allow for, an already stored header for the block.
+        let stored_block = StoredBlock { block, serialized };
+        let to_store = Arc::new(RwLock::new(stored_block));
+        let mut blocks = self.blocks.write();
+        blocks.insert(block_id, to_store);
+    }
+
+    pub fn retrieve_block(&self, block_id: &BlockId) -> Option<Arc<RwLock<StoredBlock>>> {
```
AurelienFT, line 25: I have some concerns about the return type of this. If we want to make this interface be easily changeable the fact that each time we use this function and we have to make a `.read()` after is not generic. I haven't think if there is a simple possibility to make the change here but maybe worth some discussions.

AurelienFT:
Also wanted to add that it could be cool to have logs in the retrieve and store block at info level for example so that if we want to debug it will help and will not slow down the code because it's a level that we don't activate by default.

massa-graph/src/export_active_block.rs
```diff
@@ -13,7 +13,7 @@ use serde::{Deserialize, Serialize};
 /// Fields that can be easily recomputed were left out
 #[derive(Debug, Clone, Serialize, Deserialize)]
 pub struct ExportActiveBlock {
-    /// The block itself, as it was created
+    /// The id of the block.
```
AurelienFT, line 16: The comment has been changed but it still hold the whole block.

massa-graph/src/export_active_block.rs
AurelienFT: It's needed in the tests. Is that possible to have it maybe with a cfg tests ? Also I don't understand why you need shared storage to make a conversion ?

AurelienFT: I have made the modification on my PR for tests because I needed it you can check and merge it here even if it's not finish so that you have the blockId instead of block here. PR : https://github.com/massalabs/massa/pull/2383 @gterzian

massa-graph/src/export_active_block.rs
AurelienFT: Changed here : https://github.com/massalabs/massa/pull/2233#discussion_r821700545

massa-graph/src/export_active_block.rs
gterzian: great, thanks

massa-models/src/storage.rs
gterzian: Regarding our discussion in https://github.com/massalabs/massa/issues/2344, the change would be at the level of the top-level map, not the individual block. So that will not affect it. 

Did you have another potential problem in mind?

AurelienFT: Oh yeah. My bad It's ok then :)

massa-graph/src/block_graph.rs
```diff
@@ -451,8 +443,9 @@ impl BlockGraph {
                     addresses_to_operations: Map::with_capacity_and_hasher(0, BuildMap::default()),
                     roll_updates: RollUpdates::default(), // no roll updates in genesis blocks
                     production_events: vec![],
-                    block,
+                    block_id,
```
AurelienFT, line 446: As you only pass the id you shouldn't store the block just above ?

gterzian: Correct, thanks

massa-graph/src/block_graph.rs
```diff
@@ -560,7 +604,21 @@ impl BlockGraph {
             Map::with_capacity_and_hasher(required_active_blocks.len(), BuildMap::default());
         for b_id in required_active_blocks {
             if let Some(BlockStatus::Active(a_block)) = self.block_statuses.get(&b_id) {
-                active_blocks.insert(b_id, ExportActiveBlock::from(&**a_block));
+                let block = self.storage.retrieve_block(&b_id).unwrap();
```
AurelienFT, line 607: This function is returning an error maybe consider map the `None` to an Error with `ok_or` instead of unwrap.

gterzian: Yes, thanks. I will address this and others in a separate PR that will close https://github.com/massalabs/massa/issues/2381

AurelienFT:
I have pushed directly here the changes for the tests of protocol so that we can have CI and will be here for the merge of main and the conflict resolution.

gterzian:
> Maybe there is something that I miss and it will be cool to make a little voice point after the discord meeting and I'll write note in a documentation somewhere of your explanations

Reposting the answer here @AurelienFT. Any questions left?

We receive and deserialize the data from the network in `binder.rs`, at https://github.com/massalabs/massa/pull/2233/files#diff-a6b49818610831ff127e7e1e9995f87490707e9f89d5ba43aea8ff86964cb024R163

And we store the block just before sending the `ReceivedBlock` event in the `node_worker.rs`  at https://github.com/massalabs/massa/pull/2233/files#diff-eb8e909fbe8730ebc4d6bdb49e041da621180bc60ccdd6a74aa2b610f7393a7aR314

AurelienFT:
> > Maybe there is something that I miss and it will be cool to make a little voice point after the discord meeting and I'll write note in a documentation somewhere of your explanations
> 
> Reposting the answer here @AurelienFT. Any questions left?
> 
> We receive and deserialize the data from the network in `binder.rs`, at https://github.com/massalabs/massa/pull/2233/files#diff-a6b49818610831ff127e7e1e9995f87490707e9f89d5ba43aea8ff86964cb024R163
> 
> And we store the block just before sending the `ReceivedBlock` event in the `node_worker.rs` at https://github.com/massalabs/massa/pull/2233/files#diff-eb8e909fbe8730ebc4d6bdb49e041da621180bc60ccdd6a74aa2b610f7393a7aR314

No it's clear thank you !

gterzian:
I am in the process of rebasing... cc @AurelienFT

gterzian:
Still rebasing...

PR #2251, opened 2022-02-07
Path completion on the client REPL.

### LINKED ISSUE #2236: Path auto-completion in client
It would be handy to have path auto completion with tab in the client, especially for SC :)

Do you think it's a lot of work @yvan-sraka ?

AurelienFT:
Fix #2236


AurelienFT:
Hello @yvan-sraka why you have put the PR back in draft ? Is there anything that you want me to change/add ?

yvan-sraka (approved):
LGTM

massa-client/src/repl.rs
```diff
@@ -135,20 +136,39 @@ impl Completion for CommandCompletion {
...
-            .collect();
-        if !suggestions.is_empty() {
-            println!();
-            for suggestion in &suggestions {
-                println!("{}", style(suggestion).dim());
+        if input.contains(' ') {
+            let mut args: Vec<&str> = input.split(' ').collect();
+            let mut default_path = "./";
+            let path_to_complete = args.last_mut().unwrap_or_else(|| &mut default_path);
+            let suggestions: Vec<String> = glob(&(path_to_complete.to_owned() + "*"))
```
damip, line 143: I see many unwraps here, in which cases would the client crash ?

damip: We don't want the client to crash when the user inputs/does invalid stuff, it makes for terrible UX.
Let's just to tell them why the command failed and stay in the REPL.
More generally, we need to be *very* careful when using things that might panic. And a good practice is to leave a comment besides every except/unwrap etc... to explain why it won't happen, and if it happens, why it is absolutely unrecoverable anyways

AurelienFT: Yes I have changed to handles the crash already I just didn't propagate the error to the client but will add it !

massa-client/src/repl.rs
```diff
@@ -135,20 +136,48 @@ impl Completion for CommandCompletion {
...
+            let mut default_path = "./";
+            let path_to_complete = args.last_mut().unwrap_or(&mut default_path);
+            if let Ok(paths) = glob(&(path_to_complete.to_owned() + "*")) {
+                let suggestions: Vec<String> = paths
+                    .filter_map(|x| {
+                        if let Ok(path) = x {
+                            Some(path.display().to_string())
+                        } else {
+                            None
+                        }
```
AureliaDolo, line 150: Here you maybe could use `x.map(|path| path.display().to_string())` for less verbosity

massa-client/src/repl.rs
```diff
@@ -135,20 +136,50 @@ impl Completion for CommandCompletion {
...
+                    for path in &suggestions {
+                        println!("{}", style(path).dim())
+                    }
+                    *path_to_complete =
+                        longest_common_prefix(suggestions.iter().map(|s| &s[..]).collect());
+                }
+                Some(args.join(" "))
+            } else {
+                println!();
+                println!("Bad syntax for path completion.");
```
AureliaDolo, line 164: I'm not sure how useful this line is for the end user

AurelienFT: It refer to the discussion above with damir but I don't know what to put as error message. Do you an idea ?

massa-client/src/repl.rs
AurelienFT: Agree !

massa-client/src/repl.rs
AureliaDolo: I'd show nothing. Damir's concern was more about the client not crashing, not really having a nice error message. In a shell when path auto-completion fails, nothing happens ? I'd go for that. Maybe In a follow up, if you want to be extra fancy you could show the available file paths in the current directory even if they don't match after tab is pressed a second time.

AurelienFT: Ah ok so I remove it !

AureliaDolo (approved):

damip (commented):
pretty cool :)

2 points tho:

* is it portable ? does it work on windows ? (I known the hardcoded "./" is not a windows thing for example)
* it only auto-completes paths in a subfolder. If you type "~/" then press tab, it won't autocomplete like other unix tools do

AurelienFT:
The package that I'm using "expanduser" don't work on windows. Will change it.

yvan-sraka (approved):

AurelienFT:
I have updated the PR. Compile and works well on Windows. The "./" is recognized by glob package and converted automatically.

damip (approved):

AurelienFT:
bors merge

AurelienFT:
bors merge

PR #2255, opened 2022-02-07
prevent socket queuing in bootstrap

damip:
Drop connections if there are too many in queue, thus preventing bootstrap saturation under heavy load.

### DIFF
--- a/massa-bootstrap/src/lib.rs
+++ b/massa-bootstrap/src/lib.rs
@@ -368,7 +368,7 @@ impl BootstrapServer {
                 }
 
                 // listener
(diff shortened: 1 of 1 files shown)


yvan-sraka:
Wow, it's like just moving 2 character are changing a lot the behavior of the whole app, this make feel like I did not understand all this really well...

yvan-sraka (approved):

AureliaDolo (commented):
How could we test that ?

damip:
@AureliaDolo I think this one is hard to test deterministically. It only happens under certain OS-dependent circumstances. It's a strange one. But now it works

damip:
bors merge

PR #2265, opened 2022-02-09
Add read only calls to client.

### LINKED ISSUE #2246: Add readonly calls to client and API wiki
Readonly calls are available in the API but have not been documented in the wiki, and are not available in the client.

Make them available in the client, and document the API

AureliaDolo:
Fix #2246


damip:
let's display the resulting events directly in the client :)

yvan-sraka (approved):

PR #2285, opened 2022-02-13
improve block propagation algo

damip:
There were inaccuracies, especially in the insert_wanted_block function

### DIFF
--- a/massa-protocol-worker/src/protocol_worker.rs
+++ b/massa-protocol-worker/src/protocol_worker.rs
@@ -159,6 +159,21 @@ mod nodeinfo {
             self.known_blocks.get(block_id)
         }
 
+        /// Remove the oldest items from known_blocks
+        /// to ensure it contains at most max_node_known_blocks_size items.
+        /// This algorithm is optimized for cases where there are no more than a couple excess items, ideally just one.
+        fn remove_excess_known_blocks(&mut self, max_node_known_blocks_size: usize) {
+            while self.known_blocks.len() > max_node_known_blocks_size {
+                // remove oldest item
+                let (&h, _) = self
+                    .known_blocks
+                    .iter()
+                    .min_by_key(|(h, (_, t))| (*t, *h))
+                    .unwrap(); // never None because is the collection is empty, while loop isn't executed.
+                self.known_blocks.remove(&h);
+            }
+        }
+
         /// Insert knowledge of a list of blocks in NodeInfo
         ///
         /// ## Arguments
@@ -177,15 +192,7 @@ mod nodeinfo {
             for block_id in block_ids {
                 self.known_blocks.insert(*block_id, (val, instant));
             }
-            while self.known_blocks.len() > max_node_known_blocks_size {
-                // remove oldest item
-                let (&h, _) = self
-                    .known_blocks
-                    .iter()
-                    .min_by_key(|(h, (_, t))| (*t, *h))
-                    .unwrap(); // never None because is the collection is empty, while loop isn't executed.
-                self.known_blocks.remove(&h);
-            }
+            self.remove_excess_known_blocks(max_node_known_blocks_size);
         }
 
         pub fn insert_known_endorsements(
--- a/massa-protocol-worker/src/protocol_worker.rs
+++ b/massa-protocol-worker/src/protocol_worker.rs
@@ -226,7 +226,7 @@ mod nodeinfo {
             self.known_operations.contains(op)
         }
 
-        /// Remove the oldest items from known_blocks
+        /// Remove the oldest items from wanted_blocks
         /// to ensure it contains at most max_node_wanted_blocks_size items.
         /// This algorithm is optimized for cases where there are no more than a couple excess items, ideally just one.
         fn remove_excess_wanted_blocks(&mut self, max_node_wanted_blocks_size: usize) {
--- a/massa-protocol-worker/src/protocol_worker.rs
+++ b/massa-protocol-worker/src/protocol_worker.rs
@@ -226,27 +226,30 @@ mod nodeinfo {
             self.known_operations.contains(op)
         }
 
-        /// insert a block in wanted list of a node.
-        /// Note that it also insert the block as a not known block for that node.
-        pub fn insert_wanted_blocks(
+        /// insert a block in the wanted list of a node.
+        /// Also lists the block as not known by the node
+        pub fn insert_wanted_block(
             &mut self,
             block_id: BlockId,
             max_node_wanted_blocks_size: usize,
+            max_node_known_blocks_size: usize,
         ) {
-            self.wanted_blocks.insert(block_id, Instant::now());
-            while self.known_blocks.len() > max_node_wanted_blocks_size {
+            let now = Instant::now();
+            self.wanted_blocks.insert(block_id, now);
+            while self.wanted_blocks.len() > max_node_wanted_blocks_size {
                 // remove oldest item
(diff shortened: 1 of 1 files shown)


massa-protocol-worker/src/protocol_worker.rs
```diff
@@ -226,27 +226,30 @@ mod nodeinfo {
...
+            while self.wanted_blocks.len() > max_node_wanted_blocks_size {
                 // remove oldest item
                 let (&h, _) = self
-                    .known_blocks
+                    .wanted_blocks
                     .iter()
                     .min_by_key(|(h, t)| (*t, *h))
                     .unwrap(); // never None because is the collection is empty, while loop isn't executed.
-                self.known_blocks.remove(&h);
+                self.wanted_blocks.remove(&h);
```
AureliaDolo, line 246: Maybe put that in a remove exceeding blocks method for clarity (and maybe not only here but also when ever this pattern is used)

AureliaDolo (approved):

massa-protocol-worker/src/protocol_worker.rs
```diff
@@ -226,27 +226,30 @@ mod nodeinfo {
...
             }
+            self.insert_known_blocks(&[block_id], false, now, max_node_known_blocks_size);
         }
 
-        /// If given node previously manifested it wanted given block.
-        pub fn contains_wanted_block(&mut self, block_id: &BlockId) -> bool {
+        /// returns wheteher a node wants a block, and if so, updates the timestamp of that info to now()
```
gterzian, line 251: typo wheteher => whether

massa-protocol-worker/src/protocol_worker.rs
```diff
@@ -226,27 +226,30 @@ mod nodeinfo {
...
                 let (&h, _) = self
-                    .known_blocks
+                    .wanted_blocks
                     .iter()
                     .min_by_key(|(h, t)| (*t, *h))
                     .unwrap(); // never None because is the collection is empty, while loop isn't executed.
-                self.known_blocks.remove(&h);
+                self.wanted_blocks.remove(&h);
             }
+            self.insert_known_blocks(&[block_id], false, now, max_node_known_blocks_size);
```
gterzian, line 248: A comment explaining why this is necessary would be helpful here.

gterzian (approved):
LGTM with nits

massa-protocol-worker/src/protocol_worker.rs
damip: done

massa-protocol-worker/src/protocol_worker.rs
damip: done

massa-protocol-worker/src/protocol_worker.rs
damip: done

massa-protocol-worker/src/protocol_worker.rs
AureliaDolo: I don't see the change you're talking about

gterzian (approved):

massa-protocol-worker/src/protocol_worker.rs
damip: look at the absolute code. Since the code extracted by github for the preview didn't change and just moved to a new function, the change is not visible in the preview

damip:
I corrected a typo

yvan-sraka (changes requested):
What's the improvement proposed by this PR? I understand it's about fixing some correctness issue (and not a performance one). Do you have an edge case that would pass which this version of the algorithm and failed with the previous one? Could you turn it into a test?

damip:
> What's the improvement proposed by this PR? I understand it's about fixing some correctness issue (and not a performance one). Do you have an edge case that would pass which this version of the algorithm and failed with the previous one? Could you turn it into a test?

Correctness is important ;)

Otherwise, 

> What's the improvement proposed by this PR? I understand it's about fixing some correctness issue (and not a performance one). Do you have an edge case that would pass which this version of the algorithm and failed with the previous one? Could you turn it into a test?

* correctness
* network optimization: not trying to ask for a block B from a node N if N has asked for this block recently (implying that they didn't have it)
* wanted_blocks was never cleared and only grew indefinitely until RAM exhaustion

AureliaDolo:
> * wanted_blocks was never cleared and only grew indefinitely until RAM exhaustion

It was cleared when we receive a block, we sent it to the node that want it and remove that block from the wanted list.

massa-protocol-worker/src/protocol_worker.rs
```diff
@@ -226,27 +226,42 @@ mod nodeinfo {
...
-        /// insert a block in wanted list of a node.
-        /// Note that it also insert the block as a not known block for that node.
-        pub fn insert_wanted_blocks(
-            &mut self,
-            block_id: BlockId,
-            max_node_wanted_blocks_size: usize,
-        ) {
-            self.wanted_blocks.insert(block_id, Instant::now());
-            while self.known_blocks.len() > max_node_wanted_blocks_size {
+        /// Remove the oldest items from known_blocks
```
AureliaDolo, line 229:
```suggestion
        /// Remove the oldest items from wanted_blocks
```

massa-protocol-worker/src/protocol_worker.rs
```diff
@@ -226,27 +226,42 @@ mod nodeinfo {
...
+        ) {
+            // Insert into known_blocks
+            let now = Instant::now();
+            self.wanted_blocks.insert(block_id, now);
+            self.remove_excess_wanted_blocks(max_node_wanted_blocks_size);
+
+            // If the node wants a block, it means that it doesn't have it.
+            // To avoid asking the node for this block in the meantime,
+            // mark the node as not knowing the block.
+            self.insert_known_blocks(&[block_id], false, now, max_node_known_blocks_size);
```
AureliaDolo, line 260: You don't prune known blocks ?

AureliaDolo (approved):

massa-protocol-worker/src/protocol_worker.rs
damip: the insert_known_blocks function does

damip: here I also separated remove_excess_known_blocks for more clarity

damip:
> > * wanted_blocks was never cleared and only grew indefinitely until RAM exhaustion
> 
> It was cleared when we receive a block, we sent it to the node that want it and remove that block from the wanted list.
> 
> I think the assumption was that as everything goes wanted blocks is by design emptying itself, whereas known blocks are growing indefinitely
> 
> I think this is mostly useful to avoid adversarial nodes flooding the wanted blocks with random hashes.


I think there is a confusion between the blocks that we want, and the blocks that a node wants.

```
        /// The blocks the node asked for.
        wanted_blocks: Map<BlockId, Instant>,
```

This is out of our control, the node can ask for millions of random block IDs all the time and flood our RAM

AureliaDolo (dismissed):

yvan-sraka:
Could you write a test that discriminate previous and current behavior to prevent any regression?

adrien-zinger (dismissed):

damip:
> Could you write a test that discriminate previous and current behavior to prevent any regression?

added a regression test

AureliaDolo (approved):

yvan-sraka (approved):

damip:
bors merge

PR #2287, opened 2022-02-14
Create `CONTRIBUTING.md`

yvan-sraka:


CONTRIBUTING.md
```diff
@@ -0,0 +1,138 @@
+# Code review checklist (TODO: should be copy-pasted in PR template)
+
+- [ ] code should be formatted using `rustfmt` (:warning: this also applied to code in macro like `tokio::select!` and `lazy_static!`
```
AureliaDolo, line 3: I would rather say something like : the code in macros is not formatted, do your best to reduce the code size in macro arguments.

CONTRIBUTING.md
```diff
@@ -0,0 +1,138 @@
...
+- [ ] code should be exempt of `clippy` lints (and function size should not exempt 50 lines)
```
AureliaDolo, line 4: Why 50 lines ? Should we refactor every thing right now ?How to enforce that ? (I won't count your function's lines)

CONTRIBUTING.md
```diff
@@ -0,0 +1,138 @@
...
+- [ ] code should not `panic!` an `Err` variant should be propagated and capture by top-level `async main() -> Result<()>`
```
AureliaDolo, line 5: Maybe add a link to a doc about how we manage errors. (I'll try to come back later and update this comment with a link)

AureliaDolo (commented):
For the community guidelines, even if it quite standard, I'll tag @Kryptopaid to have a quick look at that

AureliaDolo:
> Always look at the consequences of what you return, access, modify and call. More generally, always be aware about how the function you are writing or reviewing interacts with its context.

This seems like a great idea, but in my opinion expecting that is la porte ouverte à toutes les fenêtres 🪟 . You can't guarantee that even reading the code you won't miss something. Moreover as the architecture is not quite standard and tough to understand it will take a lot of time, that will be shortened as deadlines approach. 😅  And again in my opinion, if we had enough separation of concern, this wouldn't be an issue. 🥲

adrien-zinger:
> Threre are much more important and specific points BOTH FOR WRITING AND REVIEW:
> 
>     * every time some code line may panic, there should be a comment describing why it will not panic, or why it would be an unrecoverable error anyway
> 
>     * in an async context, there shouldn't be any panic (except at top task level) because stack traces are unusable: explicit error propagation is always required
> 
>     * wherever you are in the code, returning an error from a given function might crash the calling function and maybe the whole program. Always look at the consequences of what you return, access, modify and call. More generally, always be aware about how the function you are writing or reviewing interacts with its context.
> 
>     * user and network input is UNTRUSTED: always parse/validate it, and do not crash the program if validation fails. The program should be robust to attacks coming from outside input.

This is more the symptom of a sequence of bad practices in the development in my opinion, at least these are the consequences of a bad design. We should better describe good practices to make errors and bug almost impossible.

For example, the 3rd comment just means "be a good developer, read all the project, and understand it". We can change that into a guide of *what should be written in the function documentation* I'll describe what I mean later, a *size limit of a routine* for the separation of contexts that make the code more readable. Make sure that adding a functionality in a feature won't move the all code base (force if guards and other design patterns associated).

## Routines size and documentation, 1, 2 and 3
In my opinion, and that's the Epita's school to do it, we should write short functions and trust subroutines. The code documentation should be mainly upside the function (some notes obviously can be inside the function when they are an important complexity). And the documentation should contain:

- What the function does with the input/output parameters (no need to format that)
- Who use it eventually (if it's internal, for example, to keep impacts in mind)
- What it returns and, if it's an error, how it should be managed and by which part of the code.

We could also aggregate for the massa project:
- (Where) If the current context is in the min loop or not
- (How) A tutorial eventually (I'm thinking about the main loop of the execution module)

All of that is described here https://github.com/rust-dev-tools/fmt-rfcs/tree/master/guide, but that would be too long t read if we want to code fast. Fortunately, _clippy_ can be configured and simplify also the review!! It can check a function overflow, missing documentations and cognitive complexity checks.

With that, if I develop a feature, I'm almost sure that I'll not destroy everything with a simple `?` :rofl: 

## Robustiffy

For the 4th, obviously, we can simply avoid usage of `unwrap()` and `panic!` outside unit tests. We can also ask to verify if there is a possible flood attack and develop the feature with a threshold. I can't say _"think of every possible bugs/attack"_ that's too difficult to see, for me and for other devs..

## Divers

We can also say to always get a value from a configuration or a setting and never write a magic value.
- Use settings for a private variable and config for a .toml configurable one.
- Write rust adapter `From` when possible and profit of `into()`
- Check documentation and wiki before development a concern
- No warnings (check clippy at least)

damip:
> > Always look at the consequences of what you return, access, modify and call. More generally, always be aware about how the function you are writing or reviewing interacts with its context.
> 
> This seems like a great idea, but in my opinion expecting that is la porte ouverte à toutes les fenêtres window . You can't guarantee that even reading the code you won't miss something. Moreover as the architecture is not quite standard and tough to understand it will take a lot of time, that will be shortened as deadlines approach. sweat_smile And again in my opinion, if we had enough separation of concern, this wouldn't be an issue. smiling_face_with_tear

If separation of concerns is the concern, it should be a task in the big refactoring :)

I somehow believe that devs should be aware of what they are doing, at least locally (eg. what are the immediate surroundings of the function they are creating, and how it fits in there), as well as a more vague idea of how the module works, and an even vaguer idea of how the project works. Otherwise they will rely solely on the issue specification, which is *never* perfect and would take more time to write than the actual code solving it if it doesn't leave anything to the dev's interpretation.
This doesn't mean that specs shouldn't be clearly written and sufficiently detailed !

damip:
My main concern is that we have seen flaws in both the code and review process, and we need to think in the following way: "what could have prevented me from writing/LGTM-ing this"

yvan-sraka:
We could also document somewhere that we require signed commits, linear history and two review (and review by code owners) by PR, and that new commits dismiss reviews?! + Checklist on how to write a good commit message / how versioning works?

CONTRIBUTING.md
yvan-sraka: @adrien-zinger discover that we can configure that into `clippy` :sparkles:

AureliaDolo:
And as I see it hasn't been checked for this testnet, make sure that changes in the api are tracked and that the wiki is updated when a new testnet is released. 

@yvan-sraka is there a way to enforce that ?

qdrn:
I'd say we can merge this and iterate over time! Thanks Yvan.

CONTRIBUTING.md
adrien-zinger:
```suggestion
- [ ] code should be formatted using `rustfmt` (:warning: this also applied to code in macro like `tokio::select!` and `lazy_static!`
- [ ] reduce the code size in macro arguments and in matching patterns
```

CONTRIBUTING.md
adrien-zinger: Yes we can add that in clippy and fix warnings in follow ups. Moreover, can we speak about that in a meeting and finally merge that? :rofl:

CONTRIBUTING.md
adrien-zinger:
```suggestion
- [ ] code should not `panic!` an `Err` variant should be propagated and capture by top-level `async main() -> Result<()>`

In addition, as a reviewer, if you can make a code suggestion to accelerate the process it'll be greatly appreciated!
```

(so meta)

CONTRIBUTING.md
qdrn: @adrien-zinger I've taken the liberty to edit you suggestion as it contained only typos.

CONTRIBUTING.md
qdrn: Same remark

CONTRIBUTING.md
```diff
@@ -0,0 +1,139 @@
...
+We as members, contributors, and leaders pledge to make participation in our
+community a harassment-free experience for everyone, regardless of age, body
+size, visible or invisible disability, ethnicity, sex characteristics, gender
+identity and expression, level of experience, education, socio-economic status,
+nationality, personal appearance, race, caste, color, religion, or sexual
+identity and orientation.
+
+We pledge to act and interact in ways that contribute to an open, welcoming,
+diverse, inclusive, and healthy community.
+
```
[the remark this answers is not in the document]
qdrn: Did you take inspiration from another repo or you just wrote that yourself ? If we can write something standard here it's better.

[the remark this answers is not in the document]
aoudiamoncef: Hi,
You could use [cla-bot](https://github.com/marketplace/cla-bot-cla-automation) for CLA automation

CONTRIBUTING.md
AurelienFT: @yvan-sraka I think you can lick the upper button to merge it if you are agree !

CONTRIBUTING.md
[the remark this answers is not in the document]
AurelienFT: Can we merge it even without contribution agreement as there is no before it will change nothing if we still don't have it ?

.github/PULL_REQUEST_TEMPLATE/pull_request_template.md
```diff
@@ -0,0 +1,9 @@
+## Describe your changes
+
+## Issue ticket number and link
+
+## Checklist before requesting a review
+- [ ] I have performed a self-review of my code
+- [ ] Code is formatted using `rustfmt` (:warning: this also applies to code in macro like `tokio::select!` and `lazy_static!`)
+- [ ] Reduce the code size in macro arguments and in matching patterns
+- [ ] Code should be exempt of `clippy` lints (and function size should not exempt 50 lines)
```
AurelienFT, line 9:
```suggestion
- [ ] Code should be exempt of `clippy` lints (and function size should not exempt 50 lines)
- [ ] I have used a spell checker on my code and comments.
```

CONTRIBUTING.md
```diff
@@ -0,0 +1,141 @@
+# Code review checklist
+
+- [ ] code should be formatted using `rustfmt` (:warning: this also applies to code in macro like `tokio::select!` and `lazy_static!`)
+- [ ] reduce the code size in macro arguments and in matching patterns
+- [ ] code should be exempt of `clippy` lints (and function size should not exempt 50 lines)
+- [ ] code should not `panic!` an `Err` variant should be propagated and captured by top-level `async main() -> Result<()>`
```
AurelienFT, line 6:
```suggestion
- [ ] code should not `panic!` an `Err` variant should be propagated and captured by top-level `async main() -> Result<()>`
- [ ] I have used a spell checker on my code and comments.
```

AurelienFT (commented):
I think it could be cool to add a line about spell checking. I have made a proposal for this.

AurelienFT (approved):
The document helps new developers and ensure quality for our code and then answer to his primary goal.

I don't think we should merge with only my approve and should wait for one of the founders. @damip @sebastien-forestier @qdrn

PR #2290, opened 2022-02-14
Add docs and fix latest warnings

adrien-zinger:
Add documentation for the worker execution, to complete later, it will help me to do the execution tests later.
(This commit also fix new/last warnings in the code.)

@damip I'm working on the execution tests and walking through the code to understand I profit of that PR to add some documentation on it, one line is still a todo (`vm.rs` line: 500) if you can help me to understand please :slightly_smiling_face: 

@gterzian we spoke about documentation today, I wanted to have your review because you have worked on the execution module and your comments on the execution worker can be really useful :+1:


damip:
@adrien-zinger 

step_history contains the execution results of CONSECUTIVE active slots that have been executed (front = oldest, back = most recent). Therefore, if the VM gets a new active step (slot) to execute, and that slot number breaks the slot-sorting of step_history when pushed back, it means that there was a history rewrite (eg. clique change) and the step_history needs to be truncated just before the incoming slot before executing it and appending its result to step_history.

yvan-sraka (approved):

yvan-sraka (approved):

adrien-zinger:
Duplicated, look at that https://github.com/massalabs/massa/pull/2296 it's better

PR #2292, opened 2022-02-15
correct crash on node close

damip:
Node workers might close just before we ask them to do so on cleanup => ignore Stop send errors.


adrien-zinger (approved):
Really nice comment!! :love_you_gesture:

AureliaDolo (approved):

PR #2294, opened 2022-02-15
document Amout

damip:


massa-models/src/amount.rs
```diff
@@ -9,32 +9,47 @@ use std::str::FromStr;
 
 pub const AMOUNT_ZERO: Amount = Amount::from_raw(0);
```
yvan-sraka, line 10: Why? How? Where? Is that constant used?

massa-models/src/amount.rs
```diff
@@ -9,32 +9,47 @@ use std::str::FromStr;
...
+    /// obtains the underlying raw u64 representation
+    /// Warning: do not use this unless you know what you are doing
+    /// because the raw value does not take the AMOUNT_DECIMAL_FACTOR into account
     pub fn to_raw(&self) -> u64 {
         self.0
     }
 
+    /// constructs an Amount from the underlying raw u64 representation
+    /// Warning: do not use this unless you know what you are doing
+    /// because the raw value does not take the AMOUNT_DECIMAL_FACTOR into account
```
AureliaDolo, line 30: Maybe add some thing along the lines of "prefer from_string whenever you can"

massa-models/src/amount.rs
```diff
@@ -94,6 +113,7 @@ impl Amount {
     }
 }
 
+/// display an Amount in decimal string form (like "10.33")
```
AureliaDolo, line 116: Maybe add some code examples

massa-models/src/amount.rs
```diff
@@ -105,6 +125,9 @@ impl fmt::Display for Amount {
     }
 }
 
+/// build an Amount from decimal string form (like "10.33")
+/// note that this will fail if the string format is invalid
+/// or if the conversion would cause an overflow or precision loss
```
AureliaDolo, line 130: maybe add some code examples

massa-models/src/amount.rs
damip: we could replace it with a zero() function constructor if you wish (like here https://docs.rs/num/latest/num/traits/trait.Zero.html#tymethod.zero )... but the result is kinda the same

massa-models/src/amount.rs
damip: done

massa-models/src/amount.rs
damip: done

AureliaDolo (dismissed):

massa-models/src/amount.rs
damip: Note that it could also be defined as a Default. But when you want a zero somewhere, I find using AMOUNT_ZERO much more explicit and readable than "Default::default()".

adrien-zinger (dismissed):
For me that's great

massa-models/src/amount.rs
damip: Maybe a const function Amount::zero() would actually be nicer than a random constant in the config

adrien-zinger: good idea, with `const fn` !

damip:
bors merge

massa-models/src/amount.rs
adrien-zinger: I close that and resolve it in a minute forward

adrien-zinger (approved):

yvan-sraka (approved):

PR #2297, opened 2022-02-15
Refactor to prepare whitelist peers

### LINKED ISSUE #1946: Add whitelist support
At some point it would be nice to add a whitelist option. You could provide a list of IP and connections from nodes corresponding to these IP would always be accepted.

AureliaDolo:
To prepare for #1946

I mostly created the PeerType variant and refactored a lot accordingly. 

In a nutshell, instead of having many not so related indexes, we have now one index and one config per peer type. I tried my best match on peer types and avoid if as much as possible (if you see something other than stuff == PeerType::default() , there may be an underlying issue), and hide complexity is smaller methods.


AureliaDolo:
A few points before you take a look:
- I didn't worked on the tests yet
- There is still a lot of copy pasting, I had to review in order to see if 1/ I didn't made mistakes 2/ there are some factorization opportunities and how to do them
- I'm not sure AT ALL with what I did with the banned peers
- good luck with the review :four_leaf_clover:

massa-network/src/peer_info_database.rs
```diff
@@ -349,114 +382,60 @@ impl PeerInfoDatabase {
...
+        } else {
+            Ok(Vec::new().into_iter().take(avaible_slots).map(ip))
         }
-
-        Ok(res_ips)
+    }
+    /// Sorts peers by ( last_failure, rev(last_success) )
+    /// and returns as many peers as there are available slots to attempt outgoing connections to.
+    pub fn get_out_connection_candidate_ips(&self) -> Result<Vec<IpAddr>, NetworkError> {
+        // TODO why is this comment here ??
```
AureliaDolo, line 417: I wasn't really able to find when it was written :/

AureliaDolo (commented):
I'm considedring having more of a hashmap peertype -> peerConnectionCount  what do you think ? @massalabs/core-team 

But first making things as generic as possible

AureliaDolo:
And I'm really wondering what to do with the banned variant, I'd happily take any suggestions

massa-network/src/peer_info_database.rs
adrien-zinger: Remove it, it looks like an old fossil out of time

AureliaDolo:
I'm working on fixing the tests, but let me know if you have any suggestions, especially on code readability

AureliaDolo:
If you have any scenario ideas with whitelist peers, I'd take them.

To do next (maybe in some follow ups, maybe here)
* [ ] add a whitelist peer file in the config
* [ ] add a whitelist client/api command
* [ ] document that in the wiki
* [ ] get a cup of tea
* [ ] think about how it could be even more generic
* [ ] maybe break down the stats by peer type
* [ ] if a peer is in the bootstrap list provided to the bootstrap module, it must be bootstrap in peer info db

yvan-sraka (dismissed):
My personal belief is we should merge that PR as it since a lot of good refactorings was already done here, and all the items of the checklist you gave should be done as followups issues/PRs!

One last tweak/code change before merging could be to annotate function that share the same pattern but only differ by their types as "mostly duplicated" since they could be refactored in a second pass using generics :)

AureliaDolo:
> One last tweak/code change before merging could be to annotate function that share the same pattern but only differ by their types as "mostly duplicated" since they could be refactored in a second pass using generics :)

I'll move them around yet again to make more sense

massa-network/src/peer_info_database.rs
```diff
@@ -92,20 +113,21 @@ async fn dump_peers(
     peers: &HashMap<IpAddr, PeerInfo>,
     file_path: &Path,
 ) -> Result<(), NetworkError> {
-    let peer_vec: Vec<Value> = peers
+    let peer_vec: Vec<_> = peers
         .values()
-        .filter(|v| v.banned || v.advertised || v.bootstrap)
+        .filter(|v| v.advertised || v.peer_type != PeerType::Standard)
+        // TODO We were really serializing by hand ? :confused:
```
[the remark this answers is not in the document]
AureliaDolo: It's working like a charm as it is

adrien-zinger (approved):

yvan-sraka (approved):
Great improvement of codebase quality here :sparkles:!

adrien-zinger (dismissed):

massa-network/src/settings.rs
```diff
@@ -90,6 +95,21 @@ mod tests {
                 initial_peers_file: std::path::PathBuf::new(),
                 peer_list_send_timeout: MassaTime::from(500),
                 max_in_connection_overflow: 2,
+                bootstrap_peers_config: PeerTypeConnectionConfig {
+                    target_out: 1,
+                    max_out_attempts: 1,
+                    max_in: 1,
+                },
+                whitelist_peers_config: PeerTypeConnectionConfig {
```
damip, line 103:
```
whitelist_peers_config:
  max_in: 10,
  target_out: 10,
  max_out_attempts: 10,
```

Let's be more explicit:
* max_in => max_in_connections
* target_out => target _out_connections

Also maybe it's worth specifying the max number of open nodes per IP address for each type to give more granularity.

massa-network/src/peer_info_database.rs
```diff
@@ -17,22 +15,41 @@ use tokio::task::JoinHandle;
...
+/// There is a defined number af slots for each category.
+#[derive(Clone, Copy, Serialize, Deserialize, Debug, PartialEq, Eq)]
+pub enum PeerType {
+    /// if the peer is in bootstrap servers list
+    /// for now it is decoupled from the real bootstrap sever list, it's just parsed
+    /// TODO: https://github.com/massalabs/massa/issues/2320
+    Bootstrap,
+    /// Connection from these nodes are always accepted
+    WhiteListed,
+    /// Just a peer :pear:
```
damip, line 28: i's pronounced quite differently :p

AureliaDolo:
> > I'm considedring having more of a hashmap peertype -> peerConnectionCount what do you think ? @massalabs/core-team
> > But first making things as generic as possible
> 
> Having a generic string for each peer category sounds like a good idea, so that people running the node in special environments can have more flexibility and add more options than the predefined categories (bootstrap, whitelist, normal),.

In a follow up issue ?

massa-network/src/peer_info_database.rs
AureliaDolo: It sounds a little bit more similar in british english :p 

Anyway do we keep the joke ? :cactus:

massa-network/src/settings.rs
AureliaDolo: > Also maybe it's worth specifying the max number of open nodes per IP address for each type to give more granularity.

You mean the max_in_connections_per_ip in Network settings would become PeerType specific ?

AureliaDolo: > ```rust
> whitelist_peers_config:
>   max_in: 10,
>   target_out: 10,
>   max_out_attempts: 10,
> ```

it's already the current config values in this file, I update it in the config.toml

adrien-zinger (approved):

yvan-sraka (dismissed):

massa-network/src/peer_info_database.rs
damip: I would remove this one because it's not flattering to our English skills :p

massa-network/src/peer_info_database.rs
```diff
@@ -17,22 +15,41 @@ use tokio::task::JoinHandle;
...
 #[derive(Clone, Copy, Serialize, Deserialize, Debug)]
 pub struct PeerInfo {
     /// Peer ip address.
     pub ip: IpAddr,
-    /// If peer is banned.
-    pub banned: bool,
-    /// If peer is boostrap, ie peer was in initial peer file
-    pub bootstrap: bool,
+    /// The category the peer is in affects how it's treated.
+    pub peer_type: PeerType,
```
damip, line 44: in the future, this could be made a String, maybe as a followup

massa-network/src/peer_info_database.rs
```diff
@@ -49,13 +66,66 @@ pub struct PeerInfo {
...
+    /// one active connection in either direction
     /// with this peer
+    #[inline]
     pub(crate) fn is_active(&self) -> bool {
         self.active_out_connection_attempts > 0
             || self.active_out_connections > 0
             || self.active_in_connections > 0
     }
+
+    pub fn new(ip: IpAddr, advertised: bool) -> PeerInfo {
```
damip, line 78: let's document the parameters, especially "advertised"

massa-network/src/peer_info_database.rs
```diff
@@ -49,13 +66,66 @@ pub struct PeerInfo {
...
+                > MassaTime::from(0u64);
+        }
+        true
+    }
+}
+
+/// Connection count for a category
+#[derive(Default, Debug)]
+pub(crate) struct ConnectionCount {
+    pub(crate) active_out_connection_attempts: usize,
```
damip, line 112: let's document each field :)

massa-network/src/peer_info_database.rs
```diff
@@ -93,20 +159,9 @@ async fn dump_peers(
...
-            json!({
-                "ip": peer.ip,
-                "banned": peer.banned,
-                "bootstrap": peer.bootstrap,
-                "last_alive": peer.last_alive,
-                "last_failure": peer.last_failure,
-                "advertised": peer.advertised,
-            })
-        })
+        .filter(|v| v.advertised || v.peer_type != PeerType::Standard)
```
damip, line 164: let's not forget banned peers !

massa-network/src/peer_info_database.rs
```diff
@@ -120,9 +175,12 @@ async fn dump_peers(
 ///
 /// Note: only non-active, non-bootstrap peers are counted when clipping to size limits.
```
damip, line 176: only standard non-active peers should be counted when clipping

massa-network/src/peer_info_database.rs
```diff
@@ -321,169 +407,40 @@ impl PeerInfoDatabase {
...
-        let mut sorted_ips: Vec<IpAddr> = sorted_peers
-            .into_iter()
-            .take(MAX_ADVERTISE_LENGTH as usize)
-            .map(|p| p.ip)
-            .collect();
-        if let Some(our_ip) = self.network_settings.routable_ip {
-            sorted_ips.insert(0, our_ip);
-            sorted_ips.truncate(MAX_ADVERTISE_LENGTH as usize);
+    /// Unbans a list of ip
+    pub async fn unban(&mut self, ips: Vec<IpAddr>) -> Result<(), NetworkError> {
```
damip, line 434: this function needs to request_dump if there was a change

massa-network/src/peer_info_database.rs
```diff
@@ -321,169 +407,40 @@ impl PeerInfoDatabase {
...
+        for ip in ips.into_iter() {
+            if let Some(peer) = self.peers.get_mut(&ip) {
+                peer.peer_type = Default::default();
```
damip, line 437: why is unbanning changing the peer type ?

massa-network/src/peer_info_database.rs
```diff
@@ -321,169 +407,40 @@ impl PeerInfoDatabase {
...
+                peer.banned = false;
+            } else {
+                return Ok(());
```
damip, line 440: this will quit the whole loop whenever an IP is not found

massa-network/src/peer_info_database.rs
```diff
@@ -494,50 +451,27 @@ impl PeerInfoDatabase {
...
+        let peer_type = self.get_peer_type(ip).ok_or({
+            NetworkError::PeerConnectionError(NetworkConnectionErrorType::PeerInfoNotFoundError(
+                *ip,
+            ))
+        })?;
+
+        self.update_global_active_out_connection_attempt_count(
+            peer_type,
+            true,
+            NetworkConnectionErrorType::TooManyConnectionAttempt(*ip),
```
damip, line 464: toomanyconnectionattemptS (plural)

granted, the previous version was even more wrong xD

massa-network/src/peer_info_database.rs
```diff
@@ -494,50 +451,27 @@ impl PeerInfoDatabase {
         if !ip.is_global() {
             return Err(NetworkError::InvalidIpError(*ip));
         }
-        let (available_bootstrap_conns, available_nonbootstrap_conns) =
-            self.get_available_out_connection_attempts();
+
```
damip, line 454: I think I see a problem here (that was probbably there even before).

Scenario;
* we ask for a peer list to connect to
* something causes peer cleanup, removing some of the peers from that list
* we notify new_out_connection_attempt to one of the peers we receeived at the 1st step => it will be refused

I think here, if the peer is not found, we should add it as a standard peer with 1 out connection attempt and call cleanup() at the end of the function (to remove any excess peers now that we have added one)

what do you think ?

massa-network/src/peer_info_database.rs
```diff
@@ -616,34 +537,32 @@ impl PeerInfoDatabase {
...
+            ))
+        })?;
+
+        self.update_global_active_out_connection_count(
+            peer_type,
+            false,
+            NetworkConnectionErrorType::CloseConnectionWithNoConnectionToClose(*ip),
+        )?;
+
         let peer = self.peers.get_mut(ip).ok_or({
```
damip, line 552: it's always a bit srange to do get_peer_type and then get the whole peer with get_mut. Why not get_mut the peer once, and just check its type as a first step to improve performance ? (also elsewhere in the code)

damip (changes requested):
Some good refactoring and ideas in this PR ! :) 

Note: the review is not finished yet

massa-network/src/peer_info_database.rs
```diff
@@ -823,19 +738,25 @@ impl PeerInfoDatabase {
...
         }
+        let peer = if let Some(p) = self.get_peer_type(ip) {
+            p
+        } else {
+            let p = PeerInfo::new(*ip, false);
+            self.peers.insert(*ip, p);
+            self.get_peer_type(ip).unwrap()
+        };
+
+        if self.is_max_in_connection_count_reached(peer) {
```
damip, line 749: this should be checked first, otherwise we might have already inserted a new peer (3 lines above)

massa-network/src/peer_info_database.rs
```diff
@@ -823,19 +738,25 @@ impl PeerInfoDatabase {
...
+            return Err(NetworkError::PeerConnectionError(
+                NetworkConnectionErrorType::MaxPeersConnectionReached(*ip),
+            ));
+        }
 
         let res = match self.peers.entry(*ip) {
             hash_map::Entry::Occupied(mut occ) => {
```
damip, line 756: a few lines above, when the peer is not found you do `self.peers.insert(*ip, p);`  so this will always be occupied and never Vacant

damip: I will re-review this whole function when this part is corrected

massa-network/src/peer_info_database.rs
```diff
@@ -847,47 +768,345 @@ impl PeerInfoDatabase {
...
+                &self.whitelist_connection_count,
+                &self.network_settings.whitelist_peers_config,
+            )?)
+            .chain(self.get_candidate_ips_for_type(
+                PeerType::Standard,
+                &self.standard_connection_count,
+                &self.network_settings.standard_peers_config,
+            )?)
+            .collect();
+        res.sort_unstable_by_key(|&p| (p.last_failure, std::cmp::Reverse(p.last_alive)));
```
damip, line 828: let's prioritize IPs of special type over standard IPs in this sorting

massa-network/src/peer_info_database.rs
```diff
@@ -847,47 +768,345 @@ impl PeerInfoDatabase {
...
+
+    ////////////////////
+    // public getters //
+    ////////////////////
+
+    /// Sorts peers by ( last_failure, rev(last_success) )
+    /// and returns as many peers as there are available slots to attempt outgoing connections to.
+    pub fn get_out_connection_candidate_ips(&self) -> Result<Vec<IpAddr>, NetworkError> {
+        let mut res: Vec<_> = self
+            .get_candidate_ips_for_type(
```
damip, line 812: let's call it get_out_connection_candidate_ips_for_type for consistency (yes it's a bit long but makes it clearer :))

damip (changes requested):

massa-network/src/settings.rs
damip: > > Also maybe it's worth specifying the max number of open nodes per IP address for each type to give more granularity.
> 
> You mean the max_in_connections_per_ip in Network settings would become PeerType specific ?

maybe as a followup

damip: > > ```rust
> > whitelist_peers_config:
> >   max_in: 10,
> >   target_out: 10,
> >   max_out_attempts: 10,
> > ```
> 
> it's already the current config values in this file, I update it in the config.toml

? I didn't understand this comment.

What I meant is just to rename the fields to make them more explicit (max_in_connections instead of max_in)

massa-network/src/peer_info_database.rs
AureliaDolo: because when we update the global counters, we take mutable ownership of self. PeerType is Copy so getting peer type does not borrow self. Whereas taking the peer mutably once takes mutable ownership of self, and then I cannot update the global counters.

massa-network/src/settings.rs
AureliaDolo: > What I meant is just to rename the fields to make them more explicit (max_in_connections instead of max_in)

My concern was about the config values in themselves. Are they the expected ones ? (especially for the whitelist ones as there was no reference value)

AureliaDolo: > > > 
> > 
> > 
> > You mean the max_in_connections_per_ip in Network settings would become PeerType specific ?
> 
> maybe as a followup

See #2333

massa-network/src/peer_info_database.rs
AureliaDolo: see #2334

massa-network/src/peer_info_database.rs
AureliaDolo: It's an artifact. I tried having Banned as a peerType variant, it was much more painful than I expected.

massa-network/src/peer_info_database.rs
AureliaDolo: Why is it wrong ?

massa-network/src/peer_info_database.rs
AureliaDolo: I can move the sorting part in get_candidate_ips_for_type and then just chain the iterators without sorting. Is that the expected behaviour ?

massa-network/src/peer_info_database.rs
AureliaDolo: That makes sense.

AureliaDolo: it appears that it's a tested behavior, so I don't know what to do about that, either revert the last commit 12c2f7397e91343ab0e95d9037fe1da40abe3203 and keep that reflection for a follow up or update the test according to the new behavior.

massa-network/src/settings.rs
damip: I would go for max_in: 3, target_out: 2, max_out_attempts: 2

massa-network/src/peer_info_database.rs
```diff
@@ -494,50 +460,35 @@ impl PeerInfoDatabase {
...
+            p
+        } else {
+            let mut p = PeerInfo::new(*ip, false);
+            p.active_out_connection_attempts += 1;
+            self.peers.insert(*ip, p);
+            update_happened = true;
+            self.get_peer_type(ip).unwrap()
+        };
+
+        self.update_global_active_out_connection_attempt_count(
```
damip, line 474: here is a logical problem: if there are no more out connection slots, this call will bail, but the peer could already have been created + inserted into self.peers (see "else" branch just above), which will leave the state inconsistent

massa-network/src/peer_info_database.rs
```diff
@@ -494,50 +460,35 @@ impl PeerInfoDatabase {
...
             return Err(NetworkError::InvalidIpError(*ip));
         }
-        let (available_bootstrap_conns, available_nonbootstrap_conns) =
-            self.get_available_out_connection_attempts();
+        let mut update_happened = false;
+        let peer_type = if let Some(p) = self.get_peer_type(ip) {
+            p
+        } else {
+            let mut p = PeerInfo::new(*ip, false);
+            p.active_out_connection_attempts += 1;
```
damip, line 468: this value is already incremented below, so the peer's active_out_connection_attempts will be 2 at the end of the function if the peer didn't exist before

massa-network/src/peer_info_database.rs
```diff
@@ -494,50 +460,35 @@ impl PeerInfoDatabase {
...
-                return Err(NetworkError::PeerConnectionError(
-                    NetworkConnectionErrorType::ToManyConnectionAttempt(*ip),
-                ));
-            }
-        } else if available_nonbootstrap_conns == 0 {
-            return Err(NetworkError::PeerConnectionError(
-                NetworkConnectionErrorType::ToManyConnectionAttempt(*ip),
-            ));
-        }
         peer.active_out_connection_attempts += 1;
```
damip, line 485: thisis also an update, and should set "update_happened" to true

massa-network/src/peer_info_database.rs
```diff
@@ -494,50 +460,35 @@ impl PeerInfoDatabase {
...
-            self.get_available_out_connection_attempts();
+        let mut update_happened = false;
+        let peer_type = if let Some(p) = self.get_peer_type(ip) {
+            p
+        } else {
+            let mut p = PeerInfo::new(*ip, false);
+            p.active_out_connection_attempts += 1;
+            self.peers.insert(*ip, p);
+            update_happened = true;
+            self.get_peer_type(ip).unwrap()
```
damip, line 471: comment the unwrap

massa-network/src/peer_info_database.rs
damip: okay

massa-network/src/peer_info_database.rs
damip: because if config.max_in_connections = 0 and if the peer didn't exist in self.peers before,  this function will fail because max peers connections is reached, but the peer will remain in self.peers (without pruning nor anything).

The cleanest way is always to check if we can create/insert something before doing so

massa-network/src/peer_info_database.rs
damip: I guess just adding `p.type == standard` as the first element of the sorting tuple should do the trick

damip (changes requested):

massa-network/src/peer_info_database.rs
```diff
@@ -823,71 +755,365 @@ impl PeerInfoDatabase {
...
+        } else if peer.active_in_connections >= self.network_settings.max_in_connections_per_ip {
+            self.request_dump()?;
+            return Err(NetworkError::PeerConnectionError(
+                NetworkConnectionErrorType::MaxPeersConnectionReached(*ip),
+            ));
+        } else {
+            peer.active_in_connections += 1;
+        }
+
+        let peer = self.get_peer_type(ip).ok_or({
```
damip, line 795: I would do this:
```rust
let peer_type = peer.type;
std::mem::drop(peer);

// here we have peer_type, but not the peer reference in scope anymore
```

damip (dismissed):

massa-network/src/peer_info_database.rs
damip: also we have the peer at hand, so you could just do:
* peer_type = peer.type  (get a copy of the peer type)
* insert the peer
* return the peer_type from the scope

That way we don't have unwraps nor useless extra lookups

massa-network/src/peer_info_database.rs
damip: still a problem

massa-network/src/peer_info_database.rs
```diff
@@ -321,169 +412,44 @@ impl PeerInfoDatabase {
...
-            sorted_peers
-                .sort_unstable_by_key(|&p| (p.last_failure, std::cmp::Reverse(p.last_alive)));
-            res_ips.extend(
-                sorted_peers
-                    .into_iter()
-                    .take(available_slots_nonbootstrap)
-                    .map(|p| p.ip),
-            )
+    /// Unbans a list of ip
+    pub async fn unban(&mut self, ips: Vec<IpAddr>) -> Result<(), NetworkError> {
```
damip, line 439: does the function neeed to be async ?

massa-network/src/peer_info_database.rs
damip: ok let's keep it for a followup

damip:
Note that I requested a new review from people who already approved because there are many critical problems that have been missed

massa-network/src/peer_info_database.rs
adrien-zinger: Yes, you can do that, you also can use scopes.

massa-network/src/peer_info_database.rs
AureliaDolo: #2347

massa-network/src/peer_info_database.rs
AureliaDolo: > but the peer will remain in self.peers (without pruning nor anything).

Why it's bad ?

massa-network/src/peer_info_database.rs
```diff
@@ -321,169 +412,44 @@ impl PeerInfoDatabase {
...
+        let mut update_happened = false;
+        for ip in ips.into_iter() {
+            if let Some(peer) = self.peers.get_mut(&ip) {
+                peer.banned = false;
+                update_happened = true;
+            } else {
+                continue;
```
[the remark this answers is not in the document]
AureliaDolo: It depends if we insert the ip as a default peer

massa-network/src/peer_info_database.rs
AureliaDolo: How much is it critical, if the local counter of the peer is not updated, and we just keep the ip in the db ?

massa-network/src/peer_info_database.rs
AureliaDolo: so if we reached that point there is always an update so no need for update_happened. I was thinking that we just call self.update() when an ip was added, removed or ip type update, not every time we update a counter ?

massa-network/src/peer_info_database.rs
```diff
@@ -321,169 +412,44 @@ impl PeerInfoDatabase {
...
-        let mut sorted_ips: Vec<IpAddr> = sorted_peers
-            .into_iter()
-            .take(MAX_ADVERTISE_LENGTH as usize)
-            .map(|p| p.ip)
-            .collect();
-        if let Some(our_ip) = self.network_settings.routable_ip {
-            sorted_ips.insert(0, our_ip);
-            sorted_ips.truncate(MAX_ADVERTISE_LENGTH as usize);
+        if update_happened {
+            self.request_dump()?
```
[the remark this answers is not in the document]
AureliaDolo: I think I'm a little bit confused about when to call update and when to call request dump

massa-network/src/peer_info_database.rs
AureliaDolo: Indeed

massa-network/src/peer_info_database.rs
```diff
@@ -823,71 +755,365 @@ impl PeerInfoDatabase {
...
+            peer.active_in_connections += 1;
+        }
+
+        let peer = self.get_peer_type(ip).ok_or({
+            NetworkError::PeerConnectionError(NetworkConnectionErrorType::PeerInfoNotFoundError(
+                *ip,
+            ))
+        })?;
+
+        self.update_global_active_in_connection_count(
```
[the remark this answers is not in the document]
AureliaDolo: 06c78f744d735a18a9864b097d752dec5517572b

massa-network/src/peer_info_database.rs
AureliaDolo: I moved the sorting part to the for_types functions.

massa-network/src/peer_info_database.rs
damip: always check if we can insert something before actually inserting it

massa-network/src/peer_info_database.rs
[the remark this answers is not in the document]
damip: update = whenever something might need pruning

request_dump = whenever something changes the contents of the dumped file on disk

massa-network/src/peer_info_database.rs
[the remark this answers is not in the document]
damip: :/

massa-network/src/peer_info_database.rs
damip: the counters were inconsistent. Now you added a rollback mechanism, but I'm not a big fan of this.

Would be nicer to always first test if we can insert/update a peer BEFORE actually doing it. Also elsewhere in this PR

massa-network/src/peer_info_database.rs
damip: see comment above

massa-network/src/peer_info_database.rs
damip: if that's the case, yes

massa-network/src/peer_info_database.rs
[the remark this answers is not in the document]
AureliaDolo: maybe I need a little bit more sleep x)

massa-network/src/peer_info_database.rs
```diff
@@ -706,56 +671,74 @@ impl PeerInfoDatabase {
...
+            peer.active_out_connections += 1;
+            peer.peer_type
+        };
+
+        if let Err(e) = self.update_global_active_out_connection_count(
+            peer_type,
+            true,
+            NetworkConnectionErrorType::UnexpectedError,
+        ) {
+            //roll back
```
damip, line 722: I'm against rollbacks :p see comments above

massa-network/src/peer_info_database.rs
```diff
@@ -767,34 +750,39 @@ impl PeerInfoDatabase {
...
+                self.update()?;
+            }
+            pt
+        };
+        if let Err(e) = self.update_global_active_out_connection_attempt_count(
+            peer_type,
+            false,
+            NetworkConnectionErrorType::TooManyConnectionFailure(*ip),
+        ) {
+            //roll back
```
damip, line 777: another one here

AureliaDolo:
Maybe we could open a follow up issue to tackle the roll backs, as we don't have a solution right now, and it's mostly a matter of personal preference.

@damip you shared this link in the discord, https://github.com/rust-lang/rfcs/issues/1215 And from what I understand, view types are not (yet ?) part of the language but some crates exist.

AurelienFT:
@AureliaDolo I have made a proposition to resolve some problems and improve genericity here : https://github.com/massalabs/massa/pull/2379 it's based on your PR. I have made a new PR instead of review because the changes suggested are too big.

AurelienFT:
@AureliaDolo It's possible that my refactor broke some tests. Do you want me to fix this or you will do it ?

AurelienFT (approved):
As I have work on that I'm agree on this new design. Well tested. Nice refactor :)

yvan-sraka (approved):
LGTM! That's always a nice thing to use an `Enum` rather than several conflicting boolean fields, nice refactoring here :sparkles:

AureliaDolo:
bors merge

adrien-zinger (approved):

AurelienFT (approved):
The merge don't change anything related to this PR.

AureliaDolo:
bors merge

PR #2305, opened 2022-02-17
Use shared storage in Network

gterzian:
Part of https://github.com/massalabs/massa/issues/2178

### DIFF
--- a/massa-models/src/storage.rs
+++ b/massa-models/src/storage.rs
@@ -6,6 +6,7 @@ use std::sync::Arc;
 pub struct StoredBlock {
     pub block: Block,
     pub serialized: Vec<u8>,
+    pub serialized_header: Option<Vec<u8>>,
 }
 
 #[derive(Clone, Default)]
@@ -16,7 +17,11 @@ pub struct Storage {
 impl Storage {
     pub fn store_block(&self, block_id: BlockId, block: Block, serialized: Vec<u8>) {
         // TODO: first check, and allow for, an already stored header for the block.
-        let stored_block = StoredBlock { block, serialized };
+        let stored_block = StoredBlock {
+            block,
+            serialized,
+            serialized_header: None,
+        };
         let to_store = Arc::new(RwLock::new(stored_block));
         let mut blocks = self.blocks.write();
         blocks.insert(block_id, to_store);
--- a/massa-network/src/binders.rs
+++ b/massa-network/src/binders.rs
@@ -1,7 +1,7 @@
 // Copyright (c) 2022 MASSA LABS <info@massa.net>
 
 //! Flexbuffer layer between raw data and our objects.
-use super::messages::Message;
+use super::messages::{deserialize_message_with_optional_serialized_object, Message};
 use crate::error::NetworkError;
 use crate::establisher::{ReadHalf, WriteHalf};
 use massa_models::{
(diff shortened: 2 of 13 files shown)


gterzian:
@damip What do you think of the basic use of storage and the re-use of serialized in this PR so far?

massa-network/src/node_worker.rs
```diff
@@ -26,14 +30,21 @@ use tokio::{
 };
 use tracing::{debug, trace, warn};
 
+#[derive(Debug, Serialize, Deserialize)]
+pub enum ToSend {
+    Msg(Message),
+    Block(BlockId),
```
AurelienFT, line 36: Maybe call this variant BlockId ?

gterzian: I think `Block` is more descriptive, because the point of the message is to actually send the block over the network.

AurelienFT: Ok

AurelienFT (approved):

massa-models/src/storage.rs
```diff
@@ -6,6 +6,7 @@ use std::sync::Arc;
 pub struct StoredBlock {
     pub block: Block,
     pub serialized: Vec<u8>,
+    pub serialized_header: Option<Vec<u8>>,
```
AureliaDolo, line 9: Why is it an option  ?

massa-network/src/binders.rs
```diff
@@ -96,7 +95,7 @@ impl ReadBinder {
     /// or =0 if there is no more data.
     /// We can't use read_exact and similar because they are not cancel-safe:
     /// https://docs.rs/tokio/latest/tokio/io/trait.AsyncReadExt.html#cancel-safety-2
-    pub async fn next(&mut self) -> Result<Option<(u64, Message)>, NetworkError> {
+    pub async fn next(&mut self) -> Result<Option<(u64, Message, Option<Vec<u8>>)>, NetworkError> {
```
AureliaDolo, line 98: What's the option you added for ?

massa-models/src/storage.rs
gterzian: Currently we only compute it when sending out the header, and then cache it. So when first receiving the block it is set to `None`. Not sure that is the right way by the way.

massa-network/src/binders.rs
gterzian: That is the serialized object, would need some documentation...

massa-network/src/messages.rs
```diff
@@ -50,9 +50,32 @@ pub enum Message {
...
+            Ok((Message::Block(block), Some(serialized)))
+        }
+        _ => Message::from_bytes_compact(buffer).map(|result| (result.0, None)),
+    }
+}
+
 #[derive(IntoPrimitive, Debug, Eq, PartialEq, TryFromPrimitive)]
 #[repr(u32)]
-enum MessageTypeId {
+pub enum MessageTypeId {
```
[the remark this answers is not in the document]
gterzian: Because it's now used elsewhere in the crate. Changed it to `pub(crate)`...

massa-models/src/storage.rs
AureliaDolo: what happens when we first receive the header and then the block ?

gterzian: Currently nothing is done with individual headers. By the way I noticed we only propagate headers after having the received the full block.

PR #2306, opened 2022-02-17
Following @damip comment: not hiding behind an abstraction how custom pre-hashed `HashSet` and `HashMap` are implemented!

yvan-sraka:


AureliaDolo:
I'm against that PR, especially right now as there is too much refactoring going on.

If you really want that kind of stuff I would rather call it PreHashedHashSet/Map I find your suggestion confusing.

yvan-sraka:
> I'm against that PR, especially right now as there is too much refactoring going on.
> 
> 
> 
> If you really want that kind of stuff I would rather call it PreHashedHashSet/Map I find your suggestion confusing.

I'm completely agreed, my opinion is the implementation choice should be transparent in module names not in type names ...

damip:
> > I'm against that PR, especially right now as there is too much refactoring going on.
> > If you really want that kind of stuff I would rather call it PreHashedHashSet/Map I find your suggestion confusing.
> 
> I'm completely agreed, my opinion is the implementation choice should be transparent in module names not in type names ...

@yvan-sraka  @AureliaDolo 

Ok for not implementing it because there are other refactorings ongoing. However the other arguments are flawed. There are good reasons why std::collections::hash_map::HashMap is not  std::hash_map::Map. (See https://doc.rust-lang.org/std/collections/hash_map/struct.HashMap.html )

Module name is not a good conveyor of information. In practice, people don't want to carry the module name around, so they "use" at the top of the file. That way they only use the typename through the file. This causes problems if you have multiple types of say "map" (for example BTreeMap and HashMap) in the same file: you will have to change behavior and specify the module name at every use, or you will end up creating an ambiguity. And even when only one type of map is used in a given file, it helps readability if the reader doesn't have to jump at the top of the file or rely on advanced ide features to figure out what type Map is used. That's exactly what happened to Sebastien: he was confused because he just saw the name "Map" used at a random place in the file where a prehashed map is expected.

PR #2307, opened 2022-02-17
Change the insertion of known blocks algos

adrien-zinger:
I'm pretty sure that this implementation is better about time complexity. About space complexity, it makes it N in the worst case. @yvan-sraka @AureliaDolo  I'm curious, what do you think about it?

### DIFF
--- a/massa-protocol-worker/src/protocol_worker.rs
+++ b/massa-protocol-worker/src/protocol_worker.rs
@@ -94,6 +94,7 @@ pub async fn start_protocol_controller(
 
 //put in a module to block private access from Protocol_worker.
 mod nodeinfo {
+    use itertools::Itertools;
     use massa_models::prehash::{BuildMap, Map, Set};
     use massa_models::{BlockId, EndorsementId, OperationId};
     use massa_protocol_exports::ProtocolSettings;
@@ -161,6 +162,9 @@ mod nodeinfo {
 
         /// Insert knowledge of a list of blocks in NodeInfo
         ///
+        /// If there is just one item to pop, just find the min, otherwise,
+        /// sort the items and take the overflow from the first item.
+        ///
         /// ## Arguments
         /// - self: node info
         /// - block_ids: list of blocks
@@ -177,14 +181,27 @@ mod nodeinfo {
             for block_id in block_ids {
                 self.known_blocks.insert(*block_id, (val, instant));
             }
-            while self.known_blocks.len() > max_node_known_blocks_size {
-                // remove oldest item
-                let (&h, _) = self
+            let len = self.known_blocks.len();
+            if len == max_node_known_blocks_size {
+                /* Fast case where there is just one to pop */
+                let block_id = *self
                     .known_blocks
                     .iter()
                     .min_by_key(|(h, (_, t))| (*t, *h))
-                    .unwrap(); // never None because is the collection is empty, while loop isn't executed.
-                self.known_blocks.remove(&h);
+                    .unwrap()
+                    .0;
+                self.known_blocks.remove(&block_id);
+                return;
+            }
+            if len > max_node_known_blocks_size {
+                self.known_blocks
+                    .iter()
+                    .map(|(h, (_, t))| (*h, *t))
+                    .sorted()
+                    .take(len - max_node_known_blocks_size)
+                    .for_each(|(block_id, _)| {
+                        self.known_blocks.remove(&block_id);
+                    });
             }
         }
 


damip:
> I'm pretty sure that this implementation is better about time complexity. About space complexity, it makes it N in the worst case. @yvan-sraka @AureliaDolo I'm curious, what do you think about it?

In the nominal case, known_blocks are added one by one. So the "while" is just there for robustness, it could have been an "if". The complexity of min_by_key is `O(N)`, which is the effective complexity of that part.

In the branch you added (which is never executed in the nominal case), sorting takes `O(N logN)`.

I agree this deserves a comment in the code :)

massa-protocol-worker/src/protocol_worker.rs
```diff
@@ -177,14 +181,27 @@ mod nodeinfo {
...
+                    .unwrap()
+                    .0;
+                self.known_blocks.remove(&block_id);
+                return;
+            }
+            if len > max_node_known_blocks_size {
+                self.known_blocks
+                    .iter()
+                    .map(|(h, (_, t))| (*h, *t))
+                    .sorted()
```
AureliaDolo, line 200: I'd be curious to see how does sorted behave when there is only one item, I'm not sure it's really different of what the fast case does

adrien-zinger:
> > I'm pretty sure that this implementation is better about time complexity. About space complexity, it makes it N in the worst case. @yvan-sraka @AureliaDolo I'm curious, what do you think about it?
> 
> In the nominal case, known_blocks are added one by one. So the "while" is just there for robustness, it could have been an "if". The complexity of min_by_key is `O(N)`, which is the effective complexity of that part.
> 
> In the branch you added (which is never executed) sorting takes `O(N logN)`.
> 
> I agree this deserves a comment in the code :) can you add it ?

Ok.. so is the while really useful ? I mean... If it's just for robustness, what the maximum difference ? 1, 2, N ? If there is a lot of blocks overflowing, the old algo became factorial if I get it. Is there a limit of the overflow in which case we want to do an algo that are adapted ?

massa-protocol-worker/src/protocol_worker.rs
adrien-zinger: I don't know how the compiler optimize that :rofl: but reading that, the worst case would be O(nlog(n)) and the nice case O(n) where the worst case is the sorted algorithm.

If we go through the nice case with a `while` we could have a kind of (n-m)*n complexity, but visually it looks like a `n!` in the worst case and a O(n) in the nice case.

The question of the robustness surprise me even if I know the pattern now. It's not common to make something robust with the current technics. Normally, or something should never append, so we don't write the _special case_, or it can append, so we choose the better algorithm for the situation.

If we had more than one known block overflow, doesn't that means that we get an error or an undefined behavior earlier? In that case, we should also write a debug comment in addition to a developer note.

damip:
> > > I'm pretty sure that this implementation is better about time complexity. About space complexity, it makes it N in the worst case. @yvan-sraka @AureliaDolo I'm curious, what do you think about it?
> > 
> > 
> > In the nominal case, known_blocks are added one by one. So the "while" is just there for robustness, it could have been an "if". The complexity of min_by_key is `O(N)`, which is the effective complexity of that part.
> > In the branch you added (which is never executed) sorting takes `O(N logN)`.
> > I agree this deserves a comment in the code :) can you add it ?
> 
> Ok.. so is the while really useful ? I mean... If it's just for robustness, what the maximum difference ? 1, 2, N ? If there is a lot of blocks overflowing, the old algo became factorial if I get it. Is there a limit of the overflow in which case we want to do an algo that are adapted to the situation? It could be for example if overflow > log(Len)

If there are K blocks overflowing and N blocks stored, the complexity will not be factorial but `O(K*N)`.

There should always be 0 or 1 block overflowing. I just put a "while" instead of an "if" so that the algorithm is self-contained and always does what it is supposed to do: prune overflows.

adrien-zinger:
> > > > I'm pretty sure that this implementation is better about time complexity. About space complexity, it makes it N in the worst case. @yvan-sraka @AureliaDolo I'm curious, what do you think about it?
> > > 
> > > 
> > > In the nominal case, known_blocks are added one by one. So the "while" is just there for robustness, it could have been an "if". The complexity of min_by_key is `O(N)`, which is the effective complexity of that part.
> > > In the branch you added (which is never executed) sorting takes `O(N logN)`.
> > > I agree this deserves a comment in the code :) can you add it ?
> > 
> > 
> > Ok.. so is the while really useful ? I mean... If it's just for robustness, what the maximum difference ? 1, 2, N ? If there is a lot of blocks overflowing, the old algo became factorial if I get it. Is there a limit of the overflow in which case we want to do an algo that are adapted to the situation? It could be for example if overflow > log(Len)
> 
> If there are K blocks overflowing and N blocks stored, the complexity will not be factorial but `O(K*N)`.
> 
> There should always be 0 or 1 block overflowing. I just put a "while" instead of an "if" so that the algorithm is self-contained and always does what it is supposed to do: prune overflows. It is just optimized for K∈{0, 1}.

I agree with you, and I'll write a good comment instead of the change :wink:

To be precise, if we really need to use mathematical annotations here :rofl: . Step 1, we make `Ω(N)` operation to find the min and get N-1 after the remove. Step 2, we make `Ω(N-1)` operation to find the min and get N-2 after the remove. Etc... With an overflow `K >= N`, we have now `Ω(N²)` that is not good.

With a smaller overflow... You're totally right with `O(K*N)` and with _K∈{0, 5}_ where 5 is just a random small value, we're close to `O(N)`.

After all, if I want to be clear in the comment. Is that normal to have `K > 1`? Shouldn't we throw an error because we're in an undefined behavior? Or at least write a warning because we're just ignoring a bug if there is one here.

AureliaDolo:
anyway there is something off test_node_info_know_block is failling

yvan-sraka:
I'm agree and I would add that often the best solution to an algorithmic problem is to use the right data structure!

In our case I would recommend to use an **Ordered** Map, there is a bunch of crates offering BST in Rust, and after a quick look I find out that https://crates.io/crates/teardown_tree would be a good fit 🙂

damip:
> I'm agree and I would add that often the best solution to an algorithmic problem is to use the right data structure!
> 
> In our case I would recommend using an **Ordered** Map, there are a bunch of crates offering BST in Rust, and after a quick look I find out that https://crates.io/crates/teardown_tree would be a good fit slightly_smiling_face

That structure would actually slow things down because we query much more often than we insert.
Currently, queries happen in constant time.
With the structure you propose, queries are logarithmic.

I don't know if there is a better option out there for this special purpose but perf measurements show that the existing algorithm brings a negligible CPU load.

damip:
> anyway there is something off test_node_info_know_block is failing, in my opinion we should focus on code readability and maintainability first.

true, and I think the current version is quite readable and maintainable :)

damip:
> > > > > I'm pretty sure that this implementation is better about time complexity. About space complexity, it makes it N in the worst case. @yvan-sraka @AureliaDolo I'm curious, what do you think about it?
> > > > 
> > > > 
> > > > In the nominal case, known_blocks are added one by one. So the "while" is just there for robustness, it could have been an "if". The complexity of min_by_key is `O(N)`, which is the effective complexity of that part.
> > > > In the branch you added (which is never executed) sorting takes `O(N logN)`.
> > > > I agree this deserves a comment in the code :) can you add it ?
> > > 
> > > 
> > > Ok.. so is the while really useful ? I mean... If it's just for robustness, what the maximum difference ? 1, 2, N ? If there is a lot of blocks overflowing, the old algo became factorial if I get it. Is there a limit of the overflow in which case we want to do an algo that are adapted to the situation? It could be for example if overflow > log(Len)
> > 
> > 
> > If there are K blocks overflowing and N blocks stored, the complexity will not be factorial but `O(K*N)`.
> > There should always be 0 or 1 block overflowing. I just put a "while" instead of an "if" so that the algorithm is self-contained and always does what it is supposed to do: prune overflows. It is just optimized for K∈{0, 1}.
> 
> I agree with you, and I'll write a good comment instead of the change wink
> 
> To be precise, if we really need to use mathematical annotations here rofl . Step 1, we make `Ω(N)` operation to find the min and get N-1 after the remove. Step 2, we make `Ω(N-1)` operation to find the min and get N-2 after the remove. Etc... With an overflow `K >= N`, we have now `Ω(N²)` that is not good.
> 
> With a smaller overflow... You're totally right with `O(K*N)` and with _K∈{0, 5}_ where 5 is just a random small value, we're close to `O(N)`.
> 
> After all, if I want to be clear in the comment. Is that normal to have `K > 1`? Shouldn't we throw an error because we're in an undefined behavior? Or at least write a warning because we're just ignoring a bug if there is one here.

Basically, we insert at most 2 blocks per second, so it is safe to assume that even if we start batching blocks in the future, this algorithm will always perform really well, and we won't have to change it.

adrien-zinger:
I also spot a case where we push all the parents of a block. I'm still in favor of switching algorithm for `block_ids.length == 1` and `block_ids.length > 1`.

What I wrote it in the current PR is simply an obvious optim for a small pb, there is a simple case with `overflow == 1` and another where we *just want to not be quadratic* (and the most of the cases here we're not). And with a test that don't even make the program more buggy and slower..


Then, since the total length is always minuscule, we're creating a problem where there isn't. We want to optimize something that don't take time after all. Defending an approach from another... I just spot a _wtf_ line, rude to understand because of the absurd complexity. :rofl: I'm very surprised to see the turn of this conversation.

If we're all ok with a O(N²) in the worst case (that should rarely append), and in addition don't write any warnings, and just comment that with a dev note... _**It's ok! Let's do it!**_ Finally, the said `N` is so tiny, we don't really care about that....

adrien-zinger:
> > > > > > I'm pretty sure that this implementation is better about time complexity. About space complexity, it makes it N in the worst case. @yvan-sraka @AureliaDolo I'm curious, what do you think about it?
> > > > > 
> > > > > 
> > > > > In the nominal case, known_blocks are added one by one. So the "while" is just there for robustness, it could have been an "if". The complexity of min_by_key is `O(N)`, which is the effective complexity of that part.
> > > > > In the branch you added (which is never executed) sorting takes `O(N logN)`.
> > > > > I agree this deserves a comment in the code :) can you add it ?
> > > > 
> > > > 
> > > > Ok.. so is the while really useful ? I mean... If it's just for robustness, what the maximum difference ? 1, 2, N ? If there is a lot of blocks overflowing, the old algo became factorial if I get it. Is there a limit of the overflow in which case we want to do an algo that are adapted to the situation? It could be for example if overflow > log(Len)
> > > 
> > > 
> > > If there are K blocks overflowing and N blocks stored, the complexity will not be factorial but `O(K*N)`.
> > > There should always be 0 or 1 block overflowing. I just put a "while" instead of an "if" so that the algorithm is self-contained and always does what it is supposed to do: prune overflows. It is just optimized for K∈{0, 1}.
> > 
> > 
> > I agree with you, and I'll write a good comment instead of the change wink
> > To be precise, if we really need to use mathematical annotations here rofl . Step 1, we make `Ω(N)` operation to find the min and get N-1 after the remove. Step 2, we make `Ω(N-1)` operation to find the min and get N-2 after the remove. Etc... With an overflow `K >= N`, we have now `Ω(N²)` that is not good.
> > With a smaller overflow... You're totally right with `O(K*N)` and with _K∈{0, 5}_ where 5 is just a random small value, we're close to `O(N)`.
> > After all, if I want to be clear in the comment. Is that normal to have `K > 1`? Shouldn't we throw an error because we're in an undefined behavior? Or at least write a warning because we're just ignoring a bug if there is one here.
> 
> Basically, we insert at most 2 blocks per second, so it is safe to assume that even if we start batching blocks in the future, this algorithm will always perform really well, and we won't have to change it. In any case we will be guided by performance benchmarks to pinpoint algorithms that need improvement.

I would like to say that I'm ok, but I'm only half ok, as I said, the complexity of that algo isn't a problem after all because of the small size of the known blocks list.I'm in the school that we shouldn't trash an idea without proof. There are no benchmarks here, just intuitions... I can do a snippet, if you really want, and benchmark that.

The complexity were visually absurd, the code written didn't describe the usage of it. I just wanted to change that a little, that took me 5 minutes.. and too much time to defend it, that's not constructive.

adrien-zinger:
Btw, by curiosity I did a benchmark yesterday to check the theory. And yes, here is a proof that the current algorithm is a snail and that the proposed is for all use cases fast, very fast, and also in other use cases. Measurement vs Intuition.

Before that test I change just two little things (with 2 little modif, check `< log(len)` and fix unit tests).

[criterion.zip](https://github.com/massalabs/massa/files/8095843/criterion.zip)

PR #2310, opened 2022-02-17
Quickly disable `test_run_node`

yvan-sraka:


AureliaDolo (approved):

massa-client/src/tests/mod.rs
```diff
@@ -1,3 +1,3 @@
-#[cfg(target_feature = "local")]
```
yvan-sraka, line 1: In fact the test was already disabled but in a way a bit non-obvious (and accidental?) since feature `local` didn't exist...

AureliaDolo: That's weird

AureliaDolo (approved):

massa-client/src/tests/mod.rs
adrien-zinger: I did that, it wasn't documented, sorry :slightly_smiling_face: because one day we said that we didn't want to test that on the CI but only if we specify it in local.

adrien-zinger (approved):

yvan-sraka:
bors merge

yvan-sraka:
bors merge

massa-client/src/tests/mod.rs
AureliaDolo: Indeed x)

PR #2311, opened 2022-02-17
Reduce max_operations_fill_attempts

qdrn:


damip:
@qdrn note that this comes at a risk: if there is a lot of bad operations pending (which is the case currently), good operations might not go through unless they increase their fee

qdrn:
Yep I understand. I think that's what we wanted to do as quick fix. Didn't we ?

damip:
> Yep I understand. I think that's what we wanted to do as quick fix. Didn't we ?

We would go from
```
   max_operations_fill_attempts = 102400
```
to
```
    max_operations_fill_attempts = 10
```

Maybe there is a middle-ground ! 10 is *really* low. How about 100 ?

How about decreasing the max pool size ?

PR #2312, opened 2022-02-18
Replace the const AMOUNT_ZERO by a const fn

adrien-zinger:
The question was posed initially here https://github.com/massalabs/massa/pull/2294, I think it's cleaner that way to get a zero Amount.


damip (approved):

yvan-sraka:
bors merge

Issue #2313, opened 2022-02-21
Wrong config taken into account


damip:
In the current labnet branch (based on main after merging the config refactoring):

* in massa-models/src/node_configuration/default_testing.rs:
```
    pub static ref VERSION: Version = if cfg!(feature = "sandbox") {
        "SAND.0.0"
    } else {
        "TEST.7.0"
    }
```
* in massa-models/src/node_configuration/default.rs:
```
pub static ref VERSION: Version = "LABN.0.0".parse().unwrap();
```

When compiled with cargo build --release,  the node crashes:

```
error while bootstrapping: incompatible version: remote is running incompatible version: LABN.0.0 (local node version: TEST.7.0)
```

Basically the "local node version" seems to be loaded from default_testing while the bootstrap serving provides the production default.

Here is the first log line at launch:

```
2022-02-21T00:40:44.966038Z  INFO massa_node: Node version : TEST.7.0
```

adrien-zinger:
The problem is simple, in the main branch there is an error in the default.rs

instead of 

```rust
pub static ref VERSION: Version = "LABN.0.0".parse().unwrap();
```
I found

https://github.com/massalabs/massa/blob/12817eb402d4e60acaa4054c66f40d2dfeca35da/massa-models/src/node_configuration/default.rs#L59-L66

AureliaDolo:
I vote for a labnet feature

adrien-zinger:
In a PR I have that:
```rust
if cfg!(feature = "sandbox") {
    "SAND.0.0"
} else {
   "LABN.0.0"
}
```

I can do the following that is more precise and differentiate if you want a labnet or a testnet (with exactly the same configs but a different version), by default the node will run a testnet. With `cargo run --feature labnet` we have a labnet, and with `cargo run --feature sandbox` we have a sandbox.
```rust
if cfg!(feature = "sandbox") {
    "SAND.0.0"
} else if  cfg!(feature = "labnet") {
   "LABN.0.0"
} else {
   "TEST.7.0"
}
```

adrien-zinger:
With the PR in fact I get the reverse error, but it's because I have not configured my `bootstrap_list` correctly.
```
error while bootstrapping: incompatible version: remote is running incompatible version: TEST.7.0 (local node version: LABN.0.0)
```

More generally, by replacing exactly the `VERSION` variable in the default rs and building with `cargo run --release` I don't reproduce the error, I always get the result:
```zsh
[...]

PR #2317, opened 2022-02-21
Change default version to `LABN.0.0`

adrien-zinger:
Attempts to resolve the issue https://github.com/massalabs/massa/issues/2313.

With the change I get the good version, it verifies that the dispatch work correctly.

```zsh
2022-02-21T08:58:59.847604Z  INFO massa_node: Node version : LABN.0.0
2022-02-21T08:58:59.847690Z  INFO massa_bootstrap: Start bootstrapping from 158.69.120.215:31245
2022-02-21T08:59:00.058700Z  WARN massa_bootstrap: error while bootstrapping: io error: early eof
^C2022-02-21T08:59:01.983847Z  INFO massa_node: interrupt signal received in bootstrap loop
```

### DIFF
--- a/massa-models/src/node_configuration/default.rs
+++ b/massa-models/src/node_configuration/default.rs
@@ -60,7 +60,7 @@ lazy_static::lazy_static! {
         if cfg!(feature = "sandbox") {
             "SAND.0.0"
         } else {
-            "TEST.7.0"
+            "LABN.0.0"
         }
         .parse()
         .unwrap()


yvan-sraka (approved):
So where is defined current TestNet version?

adrien-zinger:
> > So where is defined current TestNet version?
> 
> It's updated anyway with the release script

How? The there is a CD that write code here?

AureliaDolo:
> How? The there is a CD that write code here?

I'd rather ask @damip

adrien-zinger:
> > How? The there is a CD that write code here?
> 
> I'd rather ask @damip

Finally, no CD script write the real version.

I agree that the version is written manually with [that kind of commits](https://github.com/massalabs/massa/commit/852d3ed94f5db6eaef5365c3a5b10eaeecb95b52) each time we need it. I don't like bots that write code. :robot:

damip:
> > > How? The there is a CD that write code here?
> > 
> > 
> > I'd rather ask @damip
> 
> Finally, no CD script write the real version.
> 
> I agree that the version is written manually with [that kind of commits](https://github.com/massalabs/massa/commit/852d3ed94f5db6eaef5365c3a5b10eaeecb95b52) each time we need it. I don't like bots that write code. robot

@adrien-zinger 

The labnet config is automatically adjusted by a script, then pushed in the labnet branch, and deployed on 2 servers. The adjustments are:

* here: https://github.com/massalabs/massa/blob/labnet/massa-models/src/node_configuration/default.rs#L39
* here: https://github.com/massalabs/massa/blob/labnet/massa-models/src/node_configuration/default.rs#L42
* here: https://github.com/massalabs/massa/blob/labnet/massa-models/src/node_configuration/default.rs#L49
* the initial peers file
* the massa-node/base_config/config.toml file (motsly for bootstrap nodes)

AureliaDolo:
> The problem we have is that despite this config, the labnet believes it is running on TEST.7.0 which is only defined in the testing_default. I don't get what is happening. Feel free to try by compiling and running the labnet branch

The start_marker seems wrong, it's `start_marker = 'pub static ref VERSION: Version = "'` whereas now it's https://github.com/massalabs/massa/blob/c9562428ee4f62173c0616c143b9e8148c4783ad/massa-models/src/node_configuration/default.rs#L59-L66

So, as there is no quote anymore, the marker doesn't apply.

And from what I understand, you discard the features in the config ?!

damip:
Problem solved: https://github.com/massalabs/massa/pull/2321

Closing

PR #2324, opened 2022-02-22
Harmonize bootstrap bootsrap peers and network bootstrap peers.

### LINKED ISSUE #2320: Make bootstrap peers consistent across bootstrap and network
A peer in the bootstrap_peers list in the config is not considered as bootstrap unless it is also marked as bootstrap in peers.json 

My idea to tackle that was to have in the GlobalBootstrapState returned by get_state instead of 
```rust
pub struct GlobalBootstrapState {
    pub peers: Option<BootstrapPeers>,
    ..
}
```
have something like
```rust
pub struct GlobalBootstrapState {
    pub peers: Option<BootstrapPeers>,
    pub bootstrap_peers: Vec<IpAddr>, // peers in the BootstrapSettings.bootstrap_list
    ..
}
```
or 
```rust
pub struct GlobalBootstrapState {
    pub peers: Vec<(IpAddr, bool)>, // true if ip in the BootstrapSettings.bootstrap_list
    ..
}
```

where in network worker
```rust
pub struct BootstrapPeers(pub Vec<IpAddr>);
```

And make sure that these peers are returned even if we don't need to bootstrap.

AureliaDolo:
Fix #2320


massa-network/src/peer_info_database.rs
```diff
@@ -274,7 +274,11 @@ impl PeerInfoDatabase {
     /// # Argument
     /// * cfg : network configuration
     /// * clock_compensation: sync with server
-    pub async fn new(cfg: &NetworkSettings, clock_compensation: i64) -> Result<Self, NetworkError> {
+    pub async fn new(
+        cfg: &NetworkSettings,
+        clock_compensation: i64,
+        initial_peers: Vec<(IpAddr, bool)>,
```
adrien-zinger, line 280: doc it :monocle_face:

adrien-zinger (approved):
Really nice, I appreciate the improvement and the change in the unit test :clap:

massa-bootstrap/src/lib.rs
```diff
@@ -46,13 +46,25 @@ pub struct GlobalBootstrapState {
...
-    /// list of network peers
-    pub peers: Option<BootstrapPeers>,
+    /// list of network and bootstrap peers, true if bootstrap
+    pub peers: Vec<(IpAddr, bool)>,
 
     /// state of the execution state
     pub execution: Option<BootstrapExecutionState>,
 }
 
+impl GlobalBootstrapState {
```
yvan-sraka, line 56: Not sure about the naming of this trait, but more generally my logic is "if we are already in module `FooBar`, no needs to name something `FooBarState`, just State is enough" but this could be for sure tackle in another issue!

yvan-sraka (approved):
LGTM! Nice work :sparkles:

massa-bootstrap/src/lib.rs
AureliaDolo: see #2327

AureliaDolo:
By the way with this PR we can remove bootstrap peers from the initial peers.json file 
@damip @sebastien-forestier @qdrn have you a preference for that ?

massa-bootstrap/src/lib.rs
```diff
@@ -46,13 +46,25 @@ pub struct GlobalBootstrapState {
     /// timestamp correction in milliseconds
     pub compensation_millis: i64,
 
-    /// list of network peers
-    pub peers: Option<BootstrapPeers>,
+    /// list of network and bootstrap peers, true if bootstrap
```
yvan-sraka, line 49: You mean that the Boolean of the pair is true if the current node do have bootstrapped from this peer?

massa-bootstrap/src/lib.rs
```diff
@@ -46,13 +46,25 @@ pub struct GlobalBootstrapState {
...
+    pub peers: Vec<(IpAddr, bool)>,
```
yvan-sraka, line 50: I would rather use a struct with named fields than a tuple!

massa-network/src/peer_info_database.rs
```diff
@@ -1,5 +1,20 @@
 // Copyright (c) 2021 MASSA LABS <info@massa.net>
 
+//! Peer management (conection counts and storage)
+//!
+//! If you wish to provide initial peers, fillup the initial_peers.json file
+//! with that kind of objects
+//! ```javascript
```
yvan-sraka, line 7: `json`! Otherwise, nice doc comment explanation here :+1:

massa-node/base_config/initial_peers.json
```diff
@@ -1,34 +1 @@
-[
```
yvan-sraka, line 1: So glad we get rid of this file!

yvan-sraka (approved):
Congrats! Simple and nice code refactor here :rainbow:

massa-bootstrap/src/lib.rs
AureliaDolo: no it means if this ip appeared in our bootstrap list

massa-bootstrap/src/lib.rs
AureliaDolo: any suggestion on naming ?

damip:
There is a fundamental concern here: not all bootstrap nodes are protocol peers

AureliaDolo:
> There is a fundamental concern here: not all bootstrap nodes should be protocol peers

1/ why ?

2/ maybe add that in the config to have a vec of (ip, pub key, bool) where the bool says if it's a node could solve that

AureliaDolo:
bors cancel

yvan-sraka:
bors cancel

massa-bootstrap/src/lib.rs
```diff
@@ -204,8 +216,10 @@ async fn get_state_internal(
         pos: Some(pos),
         graph: Some(graph),
         compensation_millis,
-        peers: Some(peers),
+
```
yvan-sraka, line 219: Weird space here ...

massa-bootstrap/src/lib.rs
yvan-sraka:
```rust
struct Peer {
    ip: IpAddr,
    bootsraped: bool,
}
```

yvan-sraka (approved):

massa-bootstrap/src/lib.rs
AureliaDolo: https://github.com/massalabs/massa/blob/bedecb5e3b5495354aa71f99c16b1f7c62fe497a/massa-network/src/network_worker.rs#L84-L88

PR #2332, opened 2022-02-24
Execution: improve worker

gterzian:


gterzian:
In response to https://github.com/massalabs/massa/pull/2249#discussion_r812720320

gterzian:
@damip ready for review. 

The idea is to make the loop simpler by having one point at the point where one waits for the next event, and having the rest be about processing that event. 

Also a few other small things(each is a separate commit).

damip:
Ok I got it :) 
Implemented in the original PR. Thanks

gterzian:
Ok, closing this one then.

Issue #2348, opened 2022-02-28
Remove Sync bound on ExecutionController


gterzian:
It would be nice to be able to remove [this `Sync` bound](https://github.com/massalabs/massa/pull/2249/files#diff-3144876461b45d2786ae8d8cffac6c101f2685b172fe1f1f1f04693ce8371a21R18), as it appears only necessary for use in API, and currently it results in some weirdness such as the use of a lock around the mpsc in the [test mock](https://github.com/massalabs/massa/pull/2249/files#diff-e2918f638ff66327361cf37c5260a2eb9d48e0ce8c7b4a5c8a016b452df0e3daR49).

One way to solve this may be to use an `Arc<Mutex<ExecutionController>>`, as opposed to a box, inside the API.

gterzian:
For now this is too much for now, because the `sync` bound is (among others?) also required because of the `await` method calls of anything that owns a controller. 

Could be done after https://github.com/massalabs/massa/issues/2179

Issue #2179, opened 2022-01-18
Reduce Async usage


gterzian:
Part of https://github.com/massalabs/massa/issues/2162

Good candidate for this is to only use `async` in relation to networking, or other components where lots of sub-tasks are spawned, and to keep the main components running on threads.

damip:
Sounds like it makes sense :) let's see if there is any devil in the details

gterzian:
@damip See my comment above, by the way I've addressed it for now via https://github.com/massalabs/massa/pull/2233/commits/f89df90e5a23025d6d7ca894969a3a6e4265e2db

damip:
> @damip See my comment above, by the way I've addressed it for now via [f89df90](https://github.com/massalabs/massa/commit/f89df90e5a23025d6d7ca894969a3a6e4265e2db)

Very good !
I'm just a bit scared by the "unwrap()" like here for example: https://github.com/massalabs/massa/commit/f89df90e5a23025d6d7ca894969a3a6e4265e2db#diff-7fd127a5266f7201f782ea024113f453dcfe21a8f157848affd4781e16c6e8fbR1115  given that "storage" should have internal cleanup systems, we should always be robust to something appearing/changing/disappearing from storage

gterzian:
> we should always be robust to something appearing/changing/disappearing from storage

Yes, the `unwrap` is there mostly because it's a work in progress.

damip:
remaining async usages:
* network: can be optimized but makes sense there
* protocol: could maybe be removes
* consensus: should be removed, but after the ledger unification that will massively reduce the complexity of the consensus module
* pool: this module might go away with further storage unification
* api: to be reviewed
* bootstrap: to be reviewed

gterzian:
> pool: this module might go away with further storage unification

Just for clarification, I don't think `pool` will go away, just like `consensus` didn't go away once we switched to using shared storage for blocks. 

The reason is that all the logic currently in `operations_pool` should not be transferred to storage, because we want the logic to remain isolated in a module like `pool`, where the shared storage remains only about storing/retrieving of objects. 

But pool does not need async, indeed.

gterzian:
I think `protocol` is probably a good first candidate to remove async.

Also note that at the level of `network`, we could run the network worker on a thread, while keeping the individual node workers async(since that is where the "real" networking happens). 

[...]

PR #2349, opened 2022-03-01
Execution: remove sync bound on controller

gterzian:
Close https://github.com/massalabs/massa/issues/2348


gterzian:
Ok never mind this is too much for now, because the `sync` bound is (among others?) also required because of the `await` method calls of anything that owns a controller. 

Could be done after https://github.com/massalabs/massa/issues/2179

PR #2359, opened 2022-03-02
More SCE ledger info in api and client.

### LINKED ISSUE #2330: Improve SCELedgerOutput use in API
I mapped to the old API input and output to the types. But we actually get more info from the new execution system addded in this PR. I gives us not only the final but also the active balances, bytecode etc... for ledger entries. SO as a followup we could update the API to make better use of it. I just didn't want to break everything

_Originally posted by @damip in https://github.com/massalabs/massa/pull/2249#discussion_r813058984_

AureliaDolo:
Fix #2330

### DIFF
--- a/massa-api/src/public.rs
+++ b/massa-api/src/public.rs
@@ -415,28 +415,48 @@ impl Endpoints for API<Public> {
         let pool_command_sender = self.0.pool_command_sender.clone();
         let compensation_millis = self.0.compensation_millis;
 
-        // todo make better use of SCE ledger info
-
         // map SCE ledger info and check for address length
         let sce_ledger_info = if addresses.len() as u64 > api_cfg.max_arguments {
             Err(ApiError::TooManyArguments("too many arguments".into()))
         } else {
             // get SCE ledger info
-            let mut sce_ledger_info: Map<Address, SCELedgerInfo> =
+            let mut sce_ledger_info: Map<Address, (SCELedgerInfo, SCELedgerInfo)> =
                 Map::with_capacity_and_hasher(addresses.len(), BuildMap::default());
             for addr in &addresses {
                 let active_entry = match self
                     .0
                     .execution_controller
                     .get_final_and_active_ledger_entry(addr)
-                    .1
                 {
-                    None => continue,
-                    Some(v) => SCELedgerInfo {
-                        balance: v.parallel_balance,
-                        module: Some(v.bytecode),
-                        datastore: v.datastore.into_iter().collect(),
-                    },
+                    (None, None) => continue,
+                    (None, Some(candidate)) => (
+                        SCELedgerInfo::default(),
+                        SCELedgerInfo {
+                            balance: candidate.parallel_balance,
+                            module: candidate.bytecode,
+                            datastore: candidate.datastore.into_iter().collect(),
+                        },
+                    ),
+                    (Some(final_entry), None) => (
+                        SCELedgerInfo {
+                            balance: final_entry.parallel_balance,
+                            module: final_entry.bytecode,
+                            datastore: final_entry.datastore.into_iter().collect(),
+                        },
+                        SCELedgerInfo::default(),
+                    ),
+                    (Some(final_entry), Some(candidate)) => (
+                        SCELedgerInfo {
+                            balance: final_entry.parallel_balance,
+                            module: final_entry.bytecode,
+                            datastore: final_entry.datastore.into_iter().collect(),
+                        },
+                        SCELedgerInfo {
+                            balance: candidate.parallel_balance,
+                            module: candidate.bytecode,
+                            datastore: candidate.datastore.into_iter().collect(),
+                        },
+                    ),
                 };
                 sce_ledger_info.insert(*addr, active_entry);
             }
@@ -522,6 +542,7 @@ impl Endpoints for API<Public> {
             // compile everything per address
             for address in addresses.into_iter() {
                 let state = states.remove(&address).ok_or(ApiError::NotFound)?;
+                let sce = sce_ledger_info.get(&address).cloned().unwrap_or_default();
                 res.push(AddressInfo {
                     address,
                     thread: address.get_thread(cfg.thread_count),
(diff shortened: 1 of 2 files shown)


massa-api/src/public.rs
```diff
@@ -479,17 +454,23 @@ impl Endpoints for API<Public> {
...
+            let mut final_sce_ledger_info: Map<Address, SCELedgerInfo> =
+                Map::with_capacity_and_hasher(addresses.len(), BuildMap::default());
+            let mut candidate_sce_ledger_info: Map<Address, SCELedgerInfo> =
+                Map::with_capacity_and_hasher(addresses.len(), BuildMap::default());
 
             let mut concurrent_getters = FuturesUnordered::new();
             for &address in addresses.iter() {
                 let mut pool_cmd_snd = pool_command_sender.clone();
                 let cmd_snd = cmd_sender.clone();
+                let exec_snd = execution_controller.clone();
```
AurelienFT, line 466: I'm skeptical about this clone you use it only to get a specific element could you just clone the element  L.493 ?

massa-api/src/public.rs
```diff
@@ -499,24 +480,75 @@ impl Endpoints for API<Public> {
...
+
+                    let (final_sce, candidate_sce) =
+                        match exec_snd.get_final_and_active_ledger_entry(&address) {
+                            (None, None) => (SCELedgerInfo::default(), SCELedgerInfo::default()),
+                            (None, Some(candidate)) => (
+                                SCELedgerInfo::default(),
+                                SCELedgerInfo {
+                                    balance: candidate.parallel_balance,
+                                    module: candidate.bytecode,
+                                    datastore: candidate.datastore.into_iter().collect(),
```
AurelienFT, line 500: Why not having the same type in ledger info and ledger entry ? For example having a BTreeMap on LedgerInfo to avoid this conversion very costly. (except if rust optimize it)

massa-models/src/api.rs
```diff
@@ -173,7 +174,12 @@ impl std::fmt::Display for AddressInfo {
         writeln!(f, "Address: {}", self.address)?;
         writeln!(f, "Thread: {}", self.thread)?;
         writeln!(f, "Sequential balance:\n{}", self.ledger_info)?;
-        writeln!(f, "Parallel balance:\n{}", self.sce_ledger_info)?;
+        writeln!(f, "Final Parallel balance:\n{}", self.final_sce_ledger_info)?;
+        writeln!(
+            f,
+            "Candidate Parallel balance:\n{}",
```
AurelienFT, line 180: As a user when I started to use the client I didn't understand the difference between candidate and final. Now that we split more of information between this two category it will be nice to have a line to explain when making a command that returning final and candidate things.

AurelienFT (approved):
Some concerns about the optimizations and documentation. Otherwise it's doing what was made before but splitted and didn't see any typo etc... So i'm ok with that

massa-api/src/public.rs
AureliaDolo: the issue is that we need these clone because async move

massa-api/src/public.rs
AureliaDolo: I don't know, maybe it's in the api specs @damip @yvan-sraka

massa-models/src/api.rs
AureliaDolo: Where do you expect that kind of doc ?

AurelienFT: At the top of the output of the command that display candidate and final things for example wallet info.

AureliaDolo: And in the get_addresses too ?

Maybe we should have a call to discuss what concepts need some documentation and add a vocabulary/FAQ/key concepts command that would answer these questions ? I feel like having it displayed every time is a little overkill

Or display it only the first time the command is executed ?

qdrn: I agree that it is more and more confusing as it is concepts that do not exist in other blockchains. Displaying this information all the time is a bit overkill tho'.

In the long run command line client will probably mostly be used to manage nodes on servers so it needs to be practical not necessarily extremely user friendly. I think what we are mostly lacking is high level documentation on the wiki.

AurelienFT: the problem is that the new people don't look at the wiki after they successfully stake. But yeah if they are interested by understanding the terms it can be in the wiki because not everyone is interested.

massa-api/src/public.rs
yvan-sraka: I don't know either ...

massa-api/src/public.rs
```diff
@@ -499,24 +480,75 @@ impl Endpoints for API<Public> {
...
+                                },
+                            ),
+                        };
+
+                    Result::<
+                        (
```
yvan-sraka, line 526: Wrap that return value into a struct could be neat!

massa-api/src/public.rs
```diff
@@ -413,41 +413,16 @@ impl Endpoints for API<Public> {
...
-                        module: Some(v.bytecode),
-                        datastore: v.datastore.into_iter().collect(),
-                    },
-                };
-                sce_ledger_info.insert(*addr, active_entry);
-            }
-            Ok(sce_ledger_info)
-        };
-
         let closure = async move || {
```
yvan-sraka, line 419: A good follow-up issue would be to really split this enormous closure into several chunks of code ...

yvan-sraka (approved):
LGTM with comment notes on future refactoring (even if it's a feature PR that should not be stuck by refactoring concerns I guess)!

massa-api/src/public.rs
AureliaDolo: #2397

massa-models/src/api.rs
AureliaDolo: #2398

massa-api/src/public.rs
AureliaDolo: #2399

massa-api/src/public.rs
AureliaDolo: #2399

AureliaDolo:
bors merge

PR #2360, opened 2022-03-02
Feature/docker

BatiGencho:
The MR contains a Dockerfile for containerizing the massa node (labnet). 
The build is to be run as follows:

DOCKER_BUILDKIT=1 docker build --file Dockerfile --target release --tag massa-node . (target dev is probably not very adequate as we dont want running the node using cargo run). Maybe we should remove it ?

I have tested it and it works quite well. I am welcome to suggestions regarding adding further options such as ENV variables, mapping config folders etc.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -4,3 +4,6 @@
 [submodule "docs"]
 	path = docs
 	url = https://github.com/massalabs/massa.wiki.git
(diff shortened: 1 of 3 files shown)


damip:
sorry, this auto-closed with the labnet branch somehow

PR #2361, opened 2022-03-02
Correct execution truncation

damip:
Truncation only looked into active slots, but it did not look into the final slots that were ready for execution, which caused premature truncation in some cases.

### DIFF
--- a/massa-execution-worker/src/execution.rs
+++ b/massa-execution-worker/src/execution.rs
@@ -154,7 +154,11 @@ impl ExecutionState {
     /// # Arguments
     /// * active_slots: A HashMap mapping each active slot to a block or None if the slot is a miss
     /// * ready_final_slots:  A HashMap mapping each ready-to-execute final slot to a block or None if the slot is a miss
-    pub fn truncate_history(&mut self, active_slots: &HashMap<Slot, Option<(BlockId, Block)>>, ready_final_slots: &HashMap<Slot, Option<(BlockId, Block)>>) {
+    pub fn truncate_history(
+        &mut self,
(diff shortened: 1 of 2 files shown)


massa-execution-worker/src/execution.rs
```diff
@@ -152,15 +152,21 @@ impl ExecutionState {
     /// Slots after that point will need to be (re-executed) to account for the new sequence.
     ///
     /// # Arguments
-    /// * active_slots: A HashMap mapping each slot to a block or None if the slot is a miss
-    pub fn truncate_history(&mut self, active_slots: &HashMap<Slot, Option<(BlockId, Block)>>) {
+    /// * active_slots: A HashMap mapping each active slot to a block or None if the slot is a miss
+    /// * ready_final_slots:  A HashMap mapping each ready-to-execute final slot to a block or None if the slot is a miss
```
AureliaDolo, line 156: Ready to execute but not yet executed ?

AureliaDolo (approved):

massa-execution-worker/src/execution.rs
damip: I improved the PR description a bit. Basically we need t o list all blocks listed for execution (both final and active) otherwise when a block becomes final, it is removed from ACTIVE , and the truncation detects that an early active block is missing and truncates. But actually it is in ready_final_slots

adrien-zinger: Can we test that?

damip:
bors merge

qdrn (approved):

Issue #2308, opened 2022-02-17
Prune shared storage


gterzian:
Part of https://github.com/massalabs/massa/issues/2178

Currently it's mostly `graph` that is in charge of pruning the block it manages locally, for example at:

https://github.com/massalabs/massa/blob/ec8dd82a86b1d1602d342a0610d31f9dae67d89a/massa-graph/src/block_graph.rs#L3200

Presumably it could instead prune the shared storage.

damip:
I think we might need both: storage needs a way of pruning but graph will probably also maintain a local list of "active" and "final-but-still-needed" blocks

gterzian:
> also maintain a local list of "active" and "final-but-still-needed" blocks that also needs to be pruned

Yes `consensus` will keep various things that relate to its local logic, and as those are pruned I think storage can be pruned alongside, by `consensus`. 

Other than consensus, is there another component that would be involved in pruning?

damip:
> > also maintain a local list of "active" and "final-but-still-needed" blocks that also needs to be pruned
> 
> Yes `consensus` will keep various things that relate to its local logic, and as those are pruned I think storage can be pruned alongside.
> 
> Other than consensus, is there another component that would be involved in pruning?

no, I think only graph management will be responsible for pruning final blocks

PR #2365, opened 2022-03-03
Prune shared storage

gterzian:
Close https://github.com/massalabs/massa/issues/2308

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1818,6 +1818,7 @@ dependencies = [
  "num",
  "num-traits 0.2.14",
  "num_enum",
+ "parking_lot 0.11.2",
  "pretty_assertions",
  "rand 0.8.5",
  "reqwest",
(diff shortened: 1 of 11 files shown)


gterzian:
@damip I'm wondering how we should go about pruning discarded final blocks. Should we perhaps move that to execution, to after they've been executed?

gterzian:
@damip ping on the above.

gterzian:
How about the following:

1. When graph prunes an active block, it doesn't delete it(unless see below), instead it marks it as `pruned`. 
2. When execution executes a block, it marks it as `executed`. 

Then, when execution executes a blocked marked `pruned`, it deletes it. 

And when graph prunes a blocks marked `executed`, it also deletes it. 

cc @damip

AureliaDolo:
I may be missing something but where are you pruning old final blocks ?

gterzian:
Closing this, as the work has been done in https://github.com/massalabs/massa/pull/2428

gterzian:
Closing this, as the work has been done in https://github.com/massalabs/massa/pull/2428

PR #2375, opened 2022-03-04
Shard the worker, commands, events

adrien-zinger:
Move implementation of the event annd command management in dedicated
files to document later.


AureliaDolo (approved):
Great !! 

Maye some more doc but this could be left aside for now

yvan-sraka (approved):
Such nice refactoring 💖

PR #2376, opened 2022-03-04
Fix security audit - get rid of chrono and use time directly

AurelienFT:
Chrono still use a very old version of time (0.1 now it's 0.3). They have a PR running since months for updating but it seems that there is communication problems that lead to long time development. The PR : https://github.com/chronotope/chrono/pull/639

This break our CI like a lot of others projects that use `cargo audit`. A lot of projects that use chrono to do things that are now implemented in the new version of `time` has switched to use `time` directly instead of using tokio. Some examples : 
- https://github.com/brave/brave-browser/issues/20568
- https://github.com/meilisearch/milli/pull/450

So as we also only use chrono to make things that now possible in `time` which is more maintained I suggest in this PR a change to use `time` instead of `chrono`.

### DIFF
--- a/massa-time/src/lib.rs
+++ b/massa-time/src/lib.rs
@@ -1,17 +1,17 @@
 // Copyright (c) 2022 MASSA LABS <info@massa.net>
 
 mod error;
-use chrono::{self, DateTime, NaiveDateTime, Utc};
 pub use error::TimeError;
+use serde::{Deserialize, Serialize};
 use std::fmt;
 use std::time::{Duration, SystemTime, UNIX_EPOCH};
 use std::{
     convert::{TryFrom, TryInto},
     str::FromStr,
 };
+use time::format_description::well_known::Rfc3339;
+use time::OffsetDateTime;
 use tokio::time::Instant;
-
-use serde::{Deserialize, Serialize};
 
 /// Time structure used every where.
 /// Millis since 01/01/1970.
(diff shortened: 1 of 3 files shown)


AureliaDolo (approved):
Just the clippy warning left to fix and LGTM

By the way if you have an idea to better implement days_hours_mins_secs

AurelienFT:
> Just the clippy warning left to fix and LGTM
> 
> By the way if you have an idea to better implement days_hours_mins_secs

Good catch ! Fixed ! Thank you.

AureliaDolo (dismissed):
And I'm wondering why there is an unwrap in to_utc_string but eh

massa-time/src/lib.rs
```diff
@@ -338,16 +338,12 @@ impl MassaTime {
...
     /// ```
     pub fn to_utc_string(self) -> String {
-        let naive = NaiveDateTime::from_timestamp(
-            (self.to_millis() / 1000) as i64,
-            ((self.to_millis() % 1000) * 1_000_000) as u32,
-        );
-        let datetime: DateTime<Utc> = DateTime::from_utc(naive, Utc);
-        format!("{}", datetime)
+        let naive = OffsetDateTime::from_unix_timestamp(self.to_millis() as i64).unwrap();
+        naive.format(&Rfc3339).unwrap().to_string()
```
adrien-zinger, line 346: Fix that warning

adrien-zinger (changes requested):

massa-time/src/lib.rs
AurelienFT: Done. Thank you

AureliaDolo (approved):

adrien-zinger (approved):

AurelienFT:
bors merge

PR #2377, opened 2022-03-04
Split interaction with API from client to create an SDK.

### LINKED ISSUE #2280: RPC Abstraction should be usable by external projects
The client use a great interface to communicate with the RPC API. If I want to develop a tool for massa and I want to use the RPC API I would like to use this interface but it's in a binary not a library. So I have to rewrite mine (or copy paste).

The solution would be to split the interactions with the RPC API to make it a separate crate as a library and publish it so that everyone can use it as a base to code their own tool in Rust.

/

**Additional context**
/

AurelienFT:
Fix #2280 

### Motivations

Currently, if someone wants to develop a tool on massa he will have to create his own interactions with the API but we already have one that we use for the client. I think we could make those interactions as a new package that can be consider as a SDK.

### Follow-up questions

I think it's a package to be published so that people can directly refer to it easily. Two problems I see with the publish : 
- Need a better documentation
- What about the versioning ? Should we have a version for each testnet ? Auto update on CI ?

### DIFF
--- a/massa-sdk/src/lib.rs
+++ b/massa-sdk/src/lib.rs
@@ -142,14 +142,14 @@ impl RpcClient {
         self.call_method("get_status", "NodeStatus", ()).await
     }
 
-    pub async fn _get_cliques(&self) -> RpcResult<Vec<Clique>> {
+    pub(crate) async fn _get_cliques(&self) -> RpcResult<Vec<Clique>> {
         self.call_method("get_cliques", "Vec<Clique>", ()).await
     }
 
     // Debug (specific information)
 
     /// Returns the active stakers and their roll counts for the current cycle.
-    pub async fn _get_stakers(&self) -> RpcResult<Map<Address, u64>> {
+    pub(crate) async fn _get_stakers(&self) -> RpcResult<Map<Address, u64>> {
         self.call_method("get_stakers", "Map<Address, u64>", ())
             .await
     }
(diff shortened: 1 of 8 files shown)


yvan-sraka (dismissed):
I'm not sure of the name `massa-sdk`, otherwise LGTM!

massa-sdk/src/lib.rs
```diff
@@ -172,36 +176,30 @@ impl RpcClient {
...
     /// Get the block graph within the specified time interval.
     /// Optional parameters: from <time_start> (included) and to <time_end> (excluded) millisecond timestamp
-    pub(crate) async fn _get_graph_interval(
+    pub async fn _get_graph_interval(
         &self,
         time_interval: TimeInterval,
     ) -> RpcResult<Vec<BlockSummary>> {
         self.call_method("get_graph_interval", "Vec<BlockSummary>", time_interval)
             .await
     }
```
adrien-zinger, line 192: Not sure. Maybe functions like that doesn't have to begin with a _ if it's pub ?

adrien-zinger (commented):
We can talk about versioning yes and also about a design more "auto-generated" with proc macros.

I didn't test but it looks great!

massa-sdk/src/lib.rs
AurelienFT: Agree. Just made the modification.

yvan-sraka:
I think the two interesting question @AurelienFT tackled here are: 

- how do we version API?

A quick answer would be using crate versioning of both API and SDK that would not be the same number as TestNet version (since a Protocol upgrade for example de not necessarily break the API contract).

- do we start uploading pieces of our software on crates.io to let community (and core team) more easily hack with it?

I have in mind of course the client and wallet crates!

What do you think about it @damip?

AureliaDolo (approved):

AureliaDolo:
> * Need a better documentation

The api part is quite documented in the wiki, but the client is only documented in the help. Do you have any idea where it would be better to document it ?

Most command are only using an api endpoint

AurelienFT:
bors merge

PR #2384, opened 2022-03-07
Operation batch quest

adrien-zinger:
I'm storing here the specification and the main branch for the feature integration "OperationBatch" that should optimize the communication through nodes.

---
- Reviewing the technical specification (start with [massa-network-exports/src/commands.rs](https://github.com/massalabs/massa/blob/feature/operations_batch/massa-network-exports/src/commands.rs))
https://github.com/massalabs/massa/blob/441cdab95ba10a6f12f751bc9a859c9826de0843/massa-network-exports/src/commands.rs#L4-L13
---

- Protocol  worker behavior
https://github.com/massalabs/massa/blob/441cdab95ba10a6f12f751bc9a859c9826de0843/massa-protocol-worker/src/node_info.rs#L47-L74
- Implement functions like `insert_wanted_block`, `insert_known_block`, etc. (node_info.rs)
- Implement the propagation
https://github.com/massalabs/massa/blob/441cdab95ba10a6f12f751bc9a859c9826de0843/massa-protocol-worker/src/protocol_worker.rs#L467-L493
- Update the local status and the `node(s)_info` on receive the operations
https://github.com/massalabs/massa/blob/441cdab95ba10a6f12f751bc9a859c9826de0843/massa-protocol-worker/src/protocol_worker.rs#L1267-L1280

---

- Replace unit tests with the new configuration.

### DIFF
--- a/massa-client/src/settings.rs
+++ b/massa-client/src/settings.rs
@@ -24,3 +24,9 @@ pub struct DefaultNode {
     pub private_port: u16,
     pub public_port: u16,
 }
+
+#[cfg(test)]
+#[test]
+fn test_load_client_config() {
+    let _ = *SETTINGS;
+}
--- a/massa-execution-worker/src/execution.rs
+++ b/massa-execution-worker/src/execution.rs
@@ -154,7 +154,11 @@ impl ExecutionState {
     /// # Arguments
     /// * active_slots: A HashMap mapping each active slot to a block or None if the slot is a miss
     /// * ready_final_slots:  A HashMap mapping each ready-to-execute final slot to a block or None if the slot is a miss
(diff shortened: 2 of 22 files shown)


AureliaDolo:
Not sure about it but maybe we could try to have some generic struct
```rust
struct Snapshot<U: impl PreHashed>  {
     known: Map<U, (bool, instant)>,
    wanted: Map<U, instant>, 
    asked: Map<U, instant>, 
}

impl SnapShot<U: impl Prehashed> {
    /// if there are too many items, prune them
    fn prune(&mut self, cfg: SnapshotCOnfig) {}
    
    /// these three function would update the timestamps if needed and all the other updates too
    fn insert_known(&mut self, item: U, knows: bool) {}
    fn insert_wanted(&mut self, item: U) {}
    fn insert_asked(&mut self, item: U) {} 
    
    // the remove functions too
    
    fn knows(&self, item: U) -> bool {}
}

struct  SnapshotConfig {
    max_known_item_count: usize,
    max_wanted_item_count: usize,
    max_asked_item_count: usize,
}
```

Issue #2373, opened 2022-03-04
Change massa cryptographic functions


Eitu33:
## Context

We have been studying alternative crypto schemes in an effort to determine if they could improve the performances.

You can check the benchmarks results [here](https://github.com/massalabs/massa/issues/1841#issuecomment-1057923069).

It was concluded that using Schnorr (from [secp256k1](https://docs.rs/secp256k1/latest/secp256k1/)) as signature and [Blake3](https://docs.rs/blake3/latest/blake3/) as hash would result in an improvement when compared to what is currently used, Secp256k1 and Sha256.

### Schnorr vs Secp256k1
- no loss in performance
- gain in features

### Blake3 vs Sha256
- no loss in security
- gain in performance

## To Do

Update `massa-hash`

- [x] functions
- [x] tests
- [x] doctests

Update `massa-signature`

- [x] functions
- [x] tests
- [x] doctests

## Related Issues

- https://github.com/massalabs/massa/issues/1841
- https://github.com/massalabs/massa/issues/2370

AureliaDolo:
for reference https://github.com/massalabs/massa/pull/1584

qdrn:
bump

PR #2241, opened 2022-02-03
Prefix hash string representation.

AureliaDolo:
Fix #1105

If there is no prefix, it will work just as it used to.

* [x] ADR - Address
* [x] PRI - Private key
* [x] PUB - Public Key
* [x] SIG - Signature
* [x] END - Endorsement Id
* [x] OPE - Operation Id
* [x] BLO - Block id
* [x] SCE - SC output event id
* [x] NOD - Node id

## Example outputs
`cargo run --features hash-prefix -- wallet_info`
```shell
Private key: PRI-xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx
Public key: PUB-xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx
Address: ADR-xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx
```

 `cargo run -- wallet_info`
``` shell
Private key: xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx
Public key: xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx
Address: xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx
```

`cargo run --features hash-prefix --  --json wallet_info`
```
{
  "keys": {
    "xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx": [
      "xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx",
      "xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx"
    ],
    "xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx": [
      "xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx",
      "xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx"
    ],
[...]

PR #2402, opened 2022-03-10
2373 change massa cryptographic functions

Eitu33:
https://github.com/massalabs/massa/issues/2373


massa-consensus-worker/src/tests/scenario_roll.rs
```diff
@@ -457,7 +457,6 @@ async fn test_roll() {
     .await;
 }
 
-#[ignore]
```
AureliaDolo, line 460: :+1:

massa-signature/src/signature_impl.rs
```diff
@@ -722,92 +722,68 @@ impl<'de> ::serde::Deserialize<'de> for Signature {
...
+/// let public_key: PublicKey = derive_public_key(&private_key);
+/// let data = Hash::compute_from("Hello World!".as_bytes());
+/// let signature = sign(&data, &private_key).unwrap();
+/// let verification: bool = verify_signature(&data, &signature, &public_key).is_ok();
+/// ```
 pub fn verify_signature(
     hash: &Hash,
     signature: &Signature,
     public_key: &PublicKey,
 ) -> Result<(), MassaHashError> {
```
damip, line 774: it's a bit strange that we return an error type named "MassaHashError" on a signature function

massa-hash/src/hash.rs
```diff
@@ -105,11 +116,7 @@ impl Hash {
     /// let deserialized: Hash = Hash::from_bytes(&serialized).unwrap();
     /// ```
     pub fn from_bytes(data: &[u8; HASH_SIZE_BYTES]) -> Result<Hash, MassaHashError> {
```
damip, line 118: Followup: this shouldn't return a Result

massa-hash/src/hash.rs
```diff
@@ -54,8 +65,9 @@ impl Hash {
     /// let serialized = hash.to_bytes();
     /// ```
     pub fn to_bytes(&self) -> [u8; HASH_SIZE_BYTES] {
-        use bitcoin_hashes::Hash;
-        *self.0.as_inner()
+        // note: this should return &[u8; HASH_SIZE_BYTES]
```
damip, line 68: Followup: return a reference

massa-signature/src/signature_impl.rs
```diff
@@ -78,7 +77,8 @@ impl PrivateKey {
     /// let serialized = private_key.to_bytes();
     /// ```
     pub fn to_bytes(&self) -> [u8; PRIVATE_KEY_SIZE_BYTES] {
-        *self.0.as_ref()
+        // note: should return a ref to respect conventions
```
damip, line 80: followup:: return a ref

massa-signature/src/signature_impl.rs
```diff
@@ -545,7 +545,7 @@ impl Signature {
     /// let serialized = signature.to_bytes();
     /// ```
     pub fn to_bytes(&self) -> [u8; SIGNATURE_SIZE_BYTES] {
```
damip, line 547: followup: return ref

damip (dismissed):

massa-signature/Cargo.toml
```diff
@@ -11,7 +11,7 @@ bitcoin_hashes = "0.10"
 bs58 = { version = "0.4", features = ["check"] }
 displaydoc = "0.2"
 rand = "0.8"
-secp256k1 = { version = "0.20", features = ["rand-std"] }
+secp256k1 = { version = "0.21.3", features = ["rand-std", "global-context"] }
```
AurelienFT, line 14: We avoid to specify patch versions

AurelienFT (dismissed):

AurelienFT (dismissed):

AurelienFT (dismissed):

massa-hash/src/hash.rs
```diff
@@ -102,14 +111,10 @@ impl Hash {
...
-    /// let deserialized: Hash = Hash::from_bytes(&serialized).unwrap();
+    /// let deserialized: Hash = Hash::from_bytes(&serialized);
     /// ```
-    pub fn from_bytes(data: &[u8; HASH_SIZE_BYTES]) -> Result<Hash, MassaHashError> {
-        use bitcoin_hashes::Hash;
-        Ok(Hash(
-            bitcoin_hashes::sha256::Hash::from_slice(&data[..])
-                .map_err(|err| MassaHashError::ParsingError(format!("{}", err)))?,
-        ))
+    pub fn from_bytes(data: &[u8; HASH_SIZE_BYTES]) -> Hash {
```
AureliaDolo, line 116: What happens if we receive some random bytes that we try to parse as a hash ?

Eitu33: The function will parse a `Hash` no matter what, if you wonder why the previous implementation returned a `Result<T, E>` it was because `bitcoin_hashes::sha256::Hash::from_slice` does not take a fixed-sized array and therefore has to check for the length (see [here](https://docs.rs/bitcoin_hashes/latest/src/bitcoin_hashes/sha256.rs.html#133-141)). Since Blake3 conversion takes a fixed sized array as input it is not necessary to have that kind of check or error.

massa-signature/src/error.rs
```diff
@@ -0,0 +1,17 @@
...
+use displaydoc::Display;
+use thiserror::Error;
+
+#[non_exhaustive]
+#[derive(Display, Error, Debug)]
+pub enum MassaSignatureError {
+    /// parsing error : {0}
+    ParsingError(String),
+
+    /// error forwarded by engine: {0}
```
AureliaDolo, line 12:
```suggestion
    /// secp256k1 engine error: {0}
```

AurelienFT: It's schnorr engine no ?

AureliaDolo: good question I took the crate name

AurelienFT (approved):

massa-signature/src/error.rs
Eitu33: saying secp256k1 engine error is the right thing to do, schnorr is just a part of what is handled by the crate engine

damip (changes requested):
do not merge yet until the tooling is ready

AureliaDolo:
> do not merge yet until the tooling is ready

Or we could hide this change behind a feature to avoid the heavy cost of maintaining a PR up to date

damip:
> > do not merge yet until the tooling is ready
> 
> Or we could hide this change behind a feature to avoid the heavy cost of maintaining a PR up to date

The "feature" approach has heavy drawbacks as well:
* this PR touches many places in the codebase. Having a `#cfg(XXX)` at every change would make the codebase heavier, more complicated, less readable, harder to maintain
* the "deactivated" feature parts will still have to be maintained, and new crypto uses must not forget to use branched  `#cfg(XXX)`  when necessary

AureliaDolo:
>     * this PR touches many places in the codebase. Having a `#cfg(XXX)` at every change would make the codebase heavier, more complicated, less readable, harder to maintain

Quite a lot of changes are due to a change in a method signature, not the cryptographic functions change itself. I would keep it as it was for now. So the switches would apply to :
* the imports
* the methods content
* the harcoded tests values (I would even refactor to limit the number of hardcoded values)
* maybe I'm missing something  

>     * the "deactivated" feature parts will still have to be maintained, and new crypto uses must not forget to use branched  `#cfg(XXX)`  when necessary

I don't see your point. The current crypto methods hasn't been touched in quite a while so I don't see how this would be different. 

>     * it requires a heavier setup

In my opinion the hassle of maintaining a pr up to date is heavier. Here it would allow to test the tooling with the new version while still having the old one by default

>     * it will require a followup refactoring to remove the feature

Just adding the "new_hash" feature to default feature. The code cleaning part could be done little by little with no real pros as it won't be compiled nor used anywhere.

damip:
@Eitu33  can you correct conflicts and merge main into this branch ?
I will launch a labnet so we can start the tooling migration

Eitu33:
@damip Will do so on Monday. I'll also take some time just to make sure everything works properly.

damip:
@Eitu33  @qdrn  While we are at it, let's add prefixing and version encoding. Otherwise there will be more tooling updates later, it's a big overhead.

damip:
@Eitu33  the CI seems to fail

Eitu33:
By the way @damip, about the address prefix, isn't it already handled by https://github.com/massalabs/massa/pull/2241 ?

massa-models/src/address.rs
```diff
@@ -14,40 +14,62 @@ use std::str::FromStr;
...
-            write!(f, "{}", self.0.to_bs58_check())
-        }
+        // note: might want to allocate the vector with capacity
+        // in order to avoid re-allocation
+        let mut bytes: Vec<u8> = ADDRESS_VERSION.to_varint_bytes();
+        bytes.extend(self.0.to_bytes());
+        write!(
+            f,
+            "{}{}",
+            ADDRESS_PREFIX,
```
AurelienFT, line 30: I think just an A in front of the address will not help people to distinguish when reading fast. Maybe add a dash between the A and the address can help.

massa-models/src/address.rs
```diff
@@ -114,9 +136,7 @@ impl Address {
     /// assert_eq!(address, res_addr);
     /// ```
     pub fn from_bytes(data: &[u8; ADDRESS_SIZE_BYTES]) -> Result<Address, ModelsError> {
```
AurelienFT, line 138: The signature of `Hash::from_bytes` doesn't return errors anymore so there is no case where this function return error.  (and same for all ids in the other files)

damip (changes requested):
preventing accidental merge for now

massa-models/src/address.rs
AurelienFT: My concern was that we can have A in the address so people can just consider it as a part of the address. But I see your point

damip (approved):

damip:
bors merge

damip (changes requested):
breaking config files

damip:
bors stop

damip (approved):

damip:
bors merge

PR #2408, opened 2022-03-11
Remove static log limitation

AurelienFT:
I saw that you added log limitation on all crates in this commit @damip :    https://github.com/massalabs/massa/commit/051a3a3a52efad46b9ca850da7c2bef85f653cc4

It has been removed by @yvan-sraka here : https://github.com/massalabs/massa/commit/777d09c5537989e93f524ce00391ddb4716f0152

But it's still in two crate that prevent the whole node to have access to trace log.

Do we remove the last two crate that have this by merging this PR or we add them add back eveywhere and why ? And also we have documentation to tell that it's not possible to have trace anymore

### DIFF
--- a/massa-ledger/Cargo.toml
+++ b/massa-ledger/Cargo.toml
@@ -26,5 +26,5 @@ pretty_assertions = "1.0"
 serial_test = "0.5"
 
 [features]
-instrument = ["tokio/tracing", "massa_consensus_exports/instrument", "massa_graph/instrument", "massa_models/instrument", "massa_pool/instrument", "massa_proof_of_stake_exports/instrument", "massa_protocol_exports/instrument", "massa_time/instrument"]
+instrument = ["tokio/tracing", "massa_models/instrument", "massa_time/instrument"]
 testing = []


adrien-zinger (commented):
It looks like something we forgot when we have add the tokio tracing. What do you think about that @yvan-sraka ? :thinking:

massa-ledger/Cargo.toml
```diff
@@ -13,7 +13,7 @@ serde = { version = "1.0", features = ["derive"] }
 serde_json = "1.0"
 thiserror = "1.0"
 tokio = { version = "1.11", features = ["full"] }
```
yvan-sraka, line 15: Since here we depend on `tokio`, we need to propagate the `instrument` feature in this crate too!

yvan-sraka (changes requested):
LGTM! Just need a little change to complete the fix :+1:

massa-ledger/Cargo.toml
AurelienFT: Done :)

AureliaDolo (approved):

AurelienFT:
bors merge

PR #2323, opened 2022-02-22
Implement whitelist api/client commands

AureliaDolo:
Fix #1946 

* [x] api
* [x] client

PR #2409, opened 2022-03-14
Feature/api/whitelist

AureliaDolo:
related #2323


massa-client/src/cmds.rs
```diff
@@ -73,6 +73,20 @@ pub enum Command {
     )]
     node_testnet_rewards_program_ownership_proof,
 
+    #[strum(
+        ascii_case_insensitive,
+        props(args = "[IpAddr]"),
+        message = "whitelist a given IP addresses"
```
AurelienFT, line 79:
```suggestion
        message = "whitelist a given IP address"
```

massa-client/src/cmds.rs
```diff
@@ -73,6 +73,20 @@ pub enum Command {
...
+    )]
+    node_whitelist,
+
+    #[strum(
+        ascii_case_insensitive,
+        props(args = "[IpAddr]"),
+        message = "remove from whitelist a given IP addresses"
```
AurelienFT, line 86:
```suggestion
        message = "remove from whitelist a given IP address"
```

massa-network-worker/src/peer_info_database.rs
```diff
@@ -336,6 +336,78 @@ impl PeerInfoDatabase {
...
+            };
+
+            if old_pt == PeerType::WhiteListed {
+                continue;
+            }
+            // update global connection counts by peer type
+            let peer = *self.peers.get(&ip).unwrap(); // filled just before
+            if peer.active_out_connection_attempts > 0 {
+                self.decrease_global_active_out_connection_attempt_count(old_pt, &ip)?;
+                self.increase_global_active_out_connection_attempt_count(
```
AurelienFT, line 361: When you create a new peer L.349 you want to call this function ? Because here it's not trigger and the next ones that increase the whitelist count also.

massa-network-worker/src/peer_info_database.rs
```diff
@@ -336,6 +336,78 @@ impl PeerInfoDatabase {
...
+            } else {
+                return Ok(());
+            };
+
+            if old_pt != Default::default() {
+                // update global connection counts by peer type
+                let peer = *self.peers.get(&ip).unwrap(); // filled just before
+                if peer.active_out_connection_attempts > 0 {
+                    self.decrease_global_active_out_connection_attempt_count(old_pt, &ip)?;
+                    self.increase_global_active_out_connection_attempt_count(
```
AurelienFT, line 393: I don't understand why you are updating the count for the default of peertype (which is standard if I remember well) if you filtered just above that type of peertype. Maybe worth some comments on why you filter to edit the counts and why you edit the default ones.

massa-client/src/cmds.rs
AureliaDolo: same as above

massa-network-worker/src/peer_info_database.rs
AureliaDolo: I don't understand your comment

massa-network-worker/src/peer_info_database.rs
AureliaDolo: because the peer isn't whitelist anymore, so as it became standard, we need to update the global counters

massa-client/src/cmds.rs
```diff
@@ -35,14 +35,14 @@ pub enum Command {
     #[strum(
         ascii_case_insensitive,
         props(args = "[IpAddr]"),
-        message = "unban a given IP addresses"
+        message = "unba given IP addresses"
```
AurelienFT, line 38:
```suggestion
        message = "unban given IP addresses"
```

massa-network-worker/src/peer_info_database.rs
AurelienFT: When you create a peer in this function you set the old peer type to Whitelisted so we will not go into this condition but if you have a new Whitelist peer you want to increase the count of whitelist peers no ?

AureliaDolo: old peer type is never set to whitelist. If the peer was whitelist before, there is no need to update the global counters but if it was something else we want to decrease the old peer type counters and increase the whitelist ones

massa-network-worker/src/peer_info_database.rs
AurelienFT: Remove node_id from whitelist.

massa-network-worker/src/peer_info_database.rs
AurelienFT: L.351 after create a new peer if it don't exist the old peer type is set to whitelist.

massa-network-worker/src/peer_info_database.rs
```diff
@@ -336,6 +336,76 @@ impl PeerInfoDatabase {
...
+                if pt == PeerType::WhiteListed {
+                    continue;
+                }
+                peer.peer_type = PeerType::WhiteListed;
+                pt
+            } else {
+                let mut p = PeerInfo::new(ip, false);
+                p.peer_type = PeerType::WhiteListed;
+                self.peers.insert(ip, p);
+                continue;
```
AurelienFT, line 352: I don't understand why you continue here. You don't want to call : `self.increase_global_active_out_connection_attempt_count(PeerType::WhiteListed, &ip)?` to increase the count now that there is a new whitelist peer ?

AureliaDolo: Because we didn't know this peer so there is no connection counter to update

yvan-sraka (approved):

AureliaDolo:
bors merge

PR #2428, opened 2022-03-16
Rebase storage

gterzian:
Rebasing https://github.com/massalabs/massa/pull/2233

### DIFF
--- a/massa-consensus-worker/src/consensus_worker.rs
+++ b/massa-consensus-worker/src/consensus_worker.rs
@@ -147,8 +147,18 @@ impl ConsensusWorker {
         // notify execution module of current blockclique and final blocks
         // we need to do this because the bootstrap snapshots of the executor vs the consensus may not have been taken in sync
         // because the two modules run concurrently and out of sync
-
-        // TODO: use shared-storage.
+        channels.execution_controller.update_blockclique_status(
+            block_db.get_all_final_blocks(),
+            block_db
+                .get_blockclique()
+                .into_iter()
+                .filter_map(|block_id| {
+                    block_db
+                        .get_active_block(&block_id)
+                        .map(|a_block| (a_block.slot.clone(), block_id))
+                })
+                .collect(),
+        );
 
         Ok(ConsensusWorker {
             genesis_public_key,
@@ -1180,10 +1190,32 @@ impl ConsensusWorker {
         // get blockclique
         let blockclique_set = self.block_db.get_blockclique();
 
-        // notify Execution
-        {
-            // TODO: use shared storage.
-        }
+        // notify execution
+        self.channels
+            .execution_controller
+            .update_blockclique_status(
+                new_final_block_ids
+                    .clone()
+                    .into_iter()
+                    .filter_map(|b_id| {
+                        if let Some(a_b) = self.block_db.get_active_block(&b_id) {
+                            if a_b.is_final {
+                                return Some((a_b.slot.clone(), b_id));
+                            }
+                        }
+                        None
+                    })
+                    .collect(),
+                blockclique_set
+                    .clone()
+                    .into_iter()
+                    .filter_map(|block_id| {
+                        self.block_db
+                            .get_active_block(&block_id)
+                            .map(|a_block| (a_block.slot.clone(), block_id))
+                    })
+                    .collect(),
+            );
 
         // Process new final blocks
         let mut new_final_ops: Map<OperationId, (u64, u8)> = Map::default();
--- a/massa-consensus-worker/src/consensus_worker.rs
+++ b/massa-consensus-worker/src/consensus_worker.rs
@@ -1113,7 +1113,7 @@ impl ConsensusWorker {
                     "consensus.consensus_worker.process_protocol_event.get_blocks",
                     { "list": list }
                 );
-                // TODO: use shared storage.
+                // TODO: remove, see https://github.com/massalabs/massa/issues/2299
             }
         }
         Ok(())
--- a/massa-consensus-worker/src/consensus_worker.rs
+++ b/massa-consensus-worker/src/consensus_worker.rs
@@ -680,7 +680,19 @@ impl ConsensusWorker {
                     "consensus.consensus_worker.process_consensus_command.get_active_block",
                     {}
                 );
-                // TODO: use shared storage.
+                if response_tx
+                    .send(self.block_db.get_active_block(&block_id).and_then(|v| {
+                        if let Some(block) = self.block_db.storage.retrieve_block(&v.block_id) {
+                            let stored_block = block.read();
+                            Some(stored_block.block.clone())
+                        } else {
+                            None
+                        }
+                    }))
+                    .is_err()
+                {
+                    warn!("consensus: could not send GetBlock answer");
+                }
                 Ok(())
             }
             // return full block and status with specified hash
--- a/massa-consensus-worker/src/consensus_worker.rs
+++ b/massa-consensus-worker/src/consensus_worker.rs
@@ -756,7 +756,8 @@ impl ConsensusWorker {
                 );
                 let resp = (
                     ExportProofOfStake::from(&self.pos),
-                    self.block_db.export_bootstrap_graph()?,
+                    self.block_db
+                        .export_bootstrap_graph(self.block_db.storage.clone())?,
                 );
                 if response_tx.send(resp).is_err() {
                     warn!("consensus: could not send GetBootstrapState answer");
--- a/massa-consensus-worker/src/consensus_worker.rs
+++ b/massa-consensus-worker/src/consensus_worker.rs
@@ -583,6 +583,13 @@ impl ConsensusWorker {
             creator_private_key,
         )?;
         let block = Block { header, operations };
+        let serialized_block = block.to_bytes_compact()?;
+
+        // Add to shared storage
+        // TODO: remove clone, in https://github.com/massalabs/massa/pull/2304
+        self.block_db
+            .storage
+            .store_block(block_id.clone(), block.clone(), serialized_block);
 
         massa_trace!("create block", { "block": block });
         info!(
--- a/massa-consensus-worker/src/consensus_worker.rs
+++ b/massa-consensus-worker/src/consensus_worker.rs
@@ -1137,7 +1137,10 @@ impl ConsensusWorker {
             massa_trace!("consensus.consensus_worker.block_db_changed.integrated", {
                 "block_id": block_id
             });
-            // TODO: propagate using protocol.
+            self.channels
+                .protocol_command_sender
+                .integrated_block(block_id, op_ids, endo_ids)
+                .await?;
         }
 
         // Notify protocol of attack attempts.
--- a/massa-consensus-worker/src/consensus_worker.rs
+++ b/massa-consensus-worker/src/consensus_worker.rs
@@ -147,18 +147,8 @@ impl ConsensusWorker {
         // notify execution module of current blockclique and final blocks
         // we need to do this because the bootstrap snapshots of the executor vs the consensus may not have been taken in sync
         // because the two modules run concurrently and out of sync
-        channels.execution_controller.update_blockclique_status(
-            block_db.clone_all_final_blocks(),
-            block_db
-                .get_blockclique()
-                .into_iter()
-                .filter_map(|block_id| {
-                    block_db
-                        .get_active_block(&block_id)
-                        .map(|a_block| (block_id, a_block.block.clone()))
-                })
-                .collect(),
-        );
+
+        // TODO: use shared-storage.
 
         Ok(ConsensusWorker {
             genesis_public_key,
@@ -491,7 +481,7 @@ impl ConsensusWorker {
                         ancestor_id
                     ))
                 })?;
-            if ancestor.block.header.content.slot.period < stop_period {
+            if ancestor.slot.period < stop_period {
                 break;
             }
             exclude_operations.extend(ancestor.operation_set.keys());
@@ -674,16 +664,7 @@ impl ConsensusWorker {
                     "consensus.consensus_worker.process_consensus_command.get_active_block",
                     {}
                 );
-                if response_tx
-                    .send(
-                        self.block_db
-                            .get_active_block(&block_id)
-                            .map(|v| v.block.clone()),
-                    )
-                    .is_err()
-                {
-                    warn!("consensus: could not send GetBlock answer");
-                }
+                // TODO: use shared storage.
                 Ok(())
             }
             // return full block and status with specified hash
@@ -1125,28 +1106,7 @@ impl ConsensusWorker {
                     "consensus.consensus_worker.process_protocol_event.get_blocks",
                     { "list": list }
                 );
-                let mut results = Map::default();
-                for block_hash in list {
-                    if let Some(a_block) = self.block_db.get_active_block(&block_hash) {
-                        massa_trace!("consensus.consensus_worker.process_protocol_event.get_block.consensus_found", { "hash": block_hash});
-                        results.insert(
-                            block_hash,
-                            Some((
-                                a_block.block.clone(),
-                                Some(a_block.operation_set.keys().copied().collect()),
-                                Some(a_block.endorsement_ids.keys().copied().collect()),
-                            )),
-                        );
-                    } else {
-                        // not found in consensus
-                        massa_trace!("consensus.consensus_worker.process_protocol_event.get_block.consensu_not_found", { "hash": block_hash});
-                        results.insert(block_hash, None);
-                    }
-                }
-                self.channels
-                    .protocol_command_sender
-                    .send_get_blocks_results(results)
-                    .await?;
+                // TODO: use shared storage.
             }
         }
         Ok(())
@@ -1165,14 +1125,11 @@ impl ConsensusWorker {
         massa_trace!("consensus.consensus_worker.block_db_changed", {});
 
         // Propagate new blocks
-        for (block_id, (block, op_ids, endo_ids)) in
-            self.block_db.get_blocks_to_propagate().into_iter()
-        {
-            massa_trace!("consensus.consensus_worker.block_db_changed.integrated", { "block_id": block_id, "block": block });
-            self.channels
-                .protocol_command_sender
-                .integrated_block(block_id, block, op_ids, endo_ids)
-                .await?;
+        for (block_id, (op_ids, endo_ids)) in self.block_db.get_blocks_to_propagate().into_iter() {
+            massa_trace!("consensus.consensus_worker.block_db_changed.integrated", {
+                "block_id": block_id
+            });
+            // TODO: propagate using protocol.
         }
 
         // Notify protocol of attack attempts.
@@ -1194,25 +1151,7 @@ impl ConsensusWorker {
 
         // notify Execution
         {
-            let finalized_blocks = new_final_block_ids
-                .iter()
-                .filter_map(|b_id| {
-                    self.block_db
-                        .get_active_block(b_id)
-                        .map(|b| (*b_id, b.block.clone()))
-                })
-                .collect();
-            let blockclique = blockclique_set
-                .iter()
-                .filter_map(|b_id| {
-                    self.block_db
-                        .get_active_block(b_id)
-                        .map(|b| (*b_id, b.block.clone()))
-                })
-                .collect();
-            self.channels
-                .execution_controller
-                .update_blockclique_status(finalized_blocks, blockclique);
+            // TODO: use shared storage.
         }
 
         // Process new final blocks
@@ -1224,9 +1163,10 @@ impl ConsensusWorker {
             if let Some(a_block) = self.block_db.get_active_block(&b_id) {
                 // List new final ops
                 new_final_ops.extend(
-                    a_block.operation_set.iter().map(|(id, (_, exp))| {
-                        (*id, (*exp, a_block.block.header.content.slot.thread))
-                    }),
+                    a_block
+                        .operation_set
+                        .iter()
+                        .map(|(id, (_, exp))| (*id, (*exp, a_block.slot.thread))),
                 );
                 // List final block
                 new_final_blocks.insert(b_id, a_block);
@@ -1234,7 +1174,7 @@ impl ConsensusWorker {
                 self.final_block_stats.push_back((
                     timestamp,
                     a_block.operation_set.len() as u64,
-                    Address::from_public_key(&a_block.block.header.content.creator),
+                    a_block.creator_address,
                 ));
             }
         }
@@ -1289,7 +1229,7 @@ impl ConsensusWorker {
             // iterate on all blockclique blocks
             for block_id in blockclique_set.into_iter() {
                 let block_slot = match self.block_db.get_active_block(&block_id) {
-                    Some(b) => b.block.header.content.slot,
+                    Some(b) => b.slot,
                     None => continue,
                 };
                 if self.endorsed_slots.contains(&block_slot) {
--- a/massa-consensus-worker/src/consensus_worker.rs
+++ b/massa-consensus-worker/src/consensus_worker.rs
@@ -604,7 +604,6 @@ impl ConsensusWorker {
         // add block to db
         self.block_db.incoming_block(
             block_id,
-            block,
             operation_set,
             endorsement_ids,
             &mut self.pos,
@@ -1084,14 +1083,15 @@ impl ConsensusWorker {
         match event {
             ProtocolEvent::ReceivedBlock {
                 block_id,
-                block,
                 operation_set,
                 endorsement_ids,
             } => {
-                massa_trace!("consensus.consensus_worker.process_protocol_event.received_block", { "block_id": block_id, "block": block });
+                massa_trace!(
+                    "consensus.consensus_worker.process_protocol_event.received_block",
+                    { "block_id": block_id }
+                );
                 self.block_db.incoming_block(
                     block_id,
-                    block,
                     operation_set,
                     endorsement_ids,
                     &mut self.pos,
@@ -1114,7 +1114,27 @@ impl ConsensusWorker {
                     "consensus.consensus_worker.process_protocol_event.get_blocks",
                     { "list": list }
                 );
-                // TODO: remove, see https://github.com/massalabs/massa/issues/2299
+                let mut results = Map::default();
+                for block_hash in list {
+                    if let Some(a_block) = self.block_db.get_active_block(&block_hash) {
+                        massa_trace!("consensus.consensus_worker.process_protocol_event.get_block.consensus_found", { "hash": block_hash});
+                        results.insert(
+                            block_hash,
+                            Some((
+                                Some(a_block.operation_set.keys().copied().collect()),
+                                Some(a_block.endorsement_ids.keys().copied().collect()),
+                            )),
+                        );
+                    } else {
+                        // not found in consensus
+                        massa_trace!("consensus.consensus_worker.process_protocol_event.get_block.consensu_not_found", { "hash": block_hash});
+                        results.insert(block_hash, None);
+                    }
+                }
+                self.channels
+                    .protocol_command_sender
+                    .send_get_blocks_results(results)
+                    .await?;
             }
         }
         Ok(())
--- a/massa-consensus-worker/src/consensus_worker.rs
+++ b/massa-consensus-worker/src/consensus_worker.rs
@@ -778,7 +778,7 @@ impl ConsensusWorker {
                 let resp = (
                     ExportProofOfStake::from(&self.pos),
                     self.block_db
-                        .export_bootstrap_graph(self.block_db.storage.clone())?,
+                        .export_bootstrap_graph()?,
                 );
                 if response_tx.send(resp).is_err() {
                     warn!("consensus: could not send GetBootstrapState answer");
--- a/massa-execution-exports/src/controller_traits.rs
+++ b/massa-execution-exports/src/controller_traits.rs
@@ -7,12 +7,11 @@ use crate::types::ReadOnlyExecutionRequest;
 use crate::ExecutionError;
 use massa_ledger::LedgerEntry;
 use massa_models::output_event::SCOutputEvent;
-use massa_models::prehash::Map;
 use massa_models::Address;
-use massa_models::Block;
 use massa_models::BlockId;
 use massa_models::OperationId;
 use massa_models::Slot;
+use std::collections::HashMap;
 
 /// interface that communicates with the execution worker thread
 pub trait ExecutionController: Send + Sync {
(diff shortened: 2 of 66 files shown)


gterzian:
@AurelienFT If you have time today, could you help with the rebase? Doing this here because I want to keep the original PR for now in case anything important got lost along the way.

AurelienFT:
> @AurelienFT If you have time today, could you help with the rebase and fixing remaining errors? Doing this here because I want to keep the original PR for now in case anything important got lost along the way.
> 
> I'm done with this for today...

Let's make a relay race ! 🏃‍♂️

AurelienFT:
Tests are now compiling but some don't pass. I'll update here if I have time to fix them.

gterzian:
@AurelienFT Thanks a lot for your help!

```
---- tests::test_block_graph::test_get_ledger_at_parents stdout ----
thread 'tests::test_block_graph::test_get_ledger_at_parents' panicked at 'called `Result::unwrap()` on an `Err` value: MissingBlock', massa-consensus-worker/src/tests/test_block_graph.rs:563:10
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

It seems only this one is not passing, I guess certain blocks are not found in storage...

Looking into it, I guess we could keep doing so in parallel this morning...

AurelienFT:
> It seems only this one is not passing, I guess certain blocks are not found in storage...
> 
> Looking into it, I guess we could keep doing so in parallel this morning...

I need to write a test for #2433

gterzian:
It appears that the block related to an active block is not stored, but I haven't found in the tests where this omission might occur...

Failure is at https://github.com/massalabs/massa/pull/2428/files#diff-c4310a8d90af39a317383da61260d64eac43422f2d113b6bbfaeb7a736c8388dR2480

gterzian:
Ok this is the problem:

```
ID for active_block_p3t0: 29TVs2D49qXPTQePSYwgYoBYJAQvEBidue5HVoZsTgzfKJkDEa
Computed ID for active_block_p3t0: JnWsQucoMrEkM3sTudG4MbefTQPfNnPEAaZhriuHz98MZzv3t
trying to retrieve block JnWsQucoMrEkM3sTudG4MbefTQPfNnPEAaZhriuHz98MZzv3t
thread 'tests::test_block_graph::test_get_ledger_at_parents' panicked at 'called `Result::unwrap()` on an `Err` value: MissingBlock'
```

(This is from prints I added, I hope it's clear)

And the problem occurs because `TryFrom<ExportActiveBlock> for ActiveBlock` computes the id using `a_block.block.header.content.compute_id().unwrap(),` and this will not be the same Id as the one that was used to store the block. 

So there is a disconnect between the active blocks used in the graph, and the blocks found in storage, because storage uses the result of `get_dummy_block_id("active_block_p3t0")`, whereas graph will have an active block, created from the exported active block, whose `block_id` is the computed id, not `get_dummy_block_id("active_block_p3t0")`. 

I'm done for the day, will pick it up tomorrow, feel free to fix it in the meantime if you have time. @AurelienFT

AurelienFT:
> I'm done for the day, will pick it up tomorrow, feel free to fix it in the meantime if you have time. @AurelienFT

Nice catch I don't have time to help you sorry :/

gterzian:
@AurelienFT Ok the consensus worker tests are now all passing, it seems we still need to do the same thing(storing blocks as part of the tests) in protocol worker tests. 

I'm done for today, so if you have time you can help if you want.

AurelienFT:
I fixed all the tests and most of the clippy warning. (1 left I don't know if we want to fix or ignore it). I think we can both checks if we don't see any wired diff with main on things that we shouldn't touch or that change behavior too much and then we can submit to review to the others.

gterzian:
@massalabs/core-team ready for review.

AurelienFT (commented):
Reviewed only half of the files.

gterzian:
@AurelienFT thanks for the review, addressing those points...

gterzian:
I've replaced the `unwrap` with returning errors where possible, and changed a few into `expect`. In the execution thread this is ok, since it's a thread. In an async context it's not ideal because of the unclear backtrace but ok for those few exceptions in my opinion.

massa-execution-worker/src/worker.rs
```diff
@@ -270,7 +271,16 @@ impl ExecutionThread {
         exec_state.clear_history();
 
         // execute slot
-        let exec_out = exec_state.execute_slot(slot, exec_target);
+        let exec_out = if let Some(exec_target) = exec_target {
+            let block = self
+                .storage
+                .retrieve_block(&exec_target)
+                .expect("Missing block in storage.");
```
AurelienFT, line 278: Why not just return false in error case ?

massa-execution-worker/src/worker.rs
```diff
@@ -307,12 +317,20 @@ impl ExecutionThread {
...
+            _ => return false,
         };
 
-        // execute the slot
-        let exec_out = exec_state.execute_slot(slot, exec_target);
+        let exec_out = {
+            let block = self
+                .storage
+                .retrieve_block(&exec_target)
+                .expect("Missing block in storage.");
```
AurelienFT, line 328: Why not just return false in error case ?

massa-graph/src/block_graph.rs
```diff
@@ -1188,7 +1204,28 @@ impl BlockGraph {
...
+                BlockStatus::Incoming(_) => ExportBlockStatus::Incoming,
+                BlockStatus::WaitingForSlot(_) => ExportBlockStatus::WaitingForSlot,
+                BlockStatus::WaitingForDependencies { .. } => {
+                    ExportBlockStatus::WaitingForDependencies
+                }
+                BlockStatus::Active(active_block) => {
+                    let block = self
+                        .storage
+                        .retrieve_block(block_id)
+                        .expect("Missing block in storage.");
```
AurelienFT, line 1217: Why not return None here to avoid crash ?

massa-graph/src/block_graph.rs
```diff
@@ -1200,18 +1237,23 @@ impl BlockGraph {
         // for each active block
         for block_id in self.active_index.iter() {
             if let Some(BlockStatus::Active(active_block)) = self.block_statuses.get(block_id) {
+                // TODO: clone operations in one go as in `get_operations_involving_address`.
+                let block = self
+                    .storage
+                    .retrieve_block(block_id)
+                    .expect("Missing block in storage.");
```
AurelienFT, line 1244: Why not `continue` instead of crashing ?

massa-network-worker/src/node_worker.rs
```diff
@@ -27,14 +31,21 @@ use tokio::{
 };
 use tracing::{debug, trace, warn};
 
+#[derive(Debug, Serialize, Deserialize)]
+pub enum ToSend {
+    Msg(Message),
```
AurelienFT, line 36: Clippy warning. Do we want to address it ?

massa-network-worker/src/node_worker.rs
```diff
@@ -261,7 +313,7 @@ impl NodeWorker {
                                     "node_worker.run_loop. receive Message::Block",
                                     {"block_id": block.header.content.compute_id()?, "block": block, "node": self.node_id}
                                 );
-                                self.send_node_event(NodeEvent(self.node_id, NodeEventType::ReceivedBlock(block))).await;
+                                self.send_node_event(NodeEvent(self.node_id, NodeEventType::ReceivedBlock(block, serialized.expect("Block should come wiht its serialized form.")))).await;
```
AurelienFT, line 316:
```suggestion
                                self.send_node_event(NodeEvent(self.node_id, NodeEventType::ReceivedBlock(block, serialized.expect("Block should come with its serialized form.")))).await;
```

AurelienFT (commented):
Still some crashing that concern me.

gterzian:
> Still some crashing that concern me.

Why does it concern you? 

I am personally much more concerned about continuing to run the program in an invalid state. 

> Why not (...) instead of crashing ?

In all cases, because we wouldn't want to change an algorithm to support storage being in an inconsistent state.

It would be different in a case where for example an algorithm is not sure about whether a block is present, and we would still be unwrapping the result from retrieving a block in all case.

For example if `graph` where to check if a certain id was present in `self.block_statuses`, and we would still assume that the block would always be present in storage(by unwrapping on retrieval). 

On the other hand, for example at https://github.com/massalabs/massa/pull/2428#discussion_r831944718, `self.block_statuses.get(block_id)` returns `BlockStatus::Active`, in which case not having the block in storage means storage is for some reason in an invalid state in a way that is unrelated from the surrounding code. Crashing there is the right thing to do because there is no way to recover from this, and it should in fact never happen(and the problem would have to be fixed elsewhere, for example in the pruning algorithm).

massa-network-worker/src/node_worker.rs
gterzian: Do you think it's a problem, and if so, do you have a suggestion on have to address it?

AurelienFT: I don't if use box as clippy wants its a good idea maybe add attribute to clear the warning

gterzian: How about we leave it as a warning(as a reminder), and address it if it becomes a performance problem?

AurelienFT: Ok

massa-graph/src/block_graph.rs
```diff
@@ -849,17 +864,23 @@ impl BlockGraph {
...
             .get_cycle(self.cfg.periods_per_cycle);
 
-        let same_thread_parent_creator =
-            Address::from_public_key(&same_thread_parent.header.content.creator);
+        let same_thread_parent_creator = {
+            let creator = {
+                let block = self
+                    .storage
+                    .retrieve_block(&same_thread_parent.block_id)
+                    .unwrap();
```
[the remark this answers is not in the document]
AureliaDolo: Should be a Missing block/ graph inconsistency error

massa-graph/src/block_graph.rs
AureliaDolo: Because it's an inconsistency, this should not happen

massa-graph/src/block_graph.rs
```diff
@@ -1188,7 +1204,28 @@ impl BlockGraph {
     pub fn get_export_block_status(&self, block_id: &BlockId) -> Option<ExportBlockStatus> {
         self.block_statuses
             .get(block_id)
-            .map(|block_status| block_status.into())
+            .map(|block_status| match block_status {
```
AureliaDolo, line 1207: Is there a reason to not use the into anymore ?

massa-graph/src/block_graph.rs
```diff
@@ -1200,18 +1237,23 @@ impl BlockGraph {
         // for each active block
         for block_id in self.active_index.iter() {
             if let Some(BlockStatus::Active(active_block)) = self.block_statuses.get(block_id) {
+                // TODO: clone operations in one go as in `get_operations_involving_address`.
```
AureliaDolo, line 1240: Maybe we should directly query storage for the not graph related api queries. Maybe more in a follow up

massa-network-worker/src/binders.rs
```diff
@@ -34,11 +31,9 @@ impl WriteBinder {
     ///
     /// # Argument
     /// * msg: date to transmit.
-    pub async fn send(&mut self, msg: &Message) -> Result<u64, NetworkError> {
+    pub async fn send(&mut self, buf: &[u8]) -> Result<u64, NetworkError> {
```
AureliaDolo, line 34: Why don't you use the Message struct anymore ?

massa-network-worker/src/node_worker.rs
AureliaDolo, line 36: Maybe explicitely ignore the warning ? To show in the code that there is indeed a warning ?

massa-execution-exports/src/controller_traits.rs
```diff
@@ -23,8 +22,8 @@ pub trait ExecutionController: Send + Sync {
     /// * blockclique: new blockclique
     fn update_blockclique_status(
         &self,
-        finalized_blocks: Map<BlockId, Block>,
-        blockclique: Map<BlockId, Block>,
+        finalized_blocks: HashMap<Slot, BlockId>,
+        blockclique: HashMap<Slot, BlockId>,
```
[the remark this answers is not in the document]
gterzian: This would happen previously at the point where the message was received, so I thought I'd just put that in the message itself.

massa-graph/src/block_graph.rs
gterzian: Yes, the need to retrieve the block from storage(which is not available from inside the implementation of `into`).

massa-graph/src/block_graph.rs
gterzian: For now I don't think so, because it's the graph that has the logic about whether a block should be returned or not, for example here we only use blocks that are `BlockStatus::Active`, and storage does not, and I think should not, know about this.

massa-models/src/storage.rs
```diff
@@ -0,0 +1,55 @@
...
+pub struct Storage {
+    blocks: Arc<RwLock<Map<BlockId, Arc<RwLock<StoredBlock>>>>>,
+}
+
+impl Storage {
+    /// Store a block, along with it's serialized representation.
+    pub fn store_block(&self, block_id: BlockId, block: Block, serialized: Vec<u8>) {
+        let mut blocks = self.blocks.write();
+        match blocks.entry(block_id) {
+            Entry::Occupied(_) => {}
```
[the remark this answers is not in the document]
gterzian: yes.

massa-network-worker/src/binders.rs
gterzian: So that in the case of a block and header, we can reuse the serialized form stored in storage(without having to do that inside the binder).

massa-network-worker/src/messages.rs
```diff
@@ -52,9 +52,32 @@ pub enum Message {
     Endorsements(Vec<SignedEndorsement>),
 }
 
+pub fn deserialize_message_with_optional_serialized_object(
```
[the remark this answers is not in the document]
gterzian: yes.

massa-network-worker/src/node_worker.rs
gterzian: Why would you want to show in the code that there is warning? It's not actually a problem as far as I can tell, and leaving something in the code makes it look a bit like a TODO, and I don't think we should care about this at all unless there is a problem. 

For example here I don't think we need to box the `Message` variant just to make the warning go away, because, why would we? I actually introduced a smaller variant, previously everything was a `Message`, so this would appear not to be a possible loss in performance.

A box would mean an additional pointer and heap allocation(I think), which actually might degrade performance more. 

And actually both of these considerations in my opinion do not matter unless this particular thing were to show up on a performance test.

I am not a big fan of clippy, since it appears to often makes things look like problems when there aren't actually any.

gterzian:
> do you think there are some parts that require more attention that other ?

I think the most important part is checking that no logic was changed. 

Thank you for the review! @AureliaDolo

massa-network-worker/src/node_worker.rs
adrien-zinger: Actually the compiler will align the enum with the sizeof Message even if it gets just a BlockId, here is globally the purpose of that warning. Would you like a PR with a proposal? :slightly_smiling_face: 

Here is a video from Rust explaining why that is a performance wickness on youtube. @yvan-sraka used to broadcast that video some month ago in the team.

adrien-zinger: it's when she speaks about the stack size at perhaps 23 minutes

massa-consensus-worker/src/consensus_worker.rs
```diff
@@ -675,11 +681,14 @@ impl ConsensusWorker {
                     {}
                 );
                 if response_tx
-                    .send(
-                        self.block_db
-                            .get_active_block(&block_id)
-                            .map(|v| v.block.clone()),
-                    )
+                    .send(self.block_db.get_active_block(&block_id).and_then(|v| {
```
gterzian, line 684: I'm going to remove the `get_active_block` command, because it seems to only be used in tests and isn't necessary anymore with the use of storage.

massa-network-worker/src/node_worker.rs
gterzian: @adrien-zinger Thanks for the video.

> Actually the compiler will align the enum with the sizeof Message even if it gets just a BlockId

How can that be a performance regression, if previously all variants were `Message`?

On the other hand, boxing `Message` would mean a extra pointer access that previously wasn't there.

Also, this is an isolated case(not "all errors across the application" like in the video), so in my opinion it's irrelevant, unless it were to show up somewhere on a performance test, in which case the fix would be easy.

AurelienFT (approved):

gterzian:
@AureliaDolo @AurelienFT @damip Ok this one is ready for a final round of review/approval and merge. I just changed a few small things, see the last commits...

damip:
Can someone rebase it (it is needed since the merge of op propagation) ?

damip:
> Can someone rebase it (it is needed due to conflicts with main since the merge of op propagation) ?

I think you would be the best candidates for this task @adrien-zinger  /  @AurelienFT to ensure that it doesn't break the algo you implemented

AurelienFT:
Rebasing done I think we should make some test to be sure I didn't broke anything.

massa-models/src/active_block.rs
```diff
@@ -37,4 +37,13 @@ pub struct ActiveBlock {
     pub roll_updates: RollUpdates,
     /// list of (period, address, did_create) for all block/endorsement creation events
     pub production_events: Vec<(u64, Address, bool)>,
+    /// Slot of the block.
+    pub slot: Slot,
+}
+
+impl ActiveBlock {
```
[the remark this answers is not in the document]
gterzian: Because it appears like that method is only used in the context of an active block(and with the changes introduced in this PR, the block itself is not stored in the active block anymore).

gterzian:
@massalabs/core-team 

Since this is a big change, and it's been affected by conflicts in many parts, may I suggest we make it a priority to merge this today? 

One way to do this would be for everyone to actually take a look, focusing on the parts of the code that you may have been working on recently and that conflicted with the changes here.

I'd say it's a good idea to focus on correctness, and that small things should be addressed in follow-ups via a new issues.

massa-models/src/storage.rs
```diff
@@ -0,0 +1,55 @@
...
+    /// The serialized representation of the block.
+    pub serialized: Vec<u8>,
+    /// The serialized representation of the header, if any.
+    /// Note: the header is written as part of propagation of headers.
+    pub serialized_header: Option<Vec<u8>>,
+}
+
+/// A storage of block, shared by various components.
+#[derive(Clone, Default)]
+pub struct Storage {
```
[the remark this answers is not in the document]
AurelienFT: Done.

massa-execution-worker/src/execution.rs
```diff
@@ -446,7 +445,11 @@ impl ExecutionState {
...
     /// An `ExecutionOutput` structure summarizing the output of the executed slot
-    pub fn execute_slot(&self, slot: Slot, opt_block: Option<(BlockId, Block)>) -> ExecutionOutput {
+    pub fn execute_slot(
+        &self,
+        slot: Slot,
+        opt_block: Option<(BlockId, &Block)>,
+    ) -> ExecutionOutput {
         // get optional block ID and creator address
         let (opt_block_id, opt_block_creator_addr) = opt_block
             .as_ref()
```
Eitu33, line 455: With `execute_slot` now taking `Option<(BlockId, &Block)>` as argument, this `as_ref()` becomes unnecessary since `opt_block` won't get moved when retrieving the public key.

Eitu33:
Apart from the above comment, the execution module changes look good to me. Also because we currently lack async messages tests, I hand tested the feature and can confirm that theses changes did not break anything on that side.

massa-execution-worker/src/execution.rs
AurelienFT: Done.

gterzian:
@massalabs/core-team Thank you all for your help, can we get two approvals and a bors merge please?

AurelienFT (dismissed):

Eitu33:
Since I've only partially reviewed this PR I believe it would be better to wait for a 3rd approval.

Eitu33 (dismissed):

AureliaDolo (approved):
in #2416 I added a warning for missing doc (it applies only for pub stuff) so I opened a branch https://github.com/massalabs/massa/tree/doc_shared_storage that merges our two prs and adds some doc where it triggered warnings. 
Feel free to open a pr that target main or the rebase storage branch depending on the order everything is merged :)

damip:
I will take a last look at the consensus part !

AurelienFT:
> https://github.com/massalabs/massa/tree/doc_shared_storage

I think it's better that we merge this PR and then take your new branch for the PR doc so that it's up to date with this one.

massa-execution-exports/src/controller_traits.rs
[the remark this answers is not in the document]
Eitu33: I believe it would be better to use `Map<BlockId, Slot>` instead to match the previous behaviour.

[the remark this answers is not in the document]
Eitu33: I'm doing a branch based on this one to see if that was indeed the cause of https://github.com/massalabs/massa/issues/2469

massa-graph/src/block_graph.rs
```diff
@@ -1324,18 +1366,27 @@ impl BlockGraph {
         if self.genesis_hashes.contains(&block_id) {
             return Ok(());
         }
-        debug!(
-            "received block {} for slot {}",
-            block_id, block.header.content.slot
-        );
-        massa_trace!("consensus.block_graph.incoming_block", {"block_id": block_id, "block": block});
+
+        let slot = {
```
damip, line 1370: if this whole retrieval is just there to get the slot, we should add the slot as a parameter of the incoming_block function

massa-execution-exports/src/controller_traits.rs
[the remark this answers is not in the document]
AureliaDolo: Be careful, we only use Maps when the key is already hashed (hence the prehashed trait), that's why when block id is the key it's ok to use a map, and it's not ok when slot is a key

massa-graph/src/block_graph.rs
```diff
@@ -3122,13 +3196,26 @@ impl BlockGraph {
         for (thread, id) in latest_final_blocks.iter().enumerate() {
             let mut current_block_id = *id;
             while let Some(current_block) = self.get_active_block(&current_block_id) {
+                let parent_id = {
```
damip, line 3199: isn't the parent info available in current_block ?

damip:
Cases where I'm not sure the pruned block is removed from storage:
* prune_active (old final and stale blocks)
* prune_waiting_for_dependencies

Can someone double-check ?

damip (commented):
Cases where I'm not sure the pruned block is removed from storage:
* prune_active (old final and stale blocks)
* prune_waiting_for_dependencies

Can someone double-check ?

massa-execution-exports/src/controller_traits.rs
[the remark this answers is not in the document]
Eitu33: Yes I am aware of it as I mentioned in the 3rd point

damip (commented):
Another point about unwap() and expect() on blocks missing from storage:
* they should all be double-checked (we had such a panic during a small-scale test of this branch) to ensure that they don't happen when missing blocks can be expected (API, non-validated protocol requests etc...)
* in async contexts, they should be removed and turned into error bubbling

massa-graph/src/block_graph.rs
AurelienFT: Yes, retrieve removed.

massa-graph/src/block_graph.rs
```diff
@@ -2429,7 +2500,12 @@ impl BlockGraph {
         for thread in involved_threads.into_iter() {
             match self.block_statuses.get(&parents[thread as usize]) {
                 Some(BlockStatus::Active(b)) => {
-                    if b.block.header.content.slot.period
+                    let block = self
```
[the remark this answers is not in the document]
AurelienFT: Yes, retrieve removed.

massa-graph/src/block_graph.rs
AurelienFT: Yes, retrieve removed and slot added in the ReceiveBlock message.

massa-graph/src/block_graph.rs
```diff
@@ -849,17 +870,23 @@ impl BlockGraph {
...
 
         let same_thread_parent_cycle = same_thread_parent
-            .header
-            .content
             .slot
             .get_cycle(self.cfg.periods_per_cycle);
 
-        let same_thread_parent_creator =
-            Address::from_public_key(&same_thread_parent.header.content.creator);
+        let same_thread_parent_creator = {
```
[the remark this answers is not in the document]
AurelienFT: No it was necessary, retrieve removed.

massa-execution-exports/src/controller_traits.rs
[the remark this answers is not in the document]
gterzian: @Eitu33 Thanks for double checking. 

Re point 1: what "major behaviour change" does it create, and why is it "a risky thing to do"? I don't fully understand since even before the changes introduced in this PR the controller would end-up with a [`finalized_blocks: HashMap<Slot, BlockId>,`](https://github.com/massalabs/massa/pull/2428/files/c9dd61e7e54de844dae8fd2602aaa0a9af27b86b#diff-a78e41f7449ec5bc05f6f83364287a9ccc177125329251c608436baaf11c20c0) and a [`new_blockclique: Option<HashMap<Slot, BlockId>>`](https://github.com/massalabs/massa/pull/2428/files/c9dd61e7e54de844dae8fd2602aaa0a9af27b86b#diff-a78e41f7449ec5bc05f6f83364287a9ccc177125329251c608436baaf11c20c0R29)

Re point 2 and 3: yes that is correct, and since the controller ends up with a `HashMap<Slot, BlockId>` (and before the changes in this PR it is a `HashMap<Slot, (BlockId, Block)>`) anyway, why is that a problem? See https://github.com/massalabs/massa/pull/2428/files/c9dd61e7e54de844dae8fd2602aaa0a9af27b86b#diff-a78e41f7449ec5bc05f6f83364287a9ccc177125329251c608436baaf11c20c0L28

In both cases, the difference seems to be that in this PR the output is put directly in the message, as opposed to be mapped into it when receiving the message, hence I am wondering where the major change comes from.

gterzian:
> Cases where I'm not sure the pruned block is removed from storage:
> 
>     * prune_active (old final and stale blocks)
> 
>     * prune_waiting_for_dependencies
> 
> 
> Can someone double-check ?

For actives, it looks like they are always switched to `Discarded` in their status, so then they will be pruned along with other discarded blocks.

For those waiting for deps, it looks like we were missing those that are not switched to discarded, addressed in https://github.com/massalabs/massa/pull/2428/commits/298ba17e5fd6750317e03fedf3a860fa27d04f0d 

@damip

gterzian:
@massalabs/core-team Ok I've addressed most of the open issues, with the exception of the discussion starting at https://github.com/massalabs/massa/pull/2428#discussion_r837482711 cc @Eitu33

Eitu33:
https://github.com/massalabs/massa/pull/2473/files#diff-990c58de6c3e1194da0ee5db7bcac14cdd0816ff8682805753c594f83e304511

massa-execution-exports/src/controller_traits.rs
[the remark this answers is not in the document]
Eitu33: The changes I've made in https://github.com/massalabs/massa/pull/2473 did not fix https://github.com/massalabs/massa/issues/2469, I will be investigating this issue today.

Also I confirm that https://github.com/massalabs/massa/issues/2469 does not happen on main (since we weren't so sure).

damip:
>     * in async loop contexts, panics should be removed and turned into error bubbling

I think this was not addressed

gterzian:
> so I opened a branch https://github.com/massalabs/massa/tree/doc_shared_storage that merges our two prs and adds some doc where it triggered warnings.
Feel free to open a pr that target main or the rebase storage branch depending on the order everything is merged :)

Thanks @AureliaDolo  and yes let's open a PR when the other issues have been resolved...

gterzian:
> I think this was not addressed

Yes there are a few left, and those are all cases were we have an `BlockStatus::Active`(so the panic should never happen), and also where the function is not currently returning a `Result`. 

So I'd say we can leave it to a follow-up, since removing them would require some changes to the surrounding code and I'd rather do it as a simple follow-up.

massa-execution-worker/src/worker.rs
```diff
@@ -309,12 +319,20 @@ impl ExecutionThread {
 
         // choose the execution target
         let exec_target = match self.active_slots.get(&slot) {
-            Some(b) => b.clone(), //TODO get rid of that clone on storage refactorig https://github.com/massalabs/massa/issues/2178
-            None => return false,
+            Some(Some(b_id)) => *b_id,
```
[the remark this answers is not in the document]
AurelienFT: I have updated is it good now ?

massa-execution-worker/src/worker.rs
```diff
@@ -309,12 +319,20 @@ impl ExecutionThread {
...
+            _ => return false,
         };
 
-        // execute the slot
-        let exec_out = exec_state.execute_slot(slot, exec_target);
+        let exec_out = {
+            let block = self
```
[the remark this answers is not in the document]
AurelienFT: Done.

massa-execution-worker/src/worker.rs
```diff
@@ -573,6 +591,7 @@ impl ExecutionThread {
 pub fn start_execution_worker(
```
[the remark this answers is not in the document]
AurelienFT: Done.

massa-execution-worker/src/worker.rs
```diff
@@ -254,9 +257,7 @@ impl ExecutionThread {
 
         // check if the final slot is cached at the front of the speculative execution history
         if let Some(exec_out) = exec_state.pop_first_execution_result() {
-            if exec_out.slot == slot
-                && exec_out.block_id == exec_target.as_ref().map(|(b_id, _)| *b_id)
-            {
+            if exec_out.slot == slot && exec_out.block_id == exec_target.as_ref().copied() {
```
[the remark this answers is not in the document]
AurelienFT: Removed.

damip (approved):
LGTM

AurelienFT (approved):

AurelienFT:
bors merge

AurelienFT:
Screenshot 2022-03-31 at 16 09 27

@damip about pruning we can see that after 10 minutes the pruning start to work and the storage stabilize at 1160

PR #2429, opened 2022-03-16
Add t0 and thread_count in env variable when using sandbox feature.

### LINKED ISSUE #2423: Add a override for genesis timestamp
## Context

I want to easily run a network over and over on my local computer. I need to advance the genesis timestamp each time to relaunch the network correctly. For the moment we can only parameter it using the variable in the `.rs` which prevent to relaunch without recompiling.

## Possible solutions
- Have a environment variable that can override. It's a good solution because it's not on the configuration and the users will likely not modify it.
- Have the value in the config like before
- Pass as parameter of the node. If we document this the users will not understand why we need that if we don't it's a bit hacky.

I prefer the environment variable but allowing it only if we compile with a certain feature.

Linked questions : If we want to be able to change parameter like t0 easily we should allow using env variable also.

@damip @adrien-zinger @AureliaDolo @yvan-sraka What's your thoughts  on this ?

AurelienFT:
Fix #2423 .

A bit hacky. I think we could make a more generic switch between the two configs but for the moment it's ok imo because it's two variable and also that the config is well managed in the whole project. It should be part of a more bigger refactor on the config.

### DIFF
--- a/massa-models/src/node_configuration/default.rs
+++ b/massa-models/src/node_configuration/default.rs
@@ -69,14 +69,26 @@ lazy_static::lazy_static! {
     };
 }
 
+#[cfg(feature = "sandbox")]
+lazy_static::lazy_static! {
+    pub static ref T0: MassaTime = std::env::var("T0").map(|timestamp| timestamp.parse::<u64>().unwrap().into()).unwrap_or_else(|_|
+        MassaTime::from(16000)
+    );
+    pub static ref THREAD_COUNT: u8 = std::env::var("THREAD_COUNT").map(|timestamp| timestamp.parse::<u8>().unwrap().into()).unwrap_or_else(|_|
+        32
+    );
+}
+
 /// Price of a roll in the network
 pub const ROLL_PRICE: Amount = Amount::from_raw(100 * AMOUNT_DECIMAL_FACTOR);
 /// Block reward is given for each block creation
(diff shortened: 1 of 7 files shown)


massa-models/src/node_configuration/default.rs
```diff
@@ -148,6 +160,7 @@ pub const SLOT_KEY_SIZE: usize = 9;
 /// Size of the event id hash used in execution module, safe to import
 pub const EVENT_ID_SIZE_BYTES: usize = massa_hash::HASH_SIZE_BYTES;
 
+#[cfg(not(feature = "sandbox"))]
 // Some checks at compile time that should not be ignored!
```
AureliaDolo, line 164: These checks are in start consensus controller anyway :/

massa-models/src/serialization_context.rs
```diff
@@ -72,13 +72,43 @@ pub struct SerializationContext {
...
 }
-
+#[cfg(not(feature = "sandbox"))]
 impl Default for SerializationContext {
     fn default() -> Self {
         SerializationContext::const_default()
     }
 }
 
+#[cfg(feature = "sandbox")]
```
AureliaDolo, line 82: Could you comment what's the diff with the default with not sandbox

AurelienFT: Done.

AureliaDolo (approved):
Maybe just have a look at the doc at the top of the files and make sure it's consistent with your changes

adrien-zinger (approved):

PR #2430, opened 2022-03-16
Massa async pool

damip:


massa-async-pool/src/pool.rs
```diff
@@ -0,0 +1,178 @@
...
+            .into_iter()
+            .map(|v| (v.compute_id(), v))
+            .collect();
+
+        // Filter out all messages for which the validity end is expired.
+        // Note that the validity_end bound is NOT included in the validity interval of the message.
+        let mut eliminated: Vec<_> = self
+            .messages
+            .drain_filter(|_k, v| slot >= v.validity_end)
+            .chain(new_messages.drain_filter(|(_k, v)| slot > v.validity_end))
```
AurelienFT, line 109: Why this check is different for new messages that for the ones in the async pool ?

massa-async-pool/src/pool.rs
```diff
@@ -0,0 +1,178 @@
...
+    /// A vector of messages, sorted from the most prioritary to the least prioritary
+    pub fn take_batch_to_executte(
+        &mut self,
+        slot: Slot,
+        mut available_gas: u64,
+    ) -> Vec<AsyncMessage> {
+        let mut selected = Vec::new();
+
+        // iterate in decreasing priority order
+        for (msg_id, msg) in self.messages.iter().rev() {
```
AurelienFT, line 146: I don't understand why you iterate in the rev order.
Also if it's normal something that I found strange is that you are using reverse to build your id and then "re-reverse" to iterate isn't it a double reverse ?

massa-async-pool/src/pool.rs
```diff
@@ -0,0 +1,178 @@
...
+            available_gas -= msg.max_gas;
+
+            // if there is no more gas, quit
+            if available_gas == 0 {
+                break;
+            }
+        }
+
+        // gather all selected items and remove them from self.messages
+        let mut accumulator = Vec::with_capacity(selected.len());
```
AurelienFT, line 170: I think this is an unnecessary loop. As you already filter them above I think you can use `drain_filter` (Example: https://doc.rust-lang.org/std/vec/struct.Vec.html#examples-47) above so that you directly remove them from `messages` and have them in a separate vector without this loop.

AurelienFT (commented):
Nice implementation just the clarity could be improved a bit.

AureliaDolo:
@Eitu33 for typos, you can use code suggestion, it's a life changer :wink:

massa-async-pool/src/bootstrap.rs
```diff
@@ -0,0 +1,58 @@
...
+use serde::{Deserialize, Serialize};
+
+/// Represents a snapshot of the async pool aimed at bootstrapping an AsyncPool
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct AsyncPoolBootstrap {
+    /// list of async messages in no particular order
+    pub(crate) messages: Vec<AsyncMessage>,
+}
+
+/// Allows serializing the AsyncPoolBootstrap to a compact binary representation
```
AureliaDolo, line 18: No need to document trait impl, the trait documentation should be enough

massa-async-pool/src/changes.rs
```diff
@@ -0,0 +1,54 @@
+//! Copyright (c) 2022 MASSA LABS <info@massa.net>
+
+//! This file provides structures representing changes to the async message pool
+
+use crate::message::{AsyncMessage, AsyncMessageId};
+
+/// Enum representing a add/delete change on a value T
+#[derive(Debug, Clone)]
+pub enum AddOrDelete<T> {
```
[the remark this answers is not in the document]
AureliaDolo: Or Maybe update to be consistent with what is done in other places ?

massa-async-pool/src/changes.rs
```diff
@@ -0,0 +1,54 @@
...
+    /// adds a new value T
+    Add(T),
+
+    /// deletes the value T
+    Delete,
+}
+
+/// allows applying another AddOrDelete to the current one
+impl<T> AddOrDelete<T> {
+    fn apply(&mut self, other: Self) {
```
AureliaDolo, line 19:
```suggestion
impl<T> AddOrDelete<T> {
   /// allows applying another AddOrDelete to the current one
    fn apply(&mut self, other: Self) {
```

Comment in the wrong place

massa-async-pool/src/changes.rs
```diff
@@ -0,0 +1,54 @@
...
+        *self = other;
+    }
+}
+
+/// represents a list of additions and deletions to the async message pool
+#[derive(Default, Debug, Clone)]
+pub struct AsyncPoolChanges(pub(crate) Vec<(AsyncMessageId, AddOrDelete<AsyncMessage>)>);
```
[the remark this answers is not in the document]
AureliaDolo: Kinda like the signed struct stuff?

massa-async-pool/src/message.rs
```diff
@@ -0,0 +1,219 @@
...
+
+/// Structure defining an asynchronous smart contract message
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct AsyncMessage {
+    /// Slot at which the message was emitted
+    pub emission_slot: Slot,
+
+    /// Index of the emitted message within the emission_slot.
+    /// This is used for disambiguating the emission of multiple messages at the same slot.
+    pub emission_index: u64,
```
AureliaDolo, line 25: How is this set ?

massa-async-pool/src/message.rs
```diff
@@ -0,0 +1,219 @@
...
+
+    /// Maximum gas to use when processing the message
+    pub max_gas: u64,
+
+    /// Gas price to take into account when executing the message.
+    /// max_gas * gas_price are burned by the sender when the emessage is sent.
+    pub gas_price: Amount,
+
+    /// Coins sent from the sender to the target address of the message.
+    /// Those coins are spent by the sender address when the emessage is sent,
```
AureliaDolo, line 44:
```suggestion
    /// Those coins are spent by the sender address when the message is sent,
```

massa-async-pool/src/message.rs
```diff
@@ -0,0 +1,219 @@
...
+    /// and credited to the destination address when receiving the message.
+    /// In case of failure or discard, those coins are reimbursed to the sender.
+    pub coins: Amount,
+
+    /// Slot at which the message starts being valid (bound included in the validity range)
+    pub validity_start: Slot,
+
+    /// Slot at which the message stops being valid (bound not included in the validity range)
+    pub validity_end: Slot,
```
AureliaDolo, line 53: So it's not the same validity range concept as for operations ?

massa-async-pool/src/message.rs
```diff
@@ -0,0 +1,219 @@
...
+    pub fn compute_id(&self) -> AsyncMessageId {
+        (
+            self.gas_price.saturating_mul_u64(self.max_gas),
+            std::cmp::Reverse(self.emission_slot),
+            std::cmp::Reverse(self.emission_index),
+        )
+    }
+}
+
+/// Allow serializing the AsyncMessage into a compact binary representation
```
AureliaDolo, line 71: Trait doc should be enough

massa-async-pool/src/message.rs
```diff
@@ -0,0 +1,219 @@
...
+        res.extend(data_len.to_varint_bytes());
+
+        // data
+        res.extend(&self.data);
+
+        Ok(res)
+    }
+}
+
+/// Allow deserializing a AsyncMessage from its compact binary representation
```
AureliaDolo, line 125: same as above

massa-async-pool/src/pool.rs
```diff
@@ -0,0 +1,178 @@
...
+    /// async pool config
+    config: AsyncPoolConfig,
+
+    /// messages sorted by increasing ID (increasing priority)
+    messages: BTreeMap<AsyncMessageId, AsyncMessage>,
+}
+
+impl AsyncPool {
+    /// Creates an empty AsyncPool
+    pub fn new(config: AsyncPoolConfig) -> AsyncPool {
```
AureliaDolo, line 30: I'd rather have new take an optional bootstrap snapshot for consistency with other crates. (or open an issue to change that out there)

massa-async-pool/src/pool.rs
```diff
@@ -0,0 +1,178 @@
...
+            }
+
+            // add to selected items
+            selected.push(*msg_id);
+
+            // substract available gas
+            available_gas -= msg.max_gas;
+
+            // if there is no more gas, quit
+            if available_gas == 0 {
```
AureliaDolo, line 164: Maybe we could break if the available gas is smaller than the smallest max_gas.

massa-async-pool/src/test_exports/bootstrap.rs
```diff
@@ -0,0 +1,43 @@
...
+
+/// This file defines tools to test the async pool bootstrap
+
+/// creates an async pool bootstrap state from components
+pub fn make_bootstrap_state(messages: Vec<AsyncMessage>) -> AsyncPoolBootstrap {
+    AsyncPoolBootstrap { messages }
+}
+
+/// asserts that two instances of AsyncMessage are the same
+pub fn assert_eq_async_message(v1: &AsyncMessage, v2: &AsyncMessage) {
```
AureliaDolo, line 13: Should we directly implement Eq for AsyncMessage?

massa-async-pool/src/pool.rs
damip: this would imply going through the whole set to find the smallest max_gas first

massa-async-pool/src/bootstrap.rs
```diff
@@ -0,0 +1,58 @@
...
+    DeserializeCompact, DeserializeVarInt, ModelsError, SerializeCompact, SerializeVarInt,
+};
+use serde::{Deserialize, Serialize};
+
+/// Represents a snapshot of the async pool aimed at bootstrapping an AsyncPool
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct AsyncPoolBootstrap {
+    /// list of async messages in no particular order
+    pub(crate) messages: Vec<AsyncMessage>,
+}
```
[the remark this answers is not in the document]
AurelienFT: I don't think it's a good practice to always turn struct with one field to a new type pattern since when accessing the field you make it by 0 instead of the real name which loose clarity.

massa-async-pool/src/test_exports/bootstrap.rs
Eitu33: We should keep it separated imo, there is no use (yet?) for being able to match the content of two messages (apart from testing).

massa-async-pool/src/changes.rs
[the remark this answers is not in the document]
damip: The thing is that elsewhere we have different kinds of "change":
* AddOrDelete
* KeepOrChange
* SetUpdateOrDelete
* SetOrKeep

That's why I preferred to stay explicit here

massa-async-pool/src/changes.rs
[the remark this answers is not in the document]
damip: we're not supposed to sign messages !

massa-async-pool/src/test_exports/bootstrap.rs
damip: Exactly. I believe that testing concerns should be kept outside of the main code

massa-async-pool/src/bootstrap.rs
```diff
@@ -0,0 +1,58 @@
...
+
+/// Allows deserializing a AsyncPoolBootstrap from its compact binary representation
+impl DeserializeCompact for AsyncPoolBootstrap {
+    fn from_bytes_compact(buffer: &[u8]) -> Result<(Self, usize), massa_models::ModelsError> {
+        let mut cursor = 0usize;
+
+        // message count
+        let (message_count, delta) = u64::from_varint_bytes(&buffer[cursor..])?;
+        // TODO cap message count https://github.com/massalabs/massa/issues/1200
```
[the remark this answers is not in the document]
damip: The serialization improvements were targeting TEST9 but will need to be moved (https://github.com/massalabs/massa/milestone/19)

massa-async-pool/src/pool.rs
```diff
@@ -0,0 +1,201 @@
...
+        mut available_gas: u64,
+    ) -> Vec<AsyncMessage> {
+        // gather all selected items and remove them from self.messages
+        // iterate in decreasing priority order
+        let (removed, kept) = std::mem::take(&mut self.messages)
+            .into_iter()
+            .rev()
+            .partition(|(_, msg)| {
+                // check available gas and validity period
+                if available_gas > msg.max_gas
```
damip, line 150: available_gas >= msg.max_gas

massa-async-pool/src/pool.rs
```diff
@@ -0,0 +1,201 @@
...
+        &mut self,
+        slot: Slot,
+        mut available_gas: u64,
+    ) -> Vec<AsyncMessage> {
+        // gather all selected items and remove them from self.messages
+        // iterate in decreasing priority order
+        let (removed, kept) = std::mem::take(&mut self.messages)
+            .into_iter()
+            .rev()
+            .partition(|(_, msg)| {
```
damip, line 148: I'm afraid partition() will iterate over the whole thing. But that's ok.

My dream would be a split_off_by_key version of  https://doc.rust-lang.org/std/collections/struct.BTreeMap.html#method.split_off

damip, line 148: I have one concern with partition() => will the "kept" part be completely re-sorted ?

damip: Why not use drain_filter ?

AurelienFT: Because we have to `rev` before which need an iterator and then drain filter that don't work on iterator and we can't collect object reversed in a btreemap because they are sorted directly.

AurelienFT: Partition will keep the order so it will not need a re-sorting. I don't know if it's more costly or not

AurelienFT (approved):
Some warnings about unused but they will be resolved with the other part of the feature.

massa-async-pool/src/pool.rs
damip: in the end, we chose to keep objects in reverse order (using https://doc.rust-lang.org/std/cmp/struct.Reverse.html ) and do drain_filter i

massa-async-pool/src/bootstrap.rs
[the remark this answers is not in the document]
AureliaDolo: Maybe it would be worth having a kanban style board with the high level concepts you put in the milestone to move them around more easily ?

massa-async-pool/src/bootstrap.rs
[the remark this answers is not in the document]
AureliaDolo: As it's a vec, it seems weird to wrap it in a struct. Maybe a new type pattern
```suggestion
/// list of async messages in no particular order
pub type AsyncPoolBootstrap = Vec<AsyncMessage>;
```

massa-async-pool/src/pool.rs
AureliaDolo: indeed

massa-async-pool/src/pool.rs
```diff
@@ -0,0 +1,192 @@
...
+    ///
+    /// # arguments
+    /// * slot: used to filter out expired messages, not stored
+    /// * new_messages: list of AsyncMessage to add to the pool
+    ///
+    /// # returns
+    /// The list of (message_id, message) that were eliminated from the pool after the changes were applied, sorted in the following order:
+    /// * expired messages from the pool, in priority order (from lowest to highest priority)
+    /// * expired messages from new_messages (in the order they appear in new_messages)
+    /// * excess messages after inserting all remaining new_messages, in priority order (from lowest to highest priority)
```
damip, line 92:
```suggestion
    /// * excess messages after inserting all remaining new_messages, in priority order (from highest to lowest priority)
```

massa-async-pool/src/test_exports/bootstrap.rs
AureliaDolo: But you could hide the trait implementation behind a flag, and that would allow to use directly the == operator only in tests

massa-async-pool/src/bootstrap.rs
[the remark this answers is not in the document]
AurelienFT: I don't think it's a discussion related to this PR. it's more a organizational one.

massa-async-pool/src/message.rs
Eitu33: In send_message interface implementation: https://github.com/massalabs/massa/blob/async-pool-integration/massa-execution-worker/src/interface_impl.rs#L469

massa-async-pool/src/bootstrap.rs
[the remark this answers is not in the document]
AurelienFT: To keep harmonization with ledger and be able to have more data in the future it's also better to have a structure. Also when you use it you will directly use the vec methods which is not clear and will need changes all accross the code if we need an other field.

massa-async-pool/src/pool.rs
AurelienFT: I don't see any options in new.
In final state : https://github.com/massalabs/massa/blob/main/massa-final-state/src/final_state.rs#L34
In ledger : https://github.com/massalabs/massa/blob/main/massa-ledger/src/ledger.rs#L77 for examples

massa-async-pool/src/test_exports/bootstrap.rs
damip: That's an option, but is the effort and added complexity really worth it given that this is used only once in tests, especially since it would not change anything except giving less information when a mismatch happens (currently the panic says which sub-field mismatched) ?

AurelienFT: Done.

damip: I disagree with the changes.

AurelienFT: Ok I can revert

massa-async-pool/src/pool.rs
AureliaDolo: https://github.com/massalabs/massa/blob/932870f1df459ebb0e84a42425df5d10249aef6c/massa-graph/src/block_graph.rs#L428
https://github.com/massalabs/massa/blob/932870f1df459ebb0e84a42425df5d10249aef6c/massa-graph/src/ledger.rs#L154

massa-async-pool/src/bootstrap.rs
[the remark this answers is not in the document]
adrien-zinger: If we need another field? Like what?

If that struct evolves soon, ok, we would like a struct. But if it's just to hide a data structure, a type is enough. As `AsyncMessage` is in our crate, we also don't care about the orphan rule, and we can impl whatever we want as if it was a struct.

massa-async-pool/src/pool.rs
AurelienFT: It's only in block_graph and all the rest of the code is matching the pattern we current have. I don't think we should change it.

AurelienFT (dismissed):

massa-async-pool/src/pool.rs
AureliaDolo: https://github.com/massalabs/massa/blob/932870f1df459ebb0e84a42425df5d10249aef6c/massa-proof-of-stake-exports/src/proof_of_stake.rs#L42-L46

The issue may be that we never took the time to agree on a common ground, and as the team is kinda growing, now could be the time

In my opinion, we can't really expect even specifying streaming bootstrap before may, as there is too much huge stuff going on that need to be stabilized. I hope we reach a consensus before \o/

massa-async-pool/src/message.rs
AureliaDolo: Ok so no fixed width for that range, right ? As for the upper bound not being included, is it documented somwhere for the user ?

massa-async-pool/src/bootstrap.rs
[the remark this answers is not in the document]
AureliaDolo: #2457

massa-async-pool/src/changes.rs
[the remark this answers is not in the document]
AureliaDolo: My point was about the way Signed is implemented, with the phantom data to keep track of the id type, and maybe we could do something similar here

massa-async-pool/src/pool.rs
```diff
@@ -21,8 +21,8 @@ pub struct AsyncPool {
     /// async pool config
     config: AsyncPoolConfig,
 
-    /// messages sorted by increasing ID (increasing priority)
-    messages: BTreeMap<AsyncMessageId, AsyncMessage>,
+    /// messages sorted by decreasing ID (decreasing priority)
```
AureliaDolo, line 24: Not sure I see how the id is related to the priority

damip: we uniquely identify every message by its priority tuple :) that way we have only one type of ID to use in the BTreeMap

AureliaDolo: It feels odd calling that an id as everywhere else ids are hashes, but maybe I'm getting too mush used to this code base specificity :sauropod:

massa-async-pool/src/changes.rs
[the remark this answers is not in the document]
AurelienFT: It has been change to `Change`

massa-async-pool/src/changes.rs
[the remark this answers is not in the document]
AurelienFT: I don't see use cases here for that.

massa-async-pool/src/bootstrap.rs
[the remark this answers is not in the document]
AurelienFT: I don't think it's for hide a type as it's a structure for Bootstrap (as I deduce from the name) if in the future we want to for example save a value for the max number of messages etc...

massa-async-pool/src/changes.rs
```diff
@@ -0,0 +1,54 @@
+//! Copyright (c) 2022 MASSA LABS <info@massa.net>
+
+//! This file provides structures representing changes to the async message pool
+
```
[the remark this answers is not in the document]
Eitu33: That was done, I am resolving this discussion.

massa-async-pool/src/bootstrap.rs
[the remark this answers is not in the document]
damip: Since this debate won't end due to it being related to personal preference, I will decide "arbitrarily" to unlock the situation, but we can discuss this afterwards. 

=> let's keep it as is for now

massa-async-pool/src/pool.rs
damip: The downside of having the boot_pos is that we always end up with a big "if Some(xx) { [ a lot of code ] } else { [ a lot of code ] }" inside the "new" function. Having a new(config) function to create an empty structure and a from_bootstrap function to create from a bootstrap structure should be the preferred choice as it is more explicit and the functions are better split. That's why I did it in the execution module.

AurelienFT (approved):

Eitu33 (approved):

damip:
bors merge

PR #2437, opened 2022-03-17
Remove deserialization of phantom field.

### LINKED ISSUE #2436: API endpoints need phantom data
Now that the Signature has been refactor we use PhantomData in structure.

The problem is that we don't want user to have to provide this type as it's only a PhantomData and it break the current format of the API.

Solution : https://play.rust-lang.org/?gist=cbfadccf2dc1a3b4b4eb2279be2a4535&version=stable

AurelienFT:
Fix #2436

### DIFF
--- a/massa-models/src/signed.rs
+++ b/massa-models/src/signed.rs
@@ -14,8 +14,10 @@ where
 {
     pub content: T,
     pub signature: Signature,
+    #[serde(skip_deserializing)]
     phantom: PhantomData<U>,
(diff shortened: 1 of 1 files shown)


Eitu33 (approved):

AureliaDolo (approved):

AureliaDolo:
We REALLY need #1182

AurelienFT:
bors merge

PR #2440, opened 2022-03-18
Remove warning in massa-ledger

adrien-zinger:
Reminders: :leafy_green: 
- don't merge warnings into main.
- don't validate a code with a warning inside


AurelienFT (approved):
I let it when I saw the PR because I know the feature is still in WIP and it will be solved soon but we can resolve if it's a better practice !

AureliaDolo:
> I let it when I saw the PR because I know the feature is still in WIP and it will be solved soon but we can resolve if it's a better practice !

Maybe it's then better to have a feature branch in which we can merge the WIP

AurelienFT:
bors merge

PR #2443, opened 2022-03-18
Fix serializing of operations.

AurelienFT:


adrien-zinger (approved):
I just add some comments on that change :wink:

AurelienFT:
> I just add some comments on that change 😉

Thank you

PR #2444, opened 2022-03-18
Feature/operations batch/implement

AurelienFT:


massa-protocol-worker/src/protocol_worker.rs
```diff
@@ -144,6 +146,11 @@ pub struct ProtocolWorker {
     checked_operations: Set<OperationId>,
     /// List of processed headers
     checked_headers: Map<BlockId, BlockInfo>,
+    /// List of operations received
+    // TODO: Operation should be store when receive and we work only with id in the future.
```
qdrn, line 150: TODO: OperationS should be storeD when receiveD and we work only with id in the future.

AurelienFT: Done.

massa-protocol-worker/src/protocol_worker.rs
```diff
@@ -141,9 +143,20 @@ pub struct ProtocolWorker {
...
+    checked_operations: OperationIds,
     /// List of processed headers
     checked_headers: Map<BlockId, BlockInfo>,
+    /// List of operations received
+    // TODO: Operation should be stored when received and we work only with id in the future.
+    received_operations: OperationIds,
+    /// List of ids of operations that we asked to the nodes
+    asked_operations: HashMap<OperationId, (Instant, Vec<NodeId>)>,
+    /// Buffer for operations that we want later
+    op_batch_buffer: VecDeque<(Instant, NodeId, OperationIds)>,
```
adrien-zinger, line 155: It's better to use the type OperationBatchBuffer

PR #2445, opened 2022-03-21
Fix workers scenarios, adapt to the new algo

adrien-zinger:
Adapt the scenarios to the new algorithm of operations propagation


massa-protocol-worker/src/tests/operations_scenarios.rs
```diff
@@ -280,6 +279,10 @@ async fn test_protocol_propagates_operations_only_to_nodes_that_dont_know_about_
     .await;
 }
 
+/*
+
+TODO: the following test might be obsolete with the new algorithm
+
 #[tokio::test]
 #[serial]
 async fn test_protocol_propagates_operations_only_to_nodes_that_dont_know_about_it_block_integration(
```
adrien-zinger, line 288: Here there is some tests that a probably obsoletes.

We can just check if the protocol call correctly the network when it should without testing the full content of the response.

We can test that in followup. Likewise, we also probably don't need to start all that mocks to test f the operation propagation work correctly.

In follow up, with the new testes after the main also implementation, we should remove that part.

adrien-zinger: To avoid a lost of time I would like to add inside:

```suggestion
/*

TODO: the following test might be obsolete with the new algorithm, if you see hat comment in a month, please delete me

#[tokio::test]
#[serial]
async fn test_protocol_propagates_operations_only_to_nodes_that_dont_know_about_it_block_integration(
```

massa-protocol-worker/src/tests/operations_scenarios.rs
```diff
@@ -649,24 +654,24 @@ async fn test_protocol_propagates_operations_only_to_nodes_that_dont_know_about_
...
                 .await
             {
-                Some(NetworkCommand::SendOperations { node, operations }) => {
-                    let id = operations[0].content.compute_id().unwrap();
-                    assert_eq!(id, operation_id_2);
-                    assert_eq!(nodes[0].id, node);
+                Some(NetworkCommand::AskForOperations { to_node, wishlist }) => {
+                    assert_eq!(wishlist.len(), 1);
+                    assert!(wishlist.contains(&operation_id_2));
+                    assert_eq!(nodes[0].id, to_node);
```
adrien-zinger, line 674: Now the propagation will just ask the received batch and not "Send" it.

In follow up, we will test if the algorithm send correctly the operation after receiving the full operations.

massa-protocol-worker/src/tests/operations_scenarios.rs
adrien-zinger: outdated

massa-protocol-worker/src/tests/operations_scenarios.rs
adrien-zinger: outdated

AurelienFT (approved):

AureliaDolo (approved):

PR #2450, opened 2022-03-21
Inter epoch finalization test

damip:
# Context
Regression test for https://github.com/massalabs/massa/pull/2433
When we have the following block sequence
1 thread, periods_per_cycle = 2, delta_f0 = 1, 1 endorsement per block
cycle 0 | cycle 1 | cycle 2
 G - B1 - B2 - B3 - B4
where G is the genesis block
and B4 contains a roll sell operation
And the block B1 is received AFTER B4, blocks will be processed recursively:
* B1 is received and included
* B2 is processed
* B1 becomes final in the graph
* B3 is processed
* B2 becomes final in the graph
* B4 is processed
* B3 becomes final in the graph
* PoS is told about all finalized blocks

The problem we had is that in order to check rolls to verify B4's roll sell,
the final roll registry was assumed to be attached to the last final block known by the graph,
but that was inaccurate because PoS was the one holding the final roll registry,
and PoS was not yet aware of the blocks that finalized during recursion,
so it was actually still attached to G when B4 was checked.
The correction involved taking the point of view of PoS on where the final roll registry is attached.
This test ensures non-regression by making sure B4 is propagated when B1 is received.

### DIFF
--- a/massa-consensus-worker/src/tests/inter_cycle_batch_finalization.rs
+++ b/massa-consensus-worker/src/tests/inter_cycle_batch_finalization.rs
@@ -1,0 +1,189 @@
+//! Copyright (c) 2022 MASSA LABS <info@massa.net>
+
+use massa_models::{ledger_models::LedgerData, Address, Amount, BlockId, Slot};
+use massa_signature::{derive_public_key, generate_random_private_key};
+use massa_time::MassaTime;
+use serial_test::serial;
+use std::str::FromStr;
+
+use super::tools::*;
+use massa_consensus_exports::ConsensusConfig;
+
+/// # Context
+///
+/// Regression test for https://github.com/massalabs/massa/pull/2433
+///
+/// When there are 2 cliques C1 and C2, and a new block B arrives and makes C2 stale,
+/// a whole batch of blocks from C1 will suddenly become final.
+/// However, clique computation, and stale/final block identification is done
+/// after checking the integrity of B and adding it to the graph.
+/// And once all of that is done, PoS is informed of the newly finalized block batch.
+/// Therefore, while B is being checked, PoS is not yet aware of the blocks
+/// finalized by the addition of B to the graph.
+///
+/// On the other hand when verifying roll sales happening in B (before its addition to the graph),
+/// we need to go back to the latest final block to which the final roll registry is attached,
+/// and apply roll changes coming from subsequent active blocks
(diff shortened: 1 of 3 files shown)


massa-consensus-worker/src/tests/inter_cycle_batch_finalization.rs
```diff
@@ -0,0 +1,189 @@
...
+            // B1, B2 = empty blocks of the alternative clique, 0 endorsements
+            //
+            // When A3 arrives:
+            // * B1 and B2 become stale
+            // * A1, A2 become final
+            // * A3 should be propagated
+
+            // Create, send and propagate B1
+            let (b1_id, b1_block, _) =
+                create_block(&cfg, Slot::new(1, 0), genesis_blocks.clone(), staking_key);
```
AurelienFT, line 113: Why not use create and serd block everywhere ?

AurelienFT (approved):
Well done.

massa-consensus-worker/src/tests/inter_cycle_batch_finalization.rs
damip: corrected everything

AurelienFT (approved):

massa-consensus-worker/src/tests/inter_cycle_batch_finalization.rs
```diff
@@ -0,0 +1,158 @@
...
+
+            // create and send B2
+            tokio::time::sleep(t0.to_duration()).await;
+            let (b2_id, b2_block, _) = create_block_with_operations_and_endorsements(
+                &cfg,
+                Slot::new(2, 0),
+                &vec![b1_id],
+                staking_key,
+                vec![],
+                vec![create_endorsement(staking_key, Slot::new(1, 0), b1_id, 0)],
```
AureliaDolo, line 101: Are the endowment need for this test to be a regression test ?

AureliaDolo (approved):

massa-consensus-worker/src/tests/inter_cycle_batch_finalization.rs
damip: the endorsement allows having delta_f0 = 1 (the minimum) while ensuring that every new block finalizes the previous one (by adding a fitness of 2 instead of 1)

damip:
bors merge

PR #2453, opened 2022-03-22
correct bad child deletion (variable shadowing typo)

damip:
Correct a typo due to variable shadowing.

This PR simply renames variables to prevent the shadowing.

### DIFF
--- a/massa-graph/src/block_graph.rs
+++ b/massa-graph/src/block_graph.rs
@@ -2836,10 +2836,10 @@ impl BlockGraph {
 
                 // remove from parent's children
                 for (parent_h, _parent_period) in active_block.parents.iter() {
-                    if let Some(BlockStatus::Active(active_block)) =
+                    if let Some(BlockStatus::Active(parent_active_block)) =
                         self.block_statuses.get_mut(parent_h)
                     {
-                        active_block.children
+                        parent_active_block.children
                             [active_block.block.header.content.slot.thread as usize]
                             .remove(&stale_block_hash);
                     }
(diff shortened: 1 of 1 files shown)


AurelienFT (approved):
Nice catch

adrien-zinger (approved):

adrien-zinger (commented):
After all I don't think there is a showing problem here, you change scope and when you shadow you don't drop anything, you retrieve the good variable after the if statement. Whatever, that is a good modif for a better variable name.

Here is an example.
https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=e81f72a6cb2dc8e110d463b8b6da48c7

and another with the if let:
https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=38a394d4061f53b4e670bc7b083ad9c0

damip:
bors merge

damip:
> After all, there is no shadowing problem. Here, you change the scope, the shadow don't drop anything, then you retrieve the good variable after the if statements. Whatever, that is a good modif for a better variable name.
> 
> Here is an example. https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=e81f72a6cb2dc8e110d463b8b6da48c7
> 
> and another with the if let: https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=38a394d4061f53b4e670bc7b083ad9c0

In the first modif, the shadowing confusion gave the wrong thread number (we needed the thread coming from the outer active_block variable and everything else fron the inner one ). The second modif is purely cosmetic

Issue #2283, opened 2022-02-11
Optimize operation propagation


damip:
# Rationale

Naked (out-of-block) operations represent the bulk of the network activity under heavy load, as in the ternet currently.
We also observe that the network has no latency problems.

We need to reduce bandwidth usage by propagating operations in a way similar to what we do for blocks: send a small advert to everyone around us except nodes that already have it, and send the full operation only to who ask for it.

This will massively reduce bandwidth usage (~x10) which is desirable, and only slightly increase the latency of operation propagation which is non-critical because operation inclusion time is t0/2 = 8 seconds on average.

#  How we do it

## Setup

Have a `asking_ops: HashMap<OperationId, (NodeId, Instant)>` field in Protocol indicating to which nodes we are asking which operations and when we started doing so.

## Asking process (ask_ids)

This will figure out the best way to ask for a list of op IDs (ask_ids).

* remove all IDs that are in asking_ops from ask_ids
* remove all operation IDs we know about from ask_ids
* remove all IDs that no node knows about from ask_ids
* while ask_ids is not empty:
  * compute the intersection inter_ids of each node's operation knowledge with ask_ids
  * choose the node with the largest inter_ids to batch as much as possible (TODO also sort by how much stuff that node is already being asked for to prevent flooding ? By how recently the intersection knowledge was announced because because ops could have been honestly dropped by nodes since they announced them ?):
    * send a AskOperations(inter_ids) message to the node 
    * insert inter_ids entries in ask_ids for that node and time now()
    * remove inter_ids from ask_ids

## asking_ops management

* if a node disconnects, if it had an entries in asking_ops:
  * remove all entries matching this node ID and gather all the operation ids into a ask_ids hashset, then ask for them using the Asking process above
* if an element in asking_ops times out according to a config-defined timeout
  * ban the node and run the "node disconnects" process above

## Operations message

Remove the Operations messasge

## Advertise operations

Create the message::AdvertiseOperations message to advertise a list of operation IDs to a node:
```rust
AdvertiseOperations(Vec<OperationId>)
```
This message can contain up to `massa_network::settings::max_operations_per_message` IDs. If a request needs more elements, it must be split into multiple messages.

This messsage is sent to nodes not already knowing about an operation that we just learned about and received in full.

When we receive a AdvertiseOperations message, we mark the sender node as knowing those operations (like we do for blocks when we receive their header), then we run the Asking process for the received operations list.

## Ask for operations

[...]

PR #2454, opened 2022-03-23
Batch limitation feature

adrien-zinger:
Add the batch limitation logic from https://github.com/massalabs/massa/issues/2283#issuecomment-1040872779.

We should limit the number of batch we receive by second from node. If exceeded, ban the node.

### DIFF
--- a/massa-protocol-worker/src/worker_operations_impl.rs
+++ b/massa-protocol-worker/src/worker_operations_impl.rs
@@ -250,4 +250,37 @@ impl ProtocolWorker {
             .send_operations(node_id, operations)
             .await
     }
+
+    pub(crate) async fn send_operation_ids_batch(
+        &mut self,
+        propagate_operations_timer: &mut std::pin::Pin<&mut Sleep>,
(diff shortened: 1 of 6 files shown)


AurelienFT (approved):

massa-protocol-worker/src/worker_operations_impl.rs
```diff
@@ -23,13 +23,53 @@ use tokio::time::{sleep_until, Instant, Sleep};
 use tracing::warn;
 
 impl ProtocolWorker {
+    /// Limit the number of batch an external node can send in a second. Ban
+    /// it if it exceed self.settings.max_op_batch_per_sec_per_node
+    ///
+    /// Call that function each time we receive a batch from a `node_id`
+    ///
+    /// Return true if node has been considered as bannished, false otherwise.
+    async fn limitation_batches_by_node(&mut self, node_id: NodeId) -> bool {
```
AureliaDolo, line 32: There is something odd in my opinion with that function name maybe batches by node limitation or limit batches by node

PR #2456, opened 2022-03-23
repair bug 2386

damip:


damip:
I also used the opportunity to avoid worker.rs accessing execution.active_history directly

AurelienFT (approved):

massa-execution-worker/src/execution.rs
```diff
@@ -39,24 +39,24 @@ macro_rules! context_guard {
...
+    final_events: EventStore,
     // final state with atomic R/W access
-    pub final_state: Arc<RwLock<FinalState>>,
+    final_state: Arc<RwLock<FinalState>>,
     // execution context (see documentation in context.rs)
-    pub execution_context: Arc<Mutex<ExecutionContext>>,
+    execution_context: Arc<Mutex<ExecutionContext>>,
     // execution interface allowing the VM runtime to access the Massa context
-    pub execution_interface: Box<dyn Interface>,
+    execution_interface: Box<dyn Interface>,
```
AureliaDolo, line 59: Why this became public ? Don't we want to avoid that ?

AureliaDolo (changes requested):
Why are these panics for ? And they are not document :confused:

massa-execution-worker/src/execution.rs
AurelienFT: They don't become pub they become private.

damip: The panics are documented though their explicit error message and used to ensure some bounds on the cursors because we're not sure that this PR fully corrected the problem (we couldn't replicate it). That way, next time we get this problem, we will have more info.

The "pub" are not added but removed to improve separation of concerns.

There are further explanations in the issue.

AureliaDolo: My bad 

As for the panics, really it seems weird for me as it's the thing we want to avoid at all cost. I would at least add a "to remove before mainnet" comment

damip: Panics in are always useful. They match the error-bubbling-then-toplevel-panic we have in async contexts: we did this in async contexts because panic traceback is unreadable, but this is not the case in non-async contexts.

adrien-zinger: > Panics in are always useful. They match the error-bubbling-then-toplevel-panic we have in async contexts: we did this in async contexts because panic traceback is unreadable, but this is not the case in non-async contexts. So we should panic on non-recoverable errors within non-async contexts.

Just for that, I'm still confused about which Error we can dismiss and which will make the node crash... Maybe we could think about a naming convention like `Warning` `Error` and `Critical` in a refacto in the future. Not related but just notice that point please.

damip: for now our philosophy (also elsewhere in the code) is to crash the node. We only reboot on desync (but not on container inconsistency)

damip: > > They don't become pub they become private.
> 
> They _became_, it means why someone made that pub before ? +1 I think everybody can read a git diff in the team. wink

I messed up and it went through the review process. But no big deal ! Could clippy setup to catch things that are uselessly public this in the future ?

damip: > > Panics in are always useful. They match the error-bubbling-then-toplevel-panic we have in async contexts: we did this in async contexts because panic traceback is unreadable, but this is not the case in non-async contexts. So we should panic on non-recoverable errors within non-async contexts.
> 
> Just for that, I'm still confused about which Error we can dismiss and which will make the node crash... Maybe we could think about a naming convention like `Warning` `Error` and `Critical` in a refacto in the future. Not related but just notice that point please.

I think that in non-async contexts, an entity that emits something that is not obviously unrecoverable (otherwise it would panic) should simply name it ErrorXXX, because it is not supposed to know how this error will be handled upstream. And the upstream caller should decide what to do with the error: panic, reboot, print a warning, ignore...

adrien-zinger: > > > Panics in are always useful. They match the error-bubbling-then-toplevel-panic we have in async contexts: we did this in async contexts because panic traceback is unreadable, but this is not the case in non-async contexts. So we should panic on non-recoverable errors within non-async contexts.
> > 
> > 
> > Just for that, I'm still confused about which Error we can dismiss and which will make the node crash... Maybe we could think about a naming convention like `Warning` `Error` and `Critical` in a refacto in the future. Not related but just notice that point please.
> 
> I think that in non-async contexts, an entity that emits an error that is not obviously unrecoverable (otherwise it would panic) should simply name it ErrorXXX, because it is not supposed to know how this error will be handled upstream. And the upstream caller should decide what to do with the error: panic, reboot, print a warning, ignore... And different callers could have different error sensitivities for the same called function.

I think you don't get my point, maybe we can talk later :slightly_smiling_face: about that. It's easy to know if we need to handle an error while we're developing. But when someone else adds a feature, he doesn't know if returning an error is  critical or not.

I remember that in some PR I needed to dismiss errors, but in the workflow wasn't clear at all because the error handling was too high in the genesis caller in another future :rofl: , and fortunately you made a comment at that moment! Now, the documentation help a little more :+1:. I can show you an example on one of my project later if you want :slightly_smiling_face:

damip: Okay feel free to ping me to talk about it :) might be easier than writing. Solving this

AureliaDolo: for me it's ok, I'll let @adrien-zinger solve this once he'll have read your reply :)

AureliaDolo (approved):

damip:
bors merge

damip:
bors merge

AurelienFT:
bors merge

PR #2459, opened 2022-03-24
Async pool integration

Eitu33:


massa-async-pool/src/changes.rs
```diff
@@ -51,4 +51,16 @@ impl AsyncPoolChanges {
     pub fn push_delete(&mut self, msg_id: AsyncMessageId) {
         self.0.push(Change::Delete(msg_id));
     }
+
+    /// Retrieve only the added messages
+    pub fn get_add(&self) -> Vec<(AsyncMessageId, AsyncMessage)> {
```
AurelienFT, line 56:
```suggestion
    pub fn get_adds(&self) -> Vec<(AsyncMessageId, AsyncMessage)> {
```

Because the function gets multiple add messages

massa-execution-worker/Cargo.toml
```diff
@@ -23,10 +23,11 @@ tracing = "0.1"
 massa_execution_exports = { path = "../massa-execution-exports" }
 massa_models = { path = "../massa-models" }
 massa_hash = { path = "../massa-hash" }
-massa-sc-runtime = { git = "https://github.com/massalabs/massa-sc-runtime", tag = "v0.4.4" }
+massa-sc-runtime = { git = "https://github.com/massalabs/massa-sc-runtime", rev = "0719439066576b7c3681d9cf95c31722be6a4734" }
```
AurelienFT, line 26: Maybe add a todo or should be release a new version of massa sc runtime before merging this ?

massa-execution-worker/src/context.rs
```diff
@@ -355,7 +380,7 @@ impl ExecutionContext {
         self.speculative_ledger.has_data_entry(address, key)
     }
 
-    /// gets the bytecode of an address if it exists in the speculative ledger, or returns None
+    /// TODO: this is not what this function does, re-define this comment
```
AurelienFT, line 383: Redefine it before merging.

massa-node/src/main.rs
```diff
@@ -135,10 +136,15 @@ async fn launch() -> (
     let ledger_config = LedgerConfig {
         initial_sce_ledger_path: SETTINGS.ledger.initial_sce_ledger_path.clone(),
     };
+    let async_pool_config = AsyncPoolConfig {
+        // TODO: define max_length value, it musn't come from a settings file
+        max_length: 100,
```
AurelienFT, line 141: Why not defining in default.rs like the others values ? Is default.rs considered as a config file?

AurelienFT (commented):
Very well documented and understandable even if a little presentation on the theory would be cool. Great job !

massa-async-pool/src/changes.rs
Eitu33: @AurelienFT Not a big fan of `get_adds` what about `get_added_messages`?

massa-execution-worker/Cargo.toml
Eitu33: TODO was added, we'll see if we do end up releasing a new version of `massa-sc-runtime` before merging this.

massa-async-pool/src/changes.rs
AurelienFT: I prefer also

damip:
I made an improvement proposition as the commit: https://github.com/massalabs/massa/pull/2459/commits/f9bcd78d5df4a6e5c6a2f71ab16b8f70e3130ae1

Eitu33:
> I made an improvement proposition as the commit: [f9bcd78](https://github.com/massalabs/massa/commit/f9bcd78d5df4a6e5c6a2f71ab16b8f70e3130ae1)

@damip I find the vast majority of your changes justified and think we should keep the proposition.

That said, I do find that `AsyncPool::settle_slot` does too much. I believe it should only be computing data directly related to `AsyncPool` and not `SpeculativeAsyncPool` data. Also computing the messages IDs at the start of `AsyncPool::settle_slot` is unnecessary since we could just compute them when adding a new message. I made these changes in the following commit: [2e3ed17](https://github.com/massalabs/massa/pull/2459/commits/2e3ed1721a68d513f7454c51d23f8d6602818ea5).

I do not think that these changes are mandatory, just slightly better. I'm ok with the previous version as well.

damip:
I agree with your changes

damip (approved):

damip (dismissed):

damip (approved):

damip (approved):

AurelienFT (approved):

Eitu33:
bors merge

PR #2464, opened 2022-03-28
Add log after banned a node.

### LINKED ISSUE #2412: Disconnect from a node when we receive corrupted data
I tried to run a node on the testnet and received some corrupted data from others node but I didn't have any print about banning this node. Screenshot :

Screenshot 2022-03-14 at 12 05 03

I don't receive any corrupted data from this node in the future so maybe I have disconnected from her but I don't have any print.

AurelienFT:
Fix #2412

### DIFF
--- a/massa-network-worker/src/network_cmd_impl.rs
+++ b/massa-network-worker/src/network_cmd_impl.rs
@@ -36,7 +36,7 @@ use std::{
     net::IpAddr,
 };
 use tokio::sync::oneshot;
-use tracing::{debug, warn};
+use tracing::{info, warn};
 
 /// Remove the `ids` from the `worker`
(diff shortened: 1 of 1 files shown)


AureliaDolo (approved):

damip (approved):
Note: this does not solve the problem (the same node shouldn't be sending us invalid data every 200ms for a couple seconds without being banned + disconnected right away after the first one)

AurelienFT:
bors merge

AurelienFT:
> Note: this does not solve the problem (the same node shouldn't be sending us invalid data every 200ms for a couple seconds without being banned + disconnected right away after the first one)
> 
> EDIT: nevermind, I misread the logs

A good example of why this PR is useful 😆

PR #2473, opened 2022-03-30
Revert controller traits

Eitu33:


massa-execution-exports/src/test_exports/mock.rs
```diff
@@ -72,8 +71,8 @@ impl MockExecutionController {
 impl ExecutionController for MockExecutionController {
     fn update_blockclique_status(
         &self,
-        finalized_blocks: HashMap<Slot, BlockId>,
-        blockclique: HashMap<Slot, BlockId>,
+        finalized_blocks: Map<BlockId, Block>,
```
AurelienFT, line 74: I think it should be `Map<BlockId, Block>` like in the `controller.rs`file.

massa-execution-worker/src/controller.rs
```diff
@@ -72,13 +73,23 @@ impl ExecutionController for ExecutionControllerImpl {
...
+        // index newly finalized blocks by slot
+        let mapped_finalized_blocks: HashMap<_, _> = finalized_blocks
+            .into_iter()
+            .map(|(b_id, slot)| (slot, b_id))
+            .collect();
+        // index blockclique by slot
+        let mapped_blockclique = blockclique
+            .into_iter()
+            .map(|(b_id, slot)| (slot, b_id))
+            .collect();
```
gterzian, line 88: So to me mapping the `Map<BlockId, Slot>` to `HashMap<Slot, BlockId>` here, instead of simply having the `HashMap<Slot, BlockId>` to begin with, seems to make things more complicated and I also don't see how swapping one for the other can have an effect on the logic. 

So I'd say let's stick to what makes sense, and try to diagnose the actual problem instead.

AureliaDolo (commented):
In my opinion, mapping slots to blocks ids raises an issue only in not finalized blocks and only if double staked blocks are in the same clique.

As for now double staked blocks are not compatible, so they won't be together in the blockclique.

Eitu33:
Closing this after https://github.com/massalabs/massa/pull/2428#discussion_r838515555

Issue #2465, opened 2022-03-28
Allow tracing on the node


AurelienFT:
Currently we are locking tracing using this feature : https://github.com/massalabs/massa/blob/main/massa-ledger/Cargo.toml#L17 .  We are blocking it because if we compile with it even if we don't have the level enable it will make our node consume 3x times more CPU in calculation of ids.

We need to be able to revert this limit without using more CPU if we don't activate this log level.

damip:
@yvan-sraka  that's the regression we had (tracing parameters being serialized even when not printed). It's still there.

litchipi:
Could you edit the link passed in the issue message ? Got a 404

PR #2494, opened 2022-04-05
Add static limit logs

AurelienFT:
While https://github.com/massalabs/massa/issues/2465 is not fix we need to avoid compiling with trace to avoir mass CPU comsuption.


damip (approved):

damip:
bors merge

PR #2496, opened 2022-04-06
add CallSC support

damip:
# Problem

Being able to easily call a smart contract-exposed function without having to go through the pain of compiling a wasm ExecuteSC to do this simple and common task.

# Solution

Add support for the CallSC operation type that looks like this:

```rust
    /// Calls an exported function from a stored smart contract
    CallSC {
        /// Target smart contract address
        target_addr: Address,
        /// Target function name. No function is called if empty.
        target_func: String,
        /// Parameter to pass to the target function
        param: String,
        /// The maximum amount of gas that the execution of the contract is allowed to cost.
        max_gas: u64,
        /// Extra coins that are spent from the caller's sequential balance and transferred to the target
        sequential_coins: Amount,
        /// Extra coins that are spent from the caller's parallel balance and transferred to the target
        parallel_coins: Amount,
        /// The price per unit of gas that the caller is willing to pay for the execution.
        gas_price: Amount,
    },
```

And does the following:
* consensus substracts gas_price * max_gas + sequential_coins from the sequential balance of the sender, or else the op is invalid
* on execution, we start by crediting gas_price * max_gas to the block producer
* then we credit sequential_coins to the parallel balance of the sender (so that if everything else fails, the coins are not lost)
* then we enter the cancellable section (either everything in this list succeeds or nothing has a consequence):
  * init the call stack with only the caller address (to give it access rights to perform the following coin transfer)
  * transfer sequential_coins + parallel_coins from the sender's parallel balance to the target's parallel balance
  * add the target to the call stack
  * set context.coins = sequential_coins + parallel_coins
  * call the function target_func with the single string parameter param, under max_gas

### DIFF
--- a/massa-execution-worker/src/execution.rs
+++ b/massa-execution-worker/src/execution.rs
@@ -477,7 +477,7 @@ impl ExecutionState {
             bytecode = context.get_bytecode(&target_addr).unwrap_or_default();
 
             // save a snapshot of the context state to restore it if the op fails to execute,
-            // this reverting any changes except the coin transfers above
+            // thus reverting any changes except the coin transfers above
             context_snapshot = context.get_snapshot();
 
             // compute the total amount of coins that need to be transferred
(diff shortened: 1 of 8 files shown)


massa-execution-worker/src/execution.rs
```diff
@@ -293,6 +289,50 @@ impl ExecutionState {
             .compute_id()
             .expect("could not compute operation ID");
 
+        // call the execution process specific to th operation type
```
Eitu33, line 292:
```suggestion
        // call the execution process specific to the operation type
```

massa-execution-worker/src/execution.rs
```diff
@@ -362,6 +402,148 @@ impl ExecutionState {
...
+                    block_creator_addr, gas_fees, err
+                );
+            }
+
+            // Credit the operation sender with `sequential_coins` parallel coins.
+            // This is used to ensure that those coins are not lost in case of failure,
+            // since they have been debited by consensus beforehand.
+            // Note that errors are deterministic and do not cancel op execution.
+            if let Err(err) =
+                context.transfer_parallel_coins(None, Some(sender_addr), sequential_coins)
```
Eitu33, line 468: Why are you crediting the operation sender? Shouldn't he be debited?

Eitu33: Ah it does give back on failure because of the context snapshot, ok nevermind.

Eitu33:
I have taken extra time this once to make sure I don't miss anything but I'd rather have a second review tbh 😅

Eitu33 (approved):

damip:
bors merge

PR #2498, opened 2022-04-06
Add readsc

damip:
Add read-only smart contract function calls in the API.

TODO: update the API documentation once merged !!!

### DIFF
--- a/massa-client/src/cmds.rs
+++ b/massa-client/src/cmds.rs
@@ -3,8 +3,8 @@
 use crate::repl::Output;
 use anyhow::{anyhow, bail, Result};
 use console::style;
-use massa_models::api::ReadOnlyExecution;
 use massa_models::api::{AddressInfo, CompactAddressInfo};
+use massa_models::api::{ReadOnlyBytecodeExecution, ReadOnlyCall};
 use massa_models::prehash::Map;
 use massa_models::timeslots::get_current_latest_block_slot;
 use massa_models::{
@@ -191,6 +191,13 @@ pub enum Command {
         message = "execute byte code, address is optional. Nothing is really executed on chain"
     )]
     read_only_smart_contract,
+
+    #[strum(
+        ascii_case_insensitive,
+        props(args = "TargetAddress TargetFunction Parameter MaxGas GasPrice SenderAddress",),
+        message = "call a smar contract function, sender address is optional. Nothing is really executed on chain"
+    )]
+    read_only_call,
 
     #[strum(
         ascii_case_insensitive,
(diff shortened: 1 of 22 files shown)


damip:
@gterzian  can you take a look ?

massa-execution-exports/src/types.rs
```diff
@@ -19,17 +19,43 @@ pub struct ExecutionOutput {
     pub events: EventStore,
 }
 
-/// structure describing a read-only execution request
+/// structure describing different types of read-only execution request
+#[derive(Debug, Clone)]
+pub enum ReadOnlyRequest {
```
gterzian, line 24: I would suggest 

```rust
/// structure describing different types of read-only execution request
#[derive(Debug, Clone)]
pub enum ReadOnlyRequest {
/// Maximum gas to spend in the execution.
    pub max_gas: u64,
    /// The simulated price of gas for the read-only execution.
    pub simulated_gas_price: Amount,
    /// Call stack to simulate, older caller first
    pub call_stack: Vec<ExecutionStackElement>,
    /// Target of the request
    target: ReadOnlyTarget
}

enum ReadOnlyTarget {
    BytecodeExecution(Vec<u8>),
    FunctionCall {
        /// Target address
        pub target_addr: Address,
        /// Target function
        pub target_func: String,
        /// Parameter to pass to the target function
        pub parameter: String,
        }
}
```

That way you can also unify the two `execute_readonly_` functions in `execution.rs`(they would only differ at the end when calling into runtime)

gterzian (dismissed):
Looks good, just one design suggestion.

massa-execution-exports/src/types.rs
damip: done

damip: very good idea by the way !

massa-client/src/cmds.rs
```diff
@@ -192,6 +192,13 @@ pub enum Command {
     )]
     read_only_smart_contract,
 
+    #[strum(
+        ascii_case_insensitive,
+        props(args = "TargetAddress TargetFunction Parameter MaxGas GasPrice SenderAddress",),
+        message = "call a smar contract function, sender address is optional. Nothing is really executed on chain"
```
AurelienFT, line 198:
```suggestion
        message = "call a smart contract function, sender address is optional. Nothing is really executed on chain"
```

massa-client/src/cmds.rs
```diff
@@ -809,6 +816,37 @@ impl Command {
...
+                if parameters.len() != 5 && parameters.len() != 6 {
+                    bail!("wrong number of parameters");
+                }
+
+                let target_address = parameters[0].parse::<Address>()?;
+                let target_function = parameters[1].parse::<String>()?;
+                let parameter = parameters[2].parse::<String>()?;
+                let max_gas = parameters[3].parse::<u64>()?;
+                let simulated_gas_price = parameters[4].parse::<Amount>()?;
+                let caller_address = if let Some(adr) = parameters.get(5) {
```
AurelienFT, line 829:
```suggestion
                let caller_address = if let Some(addr) = parameters.get(5) {
```

massa-client/src/cmds.rs
```diff
@@ -809,6 +816,37 @@ impl Command {
...
+                    Some(adr.parse::<Address>()?)
```
AurelienFT, line 830:
```suggestion
                    Some(addr.parse::<Address>()?)
```

AurelienFT (approved):

AurelienFT:
Tested. Works well.

Issue #2474, opened 2022-03-30
Storage: remove last use of `expect` in async contexts


gterzian:
For example inside `block_graph`, where there are still a few left ([here](https://github.com/massalabs/massa/pull/2428/files/c9dd61e7e54de844dae8fd2602aaa0a9af27b86b#diff-c4310a8d90af39a317383da61260d64eac43422f2d113b6bbfaeb7a736c8388dR212), [here](https://github.com/massalabs/massa/pull/2428/files/c9dd61e7e54de844dae8fd2602aaa0a9af27b86b#diff-c4310a8d90af39a317383da61260d64eac43422f2d113b6bbfaeb7a736c8388dR1218), and [here](https://github.com/massalabs/massa/pull/2428/files/c9dd61e7e54de844dae8fd2602aaa0a9af27b86b#diff-c4310a8d90af39a317383da61260d64eac43422f2d113b6bbfaeb7a736c8388dR1244)). 

See https://github.com/massalabs/massa/pull/2428#issuecomment-1082899078

PR #2504, opened 2022-04-08
Remove expect in use of storage in async contexts

gterzian:
FIX https://github.com/massalabs/massa/issues/2474

### DIFF
--- a/massa-execution-worker/src/execution.rs
+++ b/massa-execution-worker/src/execution.rs
@@ -671,23 +671,23 @@ impl ExecutionState {
 
         // check if there is a block at this slot
         if let Some(block_id) = opt_block_id {
-            if let Some(block) = self.storage.retrieve_block(&block_id) {
-                let stored_block = block.read();
-                // Try executing the operations of this block in the order in which they appear in the block.
-                // Errors are logged but do not interrupt the execution of the slot.
-                for (op_idx, operation) in stored_block.block.operations.iter().enumerate() {
-                    if let Err(err) = self.execute_operation(
-                        operation,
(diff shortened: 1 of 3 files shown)


AurelienFT (dismissed):

Eitu33:
@gterzian Changes LGTM but there is still one occurence of `expect` on missing block in [`massa/execution.rs#L677`](https://github.com/massalabs/massa/blob/main/massa-execution-worker/src/execution.rs#L677). I'll make the change.

massa-execution-worker/src/execution.rs
```diff
@@ -671,23 +671,23 @@ impl ExecutionState {
...
-            for (op_idx, operation) in stored_block.block.operations.iter().enumerate() {
-                if let Err(err) = self.execute_operation(
-                    operation,
-                    Address::from_public_key(&stored_block.block.header.content.creator),
-                ) {
-                    debug!(
-                        "failed executing operation index {} in block {}: {}",
-                        op_idx, block_id, err
-                    );
+            if let Some(block) = self.storage.retrieve_block(&block_id) {
```
gterzian, line 674: Since this occurs in a threaded context, not async, the expect is ok. Also this would be an unrecoverable error which we would not want to let pass with a `debug`.

Eitu33: Oh okay I was not aware of this, Damir just explained it to me. I will revert my changes.

Eitu33: Done 👍

Eitu33 (approved):

AurelienFT (approved):

Issue #2512, opened 2022-04-08
Make version 9.0 incompatible with 9.1


Eitu33:
## Context

See #2505 

To prevent nodes on version 9.0 from eating all of the available connection from the 9.1 nodes, the two version must be incompatible.

## TO DO

- [x] Update `is_compatible` implementation, adding a match on the minor version only in the `major == 9` case

PR #2514, opened 2022-04-08
Break compatibility between 9.0 and 9.1

Eitu33:
Fix https://github.com/massalabs/massa/issues/2512

### DIFF
--- a/massa-models/src/version.rs
+++ b/massa-models/src/version.rs
@@ -120,6 +120,10 @@ impl DeserializeCompact for Version {
 impl Version {
     /// true if instance and major are the same
     pub fn is_compatible(&self, other: &Version) -> bool {
+        // TODO for next testnet: update this control statement
+        if self.major == 9 && self.minor != other.minor {
+            return false;
+        }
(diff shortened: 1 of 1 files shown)


AurelienFT (approved):

massa-models/src/version.rs
```diff
@@ -120,6 +120,10 @@ impl DeserializeCompact for Version {
 impl Version {
     /// true if instance and major are the same
     pub fn is_compatible(&self, other: &Version) -> bool {
+        // TODO for next testnet: update this control statement
+        if self.major == 9 && self.minor != other.minor {
```
damip, line 124: let's make it more specific to not break the compat with 9.2 afterwards

damip (changes requested):

massa-models/src/version.rs
damip: if self.major == 9 && other.minor == 0 {
return false
}

damip (approved):

damip:
bors merge

PR #2526, opened 2022-04-11
Reverse order events.

AurelienFT:


massa-execution-exports/src/event_store.rs
```diff
@@ -272,7 +272,9 @@ impl EventStore {
             .filter_map(|id| self.id_to_event.get(id))
             .cloned()
             .collect();
-        events.sort_unstable_by_key(|event| (event.context.slot, event.context.index_in_slot));
+        events.sort_unstable_by_key(|event| {
+            Reverse((event.context.slot, event.context.index_in_slot))
```
damip, line 276: this will sort it from the newest to the oldest... the old version was supposed to do the right thing (start with the oldest) 

For some reason it wasn't taken into account (the order was wrong: from newest to oldest)  in the client's readonly SC return

damip (changes requested):

AurelienFT:
Overlap by #2552

Issue #2522, opened 2022-04-09
Node crash


sebastien-forestier:
One of our bootstrap nodes crashed :
[logs_bug.tar.gz](https://github.com/massalabs/massa/files/8457585/logs_bug.tar.gz)

damip:
Seems to be problems around Nodes/Network.

`Failed to send NetworkEvent due to timeout: ReceivedEndorsements `

gterzian:
I see a lot of `[2m2022-04-09T18:05:35.286054Z[0m [34mDEBUG[0m [2mmassa_protocol_worker::worker_operations_impl[0m[2m:[0m re-ask operation vCPYGxTQwx8ESdntTV6a2qZ1r72Bh7yPWRwrHg29hyEvAQF9 asked for the first time 37734 millis ago.`

and apparently not a single `NetworkCommand::SendOperationAnnouncements` receipt being traced(maybe it's because of the log level?). 

I also see the node worker logging it's inablity to send events(to the network worker).

And then we see a lot of failures to send events by the network worker(many seem to be operations announcements coming from other nodes).

gterzian:
so I think one potential improve is not to immediately ask for operations when receiving announcements(same goes for other objects). 

Currently [`on_operations_announcements_received`](https://github.com/massalabs/massa/blob/0835a6175207832d4a7deba00433736b9caa0628/massa-protocol-worker/src/worker_operations_impl.rs#L62) is called both when [receiving announcements](https://github.com/massalabs/massa/blob/0835a6175207832d4a7deba00433736b9caa0628/massa-protocol-worker/src/protocol_worker.rs#L1345), and as part of [`update_ask_operation`](https://github.com/massalabs/massa/blob/0835a6175207832d4a7deba00433736b9caa0628/massa-protocol-worker/src/worker_operations_impl.rs#L176)(which runs according to a timer). 

So one problem might be sending a message to the network worker as part of handling an network event, which when the channels are filled, means each time we have to wait for the network worker to drop an event before it can handle the command. 

I propose that when receiving announcement, we add them to `op_batch_buffer`, and then process as part of `update_ask_operation` at the next timer.

That way when the network sends us an event, it is handled without sending a command back to the network worker, which means that protocol goes back to its select loop and is able to handle the next event from network. 

Eventually, the network commands are sent, but only as part when the timer hits. 

[...]

PR #2527, opened 2022-04-12
Improve operations batching

gterzian:
Possible fix for https://github.com/massalabs/massa/issues/2522

Changes:

1. `on_operations_announcements_received` now only adds to the buffer.
2. buffer is processed at each timer interval.
3. A command is only send to network if a permit can be acquired.
4. An item in the buffer is only removed if sending the message at 3 was successful.
5. If the buffer is full when receiving an announcement is received, we drop the oldest one.

### DIFF
--- a/massa-network-exports/src/network_controller.rs
+++ b/massa-network-exports/src/network_controller.rs
@@ -218,13 +218,12 @@ impl NetworkCommandSender {
         to_node: NodeId,
         wishlist: OperationIds,
     ) -> Result<(), NetworkError> {
-        self.0
-            .send(NetworkCommand::AskForOperations { to_node, wishlist })
-            .await
-            .map_err(|_| {
-                NetworkError::ChannelError("could not send AskForOperations command".into())
-            })?;
-        Ok(())
+        match self.0.try_reserve() {
+            Ok(permit) => Ok(permit.send(NetworkCommand::AskForOperations { to_node, wishlist })),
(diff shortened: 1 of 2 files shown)


gterzian:
@adrien-zinger Here is a start, please take a look. Feel free to make changes, I'm done for the day.

gterzian:
Ok I'd say we try to merge this one, and then we can also consider doing [this](https://github.com/massalabs/massa/issues/2522#issuecomment-1097684672) as a follow-up.

adrien-zinger:
I'm not sure,that change totally the original algorithm of Seb no? We should see with him :)

adrien-zinger (approved):
Yes, that can be interesting to try!!

gterzian:
Do we want to merge this as is, or first deploy the branch for testing on the labnet? @adrien-zinger @damip 

I think we should add some tests also because some should have failed with these changes.

adrien-zinger:
> Do we want to merge this as is, or first deploy the branch for testing on the labnet? @adrien-zinger @damip
> 
> I think we should add some tests also because some should have failed with these changes.

It's to @damip to take the responsibility :rofl:. In my opinion, the current change don't break the network, and we want to iterate fast. :slightly_smiling_face:

gterzian:
@AurelienFT can you take a look please? I think we can merge it after...

AurelienFT (approved):

AurelienFT:
bors merge

PR #2530, opened 2022-04-13
refactor: moving "mock_network_controller.rs" file

temiriqo:
Move "mock_network_controller.rs" from "massa-protocol-exports/src/test" to "massa-network-exports/src/test_exports"


AurelienFT:
The ci is not passing @temiriqo . If you need any help please ask your questions here :)

temiriqo:
Yes I saw because the check looks for "mock_network_controller.rs" under the "massa-protocol-exports/src/test" so It fails. Either the ci need to be reconfigured to adept new location or that file can not be moved. Any idea @AurelienFT ?

AurelienFT:
> Yes I saw because the check looks for "mock_network_controller.rs" under the "massa-protocol-exports/src/test" so It fails. Either the ci need to be reconfigured to adept new location or that file can not be moved. Any idea @AurelienFT ?

Hello, Sorry for late reply. I think you should adapt the tests to match the new path of this file :)

AurelienFT:
This PR is stale i'm closing it.

PR #2542, opened 2022-04-19
Add more size in the API bodies

AurelienFT:
Currently the API accept bodies of maximum size of 5 MiB (default value of `jsonrpc_http_server` : https://docs.rs/jsonrpc-http-server/latest/jsonrpc_http_server/struct.ServerBuilder.html#method.max_request_body_size)

In order to upload bigger SC we need to allow bigger body size in the construction of the RPC API.


Eitu33 (approved):
Very nice 👍

AurelienFT:
bors merge

PR #2546, opened 2022-04-19
add set bytecode

damip:
Adds the ABI allowing to set the bytecode of an address


AurelienFT (commented):
I agree with the changes just I think we should wait for massa-sc-runtime PR to be merged and a new tag on this so that we avoid depending on a branch here.

AurelienFT (approved):

AurelienFT:
bors merge

Issue #2553, opened 2022-04-20
Analysis of RAM usage


sebastien-forestier:
Here is a tentative analysis of what's taking so much RAM after a few days running the node, with abrupts increase and decreases in RAM.
Usually, the RAM usage is around 2GB, with small and big "random" increments up to 8GB and sometimes but not always, decrements back to ~2GB.

Here is the last 2 days on my node, gray curve is the RAM usage:
![ram](https://user-images.githubusercontent.com/11048486/164190069-ce5562fc-6b86-48f8-9444-84957acdf794.png)

During this time I launched [heaptrack](https://github.com/KDE/heaptrack) with
```
heaptrack --pid $(pidof massa-node)
heaptrack_print heaptrack.massa-node.3640803.gz --flamegraph-cost-type peak -F peak.txt
./flamegraph.pl --title "heaptrack: peak" --colors mem --countname allocations < peak.txt > heaptrack.peak.svg
```
`./flamegraph.pl` comes from [here](https://github.com/brendangregg/FlameGraph), but you can also use `heaptrack_gui` instead of print+flamegraph.

Here is the result:
![heaptrack peak](https://user-images.githubusercontent.com/11048486/164191162-8c03a68e-986a-4c7d-91d0-279ebcee40f4.svg)
You can open it with `firefox heaptrack.peak.svg` to be able to zoom etc.

If I understand well, it shows the memory usage at the peak of 4.5GB.
So it seems too much RAM is allocated in https://github.com/massalabs/massa/blob/main/massa-protocol-worker/src/node_info.rs#L133

AurelienFT:
@gterzian is working a lot on improving the operations system. I think his thought on that would be nice to hear !

damip:
Note: currently we have `max_known_ops_size = 10024`

gterzian:
The function at the top of the graph corresponds to [`reserve_rehash`](https://github.com/rust-lang/hashbrown/blob/21bfd9aff79ebcfec29143a14b18d8c7bb2bac5c/src/raw/mod.rs#L676) in the std. 

In the current code, once `max_known_ops_size` has been reached, the new operation is first inserted, and then the oldest one removed. Since at that point the capacity of the set has been reached, it will re-allocate. 

The fix I propose in https://github.com/massalabs/massa/pull/2556 is to do the removal before the insertion, keeping capacity constant and therefore avoiding allocations.

qdrn:
[...]

PR #2556, opened 2022-04-21
Protocol: prevent allocations when insert known ops

gterzian:
Fix https://github.com/massalabs/massa/issues/2553

### DIFF
--- a/massa-protocol-worker/src/node_info.rs
+++ b/massa-protocol-worker/src/node_info.rs
@@ -132,14 +132,23 @@ impl NodeInfo {
 
     pub fn insert_known_ops(&mut self, ops: Set<OperationId>, max_ops_nb: usize) {
         for operation_id in ops.into_iter() {
-            if self.known_operations.insert(operation_id) {
-                self.known_operations_queue.push_front(operation_id);
-                if self.known_operations_queue.len() > max_ops_nb {
-                    if let Some(op_id) = self.known_operations_queue.pop_back() {
-                        self.known_operations.remove(&op_id);
-                    }
-                }
+            if self.known_operations.contains(&operation_id) {
+                // Ignore operations the node already knowns about.
+                continue;
             }
+
+            // If capacity has been reached,
+            // remove the last known op before inserting the new one,
+            // to prevent memory allocations.
+            if self.known_operations.len() == max_ops_nb {
+                let to_remove = self
+                    .known_operations_queue
+                    .pop_back()
+                    .expect("Empty known operations queue, when known operations is full.");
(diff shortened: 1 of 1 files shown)


damip (dismissed):

damip:
Only a couple downsides:

* I don't know why the initial approach created such a problem (?)
* maybe we should do a preallocation (with_capacity) for both the hashmap and the deque to avoid allows when resizing
* we should reduce the known_ops_size value in the config to something like 2048
* the new algorith does 2 hashmap lookups instead of 1 ("contains" then "remove")

damip:
Here is a better option I thinnk: https://github.com/massalabs/massa/pull/2558

gterzian:
@damip 

> I don't know why the initial approach created such a problem

The initial problem was that when having reached the limit, we would first do an `insert`, and then a `remove`. Since the `insert` happens when the set is already at full capacity, it will re-allocate, and I think it will basically double it's capacity. 

> we should reduce the known_ops_size value in the config to something like 2048

Why? Also, in that case it still makes sense to avoid doubling the capacity just to insert one additional element.

> the new algorithm does 2 hashmap lookups instead of 1 ("contains" then "remove")

Yes, and the current problem is memory usage, not cpu. 

> Here is a better option I thinnk: https://github.com/massalabs/massa/pull/2558

It does the trick, although I think it is too convoluted(why add 1 to a config? That would surprise anyone looking at the code), and that complication is not worth saving one instruction(which may not be in fact a performance problem at all).

damip:
> @damip
> 
> > I don't know why the initial approach created such a problem
> 
> The initial problem was that when having reached the limit, we would first do an `insert`, and then a `remove`. Since the `insert` happens when the set is already at full capacity, it will re-allocate, and I think it will basically double it's capacity.
> 
> > we should reduce the known_ops_size value in the config to something like 2048
> 
> Why? Also, in that case it still makes sense to avoid doubling the capacity just to insert one additional element.
> 
> > the new algorithm does 2 hashmap lookups instead of 1 ("contains" then "remove")
> 
> Yes, and although the current problem is memory usage, not cpu, it is still better to avoid. This can be done as in the other PR.
> 
> > Here is a better option I thinnk: #2558
> 
> If the review at [#2558 (review)](https://github.com/massalabs/massa/pull/2558#pullrequestreview-949532415) is addressed, then indeed it is better (since it removes the additional `contains` check).
> 
> It would have been better to progress via a review here.

I replied to the comments. Yeah sorry we had to rush this one through because of the time constraints. We'll be back to a normal way of working after the hackathon

PR #2558, opened 2022-04-21
increase pipes size and reduce max known ops

damip:
increase channel sizes and reduce max known ops

### DIFF
--- a/massa-protocol-worker/src/node_info.rs
+++ b/massa-protocol-worker/src/node_info.rs
@@ -53,10 +53,12 @@ impl NodeInfo {
             asked_blocks: Default::default(),
             connection_instant: Instant::now(),
             known_operations: Set::<OperationId>::with_capacity_and_hasher(
-                pool_settings.max_known_ops_size,
+                pool_settings.max_known_ops_size.saturating_add(1),
                 BuildMap::default(),
             ),
-            known_operations_queue: VecDeque::with_capacity(pool_settings.max_known_ops_size),
+            known_operations_queue: VecDeque::with_capacity(
+                pool_settings.max_known_ops_size.saturating_add(1),
+            ),
             known_endorsements: Set::<EndorsementId>::with_capacity_and_hasher(
                 pool_settings.max_known_endorsements_size,
                 BuildMap::default(),
@@ -133,9 +135,9 @@ impl NodeInfo {
     pub fn insert_known_ops(&mut self, ops: Set<OperationId>, max_ops_nb: usize) {
         for operation_id in ops.into_iter() {
             if self.known_operations.insert(operation_id) {
-                self.known_operations_queue.push_front(operation_id);
-                if self.known_operations_queue.len() > max_ops_nb {
-                    if let Some(op_id) = self.known_operations_queue.pop_back() {
+                self.known_operations_queue.push_back(operation_id);
+                while self.known_operations_queue.len() > max_ops_nb {
+                    if let Some(op_id) = self.known_operations_queue.pop_front() {
                         self.known_operations.remove(&op_id);
                     }
                 }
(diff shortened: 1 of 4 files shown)


AurelienFT (approved):

massa-protocol-worker/src/node_info.rs
```diff
@@ -133,9 +135,9 @@ impl NodeInfo {
     pub fn insert_known_ops(&mut self, ops: Set<OperationId>, max_ops_nb: usize) {
         for operation_id in ops.into_iter() {
             if self.known_operations.insert(operation_id) {
-                self.known_operations_queue.push_front(operation_id);
-                if self.known_operations_queue.len() > max_ops_nb {
-                    if let Some(op_id) = self.known_operations_queue.pop_back() {
+                self.known_operations_queue.push_back(operation_id);
+                while self.known_operations_queue.len() > max_ops_nb {
```
gterzian, line 139: Why the `while` loop? You could just just a simple `if self.known_operations_queue.len() > max_ops_nb`.

massa-protocol-worker/src/node_info.rs
```diff
@@ -133,9 +135,9 @@ impl NodeInfo {
...
+                    if let Some(op_id) = self.known_operations_queue.pop_front() {
```
gterzian, line 140: So here you are removing the element that just was inserted. Why the change in logic(versus the `pop_back` used earlier)?

massa-protocol-worker/src/node_info.rs
```diff
@@ -53,10 +53,12 @@ impl NodeInfo {
             asked_blocks: Default::default(),
             connection_instant: Instant::now(),
             known_operations: Set::<OperationId>::with_capacity_and_hasher(
-                pool_settings.max_known_ops_size,
+                pool_settings.max_known_ops_size.saturating_add(1),
```
gterzian, line 56: I find this surprising, it would be better to keep the setting value as it is, and change the algorithm using the value. 

For example, instead of `while self.known_operations_queue.len() > max_ops_nb`, simply do `if self.known_operations_queue.len() == max_ops_nb`

gterzian (commented):
Since this relates to https://github.com/massalabs/massa/pull/2556, I would have preferred to keep it open while the discussion is still ongoing.

massa-protocol-worker/src/node_info.rs
damip: indeed no loop should be required, and this doesn't add any overhead. I just sends the following message to the developer: "after this loop you know that there is no excess element anymore". The behavior is exactly the same.

massa-protocol-worker/src/node_info.rs
damip: The behavior is the same (push_back then pop_front instead of push_front then pop_back).
It is just more canonical with the usual way of doing things with Vec (most recent element / top of the stack to the back).

massa-protocol-worker/src/node_info.rs
damip: true, but changing to `if self.known_operations_queue.len() == max_ops_nb` would actually not solve the problem (the behavior is exactly the same) and would also require adding 1 to the capacity to ensure a no-realloc policy.

The best course in my opinion would be to add a comment to explain. Might do this in a followup

gterzian: > and would also require adding 1 to the capacity to ensure a no-realloc policy

How is that? The re-allocation happens not when reaching the capacity, but when exceeding it. So if after each insert you check that capacity has been reached, and if so, remove one, the set will not re-allocate(without needing to add 1).

massa-protocol-worker/src/node_info.rs
gterzian: I don't see how this sends a clearer message to the developer then a simple if, and I find it actually confusing since a loop seems to imply some kind of iteration(of more than one element), which never actually takes place. 

The reason why a loop is used for example when using a condvar, is to allow for spurious wake-ups, which is not an issue here.

massa-protocol-worker/src/node_info.rs
gterzian: It appears to me the behavior is different, in terms of removing the oldest or newest element when capacity has been reached. I think you could write a test for the initial behavior, which would have failed after the change.

PR #2573, opened 2022-04-25
Move endorsements from block header to block.

AurelienFT:
Endorsement are now managed like operations and have a merkle root in the header.
The content is in the body of the block.


AurelienFT:
I know it's not working for now as when I test with the simulator I get : 
```
 node 5tuXKh377eV5DNp8xzLo5dhJJVZoAmRuP9CGnKAhGp3xFrhuMv sent us critically incorrect block, which may be an attack attempt by the remote node or a loss of sync between us and the remote node
```

AurelienFT:
We don't want end up doing this because it will not allow us to make the double staking protection. We can instead compress the endorsements to only a signature and public key and when PoS will draw public key instead of addresses we will be able to reduce it to only a signature.

PR #2576, opened 2022-04-26
Some tests for using nom

### LINKED ISSUE #1200: Improve consistency of checks in (de-)serialization  of objects
In GitLab by @gterzian

Often, the checks performed in `SerializeCompact` and `DeserializeCompact` are not symmetric. 

For example in [`Slot`](https://gitlab.com/massalabs/massa/-/blob/fd6d2b98339f243ef3733d8b15a448ce5cd18985/models/src/slot.rs#L184) we check the thread against the parent count in `DeserializeCompact`, but no such check is performed in `SerializeCompact`. 

This seems to assume that we trust that objects that a node serializes are valid, whereas we don't trust serialized objects received over the network.

In some ways this makes sense and it would be more consistent to apply checks symmetrically.

AurelienFT:
Fix #1200

### DIFF
--- a/massa-models/src/endorsement.rs
+++ b/massa-models/src/endorsement.rs
@@ -10,6 +10,7 @@ use crate::{
     with_serialization_context, BlockId, ModelsError, Slot,
 };
 use massa_hash::Hash;
+use massa_signature::{sign, Signature};
 use massa_signature::{PublicKey, PUBLIC_KEY_SIZE_BYTES};
 use serde::{Deserialize, Serialize};
 use std::{fmt::Display, str::FromStr};
--- a/massa-models/src/endorsement.rs
+++ b/massa-models/src/endorsement.rs
@@ -2,7 +2,7 @@
 
 use crate::constants::{BLOCK_ID_SIZE_BYTES, ENDORSEMENT_ID_SIZE_BYTES};
 use crate::prehash::PreHashed;
-use crate::serialization::DeserializeCompactV2;
+use crate::serialization::{DeserializeCompactV2, DeserializeVarIntV2};
 use crate::signed::{Id, Signable, Signed};
 use crate::{
     serialization::{
@@ -206,14 +206,20 @@ impl DeserializeCompactV2 for Endorsement {
         let max_block_endorsements =
             with_serialization_context(|context| context.endorsement_count);
         tuple((
-            // context("sender_public_key", PublicKey::from_bytes),
+            context("sender_public_key", PublicKey::from_bytes_compact_v2),
             context("slot", Slot::from_bytes_compact_v2),
+            // use varint bounded here
+            context("index", u32::from_varint_bytes_v2),
+            context("endorsed_block", BlockId::from_bytes_compact_v2),
         ))(buffer)
-        .map(|(rest, slot)| {
+        .map(|(rest, (sender_public_key, slot, index, endorsed_block))| {
             (
                 rest,
                 Endorsement {
-                    ..Default::default()
+                    sender_public_key,
+                    slot,
+                    index,
+                    endorsed_block,
                 },
             )
         })
--- a/massa-models/src/endorsement.rs
+++ b/massa-models/src/endorsement.rs
@@ -212,11 +212,11 @@ impl DeserializeCompactV2 for Endorsement {
         tuple((
             context("sender_public_key", PublicKey::from_bytes_compact_v2),
             context("slot", Slot::from_bytes_compact_v2),
-            |buffer: &'a [u8]| {
+            context("index", |buffer: &'a [u8]| {
                 let max_block_endorsements =
                     with_serialization_context(|context| context.endorsement_count);
                 u32::from_varint_bytes_v2(buffer, None, Some(max_block_endorsements))
-            },
+            }),
             context("endorsed_block", BlockId::from_bytes_compact_v2),
         ))(buffer)
         .map(|(rest, (sender_public_key, slot, index, endorsed_block))| {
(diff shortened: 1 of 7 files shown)


Eitu33:
@AurelienFT @damip 

Result of the benchmark:
```
test endorsement::tests::deserialization    ... bench:       4,214 ns/iter (+/- 38)
test endorsement::tests::deserialization_v2 ... bench:       4,216 ns/iter (+/- 23)
```

The implementation in the PR changes, not everything is there yet but I think it's enough to see what it'll look like. It would be nice if you could check it and tell me what you think.

massa-models/src/serialization.rs
```diff
@@ -87,6 +98,72 @@ impl DeserializeVarInt for u32 {
...
+    fn from_varint_bytes_v2(
+        buffer: &[u8],
+        min: Option<Self>,
+        max: Option<Self>,
+    ) -> IResult<&[u8], Self> {
+        let error = Err(nom::Err::Failure(nom::error::Error::new(
+            buffer,
+            nom::error::ErrorKind::Fail,
+        )));
+        match unsigned_varint::decode::u32(buffer) {
```
AurelienFT, line 112: I think you can simplify and bring clarity by writing : 
```rust
let min = min.unwrap_or(u32::MIN);
let min = max.unwrap_or(u32::MAX);

match unsigned_varint::decode::u32(buffer) {
  Ok((v, rest)) if v >= min && v < max => Ok((rest, v))
  _ => error
}
```

I don't know if it change something in optimisation time/CPU/memory etc but it gain clarity.

Also I don't understand why `min` is Included and `max` excluded. I think both should be the same and if you want to bring some flexibility I think you could use something like : https://doc.rust-lang.org/stable/std/ops/enum.Bound.html 

Same for all above.

massa-models/src/endorsement.rs
```diff
@@ -197,6 +200,43 @@ impl DeserializeCompact for Endorsement {
...
+// HERE
+// pub struct ShortEndorsement {
+//     pub sender_pub_key: PublicKey,
+//     pub signature: Signature,
+// }
+
+// HERE
+impl DeserializeCompactV2 for Endorsement {
+    fn from_bytes_compact_v2<'a>(buffer: &'a [u8]) -> IResult<&[u8], Self> {
+        tuple((
```
AurelienFT, line 212: I think we could write a function like macro for this tuple + map.

AurelienFT (commented):
Overall I think the error management could be done with our type `ModelsError` everywhere and have an `into::<nom::Err>` if needed because I don't think the error proposed by nom will match our error cases.

massa-models/src/slot.rs
```diff
@@ -153,6 +153,68 @@ impl Slot {
...
+                )))
+            }
+        }
+        Err(e) => Err(e),
+    }
+}
+
+// HERE
+impl DeserializeCompactV2 for Slot {
+    fn from_bytes_compact_v2<'a>(buffer: &'a [u8]) -> IResult<&[u8], Self> {
```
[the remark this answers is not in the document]
Eitu33: Yes I agree it has to be done just didn't make it in the first iteration.

massa-models/src/slot.rs
```diff
@@ -153,6 +153,68 @@ impl Slot {
...
+        tuple((
+            context("period", |buffer: &'a [u8]| {
```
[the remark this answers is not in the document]
Eitu33: Is that a question or a request?

If what you are saying is that `u64::from_varint_bytes_v2` should return a `FnOnce`, yes I agree it would be better.

massa-models/src/serialization.rs
Eitu33: I feel like min included and max excluded is the most common behaviour that's why I went for that, but maybe its better to have something different.

Also I believe my implementation is faster, too many operation are made for a `None` value with yours.

massa-models/src/slot.rs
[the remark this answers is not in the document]
AurelienFT: It's a request. I think we should be able to easily write the deserialization method and all fields should have the same format to deserialize and the same as nom.

massa-models/src/endorsement.rs
Eitu33: Yeah we could but we need to have a meeting about it. I think it's very important to be clear about what it allows and what it doesn't and for what reasons.

massa-models/src/serialization.rs
AurelienFT: Yes it's not common for me that the max is excluded.

Yes I think it's faster but I don't know if the gain here worth the reading. It's a choice. Not a problem for me to keep as you have made.

massa-models/src/endorsement.rs
```diff
@@ -197,6 +200,43 @@ impl DeserializeCompact for Endorsement {
...
 }
 
+// HERE
+// pub struct ShortEndorsement {
+//     pub sender_pub_key: PublicKey,
+//     pub signature: Signature,
+// }
+
+// HERE
+impl DeserializeCompactV2 for Endorsement {
```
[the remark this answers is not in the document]
Eitu33: True, will change.

massa-models/src/slot.rs
```diff
@@ -6,8 +6,8 @@ use super::{
     },
     with_serialization_context,
 };
-use crate::constants::SLOT_KEY_SIZE;
-use crate::error::ModelsError;
+use crate::{constants::SLOT_KEY_SIZE, node_configuration::THREAD_COUNT};
```
damip, line 9: does this interfere with the tests when we use a different number of threads for a specific test ?

damip (commented):
Looks good !

Needs a bit more documentation on the functions, and some minor improvements, but overall it's really nice.

massa-models/src/slot.rs
Eitu33: I have no idea, it needs to be verified.

PR #2577, opened 2022-04-27
Streaming bootstrap

AurelienFT:
The goal of this PR is to implement a streaming process for the bootstrap. The explanations and specs are here : https://github.com/adrien-zinger/presentations/blob/main/streaming_bootstrap/index.md  . We will go for the polling method so that we stray stateless and the deconnexions will not be a problem.

## Sub PR: 

- [x] Merge #2578
- [x] Merge #2625 

## Currently integrated : 

- [x] Message to pass ledger subset of consensus and execution in the network
- [x] Get a part of the ledger on disk by starting address and size. (could be improved).

## Need to be done

- [x] Send and receive the actual messages
- [x] Add ledgers changes pinned to a Slot in the messages
- [x] [~~Get ledger changes with cursor.~~](https://github.com/massalabs/massa/issues/2639)
- [x] Add parts of message async pool to the message
- [x] Add async pools changes to the message
- [x] Remove ledger from old bootstrap structures
- [x] Make transition with asking the final blocks
- [x] Recognize when client sent a too old slot and we need to restart bootstrap.
- [x] Refactor the code ( ledger subset could be generic)
- [x] Remove the `FinalStateBootstrap`
- [x] Document
- [x] Tests

## How to test the PR : 

You can run two sandbox nodes on your computer with two differents config and pass the same genesis timestamp using the env variable.

### Node 1:
Command : 
```
GENESIS_TIMESTAMP=1651051538000 MASSA_CONFIG_PATH=base_config/config.toml  ../target/release/massa-node
```
Config: 
[config.txt](https://github.com/massalabs/massa/files/8572092/config.txt) (Rename to .toml

### Node 2 (to launch after genesis):
Command : 
```
GENESIS_TIMESTAMP=1651051538000 MASSA_CONFIG_PATH=base_config/config2.toml  ../target/release/massa-node
```
Config: 
[config2.txt](https://github.com/massalabs/massa/files/8572095/config2.txt) (Rename to .toml)

PR #2578, opened 2022-04-28
Add send on client binders and receive on server binders of bootstrap

AurelienFT:
Part of #2577 

The goal of this PR is to be able to send data through the client and receive in the server.

The client do not have a private key so he can't sign his messages but he need to send the signature of the previous message he received.

The server need to make a signature of the message he received and save it to put it has the previous signature of his new message.

Multiple tests has been added.

### DIFF
--- a/massa-bootstrap/src/client_binder.rs
+++ b/massa-bootstrap/src/client_binder.rs
@@ -71,8 +71,6 @@ impl BootstrapClientBinder {
             Signature::from_bytes(&sig_bytes)?
         };
 
-        self.prev_message = Some(Hash::compute_from(&sig.to_bytes()));
-
         // read message length
         let msg_len = {
             let mut meg_len_bytes = vec![0u8; self.size_field_len];
@@ -83,6 +81,7 @@ impl BootstrapClientBinder {
         // read message, check signature and check signature of the message sent just before then deserialize it
         let message = {
             if let Some(prev_message) = self.prev_message {
+                self.prev_message = Some(Hash::compute_from(&sig.to_bytes()));
                 let mut sig_msg_bytes = vec![0u8; HASH_SIZE_BYTES + (msg_len as usize)];
                 sig_msg_bytes[..HASH_SIZE_BYTES].copy_from_slice(&prev_message.to_bytes());
                 self.duplex
@@ -94,6 +93,7 @@ impl BootstrapClientBinder {
                     BootstrapMessage::from_bytes_compact(&sig_msg_bytes[HASH_SIZE_BYTES..])?;
                 msg
             } else {
+                self.prev_message = Some(Hash::compute_from(&sig.to_bytes()));
                 let mut sig_msg_bytes = vec![0u8; msg_len as usize];
                 self.duplex.read_exact(&mut sig_msg_bytes[..]).await?;
                 let msg_hash = Hash::compute_from(&sig_msg_bytes);
--- a/massa-bootstrap/src/client_binder.rs
+++ b/massa-bootstrap/src/client_binder.rs
@@ -80,17 +80,26 @@ impl BootstrapClientBinder {
 
         // read message, check signature and optionally check signature of the message sent just before then deserialize it
         let message = {
-            let prev_message = self.prev_message.unwrap();
-            let mut sig_msg_bytes = vec![0u8; HASH_SIZE_BYTES + (msg_len as usize)];
-            sig_msg_bytes[..HASH_SIZE_BYTES].copy_from_slice(&prev_message.to_bytes());
-            self.duplex
-                .read_exact(&mut sig_msg_bytes[HASH_SIZE_BYTES..])
-                .await?;
-            let msg_hash = Hash::compute_from(&sig_msg_bytes);
-            verify_signature(&msg_hash, &sig, &self.remote_pubkey)?;
-            let (msg, _len) =
-                BootstrapMessage::from_bytes_compact(&sig_msg_bytes[HASH_SIZE_BYTES..])?;
-            msg
+            if let Some(prev_message) = self.prev_message {
+                let mut sig_msg_bytes = vec![0u8; HASH_SIZE_BYTES + (msg_len as usize)];
+                sig_msg_bytes[..HASH_SIZE_BYTES].copy_from_slice(&prev_message.to_bytes());
+                self.duplex
+                    .read_exact(&mut sig_msg_bytes[HASH_SIZE_BYTES..])
+                    .await?;
+                let msg_hash = Hash::compute_from(&sig_msg_bytes);
+                verify_signature(&msg_hash, &sig, &self.remote_pubkey)?;
+                let (msg, _len) =
+                    BootstrapMessage::from_bytes_compact(&sig_msg_bytes[HASH_SIZE_BYTES..])?;
+                msg
+            } else {
+                let mut sig_msg_bytes = vec![0u8; msg_len as usize];
+                self.duplex.read_exact(&mut sig_msg_bytes[..]).await?;
+                let msg_hash = Hash::compute_from(&sig_msg_bytes);
+                verify_signature(&msg_hash, &sig, &self.remote_pubkey)?;
+                let (msg, _len) =
+                    BootstrapMessage::from_bytes_compact(&sig_msg_bytes[HASH_SIZE_BYTES..])?;
+                msg
+            }
         };
 
         // save prev sig
--- a/massa-bootstrap/src/client_binder.rs
+++ b/massa-bootstrap/src/client_binder.rs
@@ -13,6 +13,7 @@ use massa_signature::{verify_signature, PublicKey, Signature, SIGNATURE_SIZE_BYT
 use rand::{rngs::StdRng, RngCore, SeedableRng};
 use tokio::io::AsyncReadExt;
 use tokio::io::AsyncWriteExt;
+use tracing::debug;
 
 /// Bootstrap client binder
 pub struct BootstrapClientBinder {
@@ -101,6 +102,8 @@ impl BootstrapClientBinder {
             }
         };
 
+        debug!("Received = {:#?}", message);
+
         // save prev sig
         self.prev_message = Some(Hash::compute_from(&sig.to_bytes()));
 
@@ -114,6 +117,8 @@ impl BootstrapClientBinder {
         let msg_len: u32 = msg_bytes.len().try_into().map_err(|e| {
             BootstrapError::GeneralError(format!("bootstrap message too large to encode: {}", e))
         })?;
+
+        debug!("Sent = {:#?}", msg);
 
         if let Some(prev_message) = self.prev_message {
             self.duplex.write_all(&prev_message.to_bytes()).await?;
--- a/massa-bootstrap/src/client_binder.rs
+++ b/massa-bootstrap/src/client_binder.rs
@@ -96,7 +96,7 @@ impl BootstrapClientBinder {
                     self.duplex.read_exact(&mut sig_bytes).await?;
                     Signature::from_bytes(&sig_bytes)?
                 };
-                // Check if old signature match
+                // Check if old signature matches
                 {
                     let old_message = &sended_message.to_bytes_compact()?;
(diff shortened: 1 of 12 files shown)


massa-bootstrap/src/client_binder.rs
```diff
@@ -84,11 +87,36 @@ impl BootstrapClientBinder {
             u32::from_be_bytes_min(&meg_len_bytes, self.max_bootstrap_message_size)?.0
         };
 
-        // read message, check signature and deserialize
+        // read message, check signature and, optionally, signature of the message we sent just before and deserialize
```
Eitu33, line 90: This sentence doesn't make much sense to me, did you mean to say "and optionally check signature of the message sent just before then deserialize it"?

massa-bootstrap/src/client_binder.rs
```diff
@@ -84,11 +87,36 @@ impl BootstrapClientBinder {
...
             let mut sig_msg_bytes = vec![0u8; SIGNATURE_SIZE_BYTES + (msg_len as usize)];
-            sig_msg_bytes[..SIGNATURE_SIZE_BYTES]
-                .clone_from_slice(&self.prev_sig.unwrap().to_bytes());
+            if let Some(sended_message) = &self.sended_message {
+                let old_message_sig_from_server = {
+                    let mut sig_bytes = [0u8; SIGNATURE_SIZE_BYTES];
+                    self.duplex.read_exact(&mut sig_bytes).await?;
+                    Signature::from_bytes(&sig_bytes)?
+                };
+                // Check if old signature match
```
Eitu33, line 99:
```suggestion
                // Check if old signature matches
```

massa-bootstrap/src/client_binder.rs
```diff
@@ -20,6 +21,7 @@ pub struct BootstrapClientBinder {
     remote_pubkey: PublicKey,
     duplex: Duplex,
     prev_sig: Option<Signature>,
+    sended_message: Option<BootstrapMessage>,
```
Eitu33, line 24:
```suggestion
   sent_message: Option<BootstrapMessage>,
```

massa-bootstrap/src/client_binder.rs
```diff
@@ -84,11 +87,36 @@ impl BootstrapClientBinder {
...
+                    old_sig_msg_bytes[..SIGNATURE_SIZE_BYTES]
+                        .clone_from_slice(&self.prev_sig.unwrap().to_bytes());
+                    old_sig_msg_bytes[SIGNATURE_SIZE_BYTES..].clone_from_slice(&old_message);
+                    let old_msg_hash = Hash::compute_from(&old_sig_msg_bytes);
+                    verify_signature(
+                        &old_msg_hash,
+                        &old_message_sig_from_server,
+                        &self.remote_pubkey,
+                    )?;
+                };
```
Eitu33, line 113: I think this scope useless

massa-bootstrap/src/client_binder.rs
```diff
@@ -101,7 +129,30 @@ impl BootstrapClientBinder {
...
+            BootstrapError::GeneralError(format!("bootstrap message too large to encode: {}", e))
+        })?;
+        if let Some(prev_sig) = self.prev_sig {
+            self.duplex.write_all(&prev_sig.to_bytes()).await?;
+        }
+        // send message length
+        {
+            let msg_len_bytes = msg_len.to_be_bytes_min(self.max_bootstrap_message_size)?;
+            self.duplex.write_all(&msg_len_bytes).await?;
+        }
```
Eitu33, line 151: Why make a scope here? No need for a change I just don't understand the purpose.

massa-bootstrap/src/client_binder.rs
```diff
@@ -101,7 +129,30 @@ impl BootstrapClientBinder {
...
         // save prev sig
         self.prev_sig = Some(sig);
+        self.sended_message = None;
 
         Ok(message)
     }
+
+    #[allow(dead_code)]
+    /// Send a message to the bootstrap server
+    pub async fn send(&mut self, msg: BootstrapMessage) -> Result<(), BootstrapError> {
```
Eitu33, line 139: You're using `BootstrapMessage` as input for the client `send`, but when I read the enum like it is made for the server only. Should the client not have a different type for its messages given he won't send the same info as the server or am I missing something? AFAIK client needs to send info such as `StartBootstrap` `ContinueFrom` etc.

massa-bootstrap/src/server_binder.rs
```diff
@@ -95,4 +106,49 @@ impl BootstrapServerBinder {
...
+        let sig = {
+            let mut sig_bytes = [0u8; SIGNATURE_SIZE_BYTES];
+            self.duplex.read_exact(&mut sig_bytes).await?;
+            Signature::from_bytes(&sig_bytes)?
+        };
+
+        if let Some(prev_sig) = self.prev_sig {
+            if sig != prev_sig || self.received_client_message_sig.is_some() {
+                return Err(BootstrapError::GeneralError(
+                    "The prev signature sent by the client doesn't match our.".to_string(),
```
Eitu33, line 123:
```suggestion
                    "The prev signature sent by the client doesn't match ours.".to_string(),
```

massa-bootstrap/src/client_binder.rs
AurelienFT: Fixed.

massa-bootstrap/src/client_binder.rs
AurelienFT: Yes. My bad. Fixed.

massa-bootstrap/src/client_binder.rs
AurelienFT: I just copied the behavior from the function to send on the server. I believe it's for split parts that are sent in the buffer and avoid a misusing of the variables after.

massa-bootstrap/src/client_binder.rs
AurelienFT: I made it to gain clarity and avoid using the variable in the rest of the function in case someone else develop something new because it can be confusing with the similar names etc.. I think it don't change anything when compiling and IMO bring clarity

massa-bootstrap/src/client_binder.rs
AurelienFT: For the moment it's server only but with #2577 it will be client and server. Splitting could be possible but in my mind it loose the """""genericity""""" (I don't know why it's like a feeling) and in the `Message` of the Network all are in the same enum two.

I don't really see a really good reason to keep it group or split it so I have follow what is already done.

Moreover I think we could change it in the other PR because if I split in this one the Client enum will have no variant and it will be a pain to write the tests then.

Eitu33: Ok, IMHO it makes more sens to split the messages.

AurelienFT: Let's see it in #2577

damip:
Nice approach. However the client messages can be tampered with.

I would recommend this:

* when the client connects, it sends [client version + 32 random bytes]
* when receiving that handshake, the server responds with [server version + server timestamp + sign(this message + hash of previous client message)] 

* when the client sends a message, it appends at the end of it:
  * the hash of the signature of the previous message if it was server -> client
  * the hash of the previous message if it was client -> server

* when the server sends a message, it appends at the end of it the signature of [the message appended to 
  * the hash of the previous message if it was client -> server]
  * the signature of the previous message if it was server -> client]

* whenever there is an error, we (client or server) send the error explanation (with a short timeout) before disconnecting to help diagnose bootstrap problems on the other side

AurelienFT:
I pass this to draft because I will add a better error management on bootstrap problems (no available slots etc)

AurelienFT:
I have tested the errors on the simulator.

massa-bootstrap/src/client_binder.rs
```diff
@@ -36,34 +37,27 @@ impl BootstrapClientBinder {
...
-            self.duplex.write_all(&random_bytes).await?;
-            let rand_hash = Hash::compute_from(&random_bytes);
+            let version = version.to_bytes_compact()?;
+            let mut version_random_bytes =
+                vec![0u8; (version.len() as usize) + BOOTSTRAP_RANDOMNESS_SIZE_BYTES];
+            version_random_bytes[..version.len()].clone_from_slice(&version);
+            StdRng::from_entropy().fill_bytes(&mut version_random_bytes[version.len()..]);
+            self.duplex.write_all(&version_random_bytes).await?;
+            let rand_hash = Hash::compute_from(&version_random_bytes);
             self.duplex.write_all(&rand_hash.to_bytes()).await?;
```
damip, line 56: this is not necessary

massa-bootstrap/src/server_binder.rs
```diff
@@ -31,17 +34,25 @@ impl BootstrapServerBinder {
...
+                vec![0u8; (version_bytes.len() as usize) + BOOTSTRAP_RANDOMNESS_SIZE_BYTES];
             self.duplex.read_exact(&mut random_bytes).await?;
+            let receive_version =
+                Version::from_bytes_compact(&random_bytes[..version_bytes.len()])?;
+            if !receive_version.0.is_compatible(&version) {
+                return Err(BootstrapError::IncompatibleVersionError(format!("Received a bad incompatible version in handshake. (excepted: {}, received: {})", version, receive_version.0)));
+            }
             let expected_hash = Hash::compute_from(&random_bytes);
             let mut hash_bytes = [0u8; HASH_SIZE_BYTES];
             self.duplex.read_exact(&mut hash_bytes).await?;
```
damip, line 58: reading + verifying the hash is not necessary (brings nothing to security)

massa-bootstrap/src/client_binder.rs
```diff
@@ -84,24 +78,55 @@ impl BootstrapClientBinder {
...
+
+        // send message length
+        {
+            let msg_len_bytes = msg_len.to_be_bytes_min(self.max_bootstrap_message_size)?;
+            self.duplex.write_all(&msg_len_bytes).await?;
+        }
+
+        // send message
+        self.duplex.write_all(&msg_bytes).await?;
+        self.prev_message = Some(Hash::compute_from(&msg.to_bytes_compact()?));
```
damip, line 129: prev_message should be the hash of [prev_hash (if any) + msg]

massa-bootstrap/src/messages.rs
```diff
@@ -123,6 +131,17 @@ impl DeserializeCompact for BootstrapMessage {
 
                 BootstrapMessage::FinalState { final_state }
             }
+            MessageTypeId::BootstrapError => {
+                let (error_len, delta) = u32::from_varint_bytes(&buffer[cursor..])?;
+                cursor += delta;
+
+                let error = String::from_utf8_lossy(&buffer[cursor..cursor + error_len as usize]);
```
damip, line 138: Critical attack vector: this will panic if error_len is longer than the buffer. Use buffer.get(cursor..cursor + xxxx) that returns None if the range is unavailable

damip (changes requested):

massa-bootstrap/src/client_binder.rs
damip: corrected

massa-bootstrap/src/server_binder.rs
damip: corrected

massa-bootstrap/src/client_binder.rs
damip: corrected

massa-bootstrap/src/lib.rs
```diff
@@ -346,6 +381,8 @@ struct BootstrapServer {
 }
 
 impl BootstrapServer {
+    // We don't want to leave on a fail of sending error to a bootstrap client.
+    #[allow(unused_must_use)]
```
damip, line 385: remove this allow, use `let_ = XXX `when we want to explictly ignore an important value

damip (approved):
very good !

damip:
bors merge

AurelienFT:
bors merge

AurelienFT:
bors merge

Issue #2579, opened 2022-04-28
Unification of Ledgers


Eitu33:
## Context

Massa core currently contains two different Ledgers:

### FinalLedger

- Is in `massa-ledger` (good)
- Exclusively used in `massa-final-state` which is interacted with in `massa-execution-worker` (good)
- Has a corresponding SpeculativeLedger (good)
- Stored on RAM (bad)
- Only operations applied to it are SC operations (bad)

### Ledger

- Is in `massa-graph` (bad)
- Exclusively used in `massa-graph` (bad)
- Stored on Disk (good)
- All the operations except SC operations are applied to it (bad)

### Unification

If you look closely, you can see that these 2 ledger complete themselves, what isn't great about one, the other has.
The goal is therefore to unify them, which is specified in the TODO list below.

### What we currently use for disk storage

[Sled](https://docs.rs/sled/latest/sled/), a pure rust embedded database.

### What [NEAR](https://github.com/near) uses

The NEAR blockchain uses a similar stack to what we use. They also had to solve the issue of saving blockchain data on Disk using rust. The crate choice they've made for that matter is [rust-rocksdb](https://github.com/rust-rocksdb/rust-rocksdb), which is a rust binding crate for the well known [RocksDB](https://github.com/facebook/rocksdb).

### Sled vs RockDB

The only advantage that Sled has over RockDS is that it is pure rust, it is slower, has less features, and is not as maintained as RocksDB.

You can check a benchmark comparison here: https://github.com/tokahuke/sled-vs-rocksdb

### TODO

- [x] Decide if we wish to use RocksDB instead of Sled for Disk storage
- [x] Move `massa-graph` ledger to `massa-ledger`
- [x] ~Save a ledger per thread in `massa-ledger` to match `massa-graph` ledger (feels like it's the best solution but is it?)~
- [x] Keep the execution block finality and save only one ledger
- [x] Store and interact with `massa-ledger` on Disk
- [x] Update the speculative ledger corresponding to the new unified ledger
- [x] Define how we'll implement the ledger updates on rolls operations (ask PoS module for the infos?)

## Related discussions, issues and PR's

- [__Ledger unification RFC__](https://github.com/massalabs/massa/blob/ledger_unif_doc/docs/rfc_ledger_unif.md)
- https://github.com/massalabs/massa/issues/2644
- https://github.com/massalabs/massa/issues/2342
- https://github.com/massalabs/massa/pull/2590
- https://github.com/massalabs/massa/pull/2761
[...]

PR #2590, opened 2022-05-05
Save execution ledger on disk with RocksDB

Eitu33:
https://github.com/massalabs/massa/issues/2579


Eitu33:
@damip I believe this PR is ready to be merged, here is what I chose to _not_ implement in this one because I believe it should be done in separate PRs:

- Save the bytecode in the datastore: requires a bit of work to get it working, I find it to be out of scope here
- Allow the speculative ledger to lazily query any sub-entry: Did do that to query address-specific entries from the execution history, but implementing it in the speculative ledger requires more work and I also find it to be out of scope
- LedgerDB::delete function: this can be done in this PR, I just need a definition of how we want to implement it

massa-ledger/src/ledger.rs
```diff
@@ -106,9 +106,10 @@ impl FinalLedger {
     /// # Arguments
     /// * configuration: ledger configuration
     /// * state: bootstrap state
-    pub fn from_bootstrap_state(config: LedgerConfig, state: FinalLedgerBootstrapState) -> Self {
+    pub fn from_bootstrap_state(config: LedgerConfig, _state: FinalLedgerBootstrapState) -> Self {
+        // dummy implementation since bootstrap streaming is currently developed
         FinalLedger {
-            sorted_ledger: state.sorted_ledger,
+            sorted_ledger: LedgerDB::new(),
```
AurelienFT, line 112: I think we should make a temporary bootstrap state method because otherwise this PR will block launching new testnets and we shouldn't merge something that prevent launching testnets.

massa-ledger/src/ledger_db.rs
```diff
@@ -0,0 +1,187 @@
+// Copyright (c) 2022 MASSA LABS <info@massa.net>
+
+use massa_hash::Hash;
+use massa_models::Address;
+use rocksdb::{ColumnFamilyDescriptor, Options, WriteBatch, DB};
+
+use crate::{ledger_changes::LedgerEntryUpdate, LedgerEntry, SetOrDelete, SetOrKeep};
+
+const DB_PATH: &str = "../.db";
```
AurelienFT, line 9: The other DB for consensus is save in the storage folder of `massa-node`. I think we should group them.

massa-ledger/src/ledger_db.rs
```diff
@@ -0,0 +1,187 @@
...
+    }
+
+    pub fn update(&mut self, addr: &Address, entry_update: LedgerEntryUpdate) {
+        let mut batch = WriteBatch::default();
+        let key = addr.to_bytes();
+        let handle = self.0.cf_handle(LEDGER_CF).expect(CF_ERROR);
+
+        // balance
+        if let SetOrKeep::Set(balance) = entry_update.parallel_balance {
+            batch.put_cf(handle, key, balance.to_raw().to_be_bytes());
```
AurelienFT, line 94: Can we save varint instead ? To have a more light ledger.

massa-ledger/src/ledger_db.rs
```diff
@@ -0,0 +1,187 @@
...
+                SetOrDelete::Delete => batch.delete_cf(handle, data_key!(addr, hash)),
+            }
+        }
+
+        // write batch
+        self.0.write(batch).expect(CRUD_ERROR);
+    }
+
+    pub fn delete(&self, _addr: &Address) {
+        // TODO: define how we want to handle this first
```
AurelienFT, line 115: I think we can fetch all the data of an address using prefix.
Here is the docs of rocksdb about it : https://github.com/facebook/rocksdb/wiki/Iterator
And an issue of rust repo that show some examples : https://github.com/rust-rocksdb/rust-rocksdb/issues/577

massa-ledger/src/ledger_db.rs
Eitu33: Yeah I'm pretty sure we can.

massa-ledger/src/ledger.rs
```diff
@@ -150,9 +150,11 @@ impl FinalLedger {
         self.sorted_ledger
             .get_entry(addr, LedgerSubEntry::Balance)
             .map(|bytes| {
-                Amount::from_raw(u64::from_be_bytes(
-                    bytes.try_into().expect("critical: invalid balance format"),
-                ))
+                Amount::from_raw(
```
damip, line 153: Amount should have its own compact deserialization routine (so that from_raw is never used)

Eitu33: This particular one was not supposed to stay, what I'm doing now is: 
```rust
    pub fn get_parallel_balance(&self, addr: &Address) -> Option<Amount> {
        self.sorted_ledger
            .get_entry(addr, LedgerSubEntry::Balance)
            .map(|bytes| {
                Amount::from_raw(
                    u64::from_varint_bytes(&bytes)
                        .expect("critical: invalid balance format")
                        .0,
                )
            })
    }
```

Is that how it should be done?

massa-execution-worker/src/execution.rs
```diff
@@ -248,7 +242,75 @@ impl ExecutionState {
...
+    /// Lazily query (from end to beginning) the active balance of an address at a given slot.
+    /// Returns None if the address balance could not be determined from the active history.
+    pub fn get_active_balance_at_slot(&self, slot: Slot, addr: &Address) -> Option<Amount> {
+        self.verify_active_slot(slot);
+
+        let iter = self
+            .active_history
+            .iter()
+            .rev()
+            .skip_while(|output| output.slot >= slot);
```
damip, line 256: this has linear complexity

note that it can be implemented with constant complexity because active_history elements represent strictly consecutive slots in increasing order

massa-api/src/public.rs
```diff
@@ -575,66 +577,36 @@ impl Endpoints for API<Public> {
...
-                                    datastore: final_entry.datastore.into_iter().collect(),
-                                },
-                                SCELedgerInfo {
-                                    balance: candidate.parallel_balance,
-                                    module: candidate.bytecode,
-                                    datastore: candidate.datastore.into_iter().collect(),
-                                },
-                            ),
-                        };
+                    let (final_balance, candidate_balance) =
```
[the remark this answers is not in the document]
Eitu33: Actually, it has been done.

massa-execution-worker/src/execution.rs
```diff
@@ -248,7 +242,75 @@ impl ExecutionState {
...
             panic!("cannot execute at a slot beyond active cursor + 1");
         }
+    }
+
+    /// Lazily query (from end to beginning) the active balance of an address at a given slot.
+    /// Returns None if the address balance could not be determined from the active history.
+    pub fn get_active_balance_at_slot(&self, slot: Slot, addr: &Address) -> Option<Amount> {
+        self.verify_active_slot(slot);
+
+        let iter = self
```
damip, line 252: better performance can be achieved with

* finding highest history index for which `output.slot > slot` using constant complexity because we know the 1st slot of the history and we know that the history is composed of strictly consecutive slots
* iterating from that index (non-included) backwards down to 0

massa-execution-worker/src/execution.rs
```diff
@@ -248,7 +242,75 @@ impl ExecutionState {
...
+                Some(SetUpdateOrDelete::Delete) => return None,
+                _ => (),
+            }
+        }
+        None
+    }
+
+    /// Lazily query (from end to beginning) the active datastore entry of an address at a given slot.
+    /// Returns None if the datastore entry could not be determined from the active history.
+    pub fn get_active_data_entry_at_slot(
```
damip, line 274: same remark as above absent vs dont-know + lookup in the final ledger

damip, line 274: same remark as above absent vs dont-know + lookup in the final ledger

massa-execution-worker/src/execution.rs
```diff
@@ -248,7 +242,75 @@ impl ExecutionState {
         if slot > max_slot {
             panic!("cannot execute at a slot beyond active cursor + 1");
         }
+    }
+
+    /// Lazily query (from end to beginning) the active balance of an address at a given slot.
+    /// Returns None if the address balance could not be determined from the active history.
+    pub fn get_active_balance_at_slot(&self, slot: Slot, addr: &Address) -> Option<Amount> {
```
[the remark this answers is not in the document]
Eitu33: But this function is only used in [get_final_and_active_parallel_balance](https://github.com/massalabs/massa/blob/ledger-unification/massa-execution-worker/src/execution.rs#L835). Searching in the final ledger would mean doing twice the same operation. I think we should either search in the final ledger if `get_active_balance_at_slot` does not succeed and not return the final balance, or the other way around but not both.

massa-execution-worker/src/execution.rs
```diff
@@ -769,50 +831,33 @@ impl ExecutionState {
...
-                v.apply(u);
-                Some(v)
-            }
-            (Some(final_v), Some(SetUpdateOrDelete::Update(u))) => {
-                let mut v = final_v.clone();
-                v.apply(u);
-                Some(v)
-            }
-        };
+        let active_balance = self.get_active_balance_at_slot(next_slot, address);
```
damip, line 844: Note that if the balance is not found in the history, this should do a 2nd lookup into the final_state.
This has a performance impact and should be avoided because we already did this lookup above

massa-execution-worker/src/execution.rs
```diff
@@ -769,50 +831,33 @@ impl ExecutionState {
...
+        &self,
+        address: &Address,
+        key: &Hash,
+    ) -> (Option<Vec<u8>>, Option<Vec<u8>>) {
+        let final_entry = self.final_state.read().ledger.get_data_entry(address, key);
+        let next_slot = self
+            .active_cursor
+            .get_next_slot(self.config.thread_count)
+            .expect("slot overflow when getting speculative ledger");
+        let active_entry = self.get_active_data_entry_at_slot(next_slot, address, key);
```
damip, line 860: same remark as above on the performance impact of reading the final ledger potentially twice

massa-execution-worker/src/execution.rs
Eitu33: I don't understand how we can find the highest history index for which `output.slot > slot` given that `active_history` is of type `VecDeque`. I don't think it's possible to achieve constant complexity search with that data structure but maybe I'm wrong?

massa-ledger/src/ledger_db.rs
```diff
@@ -0,0 +1,279 @@
...
+const OPEN_ERROR: &str = "critical: rocksdb open operation failed";
+const CRUD_ERROR: &str = "critical: rocksdb crud operation failed";
+const CF_ERROR: &str = "critical: rocksdb column family operation failed";
+
+/// Ledger sub entry enum
+pub enum LedgerSubEntry {
+    /// Balance
+    Balance,
+    /// Bytecode
+    Bytecode,
```
damip, line 25: in a future PR, we could fuse Bytecode into the datastore

massa-ledger/src/ledger.rs
```diff
@@ -106,55 +102,78 @@ impl FinalLedger {
...
 
     /// Checks if a ledger entry exists
     ///
     /// # Returns
     /// true if it exists, false otherwise.
     pub fn entry_exists(&self, addr: &Address) -> bool {
-        self.sorted_ledger.contains_key(addr)
+        // note: document the "may"
+        self.sorted_ledger
+            .entry_may_exist(addr, LedgerSubEntry::Balance)
```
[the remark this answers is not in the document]
Eitu33: I do not exactly know where the may come from, but from what've I've red (see https://github.com/facebook/rocksdb/issues/4149#issuecomment-405966488) it seems outdated. I used it in the first place to not break the Ledger interface too much, and because it was qualified as more light weight that using `get()`, but I have seen by myself that it is not trustworthy, forgot to add a comment about it.

massa-ledger/src/ledger_db.rs
```diff
@@ -0,0 +1,279 @@
...
+
+    pub fn put(&mut self, addr: &Address, ledger_entry: LedgerEntry) {
+        let mut batch = WriteBatch::default();
+        let handle = self.0.cf_handle(LEDGER_CF).expect(CF_ERROR);
+
+        // balance
+        batch.put_cf(
+            handle,
+            balance_key!(addr),
+            ledger_entry.parallel_balance.to_raw().to_varint_bytes(),
```
damip, line 105: we *always* need to avoid usig to_raw. Here Balance::to_bytes_compact should be used

massa-execution-worker/src/execution.rs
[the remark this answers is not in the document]
damip: Oh ok then I think that this function should be renamed to reflect that we are only looking into the active history (we are not getting the absolute active value).

In turn, when looking for the "active balance", get_final_and_active_parallel_balance should return the info from the active history if available, otherwise it should return the final value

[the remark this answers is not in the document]
damip: To clarify: "active balance" does not mean "modified within the history". It means "the absolute balance at the output of the current slot". That means that if the active history does not give any info on it, it needs to be returned from the final state.

massa-ledger/src/ledger.rs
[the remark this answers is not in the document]
damip: ok let's avoid using it :)

massa-ledger/src/ledger.rs
damip: from_raw / to_raw should basically never be used (except when we absolutely need it in the SC engine)

This code snippet assumes that u64 is the underlying representation of Amount, which might change.

We should always use Amount::from_bytes_compact and Amount::to_bytes_compact for serialization, and never assume anything about the underlying representation of Amount outside of the Amount module.

Eitu33: Oh I forgot there was an implemented trait that already does that, will change it now.

AurelienFT: I join here to say that now in main there is the new way to deserialize for amount also with the deserializer. Maybe use it to avoid change in the near future.

massa-ledger/src/ledger_db.rs
damip: the day we change the underlying representation of Amount, we will have to find all the places where we made assumptions instead of using the functions exposed by the Amount module whose role is exactly that.

For the Result, you might just unwrap() it for now with a comment. In the future we will stop using this trait after the work on serialization improvements :)

damip (changes requested):
Nice work !
A few comments need to be addressed before we merge but overall it's good.

Should we discuss rocksDB in-RAM caching in a followup ?

massa-ledger/src/ledger_db.rs
```diff
@@ -0,0 +1,279 @@
...
+}
+
+macro_rules! balance_key {
+    ($addr:ident) => {
+        format!("1:{}", $addr).as_bytes()
+    };
+}
+
+// NOTE: still handle separate bytecode for now to avoid too many refactoring at once
+macro_rules! bytecode_key {
```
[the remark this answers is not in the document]
Eitu33: Done.

massa-ledger/src/ledger_db.rs
```diff
@@ -0,0 +1,273 @@
...
+pub(crate) struct LedgerDB(pub(crate) DB);
+
+/// Destroy the disk ledger db and free the lock
+pub fn destroy_ledger_db() {
+    DB::destroy(&Options::default(), DB_PATH).expect(OPEN_ERROR);
+}
+
+macro_rules! balance_key {
+    ($addr:ident) => {
+        [&[0u8], &$addr.to_bytes()[..]].concat()
```
damip, line 39: as a good practice let's avoid hardcoded "0u8", "1u8" etc... and use constants declared at the top of the file :)

massa-ledger/src/ledger_db.rs
```diff
@@ -0,0 +1,273 @@
...
+    };
+}
+
+// NOTE: still handle separate bytecode for now to avoid too many refactoring at once
+macro_rules! bytecode_key {
+    ($addr:ident) => {
+        [&[1u8], &$addr.to_bytes()[..]].concat()
```
damip, line 46: the order looks wrong. Let's start with addr, then 1u8 (and not the opposite because it will screw up prefix ranges)

massa-ledger/src/ledger_db.rs
Eitu33: Prefix was implemented for the intermediate bootstrap, will use it in the delete function.

massa-ledger/src/ledger_db.rs
damip: it is also incompatible with data_prefix

massa-ledger/src/ledger_db.rs
```diff
@@ -0,0 +1,273 @@
...
+    };
+}
+
+macro_rules! data_key {
+    ($addr:ident, $key:ident) => {
+        [&$addr.to_bytes()[..], &$key.to_bytes()[..]].concat()
```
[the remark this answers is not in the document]
damip: (for example 2u8)

massa-ledger/src/ledger_db.rs
Eitu33: We can change it later but I need it to be in that order for now because of the intermediate bootstrap implementation.

Eitu33: Can add a comment on that.

massa-ledger/src/ledger_db.rs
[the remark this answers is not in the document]
Eitu33: Will leave a comment about that but it just extra space taken for nothing right now IMHO.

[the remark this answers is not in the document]
Eitu33: Done.

massa-ledger/src/ledger_db.rs
Eitu33: Done.

massa-ledger/src/ledger_db.rs
Eitu33: Done.

massa-ledger/src/ledger.rs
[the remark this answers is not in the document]
Eitu33: Added a comment about it.

massa-ledger/src/ledger.rs
```diff
@@ -14,13 +15,11 @@ use std::collections::BTreeMap;
 /// The final ledger is part of the final state which is attached to a final slot, can be bootstrapped and allows others to bootstrap.
 /// The ledger size can be very high: it can exceed 1 terabyte.
 /// To allow for storage on disk, the ledger uses trees and has `O(log(N))` access, insertion and deletion complexity.
-///
-/// Note: currently the ledger is stored in RAM. TODO put it on the hard drive with cache.
```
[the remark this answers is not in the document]
Eitu33: This will be discussed in the follow up.

massa-ledger/src/ledger.rs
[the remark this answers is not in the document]
Eitu33: Actually I just removed it.

Eitu33:
@damip All the requests have been resolved, here is what I propose for the follow up PR's:

- Fusing bytecode with datastore (estimated time between 4 and 8 hours)
- Implement lazy sub-entry query for speculative ledger (estimated time between 4 and 8 hours)
- RocksDB cache implementation (didn't investigate enough to estimate)
- Moving blockgraph ledger to massa-ledger (didn't investigate enough to estimate accurately, would say between 4 and 7 days)

What would you like to go for first?

massa-ledger/src/ledger_db.rs
Eitu33: Yes I believe that should be part of the follow up as well.

Eitu33:
@damip Apparently RocksDB uses LRU cache by default: https://github.com/facebook/rocksdb/wiki/Block-Cache#lru-cache

damip (approved):

damip:
> @damip All the requests have been resolved, here is what I propose for the follow up PR's:

Good job ! This is a very important step for the refactoring.

>     * Fusing bytecode with datastore (estimated time between 4 and 8 hours)

Let's keep this on the side for now, we need to discuss it a bit more I think.

>     * Implement lazy sub-entry query for speculative ledger (estimated time between 4 and 8 hours)

That's something you can do now, but make sure to sync with Adrien L and Evgeni to not break the tooling and web3 libs on API changes.

> 
>     * RocksDB cache implementation (didn't investigate enough to estimate)

Apparently that part is solved by rocksdb.

>     * Moving blockgraph ledger to massa-ledger (didn't investigate enough to estimate accurately, would say between 4 and 7 days)

Let's do that last (not sure it will fit in testnet 11)

massa-execution-worker/src/execution.rs
```diff
@@ -769,51 +840,52 @@ impl ExecutionState {
...
-            (Some(final_v), Some(SetUpdateOrDelete::Update(u))) => {
-                let mut v = final_v.clone();
-                v.apply(u);
-                Some(v)
-            }
-        };
+        let active_balance = self.lookup_active_balance_at_slot(next_slot, address);
+        (
+            final_balance,
+            if active_balance.is_some() {
```
damip, line 858: There is still a problem here.

In the case where active_balance = None, it might mean that either:
* we got no info on the entry by looking at the active history => in that case we need to return the final_balance
* OR that the active history deletes the entry => in that case we need to return None instead of the final_balance => the current code does not handle that correctly

massa-api/src/public.rs
```diff
@@ -544,9 +546,9 @@ impl Endpoints for API<Public> {
                 Map::with_capacity_and_hasher(addresses.len(), BuildMap::default());
             let mut endorsements: Map<Address, Set<EndorsementId>> =
                 Map::with_capacity_and_hasher(addresses.len(), BuildMap::default());
-            let mut final_sce_ledger_info: Map<Address, SCELedgerInfo> =
+            let mut final_balances: Map<Address, Amount> =
                 Map::with_capacity_and_hasher(addresses.len(), BuildMap::default());
-            let mut candidate_sce_ledger_info: Map<Address, SCELedgerInfo> =
+            let mut candidates_balances: Map<Address, Amount> =
```
damip, line 551: We need to document every API change in the API doc, and also ping Evgeni and Adrien because it will break all the tooling (eg. explorer etc...)

damip (changes requested):

massa-api/src/public.rs
Eitu33: I implemented the precedent behaviour with the new ledger so that it stays compatible.

Eitu33:
@damip ready to be merged!

massa-execution-worker/src/execution.rs
```diff
@@ -248,7 +248,56 @@ impl ExecutionState {
...
+    /// Lazily query (from end to beginning) the active balance of an address at a given slot.
+    /// Returns None if the address balance could not be determined from the active history.
+    ///
+    /// NOTE: temporary, needs to be done in the speculative ledger
+    pub fn fetch_active_history_balance(
+        &self,
+        slot: Slot,
+        addr: &Address,
+    ) -> HistorySearchResult<Amount> {
+        self.verify_active_slot(slot);
```
damip, line 270: let's make sure the panic in there cannot be triggered by someone asking for the balance at a block in the future through the api

massa-execution-worker/src/execution.rs
```diff
@@ -248,7 +248,56 @@ impl ExecutionState {
...
+
+        if let Some(n) = self.get_active_index(slot) {
+            let iter = self.active_history.iter().rev().skip(n);
```
damip, line 273: I think it's

```
let iter = self.active_history.iter().skip(n).rev();
```

massa-ledger/src/ledger_db.rs
```diff
@@ -0,0 +1,340 @@
...
+    Datastore(Hash),
+}
+
+/// Disk ledger DB module
+///
+/// Contains a RocksDB DB instance
+pub(crate) struct LedgerDB(pub(crate) DB);
+
+/// Destroy the disk ledger db and free the lock
+pub fn destroy_ledger_db() {
```
damip, line 40: for now let's make sure we destroy the db before every bootstrap (since we cannot resume yet and we don't want to keep outdated elements from the previous launch in the db)

massa-ledger/src/ledger_db.rs
```diff
@@ -0,0 +1,340 @@
...
+
+/// Datastore entry prefix formatting macro
+macro_rules! data_prefix {
+    ($addr:ident) => {
+        &$addr.to_bytes()[..]
+    };
+}
+
+/// Compute the end bound for a given prefix
+pub fn end_prefix(prefix: &[u8]) -> Option<Vec<u8>> {
```
damip, line 80: I am still confused by what this function does exactly :)

massa-execution-worker/src/execution.rs
Eitu33: Are you sure? `active_history` has the newest element A at the back, so if we want to fetch the data of a given slot B the `skip` should happen after the `rev` because we are skipping from A to B, and parcouring the slots that come before B to determine the value. Or am I getting it wrong?

massa-execution-worker/src/execution.rs
Eitu33: This function is currently never used, I left it there as a blueprint for my future implementation of speculative ledger lazy sub-queries. No chance of that happening right now, I noted it and will make sure it cannot happen when it is actually used by the API.

damip: okay

massa-execution-worker/src/execution.rs
Eitu33: FI, this depends on which implementation of `get_active_index` we are using: https://github.com/massalabs/massa/pull/2590#discussion_r883769585

massa-execution-worker/src/execution.rs
```diff
@@ -248,7 +248,56 @@ impl ExecutionState {
...
             panic!("cannot execute at a slot beyond active cursor + 1");
         }
+    }
+
+    /// Computes the index of a given slot in the active history
+    fn get_active_index(&self, slot: Slot) -> Option<usize> {
+        let current = self.active_cursor.period * (self.config.thread_count as u64)
+            + (self.active_cursor.thread as u64);
+        let asked = slot.period * (self.config.thread_count as u64) + (slot.thread as u64);
+        current.checked_sub(asked).and_then(|v| v.try_into().ok())
```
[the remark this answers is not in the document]
Eitu33: Done.

massa-execution-worker/src/execution.rs
Eitu33: Done.

massa-ledger/src/ledger_db.rs
```diff
@@ -0,0 +1,340 @@
...
+        opt.set_iterate_upper_bound(end_prefix(data_prefix!(addr)).unwrap());
+        for (key, _) in self.0.iterator_cf_opt(
+            handle,
+            opt,
+            IteratorMode::From(data_prefix!(addr), Direction::Forward),
+        ) {
+            batch.delete_cf(handle, key);
+        }
+    }
+}
```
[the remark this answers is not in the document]
Eitu33: There is no such thing as `write_slot_changes(changes: LedgerChanges, new_slot: Slot)`, only fn `apply(&mut self, changes: LedgerChanges)`. Do we want something like that? If yes for what purpose?

Also I set all the write functions as `pub(crate)`.

massa-ledger/src/ledger_db.rs
Eitu33: Improved the comment.

massa-ledger/src/ledger_db.rs
Eitu33: Done.

massa-ledger/src/ledger_db.rs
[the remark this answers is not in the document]
damip: Oh OK I see. 

Basically the idea was to not import WriteBatch nor anything rocksdb-related into ledger.rs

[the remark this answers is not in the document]
damip: Also, the slot inside the metadata column should also be updated atomically in the same batch

damip (approved):

damip:
bors merge

Eitu33:
bors merge

AurelienFT (approved):
Approving so that bors don't fail.

PR #2596, opened 2022-05-06
Graph: remove bool to option feature flag

gterzian:
Fix for https://github.com/massalabs/massa/runs/6317067411?check_suite_focus=true#step:5:610

### DIFF
--- a/massa-graph/src/lib.rs
+++ b/massa-graph/src/lib.rs
@@ -3,7 +3,6 @@
 #![warn(missing_docs)]
 #![warn(unused_crate_dependencies)]
 #![feature(async_closure)]
-#![feature(bool_to_option)]
 #![feature(hash_drain_filter)]
 #![feature(map_first_last)]
 #![feature(int_roundings)]


massa-client/src/cmds.rs
```diff
@@ -627,16 +628,14 @@ impl Command {
                 for key in parse_vec::<Address>(parameters)?.into_iter() {
                     match wallet.remove_address(key) {
                         Ok(_) => {
-                            res.push_str(&format!("Removed address {} from the wallet\n", key));
+                            let _ = writeln!(res, "Removed address {} from the wallet", key);
```
Eitu33, line 631: Why the `let _ =` ?

Eitu33 (approved):

massa-client/src/cmds.rs
AurelienFT: Because we don't use the result. It's the recommended practice by clippy here : https://rust-lang.github.io/rust-clippy/master/index.html#format_push_string

AurelienFT:
bors merge

PR #2625, opened 2022-05-17
Cursor to advance part by part in the ledger.

AurelienFT:
This PR is part of #2577 .

The goal of this PR is to load the ledger by chunks to avoid loading too much in RAM and in the network.
The ledger can then by download by chunk but must also be created by chunks.
The cursor should also be possible to be compared with another one.

### DIFF
--- a/massa-hash/src/hash.rs
+++ b/massa-hash/src/hash.rs
@@ -6,6 +6,8 @@ use crate::error::MassaHashError;
 use crate::settings::HASH_SIZE_BYTES;
 use std::{convert::TryInto, str::FromStr};
 
+use massa_serialization::{Deserializer, SerializeError, Serializer};
+
 /// SHA256 hash
 #[derive(Eq, PartialEq, Ord, PartialOrd, Copy, Clone, Hash)]
 pub struct Hash(bitcoin_hashes::sha256::Hash);
@@ -113,9 +115,21 @@ impl Hash {
                 .map_err(|err| MassaHashError::ParsingError(format!("{}", err)))?,
         ))
     }
+}
 
-    /// nom combinator to deserialize a hash
-    pub fn nom_deserialize(buffer: &[u8]) -> IResult<&[u8], Hash> {
+/// Deserializer for `Hash`
+#[derive(Default)]
+pub struct HashDeserializer;
+
+impl HashDeserializer {
+    /// Creates a deserializer for `Hash`
+    pub fn new() -> Self {
+        Self
+    }
+}
+
+impl Deserializer<Hash> for HashDeserializer {
+    fn deserialize<'a>(&self, buffer: &'a [u8]) -> IResult<&'a [u8], Hash> {
         if buffer.len() < HASH_SIZE_BYTES {
             return Err(nom::Err::Error(nom::error::Error::new(
                 buffer,
--- a/massa-hash/src/hash.rs
+++ b/massa-hash/src/hash.rs
@@ -1,12 +1,10 @@
 // Copyright (c) 2022 MASSA LABS <info@massa.net>
 
-use nom::IResult;
-
 use crate::error::MassaHashError;
 use crate::settings::HASH_SIZE_BYTES;
-use std::{convert::TryInto, str::FromStr};
-
 use massa_serialization::Deserializer;
+use nom::IResult;
+use std::{convert::TryInto, str::FromStr};
 
 /// SHA256 hash
 #[derive(Eq, PartialEq, Ord, PartialOrd, Copy, Clone, Hash)]
--- a/massa-hash/src/hash.rs
+++ b/massa-hash/src/hash.rs
@@ -6,7 +6,7 @@ use crate::error::MassaHashError;
 use crate::settings::HASH_SIZE_BYTES;
 use std::{convert::TryInto, str::FromStr};
 
-use massa_serialization::{Deserializer, SerializeError, Serializer};
+use massa_serialization::Deserializer;
 
 /// SHA256 hash
 #[derive(Eq, PartialEq, Ord, PartialOrd, Copy, Clone, Hash)]
--- a/massa-hash/src/hash.rs
+++ b/massa-hash/src/hash.rs
@@ -1,5 +1,7 @@
 // Copyright (c) 2022 MASSA LABS <info@massa.net>
 
+use nom::IResult;
+
 use crate::error::MassaHashError;
 use crate::settings::HASH_SIZE_BYTES;
 use std::{convert::TryInto, str::FromStr};
@@ -109,6 +111,27 @@ impl Hash {
         Ok(Hash(
             bitcoin_hashes::sha256::Hash::from_slice(&data[..])
                 .map_err(|err| MassaHashError::ParsingError(format!("{}", err)))?,
+        ))
+    }
+
+    /// nom combinator to deserialize a hash
+    pub fn nom_deserialize(buffer: &[u8]) -> IResult<&[u8], Hash> {
+        if buffer.len() < HASH_SIZE_BYTES {
+            return Err(nom::Err::Error(nom::error::Error::new(
+                buffer,
+                nom::error::ErrorKind::LengthValue,
+            )));
+        }
+        use bitcoin_hashes::Hash;
+        Ok((
+            &buffer[HASH_SIZE_BYTES..],
+            Hash(
+                bitcoin_hashes::sha256::Hash::from_slice(&buffer[..HASH_SIZE_BYTES]).map_err(
+                    |_| {
+                        nom::Err::Error(nom::error::Error::new(buffer, nom::error::ErrorKind::Fail))
+                    },
+                )?,
+            ),
         ))
     }
 }
--- a/massa-ledger/src/cursor.rs
+++ b/massa-ledger/src/cursor.rs
@@ -77,7 +77,10 @@ impl Deserializer<LedgerCursorStep> for LedgerCursorStepDeserializer {
 
 #[derive(Debug, PartialEq, Eq, Clone)]
 /// A cursor to iterate through the ledger with different granularity.
-pub struct LedgerCursor(pub Address, pub LedgerCursorStep);
+pub struct LedgerCursor {
+    pub address: Address,
+    pub step: LedgerCursorStep,
+}
 
 /// A serializer for the ledger cursor.
 #[derive(Default)]
@@ -97,8 +100,8 @@ impl LedgerCursorSerializer {
 impl Serializer<LedgerCursor> for LedgerCursorSerializer {
     fn serialize(&self, value: &LedgerCursor) -> Result<Vec<u8>, SerializeError> {
         let mut bytes = vec![];
-        bytes.extend(value.0.to_bytes());
-        bytes.extend(self.cursor_step_serializer.serialize(&value.1)?);
+        bytes.extend(value.address.to_bytes());
+        bytes.extend(self.cursor_step_serializer.serialize(&value.step)?);
         Ok(bytes)
     }
 }
@@ -126,7 +129,7 @@ impl Deserializer<LedgerCursor> for LedgerCursorDeserializer {
             |input| self.address_deserializer.deserialize(input),
             |input| self.cursor_step_deserializer.deserialize(input),
         ));
-        let (rest, (address, cursor)) = parser(input)?;
-        Ok((rest, LedgerCursor(address, cursor)))
+        let (rest, (address, step)) = parser(input)?;
+        Ok((rest, LedgerCursor { address, step }))
     }
 }
--- a/massa-ledger/src/cursor.rs
+++ b/massa-ledger/src/cursor.rs
@@ -3,6 +3,13 @@ use massa_models::{address::AddressDeserializer, Address};
 use massa_serialization::{Deserializer, SerializeError, Serializer};
 use nom::{sequence::tuple, IResult};
 
+/// When sending the ledger we need to split the messages so that we don't send the whole ledger or a huge part of the datastore in one message.
+/// We have 5 different state where the cursor can stop that define 5 different positions in the ledger:
+/// `Start`: At the start of a new address
+/// `Balance`: Before the balance of an address
+/// `Bytecode`: Before the bytecode of an address
+/// `Datastore`: If `Hash` is None then it's before the datastore otherwise it's after the key represented by the `Hash``
+/// `Finish`: After the encoding of the whole data of an address
 #[derive(Debug, Clone, PartialEq, Eq)]
 pub enum LedgerCursorStep {
     Start,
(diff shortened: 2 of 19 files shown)


AurelienFT:
This PR is ready for a first review moreover there is things that I don't like but I don't find idiomatic solution. If someone has ideas during the review :

- The deserializers can't be use in nom combinators such as `tuple` or `alt` because it don't support methods of differents objects to be pass. (A ressource : https://users.rust-lang.org/t/parsers-carrying-state-in-nom-7/65291)
- I think the `get_ledger_part` and `set_ledger_part` functions can be rewritten but with the granularity level of the datastore I don't find a good way without restarting a iter for each key of the datastore.
- The deserializers need to be used with import of nom. If we say that we don't find solution to problem 1 (above) then we are not forced to match the format of nom parsers and then we can change the signature of `deserialize`. It's sad because we lost the combinators power. If we resolve problem 1 then the signature will always have a nom::error::Error and then we will need to import nom in every crate where we write a deserializer.

By the time of reviewers and questions answers I'll write more tests.

massa-ledger/src/cursor.rs
```diff
@@ -0,0 +1,125 @@
...
+                }
+            }
+            4 => Ok((&bytes[1..], LedgerCursorStep::Finish)),
+            _ => Err(nom::Err::Error(nom::error::Error::new(
+                bytes,
+                nom::error::ErrorKind::Digit,
+            ))),
+        }
+    }
+}
```
Eitu33, line 73: I believe you can associate values to enum fields in rust, something like: 
```rust
enum MyEnum {
    A = 1,
    B = 2,
    C = 3,
}
```

And convert a byte directly to it's enum correspondant with https://docs.rs/num-derive/0.3.3/num_derive/

massa-models/src/serialization.rs
```diff
@@ -19,6 +21,78 @@ impl SerializeVarInt for u16 {
     }
 }
 
+macro_rules! gen_varint {
```
Eitu33, line 24: Nice 👍

massa-models/src/amount.rs
```diff
@@ -177,6 +181,45 @@ impl FromStr for Amount {
...
+pub struct AmountSerializer;
+
+impl AmountSerializer {
+    /// Create a new `AmountSerializer`
+    pub fn new() -> Self {
+        Self
+    }
+}
+
+impl Serializer<Amount> for AmountSerializer {
```
Eitu33, line 195: I don't exactly understand the reason behind every `{type}Serializer` type, what prevents us to implement `Serializer` for `Amount` directly here for example?

AurelienFT: Because we want to be able to pass him limits or things like that and if we want to not pass it as parameters and keep it generic then we need to create a structure that holds the value.

massa-ledger/src/cursor.rs
AurelienFT: I tested your crate and this one : https://github.com/illicitonion/num_enum which is more maintained and both, sadly, don't works with enums with fields on it like my field `Datastore` :/

massa-ledger/src/ledger.rs
```diff
@@ -184,4 +196,244 @@ impl FinalLedger {
...
+                        ModelsError::DeserializeError("Fail to deserialize address".to_string())
+                    })?;
+                    data = rest;
+                    self.sorted_ledger
+                        .entry(address)
+                        .or_insert_with(LedgerEntry::default);
+                    cursor.1 = LedgerCursorStep::Balance;
+                }
+                LedgerCursorStep::Balance => {
+                    let amount_deserializer = AmountDeserializer::new();
```
Eitu33, line 318: Shouldn't this deserializer be initialized at the beginning of the function like the others?

massa-ledger/src/cursor.rs
Eitu33: That's kinda sad

massa-ledger/src/ledger.rs
AurelienFT: Done.

massa-ledger/src/ledger.rs
```diff
@@ -184,4 +193,237 @@ impl FinalLedger {
...
+        Ok((data, next_cursor))
+    }
+
+    /// Set a part of the ledger
+    /// Used for bootstrap
+    /// Parameters:
+    /// * cursor: Where we stopped in the ledger
+    ///
+    /// Returns:
+    /// Nothing on success error else.
```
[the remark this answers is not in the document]
adrien-zinger: Actually on bootstrap for the both version get a chunk of the ledger (just like what you did) but he converts that into key-value entries for the client that is much more convenient. So the client never have to know where the cursor is, it's automatically done thx to the ordering of the BTreeMap.

adrien-zinger (commented):
I have difficulties to understand the presence of the cursor in the client side to honest

Edit: I think I need more context here :) @AurelienFT when you have a bit of time..

AurelienFT:
I found a way to make the deserializers with nom. Turning this PR back to draft. In client side it's because we will receive raw data that can be stop at a key of a datastore for example and we need to deserialize the good way @adrien-zinger . But we can make a call if you want after I finished my deserializers.

massa-ledger/src/cursor.rs
```diff
@@ -0,0 +1,125 @@
...
+
+impl LedgerCursorDeserializer {
+    /// Creates a new ledger cursor deserializer.
+    pub fn _new() -> Self {
+        Self {
+            bootstrap_cursor_step_deserializer: LedgerCursorStepDeserializer::_new(),
+        }
+    }
+}
+
```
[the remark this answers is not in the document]
AurelienFT: I don't think so. They will be useful in a very near future (streaming bootstrap) and they are part of the cursor. Just the cursor isn't send yet over the network. If we can add it here I think it's better to make the streaming bootstrap lighter.

massa-hash/src/hash.rs
```diff
@@ -113,6 +117,39 @@ impl Hash {
...
+/// Deserializer for `Hash`
+#[derive(Default)]
+pub struct HashDeserializer;
+
+impl HashDeserializer {
+    /// Creates a deserializer for `Hash`
+    pub fn new() -> Self {
+        Self
+    }
+}
```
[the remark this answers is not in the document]
AurelienFT: I prefer having always a new as a convention. You think it's a bad practice ?

AurelienFT:
I have understand how to use the deserializer with nom combinators and refactored the PR. I think it's ready for "final" reviews.

massa-ledger/src/ledger.rs
```diff
@@ -184,4 +195,231 @@ impl FinalLedger {
...
+                        next_cursor.1 = LedgerCursorStep::Finish;
+                    }
+                    LedgerCursorStep::Finish => {
+                        data.push(DATASTORE_END_IDENTIFIER);
+                        next_cursor.1 = LedgerCursorStep::Start;
+                        next_cursor.0 = *addr;
+                        break;
+                    }
+                }
+                if data.len() as u64 > LEDGER_PART_SIZE_MESSAGE_BYTES {
```
gterzian, line 264: Should this not be an equality check?

massa-ledger/src/cursor.rs
```diff
@@ -0,0 +1,132 @@
...
+                bytes,
+                nom::error::ErrorKind::Digit,
+            ))),
+        }
+    }
+}
+
+#[derive(Debug, PartialEq, Eq, Clone)]
+/// A cursor to iterate through the ledger with different granularity.
+pub struct LedgerCursor(pub Address, pub LedgerCursorStep);
```
gterzian, line 80: Can we make these two names fields please?

gterzian (commented):
Great work! I've had a pretty good look around, just a few comments so far. 

This is not used yet outside of the tests?

massa-ledger/src/cursor.rs
```diff
@@ -0,0 +1,132 @@
...
+            LedgerCursorStep::Balance => Ok(vec![1]),
+            LedgerCursorStep::Bytecode => Ok(vec![2]),
+            LedgerCursorStep::Datastore(key) => {
+                let mut bytes = vec![3];
+                if let Some(key) = key {
+                    bytes.extend(key.to_bytes());
+                }
+                Ok(bytes)
+            }
+            LedgerCursorStep::Finish => Ok(vec![4]),
```
damip, line 37: for what purpose  is `LedgerCursorStep::Finish` necessary ?

massa-ledger/src/cursor.rs
```diff
@@ -0,0 +1,132 @@
...
+    fn _new() -> Self {
+        Self {
+            hash_deserializer: HashDeserializer::new(),
+        }
+    }
+}
+
+impl Deserializer<LedgerCursorStep> for LedgerCursorStepDeserializer {
+    fn deserialize<'a>(&self, bytes: &'a [u8]) -> IResult<&'a [u8], LedgerCursorStep> {
+        match bytes[0] {
```
damip, line 57: will panic if  `bytes` is not long enough. Use `bytes.get(0)`

massa-ledger/src/cursor.rs
damip, line 37: when is  `LedgerCursorStep::Finish` necessary ?

massa-ledger/src/ledger.rs
```diff
@@ -184,4 +195,231 @@ impl FinalLedger {
...
+                .sorted_ledger
+                .first_key_value()
+                .ok_or_else(|| ModelsError::BufferError("Ledger empty".into()))?
+                .0,
+            LedgerCursorStep::Start,
+        ));
+        let mut data = Vec::new();
+        let amount_serializer = AmountSerializer::new();
+        for (addr, entry) in self.sorted_ledger.range(next_cursor.0..) {
+            while (data.len() as u64) < LEDGER_PART_SIZE_MESSAGE_BYTES {
```
damip, line 221: note that LEDGER_PART_SIZE_MESSAGE_BYTES is an estimate: it will often be overshot a bit by the last added entry

massa-ledger/src/ledger.rs
```diff
@@ -184,4 +195,231 @@ impl FinalLedger {
...
+    /// A subset of the ledger starting at `cursor` and of size `LEDGER_PART_SIZE_MESSAGE_BYTES` bytes.
+    pub fn get_ledger_part(
+        &self,
+        cursor: Option<LedgerCursor>,
+    ) -> Result<(Vec<u8>, LedgerCursor), ModelsError> {
+        let mut next_cursor = cursor.unwrap_or(LedgerCursor(
+            *self
+                .sorted_ledger
+                .first_key_value()
+                .ok_or_else(|| ModelsError::BufferError("Ledger empty".into()))?
```
damip, line 214: the ledger being empty should not crash the function, just return a finish

massa-models/src/address.rs
```diff
@@ -158,6 +160,30 @@ impl Address {
...
+        Self {
+            hash_deserializer: HashDeserializer::new(),
+        }
+    }
+}
+
+impl Deserializer<Address> for AddressDeserializer {
+    fn deserialize<'a>(&self, buffer: &'a [u8]) -> IResult<&'a [u8], Address> {
+        Ok((
+            &buffer[ADDRESS_SIZE_BYTES..],
```
damip, line 181: this might panic if buffer is smaller than ADDRESS_SIZE_BYTES

massa-models/src/node_configuration/default.rs
```diff
@@ -112,6 +112,8 @@ pub const PERIODS_PER_CYCLE: u64 = 128;
 pub const POS_LOOKBACK_CYCLES: u64 = 2;
 /// PoS lock cycles: when some rolls are released, we only credit the coins back to their owner after waiting `pos_lock_cycles`
 pub const POS_LOCK_CYCLES: u64 = 1;
+/// Maximum size batch of data in a part of the ledger
+pub const LEDGER_PART_SIZE_MESSAGE_BYTES: u64 = 3000;
```
damip, line 116: we can go bigger, like 1MB

AurelienFT:
> Great work! I've had a pretty good look around, just a few comments so far.
> 
> This is not used yet outside of the tests?

Yes it will be used in the streaming bootstrap. We just merge it first to make the PR of streaming bootstrap less heavy.

massa-ledger/src/cursor.rs
AurelienFT: Done

massa-hash/src/hash.rs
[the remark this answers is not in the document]
adrien-zinger: A `new` is often used when you want to keep private fields. Here the HasDeserializer is naked, so the code is redundant. That said, it's not a blocking issue, so feel free to resolve the conversation whatever your choice :wink:

massa-ledger/src/ledger.rs
adrien-zinger: I would add a warning too :slightly_smiling_face:

massa-ledger/src/ledger.rs
AurelienFT: Yes. It's will always be a bit more (see https://github.com/massalabs/massa/pull/2625#discussion_r878122100)

massa-ledger/src/cursor.rs
```diff
@@ -0,0 +1,132 @@
+use massa_hash::{Hash, HashDeserializer};
+use massa_models::{address::AddressDeserializer, Address};
+use massa_serialization::{Deserializer, SerializeError, Serializer};
+use nom::{sequence::tuple, IResult};
+
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub enum LedgerCursorStep {
```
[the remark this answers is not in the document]
AurelienFT: Done.

massa-ledger/src/ledger.rs
```diff
@@ -184,4 +195,231 @@ impl FinalLedger {
...
+    /// Parameters:
+    /// * cursor: Where we stopped in the ledger
+    ///
+    /// Returns:
+    /// A subset of the ledger starting at `cursor` and of size `LEDGER_PART_SIZE_MESSAGE_BYTES` bytes.
+    pub fn get_ledger_part(
+        &self,
+        cursor: Option<LedgerCursor>,
+    ) -> Result<(Vec<u8>, LedgerCursor), ModelsError> {
+        let mut next_cursor = cursor.unwrap_or(LedgerCursor(
```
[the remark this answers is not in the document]
AurelienFT: I have refactored it using or_else :)

massa-ledger/src/cursor.rs
AurelienFT: Done.

massa-ledger/src/ledger.rs
AurelienFT: I have turn the return type to an option of cursor and if there is no ledger it will return an empty data and none as a cursor.

massa-models/src/address.rs
AurelienFT: Refactored.

massa-models/src/node_configuration/default.rs
AurelienFT: I have changed to 30 000 000

massa-hash/src/hash.rs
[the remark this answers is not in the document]
AurelienFT: Maybe in the future we could add a u64 deserializer in hash to check the len for example and we don't want to make it public as made in `VecU8Deserializer` for example and so the new signature will not change but the `{}` will not work anymore. I prefer to continue using new.

massa-ledger/src/cursor.rs
damip: I would have imagined something like:
when you ask for stuff strictly after AddressA Datastore(last_element), the cursor jumps to AddressB Balance.
Not sure it's the best solution in practice, just wondering.

AurelienFT: Yes I have think about it but when the ledger is finish what would be the cursor because it can't be AddressB Datastore(last_element) because then it will fetch again the last element of AddressB because the for loop include the key in the cursor. If I change it to not include the key then after the bytecode the cursor will not be Datastore(None) but Datastore(the first key) but on the client side it will be impossible to build this cursor before reading because he don't know the hash and then there will be a de-synchronization.

massa-ledger/src/ledger.rs
gterzian: Ok, thanks for the info.

Eitu33 (approved):

AurelienFT:
bors merge

Issue #2624, opened 2022-05-17
Fix shared storage pruning


gterzian:
In looking at the various `MissingBlock` errors, I've noticed the following source of potential race condition, which could lead to storage not having the block when the graph status for it is `Incoming`.

In all situations, when pruning a block, there is a race condition with protocol storing incoming blocks and sending the `ProtocolEvent::ReceivedBlock` to consensus. 

Basically, when we prune a block via graph, and remove the block from storage, we may be removing not only a block that was put there "previously", but also the same block that may have been put there again by protocol and which is on it's way to graph via a `ProtocolEvent::ReceivedBlock`. 

In such a case, when the  block is processed by graph, it will be `Incoming` as status, but absent from storage because of the pruning that just occured(in relation to another block or header that had been processes previously and whose status was up for pruning). 

The fix for this would be for graph itself to store the block when first receiving it(by adding the block to the `ProtocolEvent::ReceivedBlock` event), as opposed to protocol [just before](https://github.com/massalabs/massa/blob/939cc7f68167062447214ccee0c47603bbfefa90/massa-protocol-worker/src/protocol_worker.rs#L1254) sending the `ProtocolEvent::ReceivedBlock` event. 

In other words, and this goes also for [operations](https://github.com/massalabs/massa/pull/2587), the component doing the pruning needs to be the same one that does the storing(a single "writer"), whereas multiple components can read from storage. 

----

Another problem is that in both [`prune_waiting_for_dependencies`](https://github.com/massalabs/massa/blob/939cc7f68167062447214ccee0c47603bbfefa90/massa-graph/src/block_graph.rs#L3433) and [`prune_slot_waiting`](https://github.com/massalabs/massa/blob/939cc7f68167062447214ccee0c47603bbfefa90/massa-graph/src/block_graph.rs#L3597), we don't differentiate between pruning a block or a header, and in both cases we also prune shared storage. 

This could result in a missing block when it is incoming, in the following situation:

1. Protocol receives a block, and stores it in shared storage.
2. Graph prunes the header, removing the block from shared storage.
3. Graph handles the protocol event, add the block as incoming, and processes it, finding it missing from storage. 

We should only prune storage when we ware pruning blocks, not when we are pruning headers.

gterzian:
```tla
------------------------ MODULE FailingSharedStorage ------------------------
EXTENDS Naturals, Sequences
VARIABLES incoming, outgoing, statuses, shared_storage
CONSTANT QLen, Id
[...]

PR #2629, opened 2022-05-19
Fix shared storage

gterzian:
FIX https://github.com/massalabs/massa/issues/2624

### DIFF
--- a/massa-consensus-worker/src/consensus_worker.rs
+++ b/massa-consensus-worker/src/consensus_worker.rs
@@ -1099,6 +1099,8 @@ impl ConsensusWorker {
         match event {
             ProtocolEvent::ReceivedBlock {
                 block_id,
+                block,
+                serialized,
                 slot,
                 operation_set,
                 endorsement_ids,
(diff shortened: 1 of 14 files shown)


gterzian:
Ok this one is ready for review. 

I didn't do the "only prune blocks, not headers" part, because I think that with the proposed changes it doesn't matter anymore(if the block status is header only, there will not be a block in shared storage and the removal will just be a no-op).

AurelienFT (approved):
Ok for me. Tested on the simulator. Works well.

adrien-zinger (approved):

gterzian:
@damip Do you want to give a last look? I think this is a priority for the next testnet because it fixes the various `MissingBlock` errors encountered.

massa-consensus-worker/src/consensus_worker.rs
```diff
@@ -1107,6 +1109,12 @@ impl ConsensusWorker {
                     "consensus.consensus_worker.process_protocol_event.received_block",
                     { "block_id": block_id }
                 );
+
+                // Store block in shared storage.
```
damip, line 1113: I feel like there might be an edge case here. What happens if the block is immediately rejected as invalid or stale by self.block_db.incoming_block below ? Will the block ever be deleted from storage ?

gterzian: That may be the case, and I don't think this would be introduced by this PR. 

I have filed https://github.com/massalabs/massa/issues/2637 for further investigation.

damip:
bors merge

PR #2635, opened 2022-05-23
increase cursor delay

damip:
Increase execution cursor delay from 0 to 2000 milliseconds to reduce CPU usage


sebastien-forestier (approved):

damip:
bors merge

PR #2636, opened 2022-05-23
refactor: sort list of stakers based on the number of rolls

### LINKED ISSUE #2633: Sort list of stakers based on the number of rolls
When getting stakers using get_stakers api, the list of stakers obtained is not sorted.
Sorting the list of stakers based on the number of rolls in descending order at the server side will be very useful.
This will reduce the computation time and resource at the client side, particularly for mobile devices with limited computation resources.

aoudiamoncef:
resolves #2633

### DIFF
--- a/massa-api/src/lib.rs
+++ b/massa-api/src/lib.rs
@@ -20,7 +20,7 @@ use massa_models::execution::ExecuteReadOnlyResponse;
 use massa_models::node::NodeId;
 use massa_models::operation::OperationId;
 use massa_models::output_event::SCOutputEvent;
-use massa_models::prehash::{Map, Set};
+use massa_models::prehash::Set;
 use massa_models::{Address, BlockId, EndorsementId, SignedOperation, Version};
 use massa_network_exports::{NetworkCommandSender, NetworkSettings};
 use massa_pool::PoolCommandSender;
@@ -205,7 +205,7 @@ pub trait Endpoints {
 
     /// Returns the active stakers and their active roll counts for the current cycle.
     #[rpc(name = "get_stakers")]
-    fn get_stakers(&self) -> BoxFuture<Result<Map<Address, u64>, ApiError>>;
+    fn get_stakers(&self) -> BoxFuture<Result<Vec<(Address, u64)>, ApiError>>;
 
     /// Returns operations information associated to a given list of operations' IDs.
     #[rpc(name = "get_operations")]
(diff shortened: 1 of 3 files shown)


AurelienFT:
Thanks for your contribution ❤️  Can you fix the CI please @aoudiamoncef ?

aoudiamoncef:
Hi @AurelienFT 

With `BtreeMap` we will have a natural order by keys which isn't our purpose. I did  lot of research to find a better way before changing return type to `Vec(Address, u64)>`.  

I got no issue with `cargo clippy` in local environment.

AurelienFT:
> With `BtreeMap` we will have a natural order by keys which isn't our purpose(order by value desc). I did lot of research to find a better way before changing return type to `Vec(Address, u64)>`.

My bad I misread it.

aoudiamoncef:
> I think you can change the return type of the function : get_active_stakers to a `BtreeMap` so that it's sorted and we don't have the API format.

We could do the trick by creating an `HashMap` from the sorted `Vec(Address, u64)>`, so we could keep method signature. But the question is: where we'll add this behaviors ?  get_active_stakers or get_stakers

AurelienFT:
> We could do the trick by creating an `HashMap` from the sorted `Vec(Address, u64)>`, so we could keep method signature. But the question is: where we'll add this behaviors ? get_active_stakers or get_stakers

It's ok to change the behavior because we are in testnet. Just want @qdrn to look at it because he is the one maintaining the tools that use this API endpoint.

aoudiamoncef:
We are f***d, we have to use `LinkedHashMap` structure to achieve what we need. It's frustrating to add an external crate. 

The Rust idiomatic way to solve this issue is to use `Vec(Address, u64)`.

AurelienFT:
> The Rust idiomatic way to solve this issue is to use `Vec(Address, u64)`.

I'm ok with using A `Vec` now. Just need @qdrn approval and also so that he knew what he need to change on his side.

AurelienFT (approved):
I'm ok with it.

qdrn:
> I'm ok with using A `Vec` now. Just need @qdrn approval and also so that he knew what he need to change on his side.

Either are fine for me. It will not change much the tooling in both cases.

AurelienFT:
Ok then I merge it for the next testnet

AurelienFT:
bors merge

PR #2665, opened 2022-06-06
Remove disk ledger dir on bootstrap.

Eitu33:


massa-node/src/main.rs
```diff
@@ -99,6 +99,13 @@ async fn launch() -> (
         FinalState::new(final_state_config).expect("could not init final state"),
     ));
 
+    // Remove current disk ledger if there is one
+    // NOTE: this is temporary, since we cannot currently handle bootstrap from remaining ledger
+    if SETTINGS.ledger.disk_ledger_path.exists() {
```
AurelienFT, line 104: Isn't a problem that we delete after getting a connector in the initialization of final state just above ? The new DB will still exist or we should delete it before the final state ?

AurelienFT (approved):

massa-node/src/main.rs
Eitu33: Well spotted, could've been a nasty mistake.

AurelienFT:
bors merge

AurelienFT:
bors merge

PR #2675, opened 2022-06-09
Single data entry query API call

Eitu33:


massa-execution-exports/src/test_exports/mock.rs
```diff
@@ -106,6 +106,14 @@ impl ExecutionController for MockExecutionController {
         response_rx.recv().unwrap()
     }
 
+    fn get_final_and_active_data_entry(
+        &self,
+        _addr: &Address,
+        _key: &massa_hash::Hash,
```
AurelienFT, line 112: To avoid differences in the code I think we should import massa_hash::Hash at the top of the file like the other files.

massa-execution-worker/src/execution.rs
```diff
@@ -846,12 +887,35 @@ impl ExecutionState {
...
+        let next_slot = self
+            .active_cursor
+            .get_next_slot(self.config.thread_count)
+            .expect("slot overflow when getting speculative ledger");
+        let search_result = self.fetch_active_history_data_entry(next_slot, address, key);
+        (
+            final_entry.clone(),
+            match search_result {
+                HistorySearchResult::Found(active_entry) => Some(active_entry),
+                HistorySearchResult::NotFound => final_entry,
```
AurelienFT, line 908: In documentation of the API this need to specified that in case of that the value is not in active history it return the same as final and not None.

massa-execution-worker/src/execution.rs
```diff
@@ -293,6 +296,44 @@ impl ExecutionState {
...
+    pub fn fetch_active_history_data_entry(
+        &self,
+        slot: Slot,
+        addr: &Address,
+        key: &Hash,
+    ) -> HistorySearchResult<Vec<u8>> {
+        self.verify_active_slot(slot);
+
+        if let Some(n) = self.get_active_index(slot) {
+            let iter = self.active_history.iter().rev().skip(n);
```
AurelienFT, line 310: I think the rev and skip has been inverted.
Example : 
We send slot 6 in this context : 
```
Slots: 5 6 7 8 9 10
Index: 0 1 2 3 4 5
              ^

Rev : 10 9 8 7 6 5
Skip : 9 8 7 6 5
```
but I think you want to have : 
```
Skip : 6 7 8 9 10
Rev : 10 9 8 7 6
```
The behavior is already like that in `fetch_active_history_balance`

AurelienFT (commented):
I think we should add documentation in the API doc for this endpoint and detail it because the return type is a bit tricky.
The API doc is at : `doc/api.rst`

massa-execution-worker/src/execution.rs
Eitu33: This has been already thought through, see https://github.com/massalabs/massa/pull/2590#discussion_r883771386

AurelienFT: Yes and on this PR we changed to `self.active_history.iter().skip(n).rev();` here you have : `self.active_history.iter().rev().skip(n);`

massa-execution-exports/src/test_exports/mock.rs
Eitu33: Done.

massa-execution-worker/src/execution.rs
Eitu33: Done.

massa-execution-worker/src/execution.rs
Eitu33: Done.

docs/technical-doc/api.rst
```diff
@@ -342,6 +342,34 @@ Get the block graph within the specified time interval.
         },
     ];
 
+`get_datastore_entry`
+--------------------
+
+Get a data entry both at the latest final and active executed slots.
+
+If an existing final entry is:
```
AurelienFT, line 350: For me it's unclear that you speak of the field `active_value` of the result type. I understand that you speak of the whole return type. Maybe use the field name : `final_value` could be more informative

```suggestion
If an existing final entry (`final_value`) is:
```

docs/technical-doc/api.rst
```diff
@@ -342,6 +342,34 @@ Get the block graph within the specified time interval.
...
+* found in the active history, it will return its final value
+* deleted in the active history, it will return nothing
```
AurelienFT, line 352:
```suggestion
* deleted in the active history, it will return null in `active_value` field
```

AurelienFT (commented):
Thanks for the changes. Just some unclear documentations otherwise I'm good with this :) ! Well done.

Eitu33:
bors merge

PR #2676, opened 2022-06-09
Feature/change serialization bootstrap structs

AurelienFT:
This PR includes introduce differents changes in serialization:

- Now the buffer for the serialize is passed as parameter to avoid re-allocating buffers for no reason.
- Every deserializer/serializer have doc tests to ensure it works (already spotted a bug thanks to it in version [here](https://github.com/massalabs/massa/pull/2676/commits/f63e366bf352251f2971aebc7c4745661774f00d))
- Add documentation on each structure of deserializer/serializer
- Change new deserializer and serializer for all the bootstrap structures to use the new format except for `BoostrappableGraph` that use `Block` and so it would be too much in this PR.


Eitu33:
Overall looks good to me 😄 👍

Eitu33 (approved):

PR #2688, opened 2022-06-10
Sign without random auxiliary data

Eitu33:


damip:
let's not merge this yet, we need to discuss a bit more :)

damip:
```rust
use secp256k1::{schnorr, Message, SECP256K1};
use std::{convert::TryInto, str::FromStr};

fn main() {

    use secp256k1::rand::rngs::OsRng;
    let mut rng = OsRng::new().expect("OsRng");
    let private_key = secp256k1::KeyPair::from_secret_key(
        SECP256K1,
        secp256k1::SecretKey::new(&mut rng),
    );

    let message = Message::from_slice("11111111111111111111111111111111".as_bytes()).unwrap();
    println!("{:#?}", SECP256K1.sign_schnorr(&message, &private_key));
    println!("{:#?}", SECP256K1.sign_schnorr(&message, &private_key));
}

/*
Prints:
    Signature(1403c2329c34920d3448ecf5740581da9334c3660eef8c8a8003e5d3b91150ec79bbb4684688d43e6cf95e5c01eee0e454686d27442adfb70f0aebdaa9228a38)
    Signature(c0eaaf8efd145dba0dffb86251e59422d12d18f78dc33e6ea8fb304f3ae46d4927b37d39908bc0c905470697a15705c2bdf84b1c2d4c4145b32c3264ff07a8aa)

=> non-deterministic !
*/

```

damip:
Remarks:
* if we decide that signatures are taken into account when hashing blocks, ops, etc...:
  * we need to compute the hash of the signed contents to sign them, then another hash of the signed contents + signature in order to get the hash of the object
  * we absolutely need to make sure that signatures are not malleable, otherwise attackers could "clone" existing transactions by changing their signature in order to execute them multiple times under a new hash
    * https://en.bitcoin.it/wiki/BIP_0340#Motivation
    * https://en.bitcoin.it/wiki/BIP_0062
    * https://en.bitcoin.it/wiki/BIP_0146
* if the signatures are not included in the hashing process
  * we only compute the hash of the signed contents, and this hash is also the hash of the object => more efficient
  * the same object could be signed by multiple different signatures/private keys. Including the sender pubkey in the object's signed contents solve the problem.
  * signature malleability is probably not a problem anymore
  * warning: signatures stop being consensual
    * Apart from verification, other read operations on signatures must not be used directly in any deterministic process
    * Smart contracts should not be given read access to block/op signatures
* if the signatures are non-deterministic:
  * it allows for the same operation to be sent multiple times in a row by the sender without an explicit nonce, because each one will have a different signature, and therefore a different hash
  * how do we define genesis block hashes ?
  * smart contracts should use a different signature algorithm that is fully deterministic

Eitu33:
We could just use both signatures (random aux data and none), for the corresponding use cases.

AurelienFT:
> ```rust
> secp256k1
> ```

Using the change that made @Eitu33  in this PR using : `sign_schnorr_no_aux_rand` the two signatures are the same.

damip:
> > ```rust
> > secp256k1
> > ```
> 
> Using the change that made @Eitu33 in this PR using : `sign_schnorr_no_aux_rand` the two signatures are the same.

That comes with dangers :

https://doxygen.bitcoincore.org/schnorrsig_2main__impl_8h.html#a221c526a09b8faa406e15824eb92cf1b

```
aux_rand32: 32 bytes of fresh randomness. While recommended to provide this, it is only supplemental to security and can be NULL. A NULL argument is treated the same as an all-zero one. See BIP-340 "Default Signing" for a full explanation of this argument and for guidance if randomness is expensive. 
```

Super important description: https://en.bitcoin.it/wiki/BIP_0340#Default_Signing

damip:
Let's focus on https://github.com/massalabs/massa/discussions/2690 to solve this dilemma. Closing for now.

PR #2693, opened 2022-06-12
remove submodules

damip:
...because they are not useful anymore.

we have also disabled the stale bot and would like to keep merged branches and not delete them

Also some extra cleanups


AurelienFT (approved):

AurelienFT:
bors merge

PR #2699, opened 2022-06-13
Ledger unif doc

damip:


sebastien-forestier:
Ok for me, just I think we should set n=1 (number of slashed rolls), with no need for a parameter nor a general case implementation for all n

damip:
> Ok for me, just I think we should set n=1 (number of slashed rolls), with no need for a parameter nor a general case implementation for all n

done

damip:
@adrien-zinger @eitu I made some minor updates

Eitu33:
@damip Changes look good to me, just a curiosity:
```
* if S was the last slot of cycle C
  * if the speculative production stats of an address are not satisfactory (> 70% miss):
```
Is there a reason for 70%?

damip:
> @damip Changes look good to me, just a curiosity:
> 
> ```
> * if S was the last slot of cycle C
>   * if the speculative production stats of an address are not satisfactory (> 70% miss):
> ```
> 
> Is there a reason for 70%?

🎩

damip:
@adrien-zinger @Eitu33 can you format and move this doc in a way that makes sense ? 
I'm not sure what's best but here are the possibilities:
* add it as a doc comment in the PoS module
* add it as a .md file in the PoS module
* properly add it to the docs/ folder so that it can be deployed on readthedocs, probably in some "tech specs" subfolder

qdrn:
I think we should keep the rtd separate from the "tech spec" doc as it's not the same target (core dev vs builders). Still it would make sense to start organising the "tech spec" doc (either separate documents or doc generated from doc comments) to help people dive into the code.

PR #2703, opened 2022-06-15
Wrap object in structure that contains all needed informations

AurelienFT:
## Short explanation

The goal of this PR is reduce the computation of the object ID and the serialize version of an object that is sent through the network.

Some schemas for the example of an operation:
Deserialization
Serialization

## Detailed explanation 

The `Signed` structure has been renamed to `Wrapped` and now contains `id`, `serialized_data`, `creator_public_key`, `creator_address`. This structure have a serializer and a deserializer

The serializer encode only the public key and the serialized data.
The deserializer deduces the others fields thanks to this data.

This PR implement new serializer and deserializer for Block, BlockHeader, Operation, Endorsement because they are needed for the Serializer and Deserializer of WrappedXXX structure.

The serialization and deserialization can be customized. This is needed for the `Block` because we don't want to encode to much information as a lot are shared his `header` field.

Everywhere in the code (except at the creation) we must have `WrappedXXXX` structure because it prevent to re-compute a lot of data.

To create this type of structures you need to use the constructor that is implemented for all objects that are passed in the network (currently Block,BlockHeader,Operation,Endorsement).

This PR also implement officially the RFC #2690 even if it was already the behavior in the current codebase.

This PR simplify a lot of code since the serialized version is directly in the object (especially in storage).

This PR also removes all the constraints on all serializers on the code because the verification should be made only on deserialization.


massa-models/src/block.rs
```diff
@@ -583,28 +583,27 @@ mod test {
...
-                    .1,
+                    .unwrap(),
                     Wrapped::new_wrapped(
                         Endorsement {
-                            sender_public_key: public_key,
                             slot: Slot::new(4, 0),
                             index: 3,
                             endorsed_block: BlockId(Hash::compute_from("blk2".as_bytes())),
                         },
+                        EndorsementSerializer::new(),
```
damip, line 600: could we share the EndorsementSerializer created above instead of creating a new one ?

damip:
looks OK for now :)

AurelienFT:
> looks OK for now :)

Thank you for taking a look :)

AurelienFT:
@massalabs/core-team This PR is ready for the review. I have written explanations in the PR description above. Also I ping all of you because this PR modify a lot of things across the code and each of you knows well each part of the code.

If you have any question to not hesitate.

gterzian (approved):
Looks good to me, I'm just wondering if we can hide the instantiations of the (de-) serializers inside the implementation of `Wrapped`.

Eitu33 (approved):
Very nice changes 👍

massa-models/src/block.rs
```diff
@@ -106,34 +124,194 @@ impl BlockId {
...
+}
+
+impl BlockDeserializer {
+    /// Creates a new `BlockDeserializer`
+    pub fn new() -> Self {
+        BlockDeserializer {
+            header_deserializer: WrappedDeserializer::new(BlockHeaderDeserializer::new()),
+            operation_deserializer: WrappedDeserializer::new(OperationDeserializer::new()),
+            u32_deserializer: U32VarIntDeserializer::new(
+                Included(0),
```
damip, line 251: why use 2 bounds ? why not a Range ? https://doc.rust-lang.org/std/ops/struct.Range.html

Is it to ensure that we always specify bounds ?

massa-models/src/endorsement.rs
```diff
@@ -119,128 +123,131 @@ pub struct Endorsement {
...
+impl EndorsementDeserializer {
+    /// Creates a new `EndorsementDeserializer`
+    pub fn new(endorsement_count: u32) -> Self {
+        #[cfg(feature = "sandbox")]
+        let thread_count = *THREAD_COUNT;
+        #[cfg(not(feature = "sandbox"))]
+        let thread_count = THREAD_COUNT;
+        EndorsementDeserializer {
+            slot_deserializer: SlotDeserializer::new(
+                (Included(1), Included(u64::MAX)),
```
damip, line 178: I think we should start from 0 included here

massa-models/src/block.rs
```diff
@@ -207,36 +386,178 @@ impl std::fmt::Display for Block {
...
+    /// Creates a new `BlockHeaderDeserializer`
+    pub fn new() -> Self {
+        #[cfg(feature = "sandbox")]
+        let thread_count = *THREAD_COUNT;
+        #[cfg(not(feature = "sandbox"))]
+        let thread_count = THREAD_COUNT;
+        Self {
+            slot_deserializer: SlotDeserializer::new(
+                (Included(0), Included(u64::MAX)),
+                (Included(0), Included(thread_count)),
```
damip, line 486: thread_count is excluded

massa-models/src/endorsement.rs
```diff
@@ -119,128 +123,131 @@ pub struct Endorsement {
...
+                (Included(0), Included(thread_count)),
```
damip, line 179: thread count is excluded

damip (changes requested):
Beautiful and extensive work !
some (security-critical) nits to correct  :)

massa-models/src/endorsement.rs
AurelienFT: Ok I will make the change but the old code had : 
```rust
if slot.period == 0 {
            return Err(ModelsError::DeserializeError(
                "the target period of an endorsement cannot be 0".into(),
            ));
}
```

massa-models/src/block.rs
AurelienFT: Done.

massa-models/src/endorsement.rs
AurelienFT: Done.

damip (approved):
Amazing work

AurelienFT:
bors merge

PR #2708, opened 2022-06-16
Propagate sandbox to massa-proof-of-stake-exports

adrien-zinger:
Very small fix :)

### DIFF
--- a/massa-bootstrap/Cargo.toml
+++ b/massa-bootstrap/Cargo.toml
@@ -61,4 +61,10 @@ testing = [
     "massa_ledger_worker/testing",
     "massa_async_pool/testing",
 ]
-sandbox = ["massa_async_pool/sandbox", "massa_final_state/sandbox"]
+sandbox = [
+    "massa_async_pool/sandbox",
+    "massa_consensus_exports/sandbox",
+    "massa_final_state/sandbox",
+    "massa_models/sandbox",
+    "massa_proof_of_stake_exports/sandbox",
+]
--- a/massa-proof-of-stake-exports/Cargo.toml
+++ b/massa-proof-of-stake-exports/Cargo.toml
@@ -28,3 +28,4 @@ massa_serialization = { path = "../massa-serialization" }
 # for more information on what are the following features used for, see the cargo.toml at workspace level
 [features]
 instrument = ["tokio/tracing", "massa_models/instrument"]
+sandbox = ["massa_models/sandbox"]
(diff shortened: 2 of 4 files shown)


massa-proof-of-stake-exports/Cargo.toml
```diff
@@ -28,3 +28,4 @@ massa_serialization = { path = "../massa-serialization" }
 # for more information on what are the following features used for, see the cargo.toml at workspace level
 [features]
 instrument = ["tokio/tracing", "massa_models/instrument"]
+sandbox = []
```
AurelienFT, line 31: Need to propagate here also.
```suggestion
sandbox = ["massa_models/sandbox"]
```

massa-bootstrap/Cargo.toml
```diff
@@ -61,4 +61,8 @@ testing = [
     "massa_ledger_worker/testing",
     "massa_async_pool/testing",
 ]
-sandbox = ["massa_async_pool/sandbox", "massa_final_state/sandbox"]
+sandbox = [
+    "massa_async_pool/sandbox",
+    "massa_final_state/sandbox",
+    "massa_proof_of_stake_exports/sandbox",
```
AurelienFT, line 67: Need to add all dependency that use sandbox such as massa_models

AurelienFT (commented):
Also add on dependencies of massa-node. There is a lot of crate in dependency  of massa-node that need it.

massa-proof-of-stake-exports/Cargo.toml
adrien-zinger: done

massa-bootstrap/Cargo.toml
adrien-zinger: done

AurelienFT:
In case you didn't see because the layout of GitHub isn't very good : https://github.com/massalabs/massa/pull/2708#pullrequestreview-1008974829

adrien-zinger:
> In case you didn't see because the layout of GitHub isn't very good : [#2708 (review)](https://github.com/massalabs/massa/pull/2708#pullrequestreview-1008974829)

Done :wink: . Nevertheless, it's good to know that it's a bit useless since the feature is already activated by a foreign module. You can look at the [feature unification spec](https://doc.rust-lang.org/cargo/reference/features.html#feature-unification). It's a weird behavior but... better that the npm one :rofl: 

The problem was just that nobody was activating the feature `sandbox` of "massa_proof_of_stake_exports".

AurelienFT:
> Done 😉 . Nevertheless, it's good to know that it's a bit useless since the feature is already activated by a foreign module. 

Yes but for example you build in the repository of massa-bootstrap the other crates are built as a dependency and then need to have the features passed otherwise they don't have it enable.

AurelienFT (approved):

AurelienFT:
bors merge

AurelienFT:
bors merge

PR #2709, opened 2022-06-16
Reduce operation id length in the inter-nodes communications 1/2

adrien-zinger:
That PR proposes a first step in order to send only a part of the operation's ids when we declare and ask them in our network.

The modules impacted from that proposal are only `massa-storage` and `massa-models`. It adds a _prefix_ and _suffix_ logic in both.

The `operations` structure is modified as followed in order to retrieve operations from the _operation id truncated hash_. The motivation is mainly to not store the relation `id-trucatedId` in a foreign structure that would have the same time-overhead but a bigger space-overhead. Another good solution to avoid the waste of space would be to use a `BTreeMap` instead in order to iterate through the hashes (theorically :rofl:), but it would probably be slower... I don't know.

```rust
operations: Arc<RwLock<Map<OperationPrefixId, Map<OperationSuffixId, StoredOperation>>>>,
```

Next step: modify protocol and network to apply the solution.

After a little calculation, I found that a foreign relation table to manage in protocol would have another time overhead to regularly prune it and `1.6byte` / entry as space overhead.

### DIFF
--- a/massa-models/src/operation.rs
+++ b/massa-models/src/operation.rs
@@ -27,7 +27,7 @@ const OPERATION_ID_STRING_PREFIX: &str = "OPE";
 #[derive(Clone, Copy, Eq, PartialEq, Ord, PartialOrd, Hash, Serialize, Deserialize)]
 pub struct OperationId(Hash);
 
-/// left part of the operation id hash stored in a vector of size [OPERATION_ID_PREFIX_SIZE_BYTES]
+/// Left part of the operation id hash stored in a vector of size [OPERATION_ID_PREFIX_SIZE_BYTES]
 #[derive(Clone, Eq, PartialEq, Ord, PartialOrd, Hash, Serialize, Deserialize)]
 pub struct OperationPrefixId(Vec<u8>);
 
@@ -153,8 +153,7 @@ impl OperationId {
         ))
     }
 
-    /// split into a tuple of [OperationPrefixId] and [OperationSuffixId]
-    ///
+    /// Split into a tuple of [OperationPrefixId] and [OperationSuffixId]
     pub fn split(&self) -> (OperationPrefixId, OperationSuffixId) {
         (
             OperationPrefixId(self.0.to_bytes()[..OPERATION_ID_PREFIX_SIZE_BYTES].to_vec()),
@@ -162,7 +161,7 @@ impl OperationId {
         )
     }
 
-    /// split into a tuple of [OperationPrefixId] and [OperationSuffixId]
+    /// Split into a tuple of [OperationPrefixId] and [OperationSuffixId]
     pub fn into_split(self) -> (OperationPrefixId, OperationSuffixId) {
         self.split()
     }
@@ -182,7 +181,7 @@ impl OperationPrefixId {
         }
         let mut id = self.0.clone();
         id.extend(&suffix.0);
-        // the following code is safe until we correctly check the size in the if
+        // the following code is safe because we correctly check the size in the if
         // guard and return early an error if the transmutation would fail.
         let data: &[u8; OPERATION_ID_SIZE_BYTES] = unsafe { transmute_copy(&id) };
         Ok(OperationId::from_bytes(data))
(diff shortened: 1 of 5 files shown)


massa-models/src/operation.rs
```diff
@@ -24,6 +27,14 @@ const OPERATION_ID_STRING_PREFIX: &str = "OPE";
 #[derive(Clone, Copy, Eq, PartialEq, Ord, PartialOrd, Hash, Serialize, Deserialize)]
 pub struct OperationId(Hash);
 
+/// left part of the operation id hash stored in a vector of size [OPERATION_ID_PREFIX_SIZE_BYTES]
```
gterzian, line 30: nit: upper case `Left`

massa-models/src/operation.rs
```diff
@@ -105,6 +152,41 @@ impl OperationId {
...
+                "{}:{}",
+                self.0.len() + suffix.0.len(),
+                OPERATION_ID_SIZE_BYTES
+            );
+            return Err(ModelsError::OperationPrefixJoinError);
+        }
+        let mut id = self.0.clone();
+        id.extend(&suffix.0);
+        // the following code is safe until we correctly check the size in the if
+        // guard and return early an error if the transmutation would fail.
```
gterzian, line 186: What do you mean with "until"? Isn't the check made just above? Also we don't seem to return an error if the transmutation would fail?

massa-storage/src/lib.rs
```diff
@@ -113,7 +152,11 @@ impl Storage {
             "operation_id": operation_id
         });
         let operations = self.operations.read();
-        f(&operations.get(operation_id))
+        let (prefix, suffix) = operation_id.split();
+        f(&match operations.get(&prefix) {
```
gterzian, line 156: use `operations.get(&prefix).map(|ops| ops.get(&prefix))`?

massa-storage/src/lib.rs
```diff
@@ -139,7 +189,17 @@ impl Storage {
         });
         let mut operations = self.operations.write();
         for id in operation_ids {
-            operations.remove(id);
+            let (prefix, suffix) = id.split();
+            let len = match operations.get_mut(&prefix) {
+                Some(m) => {
+                    m.remove(&suffix);
+                    m.len()
```
gterzian, line 196: use `is_empty`?

massa-storage/src/lib.rs
```diff
@@ -139,7 +189,17 @@ impl Storage {
         });
         let mut operations = self.operations.write();
         for id in operation_ids {
-            operations.remove(id);
+            let (prefix, suffix) = id.split();
+            let len = match operations.get_mut(&prefix) {
```
gterzian, line 193: maybe use `entry` so as to remove it in the same operation if necessary?

gterzian (commented):
Looks, with some nits.

gterzian:
By the way I have another question:

Why does `massa-storage` have to concern itself with the new structure? When we use storage, will we not always have the full Id at our disposal?

massa-storage/src/lib.rs
adrien-zinger: Good :+1: I'll use `operations.get(&prefix).and_then(|ops| ops.get(&prefix))` to `flatten` the nested options

massa-models/src/operation.rs
adrien-zinger: I want to say "because". Technically, I guess that the transmutation here would fail only if the vector has a different size, the value inside have the same size (`u8`). Is it right?

massa-storage/src/lib.rs
adrien-zinger: done

massa-storage/src/lib.rs
adrien-zinger: I used an entry instead

massa-models/src/operation.rs
adrien-zinger: done

adrien-zinger:
I'm closing that PR because I'll remove the storage modification and the review history will be weird :smiling_face_with_tear: goto #2712

PR #2710, opened 2022-06-16
Wallet encryption

Eitu33:


massa-client/src/main.rs
```diff
@@ -51,13 +52,33 @@ struct Args {
...
+    /// Wallet password
+    password: Option<String>,
 }
 
 #[derive(Serialize)]
 struct JsonError {
     error: String,
 }
 
+fn retrieve_password(wallet_path: &PathBuf) -> String {
```
damip, line 65: maybe rename it ask_password

massa-wallet/src/lib.rs
```diff
@@ -5,46 +5,73 @@
...
+    derive_public_key, generate_random_private_key, sign, PrivateKey, PublicKey,
+};
 use serde::{Deserialize, Serialize};
 use std::path::PathBuf;
 
 mod error;
 
+/// Nonce slice used for cypher
+/// TODO: change this? not sure if we must
+pub const NONCE_SLICE: &[u8] = b"unique nonce";
```
damip, line 27: This will break the security of the encryption (see documentation)

* the nonce should be a random 96 bits (12 bytes) without any structure (random bytes)
* the nonce should never be reused (hence the name: N ONCE = number used once)

This means that:
* a new 12-byte random nonce should be generated every time the wallet is encrypted
* the generated nonce should be saved inside the wallet file on disk so that it can then be loaded for decryption

Here is an example: https://github.com/RCasatta/firma/blob/master/lib/src/offline/decrypt.rs#L51-L84

damip (changes requested):
Nice work !

Let's just correct the nonce-related security failure, and also make sure we try the feature in real conditions to see if it works.

massa-wallet/src/lib.rs
Eitu33: Done 👍

massa-wallet/src/lib.rs
```diff
@@ -142,13 +140,18 @@ impl Wallet {
...
+        let nonce = Nonce::from_slice(&bytes);
         let keys =
             serde_json::to_string(&self.keys.iter().map(|(_, (_, pk))| *pk).collect::<Vec<_>>())?;
         let text = cipher
             .encrypt(nonce, keys.as_ref())
             .map_err(|e| WalletError::EncryptionError(e.to_string()))?;
-        std::fs::write(&self.wallet_path, text)?;
+        let mut content = bytes.to_vec();
+        content.extend(b":");
+        content.extend(text);
```
damip: note: technically the variable name is wrong (it's not text but encrypted bytes)

massa-wallet/src/lib.rs
```diff
@@ -142,13 +140,18 @@ impl Wallet {
         let cipher = Aes256GcmSiv::new(Key::from_slice(
             Hash::compute_from(self.password.as_bytes()).to_bytes(),
         ));
-        let nonce = Nonce::from_slice(NONCE_SLICE);
+        let mut bytes = [0u8; 12];
```
damip: let's call the variable "nonce"

massa-wallet/src/lib.rs
```diff
@@ -142,13 +140,18 @@ impl Wallet {
...
+        thread_rng().fill_bytes(&mut bytes);
+        let nonce = Nonce::from_slice(&bytes);
         let keys =
             serde_json::to_string(&self.keys.iter().map(|(_, (_, pk))| *pk).collect::<Vec<_>>())?;
         let text = cipher
             .encrypt(nonce, keys.as_ref())
             .map_err(|e| WalletError::EncryptionError(e.to_string()))?;
-        std::fs::write(&self.wallet_path, text)?;
+        let mut content = bytes.to_vec();
+        content.extend(b":");
```
damip: no need for the separator since the nonce is always 12 bytes

massa-wallet/src/lib.rs
```diff
@@ -40,15 +37,16 @@ pub struct Wallet {
...
     pub fn new(path: PathBuf, password: String) -> Result<Wallet, WalletError> {
-        let content = std::fs::read(&path)?;
+        let content = &std::fs::read(&path)?[..];
         let priv_keys = if !content.is_empty() {
             // If wallet contains data decipher it
             let cipher = Aes256GcmSiv::new(Key::from_slice(
                 Hash::compute_from(password.as_bytes()).to_bytes(),
             ));
-            let nonce = Nonce::from_slice(NONCE_SLICE);
+            let mut iter = content.split(|c| c == &b':');
```
damip: no need for a separator since the nonce is always 12 bytes

You can just set NONCE_SIZE = 12 then use it as `let nonce = Nonce::from_slice(content.get(0..NONCE_SIZE)?);`

and then `decrypt(content.get([NONCE_SIZE..])?)`

massa-wallet/src/lib.rs
damip: let's call it nonce_bytes

damip (approved):
LGTM with nits

massa-client/src/main.rs
```diff
@@ -51,13 +52,33 @@ struct Args {
...
+        Password::new()
+            .with_prompt("Enter wallet password")
+            .interact()
+            .expect("Input error")
+    } else {
+        let pwd = Password::new()
+            .with_prompt("Enter new password for wallet")
+            .with_confirmation("Confirm password", "Passwords mismatching")
+            .interact()
+            .expect("Input error");
```
damip, line 76: this will crash the whole program... maybe just reask for password :)

damip, line 76: * the panic message is not explicit enough
* maybe just reask for the password instead of panicking

massa-wallet/src/lib.rs
```diff
@@ -24,27 +28,39 @@ pub struct Wallet {
...
+    pub password: String,
 }
 
 impl Wallet {
     /// Generates a new wallet initialized with the provided file content
-    pub fn new(path: PathBuf) -> Result<Wallet, WalletError> {
-        let keys = if path.is_file() {
-            serde_json::from_str::<Vec<PrivateKey>>(&std::fs::read_to_string(&path)?)?
+    pub fn new(path: PathBuf, password: String) -> Result<Wallet, WalletError> {
+        let content = &std::fs::read(&path)?[..];
```
damip, line 38: the read() will fail if the file does not exist and quit the whole function

massa-wallet/src/lib.rs
```diff
@@ -24,27 +28,39 @@ pub struct Wallet {
...
+        let priv_keys = if !content.is_empty() {
```
damip, line 39: content.is_empty() is not a good check for file existence

massa-cipher/src/decrypt.rs
```diff
@@ -0,0 +1,19 @@
...
+use crate::error::CipherError;
+
+pub fn decrypt(password: &str, data: &[u8]) -> Result<Vec<u8>, CipherError> {
+    let cipher = Aes256GcmSiv::new(Key::from_slice(
+        Hash::compute_from(password.as_bytes()).to_bytes(),
+    ));
+    let mut iter = data.split(|c| c == &b':');
+    let nonce = Nonce::from_slice(iter.next().expect("Missing nonce"));
+    let text = cipher
+        .decrypt(nonce, iter.next().expect("Missing content"))
```
damip, line 16: return an error instead of panicking (since decrypt returns a Result)

massa-cipher/src/decrypt.rs
```diff
@@ -0,0 +1,19 @@
...
+use massa_hash::Hash;
+
+use crate::error::CipherError;
+
+pub fn decrypt(password: &str, data: &[u8]) -> Result<Vec<u8>, CipherError> {
+    let cipher = Aes256GcmSiv::new(Key::from_slice(
+        Hash::compute_from(password.as_bytes()).to_bytes(),
+    ));
+    let mut iter = data.split(|c| c == &b':');
+    let nonce = Nonce::from_slice(iter.next().expect("Missing nonce"));
```
damip, line 14: return an error instead of panicking (since decrypt returns a Result)

massa-cipher/src/decrypt.rs
```diff
@@ -0,0 +1,19 @@
...
+use aes_gcm_siv::{Aes256GcmSiv, Key, Nonce};
+use massa_hash::Hash;
+
+use crate::error::CipherError;
+
+pub fn decrypt(password: &str, data: &[u8]) -> Result<Vec<u8>, CipherError> {
+    let cipher = Aes256GcmSiv::new(Key::from_slice(
+        Hash::compute_from(password.as_bytes()).to_bytes(),
+    ));
+    let mut iter = data.split(|c| c == &b':');
```
damip, line 13: no need for a separator, nonce is always 12 bytes (also there should be a NONCE_SIZE=12 constant)

massa-cipher/src/encrypt.rs
```diff
@@ -0,0 +1,24 @@
...
+use massa_hash::Hash;
+use rand::{thread_rng, RngCore};
+
+use crate::error::CipherError;
+
+pub fn encrypt(password: &str, data: &[u8]) -> Result<Vec<u8>, CipherError> {
+    let cipher = Aes256GcmSiv::new(Key::from_slice(
+        Hash::compute_from(password.as_bytes()).to_bytes(),
+    ));
+    let mut bytes = [0u8; 12];
```
damip, line 14: use the NONCE_SIZE = 12 constant

massa-cipher/src/encrypt.rs
```diff
@@ -0,0 +1,24 @@
...
+    thread_rng().fill_bytes(&mut bytes);
+    let nonce = Nonce::from_slice(&bytes);
+    let text = cipher
+        .encrypt(nonce, data.as_ref())
+        .map_err(|e| CipherError::EncryptionError(e.to_string()))?;
+    let mut content = bytes.to_vec();
+    content.extend(b":");
```
damip, line 21: no need for a separator

massa-cipher/src/encrypt.rs
damip: also let's call it nonce_bytes

massa-cipher/src/encrypt.rs
```diff
@@ -0,0 +1,24 @@
...
+use crate::error::CipherError;
+
+pub fn encrypt(password: &str, data: &[u8]) -> Result<Vec<u8>, CipherError> {
+    let cipher = Aes256GcmSiv::new(Key::from_slice(
+        Hash::compute_from(password.as_bytes()).to_bytes(),
+    ));
+    let mut bytes = [0u8; 12];
+    thread_rng().fill_bytes(&mut bytes);
+    let nonce = Nonce::from_slice(&bytes);
+    let text = cipher
```
damip, line 17: let's rename it encrypted_bytes

massa-wallet/src/lib.rs
```diff
@@ -24,27 +28,39 @@ pub struct Wallet {
...
+            // If wallet contains data decipher it
+            let decrypted_content = decrypt(&password, content)?;
+            serde_json::from_slice::<Vec<PrivateKey>>(&decrypted_content[..])?
         } else {
-            Vec::new()
+            // If wallet is empty add a new private key
+            vec![generate_random_private_key()]
```
damip, line 45: this is a new behavior ?

massa-wallet/src/lib.rs
```diff
@@ -24,27 +28,39 @@ pub struct Wallet {
...
             })
             .collect::<Result<Map<Address, _>, WalletError>>()?;
-        Ok(Wallet {
+        let wallet = Wallet {
             keys,
             wallet_path: path,
-        })
+            password,
+        };
+        if content.is_empty() {
```
damip, line 59: I don't get this part

damip (changes requested):
I found a few problems

massa-wallet/src/lib.rs
Eitu33: `Wallet::new(path, password)?` is only called in `massa-client/main.rs` right after the file was tested and created if it does not exist by `ask_password()`

massa-wallet/src/lib.rs
Eitu33: this is not for file existence but only to see if the file is empty or not

massa-wallet/src/lib.rs
Eitu33: Sole purpose was the case where user does not create a key in the first time he launches the client, but I'm currently changing this to a new behaviour.

massa-wallet/src/lib.rs
Eitu33: Also this is linked to https://github.com/massalabs/massa/pull/2710#discussion_r902292731, currently changing that.

massa-cipher/src/decrypt.rs
Eitu33: Done 👍

Eitu33:
@damip everything has been treated

massa-cipher/src/decrypt.rs
```diff
@@ -0,0 +1,26 @@
...
+use crate::constants::NONCE_SIZE;
+use crate::error::CipherError;
+
+pub fn decrypt(password: &str, data: &[u8]) -> Result<Vec<u8>, CipherError> {
+    let cipher = Aes256GcmSiv::new(Key::from_slice(
+        Hash::compute_from(password.as_bytes()).to_bytes(),
+    ));
+    let nonce = Nonce::from_slice(
+        data.get(..NONCE_SIZE)
+            .ok_or(CipherError::DecryptionError("Missing nonce".to_string()))?,
```
damip, line 16: use ok_or_else

damip, line 16: use ok_or_else

massa-cipher/src/decrypt.rs
```diff
@@ -0,0 +1,26 @@
...
+    );
+    let decrypted_bytes = cipher
+        .decrypt(
+            nonce,
+            data.get(NONCE_SIZE..)
+                .ok_or(CipherError::DecryptionError("Missing content".to_string()))?,
```
damip, line 22: use ok_or_else

massa-cipher/src/encrypt.rs
```diff
@@ -0,0 +1,24 @@
...
+
+use aes_gcm_siv::aead::{Aead, NewAead};
+use aes_gcm_siv::{Aes256GcmSiv, Key, Nonce};
+use massa_hash::Hash;
+use rand::{thread_rng, RngCore};
+
+use crate::constants::NONCE_SIZE;
+use crate::error::CipherError;
+
+pub fn encrypt(password: &str, data: &[u8]) -> Result<Vec<u8>, CipherError> {
```
damip, line 11: add documentation

massa-wallet/src/lib.rs
damip: see comment above for an alternative way of doing it

damip:
@Eitu33

massa-node/base_config/config.toml
```diff
@@ -70,6 +70,7 @@
     pos_draw_cached_cycles = 10
 
     # path to your staking private keys (not encrypted)
+    # HERE 1
```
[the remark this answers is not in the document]
Eitu33: Old comment push by mistake.

damip (approved):
good job !

Eitu33:
bors merge

PR #2727, opened 2022-06-23
Add skeleton of Selector worker & PoS'

adrien-zinger:
Starting the development of the ledger unification.


massa-pos-exports/src/controller_traits.rs
```diff
@@ -0,0 +1,51 @@
...
+use anyhow::Result;
+use massa_models::Address;
+use massa_models::Slot;
+
+/// interface that communicates with the selector worker thread
+pub trait SelectorController: Send + Sync {
+    /// Feed cycle to the selector
+    ///
+    /// # Arguments
+    /// * `cycle_info`: give or regive a cycle info for a background
```
gterzian, line 17: `give or regive` meaning `idem-potently`?

massa-pos-exports/src/controller_traits.rs
```diff
@@ -0,0 +1,51 @@
...
+    fn get_selection(&self, slot: Slot) -> Result<Selection>;
+
+    /// Get [Address] of the selected block producer for a given slot
+    /// # Arguments
+    /// * `slot`: target slot of the selection
+    fn get_producer(&self, slot: Slot) -> Result<Address>;
+
+    /// Returns a boxed clone of self.
+    /// Useful to allow cloning `Box<dyn SelectorController>`.
+    fn clone_box(&self) -> Box<dyn SelectorController>;
```
gterzian, line 33: It could be cleaner to hide this method from the trait(since it is more about the "infra" behind the trait).

You can do that with another trait called for example `SelectorController` like:

```
pub trait SelectorControllerClone {
    fn clone_box(&self) -> Box<dyn SelectorController>;
}
```

and make that a contraint on `SelectorController` itself. 

Then the actual implementing struct can implement `clone_box`. 

The only difference is you remove `clone_box` from `SelectorController`, so that the trait is only about "how to use the controller", as oppose to the basic stuff like "how to clone it".

massa-pos-exports/src/types.rs
```diff
@@ -0,0 +1,63 @@
...
+    pub complete: bool,
+
+    /// number of rolls each staking address has
+    pub roll_counts: Map<Address, u64>,
+
+    /// random seed bits of all slots in the cycle so far
+    pub rng_seed: BitVec<Lsb0, u8>,
+
+    /// Per-address production statistics
+    pub production_stats: Map<Address, ProductionStats>,
```
gterzian, line 36: Maybe merge `roll_counts` and `production_stats` into one `per_address_info: Map<Address, AddressCycleInfo>` ?

massa-pos-worker/src/lib.rs
```diff
@@ -0,0 +1,16 @@
...
+
+use massa_pos_exports::CycleInfo;
+
+use parking_lot::{Condvar, Mutex};
+use std::{collections::VecDeque, sync::Arc};
+
+/// Same structure pointer that will be used by the selector controller and his
+/// thread. It will store all new CycleInfo declared by massa (in the
+/// Execution module) and will be used to compute the draws in background.
+pub(crate) type InputDataPtr = Arc<(Condvar, Mutex<VecDeque<CycleInfo>>)>;
```
gterzian, line 14: I find declaring new types to not add value, and only make you have to look-up what that type is. Why not just pass the `Arc<_>` around?

gterzian (commented):
Looks good, a few comments...

massa-pos-exports/src/pos_final_state_impl.rs
```diff
@@ -0,0 +1,24 @@
...
+    ///     extend roll_counts with changes.roll_changes
+    ///         delete all entries from roll_counts for which the roll count is zero
+    ///     add each element of changes.production_stats to the cycle's production_stats
+    /// for each changes.deferred_credits targeting cycle Ct:
+    ///     overwrite self.deferred_credits entries of cycle Ct in cycle_history with the ones from change
+    ///         remove entries for which Amount = 0
+    /// if slot S was the last of cycle C:
+    ///     set complete=true for cycle C in the history
+    ///     compute the seed hash and notifies the PoSDrawer for cycle C+3
+    pub fn settle_slot(&mut self, slot: Slot, changes: &PoSChanges) {}
```
Eitu33, line 23: I've been thinking about it and I'm not sure if `settle_slot` should be in `exports` or in `execution`, I will implement it in `execution` for now seeing as it is easier for development and will come back to this later.

Eitu33: Otherwise looks good to me, might change my mind as we progress though.

massa-pos-exports/src/types.rs
Eitu33: I agree with greg on that one, after starting the `PoSChanges` implementation in active history I've met that need. Will be using that on my PR for now but we should discuss this.

Eitu33: What I'm currently using:
```rust
#[derive(Default, Debug, Clone)]
pub struct PoSAddressInfo {
    /// set deferred credits indexed by target slot (can be set to 0 to cancel some, in case of slash)
    /// ordered structure to ensure slot iteration order is deterministic
    pub deferred_credits: BTreeMap<Slot, Amount>,

    /// updated production statistics
    pub production_stats: ProductionStats,

    /// new rolls counts
    pub roll_changes: u64,
}

#[derive(Default, Debug, Clone)]
pub struct PoSChanges {
    /// extra block seed bits added
    pub seed_bits: BitVec<Lsb0, u8>,

    /// addresses info
    pub addresses_info: Map<Address, PoSAddressInfo>,
}
```

Eitu33: FI: I rolled back to the current PoSChanges definition, but using `Map<Address, ProductionStats>` instead of `ProductionStats`. It was a mistake in the RFC, it has been updated.

damip:
Some feedback:

* blocks should be considered invalid if the total size of their operations exceeds `max_block_size`
* otherwise, as long as everything deserializes and signatures verify, blocks cannot be invalidated by their operations
* here is what is supposed to happen when executing a `block`:
  * set `remaining_block_gas = cfg.max_block_gas`
  * set `block_credits = cfg.constant_block_reward`
  * for each operation `op` in the block:
    * set `remaining_block_gas = remaining_block_gas.checked_sub(op.max_gas)`
      * if the substraction fails, ignore this operation and skip to the next one
    * set `op_fees = op.max_gas*op.gas_price + op.fee`
    * attempt to debit `op_fees + op.coins` from the SEQUENTIAL balance of the op's sender
      * if the debit fails, ignore this operation and skip to the next one
    * increment `block_credits += op_fees`
    * try to run the operation
      * if it fails:
        * rollback the execution effects
        * keep the `op_fees` debit
        * rollback the `op.coins` debit 
  * compute `block_credit_part = block_credits / (3*(1 + cfg.endorsement_count))` with normal integer division
  * set `block_remaining_credit = block_credits * (1 + block.endorsements.len()) / (1 + cfg.endorsement_count)` with a division that rounds up
  * for each included endorsement `endo` in the block:
    * credit `endo.creator` with `block_credit_part` sequential coins
    * credit `endo.endorsed_block.creator` with `block_credit_part` sequential coins
    * set `block_remaining_credit = block_remaining_credit.saturating_sub(block_credit_part * 2)`
  * credit `block.creator` with `block_remaining_credit` sequential coins

massa-pos-exports/src/pos_final_state_impl.rs
Eitu33: FI: settle_slot became apply_changes to respect the code conventions, settle_slot is now a function in execution worker.

damip:
Some major things that still need to be changed to make sure that invalid operations don't invalidate the block (we just want the invalid ops to be ignored):

* max block gas
  * this check currently invalidates the block and should be removed from protocol (https://github.com/massalabs/massa/blob/feature/add_pos_and_selector_skel/massa-protocol-worker/src/protocol_worker.rs#L1013-L1018) 
  * instead it should handled in execution by ignoring an operation if there is not enough block gas remaining to account for the max_gas of the operation 
* operation validity times
  *  this check currently invalidates the block and should be removed from protocol (https://github.com/massalabs/massa/blob/feature/add_pos_and_selector_skel/massa-protocol-worker/src/protocol_worker.rs#L1028-L1036)
  * this check should be handled in execution by ignoring ops for which the current slot is not in their validity range
* operations being in the right thread
  * this check currently invalidates the block and should be removed from protocol (https://github.com/massalabs/massa/blob/feature/add_pos_and_selector_skel/massa-protocol-worker/src/protocol_worker.rs#L1038-L1044)
  * this check should be handled by execution by ignoring operations whose creator address is not in the thread of the slot in which the block is being executed
* coins/rolls availability
  * this check currently invalidates the block and should be removed from massa-graph (https://github.com/massalabs/massa/blob/feature/add_pos_and_selector_skel/massa-graph/src/block_graph.rs#L2497-L2533)
  * instead, it should be performed inside execution, and:
    * operations whose creator has not enough sequential coins to pay fees/gas should be ignored
    * operations for which fees were paid successfully are always considered executed, even if they had no other effect (eg. not enough rolls to sell, bytecode panic etc...)
* operation reuse
  * the intra-block op reuse check currently invalidates the block and should be removed from protocol (https://github.com/massalabs/massa/blob/feature/add_pos_and_selector_skel/massa-protocol-worker/src/protocol_worker.rs#L1022-L1025)
  * the inter-block op reuse check currently invalidates the block and should be removed from massa-graph (https://github.com/massalabs/massa/blob/feature/add_pos_and_selector_skel/massa-graph/src/block_graph.rs#L2440-L2494)
  * both of those checks should be fused into the following one inside execution: ignore operations that have been executed previously. In other words:
    * ignore operations previously included in the current block or in previous blocks AND that have been executed (eg. the fees have been paid)
    *  if this condition is not respected and operations are always ignored if they were included previously (but not executed), block producers can easily censor operations by adding them beyond their max_block_gas

damip:
Solved most of my comments in https://github.com/massalabs/massa/pull/2727/commits/c449d202e130e340c652ac909d94dbc652d6d056

What is missing:
* [ ] make sure the ExecutedOps state is bootstrapped
* [ ] check the total size of the ops of each block in Protocol. If too big => block invalid
* [ ] credit the creator of the endorsed block with sequential coins (not only the creator of the endorsement)
* [ ] repair tests

gterzian (changes requested):
Just a small comment on the stop workflow for the new pos worker.

damip:
@adrien-zinger  I'm gonna merge the factory part in order for us not to diverge too much. Will update the checklist

damip:
What is missing:
* [x] keep block valid on invalid ops
* [x] remove old ledger
* [x] remove old pos
* [x] check pos draws in block graph (block creator and endorsements)
* [x] merge main
* [ ] make PoS bootstrap + ledger bootstrap work
* [x] check the total size of the ops of each block in Protocol. If too big => block invalid
* [x] credit the creator of the endorsed block with sequential coins (not only the creator of the endorsement)
* [x] adapt block creation process to properly work with the new constraints
* [x] adapt the endorsement creation process, use the opportunity to improve it
* [x] repair API endpoints (draws, get_addresses, staker_count, get_stakers, staker stats etc...)
* [ ] repair tests
* [x] review the PoS selector
* [x] put back alerts on slots missed by one of the staking addrs
* [ ] put back alerts on auto-sells on one of the staking addrs
- [x] Use wallet module for factory key management
- [x] Connect factory to pool
- [ ] Connect factory to API
- [ ] Connect factory to node
- [ ] Connect factory to consensus
- [ ] Connect factory to execution
- [ ] make everything compile

adrien-zinger:
> @adrien-zinger I'm gonna merge the factory part in order for us not to diverge too much. Will update the checklist

adrien-zinger:
I guess we should also squash all of that stuff once it compiles, it will be easier to merge then.

damip:
> I guess we should also squash all of that stuff once it compiles, it will be easier to merge then.

I agree. More generally I am in favor of always auto-squashing on merge.

adrien-zinger:
> > I guess we should also squash all of that stuff once it compiles, it will be easier to merge then.
> 
> I agree. More generally I am in favor of always auto-squashing on merge.

Sometimes it makes sens, here it's almost obvious :rofl:

massa-pool-worker/src/operation_pool.rs
```diff
@@ -0,0 +1,227 @@
...
+            {
+                continue;
+            }
+
+            // check sequential balance
+            let creator_seq_balance = sequential_balance_cache
+                .entry(op_info.creator_address)
+                .or_insert_with(|| {
+                    self.execution_controller
+                        .get_final_and_active_parallel_balance(vec![op_info.creator_address])[0]
```
damip, line 199: Seq ! Not parallel

massa-execution-exports/src/controller_traits.rs
```diff
@@ -19,12 +23,12 @@ pub trait ExecutionController: Send + Sync {
...
-    /// * `finalized_blocks`: newly finalized blocks
-    /// * `blockclique`: new blockclique
+    /// * `finalized_blocks`: newly finalized blocks. Each Storage owns refs to the block and its ops/endorsements/parents.
+    /// * `blockclique`: new blockclique. Each Storage owns refs to the block and its ops/endorsements/parents
     fn update_blockclique_status(
         &self,
-        finalized_blocks: HashMap<Slot, BlockId>,
-        blockclique: HashMap<Slot, BlockId>,
+        finalized_blocks: HashMap<Slot, (BlockId, Storage)>,
+        blockclique: HashMap<Slot, (BlockId, Storage)>,
```
adrien-zinger, line 31: Maybe use a global storage for all that stuff would be better. I feel like we need to write-lock the owned variable for each id. In that case, with a global storage we could lock only once.

massa-consensus-worker/src/consensus_worker.rs
```diff
@@ -63,10 +42,6 @@ pub struct ConsensusWorker {
     latest_final_periods: Vec<u64>,
     /// clock compensation
     clock_compensation: i64,
-    /// staking keys
-    staking_keys: Map<Address, KeyPair>,
```
adrien-zinger, line 67: @damip @Eitu33 Now which module should manage the staking keys? Should I integrate this in PoS?

Eitu33: Isnt' this already handled by [BlockFactoryWorker](https://github.com/massalabs/massa/blob/feature/add_pos_and_selector_skel/massa-factory-worker/src/block_factory.rs#L23) `wallet` field? That's what I understood in today's morning presentation.

damip: The node wallet containing the staking keys is created in node main and an arc mutex is passed to the api (for key management) and another is passed to factory which uses it for production

massa-execution-exports/src/controller_traits.rs
damip: Yes but I'm afraid that would imply losing the per-block storage that can be moved around internally by the execution worker.

PR #2734, opened 2022-06-27
Fix success message bootstrap on server side

### LINKED ISSUE #2731: Bootstrap issue
Server:

```
2022-06-24T19:33:03.198953Z DEBUG massa_bootstrap::server: bootstrap serving error for peer XXXXXXXXX:40630: io error: early eof
```

Client:

```
2022-06-24T19:33:02.704317Z  INFO massa_bootstrap::client: Start bootstrapping from XXXXXXXXXXX:31245
2022-06-24T19:33:03.150430Z DEBUG massa_bootstrap::client: Received ledger batch from None to Some(
    Some(
        A12RTRgH1rQHi289AW6AkrLS9MNsNDLdvgZehEbPtRHs93h5phnc,
    ),
), an async pool batch from None to None a batch of ledger changes of size 0 and a batch of async pool changes of size 0. for slot: Slot {
    period: 0,
    thread: 31,
}
Received ledger batch from None to Some(
    Some(
        A12RTRgH1rQHi289AW6AkrLS9MNsNDLdvgZehEbPtRHs93h5phnc,
    ),
), an async pool batch from None to None a batch of ledger changes of size 0 and a batch of async pool changes of size 0. for slot: Slot {
    period: 0,
    thread: 31,
}
2022-06-24T19:33:03.192740Z  INFO massa_bootstrap::client: Successful state bootstrap
```
Config files: [files.zip](https://github.com/massalabs/massa/files/8980279/files.zip)

AurelienFT:
Fix #2731

### DIFF
--- a/massa-bootstrap/src/client.rs
+++ b/massa-bootstrap/src/client.rs
@@ -84,7 +84,6 @@ async fn stream_final_state(
                     } = &next_bootstrap_message
                     {
                         debug!("Received ledger batch from {:#?} to {:#?}, an async pool batch from {:#?} to {:#?} a batch of ledger changes of size {:#?} and a batch of async pool changes of size {:#?}. for slot: {:#?}", old_key.clone().map(|key| get_address_from_key(&key)), last_key.clone().map(|key| get_address_from_key(&key)), old_message_id, last_last_async_id, final_state_changes.ledger_changes.0.len(), final_state_changes.async_pool_changes.0.len(), slot);
-                        println!("Received ledger batch from {:#?} to {:#?}, an async pool batch from {:#?} to {:#?} a batch of ledger changes of size {:#?} and a batch of async pool changes of size {:#?}. for slot: {:#?}", old_key.clone().map(|key| get_address_from_key(&key)), last_key.clone().map(|key| get_address_from_key(&key)), old_message_id, last_last_async_id, final_state_changes.ledger_changes.0.len(), final_state_changes.async_pool_changes.0.len(), slot);
                     }
                     // Set new message in case of disconnection
                     *next_bootstrap_message = BootstrapClientMessage::AskFinalStatePart {
@@ -303,7 +302,7 @@ async fn bootstrap_from_server(
                 break;
             }
             BootstrapClientMessage::BootstrapError { error: _ } => {
-                panic!("Should never happens")
+                panic!("The next message to send shouldn't be BootstrapError");
             }
         };
     }


massa-bootstrap/src/server.rs
```diff
@@ -432,7 +432,10 @@ async fn manage_bootstrap(
     loop {
         match tokio::time::timeout(bootstrap_settings.read_timeout.into(), server.next()).await {
             Err(_) => return Ok(()),
-            Ok(Err(e)) => return Err(e),
+            Ok(Err(e)) => match e {
+                BootstrapError::IoError(_) => return Ok(()),
```
damip, line 436: why is IoError OK ?

AurelienFT: The client doesn't send a message when the bootstrap is finish he just end the socket and the server doesn't manage any client state so the server can't know that a bootstrap is well performed, he can know that steps independently are performed. It's our design choice. So the only way the server can assume the client bootstrapped it's when he disconnect. Of course, the client can disconnect when he just have connection problem but I don't think it will yield the same error and if it the case then the server has no way to know if it's in case of an end or disconnection.

massa-bootstrap/src/server.rs
```diff
@@ -432,7 +432,10 @@ async fn manage_bootstrap(
     loop {
         match tokio::time::timeout(bootstrap_settings.read_timeout.into(), server.next()).await {
```
damip, line 433: so at every loop iteration we expect a timeout here ? Doesn't this slow things down a lot ?

AurelienFT: No we don't except a timeout at every loop iteration. We wait for a message and manage it if we timeout then we stop the connection.

massa-bootstrap/src/server.rs
damip: Do you think we could imagine adding a client message confirming successful bootstrap before disconnecting, just like we do for errors ?

massa-bootstrap/src/server.rs
damip: yes you're right :)

massa-bootstrap/src/server.rs
AurelienFT: Personally I prefer with an end message, yes. But I understood that we want to have a server that just answer to ask of client and when it don't ask anything then well we disconnect him.

AurelienFT: Ok I do it.

AurelienFT: Done @damip

massa-bootstrap/src/client.rs
```diff
@@ -278,25 +279,35 @@ async fn bootstrap_from_server(
...
+                    )
+                    .into()),
+                    Ok(Err(e)) => Err(e),
+                    Ok(Ok(_)) => Ok(()),
+                }?;
                 break;
             }
-            Some(BootstrapClientMessage::BootstrapError { error: _ }) => {
+            BootstrapClientMessage::BootstrapError { error: _ } => {
                 panic!("Should never happens")
```
damip, line 306: let's make a more explicit panic message :)

massa-bootstrap/src/server.rs
```diff
@@ -486,6 +486,7 @@ async fn manage_bootstrap(
                         Ok(Ok(_)) => Ok(()),
                     }?;
                 }
+                BootstrapClientMessage::BootstrapSuccess => return Ok(()),
```
damip, line 489: I think the server should print that the bootstrap succeeded

damip: So the BootstrapSuccess message is basically ignored server-side ?

AurelienFT: No it return from this function and go back to the upper scope which print.

damip: okay

damip (approved):
Good work !

damip:
bros merge

damip:
bors merge

PR #2737, opened 2022-06-27
Change block structure to remove full operations

AurelienFT:


AurelienFT:
While I'm thinking of this I think we should keep the full block as it exists and add a `PartialBlock` that have only the header and hashes. Because when we want to perform checks/actions on the block if we have to retrieve operations from the shared storage each time it will be too big. So I think just inside the network we should have the light version of the block and reconstruct the block when we receive it (by asking ops or fetching from shared storage). 

What do you think ? @gterzian @damip

gterzian:
> when we want to perform checks/actions on the block if we have to retrieve operations from the shared storage each time it will be too big

I think it's best to do the most simple thing first, and only optimize later when we have an actual performance problem.

AurelienFT:
Has been merged in #2753

PR #2744, opened 2022-06-27
Add keypair.

### LINKED ISSUE #2714: Use KeyPair instead of private/public key
In some signatures algorithms the public key isn't derivable from the Private key. To have more flexibility instead of having always a PrivateKey and derive the PublicKey from it we should have a KeyPair structure so that we can always have access to both

AurelienFT:
Fix #2714


massa-signature/src/signature_impl.rs
```diff
@@ -20,6 +20,22 @@ const PRIVATE_KEY_STRING_PREFIX: &str = "PRI";
...
+pub struct KeyPair {
+    /// Private key used for signing, wallet dump/load etc...
+    pub private_key: PrivateKey,
+    /// Public key used for verifications
+    pub public_key: PublicKey
+}
+
+impl KeyPair {
+    /// Create a KeyPair from a private key
+    pub fn from_private_key(private_key: PrivateKey) -> KeyPair {
```
damip, line 34: this should be removed. Everywhere we were using/storing a private key only, we should be using a keypair

massa-signature/src/signature_impl.rs
```diff
@@ -20,6 +20,22 @@ const PRIVATE_KEY_STRING_PREFIX: &str = "PRI";
 const PUBLIC_KEY_STRING_PREFIX: &str = "PUB";
 const SIGNATURE_STRING_PREFIX: &str = "SIG";
 
+/// `KeyPair` is used for signature and decrypting
+#[derive(Clone, Copy, Debug, Eq, PartialEq)]
+pub struct KeyPair {
```
damip, line 25: I think the shnorr crate has a KeyPair ! 
see the definition of our current PrivateKey, it already wraps a keypair

massa-signature/src/signature_impl.rs
```diff
@@ -899,13 +902,17 @@ pub fn verify_signature(
 }
 
 /// Generate a random private key from a RNG.
-pub fn generate_random_private_key() -> PrivateKey {
+pub fn generate_random_keypair() -> KeyPair {
     use secp256k1::rand::rngs::OsRng;
     let mut rng = OsRng::new().expect("OsRng");
-    PrivateKey(secp256k1::KeyPair::from_secret_key(
+    let keypair = secp256k1::KeyPair::from_secret_key(
```
damip, line 908: the lib has a native keypair generator: https://docs.rs/secp256k1/latest/secp256k1/struct.KeyPair.html#examples

damip: generally, we shouldn't be using key derivation anymore

massa-signature/src/signature_impl.rs
damip: It shouldn't add any constraints: the wrapped type should be private and only a couple methods should be exposed.

For example:

* struct PublicKey(lib::PubKey)
  * verify(data, signature) -> bool

* struct KeyPair(lib::KeyPair)
  * sign(data) -> Signature
  * get_public_key() -> &PublicKey

massa-signature/src/signature_impl.rs
AurelienFT: I changed to use it.

massa-signature/src/signature_impl.rs
AurelienFT: Done.

massa-signature/src/signature_impl.rs
AurelienFT: I have implemented this interface.

AurelienFT:
@damip I have made a new implementation closer to what you asked. Tell me if it's ok so that I can propagate to the rest of the code.

damip:
@AurelienFT Yes it's a welcome improvement !

Eitu33 (dismissed):

Eitu33:
Nice

damip:
don't forget to dump both the pubkey and the privkey in files as well (eg. wallet, node privkey, staking keys etc...)

Eitu33 (approved):

Eitu33 (approved):

AurelienFT:
bors merge

AurelienFT:
bors cancel

Eitu33 (approved):
Signature update LGTM 👍

massa-wallet/src/lib.rs
```diff
@@ -36,7 +36,7 @@ impl Wallet {
     pub fn new(path: PathBuf, password: String) -> Result<Wallet, WalletError> {
         if path.is_file() {
             let content = &std::fs::read(&path)?[..];
-            let decrypted_content = decrypt(&password, content)?;
+            let (__version, decrypted_content) = decrypt(&password, content)?;
```
Eitu33, line 39:
```suggestion
            let (_version, decrypted_content) = decrypt(&password, content)?;
```

Eitu33 (approved):

damip (approved):

damip:
bors merge

PR #2757, opened 2022-06-30
Use statics (or lazy) de/serializer in message.rs

adrien-zinger:
Instead of creating a new de/serializer for each message, I use static and lazy static variables.

... in message.rs


AurelienFT (commented):
It will reduce a lot of new initializer so for me it's ok.

damip:
Won't it cause problems when the serializer parameters change ?

AurelienFT:
bors merge

PR #2758, opened 2022-06-30
Increase timeout and remove check in bs58.

AurelienFT:
Now that we transfer the content of the operation in serialized data base58 encoded in the case of a SC deployment it can be quite long to encode and decode bs58 so I removed the check and increase the timeout.


damip:
I'm confused as to why it would be bigger than before

AurelienFT:
> I'm confused as to why it would be bigger than before

Removed this change.

Eitu33 (approved):

AurelienFT:
bors merge

Issue #2747, opened 2022-06-29
Restructure Protocol workflow for blocks without operations


gterzian:
Part of https://github.com/massalabs/massa/issues/2735

New workflow:
* a header is announced
* the op hash list is asked for that header
* the op hash list is sent for that block (non-truncated hashes must be used)
* the op hash list is received, deserialized, and the total op hash is checked against the one in the header
* the hash list is compared to the operations cache
  * the list of missing ops is asked to the sender (non-truncated hashes must be used)
    * this must be prioritized over normal op propagation
  * once we have all ops, the block is processed
  * if the remote node fails to send us all missing ops after a timeout (or if any is invalid):
    * ban the node and disconnect from it
    * ask for missing ops from another node that announced the block (if any, otherwise just forget about that block for now)

damip:
done for v14

PR #2759, opened 2022-07-01
Remove ops from storage.

### LINKED ISSUE #2750: Improve pruning of shared storage(use ref counters)
## Context

When the operation pool is full we remove the least interested from the `ops` vector (https://github.com/massalabs/massa/blob/main/massa-pool/src/operation_pool.rs#L201) but we don't remove it from the shared storage.
This causes a major memory leak that makes the nodes crash often.

## Problem

If a operation is in a block, we don't want to delete it from shared storage even if it overflows out of the pool !

## Solution

Have a reference counter on the operation stored to know if it's still used in block and/or operation pool

(cc @gterzian @damip  )

AurelienFT:
Fix #2750 

The blocks are still saved full (before #2747) in storage with all of their operation we just need to delete the operations from the `operations` hashmap in storage and they will still be accessible in the full blocks for graph etc...

### DIFF
--- a/massa-pool/src/operation_pool.rs
+++ b/massa-pool/src/operation_pool.rs
@@ -187,6 +187,7 @@ impl OperationPool {
                     self.ops_by_address.insert_op(*addr, *op_id);
                 });
             self.ops.insert(*op_id, wrapped_op);
+            self.storage.store_operation(op.clone());
         }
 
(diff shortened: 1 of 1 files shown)


damip (approved):

damip:
bors merge

PR #2761, opened 2022-07-01
PoS execution update 2

adrien-zinger:


massa-execution-worker/src/execution.rs
```diff
@@ -263,6 +268,16 @@ impl ExecutionState {
             .compute_id()
             .expect("could not compute operation ID");
 
+        // transfer the fee from sender to block creator
+        {
+            let mut context = context_guard!(self);
+            context.transfer_parallel_coins(
```
damip, line 274: sequential coins :) not parallel

Issue #2742, opened 2022-06-27
Batch `get_datastore_entry `


qdrn:
It would be nice to be able to batch requests for datastore entries.

Issue #2739, opened 2022-06-27
Correct doc typo


qdrn:
https://massa.readthedocs.io/en/latest/technical-doc/api.html#get-datastore-entry

PR #2765, opened 2022-07-05
Datastore API update

Eitu33:
closes https://github.com/massalabs/massa/issues/2742
closes https://github.com/massalabs/massa/issues/2740
closes https://github.com/massalabs/massa/issues/2739

### DIFF
--- a/docs/technical-doc/api.rst
+++ b/docs/technical-doc/api.rst
@@ -346,9 +346,7 @@ Get the block graph within the specified time interval.
 
 Get a data entry both at the latest final and active executed slots.
 
-If an existing final entry (`final_value`) is:
-* found in the active history, it will return its final value in `active_value` field
-* deleted in the active history, it will return null in `active_value` field
+If an existing final entry (`final_value`) is found in the active history, it will return its final value in `active_value` field. If it was deleted in the active history, it will return null in `active_value` field
 
 -   Parameters:
 


damip:
> @damip we should probably update `massa-sc-runtime` interface to take slices instead of str's when asking for keys, see [`raw_set_data_for`](https://github.com/massalabs/massa/blob/datastore-api-update/massa-execution-worker/src/interface_impl.rs#L195-L208) for example.

For now let's just convert the str to bytes for retro compatibility.

This will come when we switch from strings to bytes everywhere in the runtime: https://github.com/massalabs/massa-sc-runtime/issues/134

massa-execution-worker/src/execution.rs
```diff
@@ -762,6 +761,17 @@ impl ExecutionState {
...
     }
 
+    /// Get every datastore key of the given address.
+    pub fn get_every_final_datastore_key(&self, addr: &Address) -> Vec<Vec<u8>> {
+        self.final_state
+            .read()
+            .ledger
+            .get_entire_datastore(addr)
+            .into_iter()
+            .map(|v| v.0)
```
AurelienFT, line 771: I don't understand if we have all the keys and the values here hy we return only keys.

massa-execution-worker/src/speculative_ledger.rs
```diff
@@ -294,6 +293,15 @@ impl SpeculativeLedger {
             )));
         }
 
+        // check key correctness
+        let key_length = key.len();
+        if key_length > MAX_DATASTORE_KEY_LENGTH as usize {
```
AurelienFT, line 298: I think we should check if key is empty also

massa-ledger-exports/src/key.rs
```diff
@@ -81,33 +83,45 @@ impl Serializer<Vec<u8>> for KeySerializer {
...
-        buffer: &mut Vec<u8>,
-    ) -> Result<(), massa_serialization::SerializeError> {
-        buffer.extend(value);
+    fn serialize(&self, value: &Vec<u8>, buffer: &mut Vec<u8>) -> Result<(), SerializeError> {
+        let limit = ADDRESS_SIZE_BYTES + 1;
+        buffer.extend(&value[..limit]);
+        if value[ADDRESS_SIZE_BYTES] == DATASTORE_IDENT {
+            if value.len() > limit {
+                self.vec_u8_serializer
+                    .serialize(&value[(ADDRESS_SIZE_BYTES + 1)..].to_vec(), buffer)?;
```
AurelienFT, line 96: I think you can use your variable `limit`

massa-ledger-exports/src/ledger_entry.rs
```diff
@@ -77,35 +71,33 @@ impl Serializer<BTreeMap<Hash, Vec<u8>>> for DatastoreSerializer {
...
 }
 
 impl DatastoreDeserializer {
     /// Creates a new `DatastoreDeserializer`
     pub fn new() -> Self {
         Self {
             u64_deserializer: U64VarIntDeserializer::new(Included(u64::MIN), Included(u64::MAX)),
-            hash_deserializer: HashDeserializer::new(),
-            value_deserializer: VecU8Deserializer::new(Included(u64::MIN), Included(u64::MAX)),
+            vec_u8_deserializer: VecU8Deserializer::new(Included(u64::MIN), Included(u64::MAX)),
```
AurelienFT, line 82: I think everywhere we can put MAX_DATASTORE_KEY_SIZE.

massa-execution-worker/src/execution.rs
Eitu33: I thought it's because it would be heavy as a return payload to have everything but maybe I'm wrong.

massa-ledger-exports/src/ledger_changes.rs
```diff
@@ -83,7 +84,7 @@ impl DatastoreUpdateDeserializer {
     pub fn new() -> Self {
         Self {
             u64_deserializer: U64VarIntDeserializer::new(Included(u64::MIN), Included(u64::MAX)),
-            hash_deserializer: HashDeserializer::default(),
+            key_deserializer: VecU8Deserializer::new(Included(u64::MIN), Included(u64::MAX)),
```
AurelienFT, line 87: Here also fro u64 and it's ok

massa-api/src/public.rs
```diff
@@ -510,18 +510,22 @@ impl Endpoints for API<Public> {
...
-            let data =
-                execution_controller.get_final_and_active_data_entry(&entry.address, &entry.key);
-            Ok(DatastoreEntryOutput {
-                final_value: data.0,
-                active_value: data.1,
-            })
+            let mut result = Vec::new();
+            for entry in entries {
+                let data = execution_controller
+                    .get_final_and_active_data_entry(&entry.address, &entry.key);
```
damip, line 522: could be cool to batch this, because otherwise we read-lock at every loop iteration which means:
* performance problems
* different entries returned can come from different slots

damip:
@Eitu33 can you check that smart contracts writing things with too-long keys, or zero-length keys don't crash the system ?

Eitu33:
@damip I've tested every case, everything works as expected. This is ready to be merged 👍

damip (dismissed):
Amazing work !

Eitu33:
bors merge

Eitu33:
@damip I forgot to update the `get_addresses` doc 😅 would you mind approving again?

Eitu33:
bors merge

PR #2793, opened 2022-07-12
Modify AddressInfo format

adrien-zinger:
Format the display of `AddressInfo`. (align 2 by 2 elements for big vectors) 

I just realized how much it would be great to add an option `-n` when we dump information like `AddressInfo` to get the n latest blocks, endorsement draws, block draws, etc. I can do it here.

Also, I think the `fold` is a bit hard to use for just that simple use case.

### DIFF
--- a/massa-models/src/api.rs
+++ b/massa-models/src/api.rs
@@ -232,42 +232,107 @@ impl std::fmt::Display for AddressInfo {
             "Final datastore keys:\n{:?}\n",
             self.final_datastore_keys
         )?;
+        let mut b = false;
         writeln!(
             f,
-            "Block draws: {}",
+            "\nBlock draws:{}",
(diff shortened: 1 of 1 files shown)


aoudiamoncef:
It's a temporary solution. 

We should think about paging our endpoints

AurelienFT (commented):
Cool change but I think you could use the chunk method instead of the variable and will simplify a lot.

damip:
Is it still relevant with the current format of get address info ?

adrien-zinger:
> Is it still relevant with the current format of get address info ?

This will still be an enhancement, but we can wait for the merge of the release 14 and then adapt it. :slightly_smiling_face:

AurelienFT:
Not relevant anymore.

Issue #2804, opened 2022-07-13
Warning in sandbox mode


damip:
![massa](https://user-images.githubusercontent.com/10211103/178790310-fc8d802a-c544-4b19-a220-20c1d0711b29.png)

Issue #2803, opened 2022-07-13
Sandbox problems


damip:
Sandbox mode does not build in TEST.12.1

```
~/massa$ cargo build --features sandbox

warning: /home/damip/massa/massa-node/Cargo.toml: unused manifest key: build
   Compiling massa_models v0.1.0 (/home/damip/massa/massa-models)
error[E0013]: constant functions cannot refer to statics
   --> massa-models/src/block.rs:480:29
    |
480 |         let thread_count = *THREAD_COUNT;
    |                             ^^^^^^^^^^^^
    |
    = help: consider extracting the value of the `static` to a `const`, and referring to that

error[E0015]: cannot perform deref coercion on `node_configuration::default::THREAD_COUNT` in constant functions
   --> massa-models/src/block.rs:480:28
    |
480 |         let thread_count = *THREAD_COUNT;
    |                            ^^^^^^^^^^^^^
    |
    = note: attempting to deref into `u8`
note: deref defined here
   --> massa-models/src/node_configuration/default.rs:75:1
    |
75  | / lazy_static::lazy_static! {
76  | |     /// t0
77  | |     pub static ref T0: MassaTime = std::env::var("T0").map(|timestamp| timestamp.parse::<u64>().unwrap().into()).unwrap_or_else(|_|
78  | |         MassaTime::from(16000)
...   |
83  | |     );
84  | | }
    | |_^
note: impl defined here, but it is not `const`
   --> massa-models/src/node_configuration/default.rs:75:1
    |
75  | / lazy_static::lazy_static! {
76  | |     /// t0
77  | |     pub static ref T0: MassaTime = std::env::var("T0").map(|timestamp| timestamp.parse::<u64>().unwrap().into()).unwrap_or_else(|_|
78  | |         MassaTime::from(16000)
...   |
83  | |     );
84  | | }
    | |_^
    = note: calls in constant functions are limited to constant functions, tuple structs and tuple variants
    = note: this error originates in the macro `__lazy_static_internal` (in Nightly builds, run with -Z macro-backtrace for more info)

error[E0013]: constant functions cannot refer to statics
   --> massa-models/src/endorsement.rs:173:29
    |
173 |         let thread_count = *THREAD_COUNT;
    |                             ^^^^^^^^^^^^
    |
    = help: consider extracting the value of the `static` to a `const`, and referring to that

error[E0015]: cannot perform deref coercion on `node_configuration::default::THREAD_COUNT` in constant functions
[...]

PR #2806, opened 2022-07-14
Improve config

damip:
* closes https://github.com/massalabs/massa/issues/2804
* closes https://github.com/massalabs/massa/issues/2803
* removes dirty THREAD_COUNT and T0 usages in sandbox vs non-sandbox modes
* solves a dangerous float rounding being done for the amounts in the config
* removes amount from_raw dangerous usages


AurelienFT (commented):
I don't like the fact that you just removed the possibility to change T0 and THREAD COUNT in runtime in sandbox. I know it was messy but I think we can find a good way to make the switch one time in the code and clean rather then removing the feature that we discussed that could be cool to have to make experiments etc

AurelienFT (dismissed):

AurelienFT:
bors merge

Eitu33 (approved):

Eitu33:
bors merge

PR #2822, opened 2022-07-18
Pos fix public api

adrien-zinger:


massa-pos-worker/src/controller.rs
```diff
@@ -58,6 +60,38 @@ impl SelectorController for SelectorControllerImpl {
         Ok(selection.producer)
     }
 
+    /// Return a list of slots where `address` has been choosen to produce a
+    /// block and a list where he is choosen for the endorsements.
+    /// Look from the `start` slot to the `end` slot.
+    fn filter_selection_by_address(
```
Eitu33, line 66: This function name don't feel too appropriated to me, maybe something like `get_address_selections` would be better?

Eitu33 (approved):

PR #2829, opened 2022-07-20
Modify selector stop handler

adrien-zinger:
Ensure that the thread stop correctly: Before it was a possible deadlock when the selector controller stopped the thread after the execution thread.

Profit of the input data pointer to pass a `Command::Stop`

### DIFF
--- a/massa-pos-worker/src/lib.rs
+++ b/massa-pos-worker/src/lib.rs
@@ -17,10 +17,15 @@ use std::{
     sync::Arc,
 };
 
+pub(crate) enum Command {
+    CycleInfo(CycleInfo),
+    Stop,
+}
+
 /// Same structure pointer that will be used by the selector controller and his
 /// thread. It will store all new CycleInfo declared by massa (in the
 /// Execution module) and will be used to compute the draws in background.
-pub(crate) type InputDataPtr = Arc<(Condvar, Mutex<VecDeque<CycleInfo>>)>;
+pub(crate) type InputDataPtr = Arc<(Condvar, Mutex<VecDeque<Command>>)>;
 
 /// Structure of the shared pointer to the computed draws.
 pub(crate) type DrawCachePtr = Arc<RwLock<BTreeMap<u64, HashMap<Slot, Selection>>>>;
--- a/massa-pos-worker/src/worker.rs
+++ b/massa-pos-worker/src/worker.rs
@@ -1,7 +1,6 @@
 // Copyright (c) 2022 MASSA LABS <info@massa.net>
 
 use std::collections::BTreeMap;
-use std::sync::{atomic::AtomicBool, Arc};
 use std::thread::JoinHandle;
 
 use massa_hash::Hash;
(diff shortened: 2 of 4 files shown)


massa-pos-worker/src/worker.rs
```diff
@@ -61,23 +58,22 @@ impl SelectorThread {
...
-            }
             let input_data = self.input_data.clone();
             let mut data = input_data.1.lock();
-            if let Some(cycle_info) = data.pop_front() {
-                self.draws(cycle_info)?
+            match data.pop_front() {
+                Some(Command::CycleInfo(cycle_info)) => self.draws(cycle_info)?,
+                Some(Command::Stop) => break,
+                None => {}
             }
```
gterzian, line 77: You could do something like:
```rust
let cycle_info = {
    // do stuff with the lock and condvar
};
self.draws(cycle_info);
```
so as not to hold the lock while performing the draw(unless the draw should be part of the criticial section)

massa-pos-worker/src/lib.rs
```diff
@@ -17,10 +17,15 @@ use std::{
     sync::Arc,
 };
 
+pub(crate) enum Command {
+    CycleInfo(CycleInfo),
+    Stop,
+}
```
gterzian, line 23: You could add some docs.

gterzian (approved):
LGTM, with just a few comments.

massa-pos-worker/src/lib.rs
adrien-zinger: done

massa-pos-worker/src/worker.rs
adrien-zinger: done

adrien-zinger:
> LGTM, with just a few comments.

Thanks for your advices @gterzian, I would be grateful if you verify again the modification before merging :slightly_smiling_face:

gterzian:
Yes, I actually have one more comment, because I was not clear enough earlier, just trying it out locally first to make sure it makes sense before posting, stay tuned... @adrien-zinger

massa-pos-worker/src/worker.rs
```diff
@@ -61,33 +58,38 @@ impl SelectorThread {
...
+            };
+
+            if let Some(cycle_info) = cycle_info {
+                self.draws(cycle_info)?;
             }
+
             // Wait to be notified of new input
             // The return value is ignored because we don't care what woke up the condition variable.
-            self.input_data.0.wait(&mut data);
+            self.input_data.0.wait(&mut self.input_data.1.lock());
```
gterzian, line 86: This in fact creates a second critical section, separate from the one above. It doesn't actually matter much in this particular case, and it's still better to consolidate the logic into one. 

I think the below run loop is much clearer:

```rust
    fn run(mut self) -> PosResult<()> {
        loop {
            let cycle_info = {
                let mut data = self.input_data.1.lock();
                loop {
                    match data.pop_front() {
                        Some(Command::CycleInfo(cycle_info)) => break cycle_info,
                        Some(Command::Stop) => return Ok(()),
                        None => self.input_data.0.wait(&mut data),
                    }
                }
            };
            self.draws(cycle_info)?;
        }
    }
```

gterzian (approved):
Ok, last comment on the run loop. LGTM otherwise so I'll approve already and let you make the change.

massa-pos-worker/src/worker.rs
adrien-zinger: Great! Very nice :+1:

PR #2837, opened 2022-07-22
Add module factory

adrien-zinger:
- [x] Skeleton
- [ ] Production thread
- [ ] Controller interface


adrien-zinger:
bad targets, I'll reopen a correct PR

PR #2851, opened 2022-07-30
Add module storage usage counters

damip:
## General idea

Add module-level references to objects in Storage so that they are removed whenever not needed by any module anymore.

## Constraints

Atomics and mutexes are performance-heavy. We therefore should not have `Arc<Mutex<..>>` on every operation, for example inside blocks. Processing should be batched as much as possible because we have 10000 operations/sec. 

## Implementation

Every module owns an instance of Storage. In this PR, we extend Storage to handle locally (module-wide) owned object references.

To claim that a list of already-stored operation references are in use by the module from now on, use `self.storage.claim_operation_refs(operation_ids)`.

To signal that the current module won't need operation references anymore, use `self.storage.drop_operation_refs(operation_ids)`. When this function is called, storage removes all objects for which not a single module has a reference anymore.

When a Storage instance is dropped, all local references are cleanly released following the drop_operation_refs behavior described above.

When adding an object (for example an operation) to storage from a given module, use:
```
self.storage.store_operations(vec![op1, op2]);
```
note that this will automatically claim a local reference for the object.


damip:
@gterzian  @adrien-zinger can you take a look ? what do you think of this approach ?

massa-storage/src/lib.rs
```diff
@@ -1,66 +1,195 @@
...
-        let mut blocks = self.blocks.write();
-        match blocks.entry(block.id) {
-            Entry::Occupied(_) => {}
-            Entry::Vacant(entry) => {
-                let to_store = Arc::new(RwLock::new(block));
-                entry.insert(to_store);
+    /// internal helper to locally claim a reference to an object
+    fn internal_claim_refs<IdT: Id + PartialEq + Eq + Hash + PreHashed + Copy>(
+        ids: &[IdT],
+        mut owners: RwLockWriteGuard<Map<IdT, usize>>,
```
gterzian, line 61: nit: in general instead of passing the guard, you can pass a `&mut`, and do `&mut guard` at the call site.

gterzian:
Thanks for the proposal, I think it's a good idea, and it will be problematic in practice, for the simple reason that there is nothing keeping an object alive until the "using module" has been able to claim it, which it can only do after having received some kind of message from the "storing module". 

For example:

1. Graph creates a new block, stores it. 
2. Send the block for processing to Execution(via a message queue). 
3. Block for some  reason is pruned by Graph. 
4. Execution receives message sent at 2, tries to claim local ownership, but the block has already been pruned. 

Such a problem would be solved by sending a message with Arcs, which would keep the reference alive while the message is in the queue. 

Not saying this cannot be addressed in some way with additional logic, however at some point it just becomes too complicated(and error prone). 

By the way Arcs and Mutexes are not actually that expensive, and parking lot is designed specifically around the idea of having lots of them, see the article that is the source of inspiration for that design: https://webkit.org/blog/6161/locking-in-webkit/

My conclusion: let's just stick to passing arcs around, it's unlikely that this will be a performance problem and it would help us avoiding re-inventing the wheel. 

I do agree with the changes batching the storage operations.

massa-storage/src/lib.rs
```diff
@@ -1,66 +1,195 @@
...
+        local_used_ids: &mut Set<IdT>,
+    ) {
+        for &id in ids {
+            if local_used_ids.insert(id) {
```
AurelienFT, line 65: I don't understand this if. It should be "if not" instead ? Because here I understand that we increase only if it's the second time we locally claim the id. @damip

damip: basically, if we have already claimed this ID, we don't increment counters on a second claim from the same module

AurelienFT: Ah yes it returns true if there is no value before it's the contrary of insert with hashmap... My bad.

massa-storage/src/lib.rs
damip: true, we can pass a reference, will improve that

damip:
Merged into feature/factory that needed it, and took all the comments into account there.

Issue #2462, opened 2022-03-27
OpenRPC Specification


aoudiamoncef:
Related to #1058 

Example with [Ethereum ](https://playground.open-rpc.org/)

May be convert Wiki documentation to OpenRPC spec ? 

![openrpc](https://user-images.githubusercontent.com/22281426/160298385-b2db7a93-2dce-4f17-b65e-69a2e576496f.png)

I'm waiting for feedback

Thanks

damip:
Thanks for the idea.

@yvan-sraka  you were mentionning using a swagger-like description. Would that be of interest?

aoudiamoncef:
It will be my next contribution to Massa. It's in progress

qdrn:
> It will be my next contribution to Massa. It's in progress . Its painfull to write it in raw JSON. We could cheat and made it with existing OpenAPI GUI tools like [ApiCurio](https://www.apicur.io/studio/) . At the end, OpenRPC is similare to a POST based OpenAPI spec.

Nice. Thanks for the update!

aoudiamoncef:
Hi,

With the help of @rykcod, we finished the first step which consists of translating current documentation to OpenAPI format.

Now we should validate it together before translating to OpenRPC. 

@qdrn if you are interested to review the current work, I'll send you an invitation link to join us.

damip:
Let's make sure the spec is hosted in the massa repo, docs folder and that the CI updates everything auomatically.

aoudiamoncef:
We have an draft OpenRPC spec, we should enrich it to be complete. 

Some useful links: 
[OpenRPC spec](https://github.com/Mermade/open-rpc-spec)
[Ethereum OpenRPC spec](https://github.com/etclabscore/ethereum-json-rpc-specification/blob/master/openrpc.json)
[Playground OpenRPC](https://playground.open-rpc.org/)

qdrn:
> Hi,
> 
> With the help of @rykcod, we finished the first step which consists of translating current documentation to OpenAPI format.
> 
> Now we should validate it together before translating to OpenRPC.
> 
> @qdrn if you are interested to review the current work, I'll send you an invitation link to join us.

That would be nice.

aoudiamoncef:
Hi @qdrn,

[...]

PR #2856, opened 2022-08-02
Update api.rst

altai0:
The get_addresses return value is outdated.

### DIFF
--- a/docs/technical-doc/api.rst
+++ b/docs/technical-doc/api.rst
@@ -388,53 +388,59 @@ Get addresses.
 
 .. code-block:: javascript
 
-    [
-        {
-            address: String,
-            balance: {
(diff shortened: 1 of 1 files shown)


aoudiamoncef:
Hi @altai0, 

Please, could you keep the same order for fields. It'll help us to see the changes.

This documentation will be replaced ASAP with an OpenRPC contract. Don't hesitate to contribute on it. 

See #2462 

Thank you for your contribution ❤️

altai0:
Hi, I checked most of them, as far as I can see there is only change in get_addresses @aoudiamoncef

aoudiamoncef:
Thanks for checking all methods.

Could you make a commit by keeping the order of keys please ?

PR #2772, opened 2022-07-06
Use new serialization everywhere.

AurelienFT:

PR #2753, opened 2022-06-30
Restructure Protocol workflow for blocks without operations

gterzian:
Close https://github.com/massalabs/massa/issues/2747

PR #2862, opened 2022-08-08
Release testnet 14

AurelienFT:
PR rollup of 

- [x] #2772 
- [x] #2753 
- [x] #2727


massa-consensus-worker/src/consensus_worker.rs
```diff
@@ -610,6 +610,7 @@ impl ConsensusWorker {
                     finished = true;
                     break;
                 }
+                self.block_db.storage.store_operation(op);
```
gterzian, line 613: Note: I think we can remove this because the operation should have been stored previously by pool.

gterzian:
Note: we're not using `NodeInfo.wanted_blocks` anymore. Is this a problem, as in we removed some check that is necessary, or can we simply remove that concept(I think we can because we currently always reply "immediately" to a request, by using storage directly, as opposed to first asking consensus for info).

gterzian:
Note: we're not using `ProtocolWorker. max_block_gas` anymore, remove it, or actually re-integrate it's use within the current structure? It may be unnecessary now because of the way we ask for info on a block, and check it.

gterzian:
@AurelienFT Currently the tests are not compiling, I will let you address it while I look into the above two points.

AurelienFT:
> @AurelienFT Currently the tests are not compiling, I will let you address it while I look into the above two points.
> 
> Also, perhaps you can make the suggested serializer changes at [#2753 (comment)](https://github.com/massalabs/massa/pull/2753#discussion_r939970842) ?
> 
> After we can both look into the various tests that have been ignored, and see if we can make them pass. I'd like to focus on the tests in `protocol_worker` so perhaps you can look at the others?

Hello,

Ok I will make the compilation of the tests, the serializer change. I think before fixing the tests I want to merge the ledger unification one.

massa-network-worker/src/messages.rs
```diff
@@ -98,277 +78,474 @@ pub enum Message {
...
+                    };
+                    self.u32_serializer
+                        .serialize(&u32::from(info_type), buffer)?;
+                    if let AskForBlocksInfo::Operations(ids) = info {
+                        let op_ids_serializer = OperationIdsSerializer::new();
+                        op_ids_serializer.serialize(&ids, buffer)?;
+                    }
+                }
+            }
+            Message::ReplyForBlocks(list) => {
```
AurelienFT, line 180: I have implemented it following the deserializer and it don't need the shared storage. Did I miss something ? @gterzian

AurelienFT:
I have checked a bit the tests that fail I think there is a problem with block creation which is fully refactored in #2727 I think I will start the merge.

massa-network-worker/src/messages.rs
gterzian: I'm wondering if we need this here(instead just having a `panic` as a placeholder and to make it clear it's not actually used), because of how the message is constructed in practice in `node_worker`(see `ToSend::ReplyForBlocksInfo`). 

Basically, in practice we don't actually serialize the `Message::ReplyForBlocks`, instead we "map" the `ToSend::ReplyForBlocksInfo` to the serialized form of the message. 

Actually maybe you can review whether that actually makes sense?

gterzian:
@AurelienFT awesome, thanks!  I'll continue reviewing the protocol tests then...

massa-network-worker/src/messages.rs
AurelienFT: I prefer having it in case we refactor but that personal decision I think

AurelienFT:
I'm in the middle of the emerge all conflict resolved but I have some compilations errors

damip:
> I'm in the middle of the emerge all conflict resolved but I have some compilations errors

Yes, the ledger fusion branch does not compile currently, mostly because it relies on some things from separate operation propagation. But I think we should merge them and finish repairing compilation here on release_testnet_14

gterzian:
> But I think we should merge them and finish repairing compilation here on release_testnet_14

Ok, then I will let you fix the compilation. 

> I have checked a bit the tests that fail I think there is a problem with block creation

@AurelienFT For the two failing consensus tests, the fix is to store operations created in the test manually in storage.

damip:
To make the branch compile:

* [ ] `get_block_slot` should be replaced by just getting the slot from the stored block
* [ ] `get_blockclique_block_at_slot` should be implemented in consensus to get the block present at a given slot within the blockclique (or none). It can be set to a skeleton for now to make things compile and open an issue to implement it properly
* [ ] `get_endorsements_by_id` and  `pool_command_sender.get_operations` and `pool_cmd_snd.get_operations_involving_address`  and the corresponding API endpoints should be commented for now to compile and rethought as a followup. Eventually, we want the storage to do the indexing, so that the API can just interrogate storage directly without going through consensus/pool. Let's open an issue on that implem.
* [ ] timeslots::get_closest_slot_to_timestamp should be implemented: it should return the slot that is the closest to a given timestamp. To compile, we can just put a skeleton for now and create an issue to implement it
* [ ] consensus.get_best_parents => should be implemented to return the current consensus.block_graph.get_best_parents()
* [ ] `cmd_sender.register_staking_keys, remove_staking_addresses, get_staking_addresses` should be removed from consensus, and commented out in the API. Let's open an issue on passing Arc<RwLock<Wallet>> to the API so that it can manage the staking keys directly without going though consensus
* [ ] `pool_command_sender.get_pool_stats` should be commented out for now (with the corresponding API endpoint) and an issue should be opened to reimplement it with the new pool

massa-protocol-worker/src/protocol_worker.rs
```diff
@@ -1234,50 +1013,231 @@ impl ProtocolWorker {
...
+                                    #[cfg(not(feature = "sandbox"))]
+                                    let thread_count = THREAD_COUNT;
+                                    let wrapped: WrappedBlock = Wrapped {
+                                        signature: info.header.signature,
+                                        creator_public_key: info.header.creator_public_key,
+                                        creator_address: info.header.creator_address,
+                                        thread: info
+                                            .header
+                                            .creator_address
+                                            .get_thread(thread_count),
```
damip, line 1138:
```suggestion
                                        thread: info
                                            .header
                                            .thread;
```


Important reminder:

* an address belongs to the thread index defined by taking the first 5 bits of the enclosed hash
* an operation belongs to the thread of its sender's address
* the thread to which a block belongs is not defined by its hash, nor by the thread of the address of its creator, but by an explicit `slot` property inside its header
* a block belonging to the thread T can only include operations that belong to thread T
* any address can produce blocks that belong to any thread
* an operation belonging to thread T can spend and receive parallel coins to and from addresses belonging to any thread
* an operation belonging to thread T can make an address belonging to any thread receive sequential coins
* an operation belonging to thread T can make addresses belonging to thread T spend sequential coins.
* an operation that belongs to the thread T cannot cause sequential coin spending from addresses that do not belong to thread T.

damip: @gterzian

AurelienFT:
All sub-module compilation is fixed. Only massa-node left.

AurelienFT:
The code now compile there is some placeholder (checkout the comment of @damir and you can take an issue to replace one of the placeholders). I will focus tomorrow on compiling the tests.

gterzian:
>      consensus.send_block: should call consensus.block_db.incoming_block. Note that the async problem can be worked around using https://docs.rs/tokio/latest/tokio/sync/mpsc/struct.Sender.html#method.blocking_send (implemented as a skeleton for now)

@damip What is "the async problem" that you mention? For now I've done https://github.com/massalabs/massa/pull/2862/commits/3604f2df15ed1fae4e8ea2d81720f0dc81e04876

damip:
> > ```
> >  consensus.send_block: should call consensus.block_db.incoming_block. Note that the async problem can be worked around using https://docs.rs/tokio/latest/tokio/sync/mpsc/struct.Sender.html#method.blocking_send (implemented as a skeleton for now)
> > ```
> 
> @damip What is "the async problem" that you mention? For now I've done [3604f2d](https://github.com/massalabs/massa/commit/3604f2df15ed1fae4e8ea2d81720f0dc81e04876)

Just the fact that some CommandSenders use async mpsc.send() and we need to change it ti blocking_send when called from non-async contexts

adrien-zinger:
[followup list
](https://github.com/massalabs/massa/pull/2862#issuecomment-1210462093)

massa-protocol-worker/src/protocol_network.rs
```diff
@@ -0,0 +1,426 @@
...
+            }
+
+            // Update ask block
+            let mut set = Set::<BlockId>::with_capacity_and_hasher(1, BuildMap::default());
+            set.insert(block_id.clone());
+            self.stop_asking_blocks(set)?;
+
+            // Re-add to wishlist with new state.
+            self.block_wishlist
+                .insert(block_id, AskForBlocksInfo::Operations(missing_operations));
```
adrien-zinger, line 281: It is probably possible to already have all operations for that block. In that case, we should already forward the block to the graph. Or I'm missing something else @gterzian ?

gterzian: Yes that's a good point, thanks.

massa-pool-worker/src/controller_impl.rs
```diff
@@ -0,0 +1,76 @@
...
+        Box::new(self.clone())
+    }
+
+    fn get_operations_involving_address(&self, address: &Address) -> Operations {
+        self.operation_pool
+            .read()
+            .expect("could not r-lock operation pool")
+            .storage
+            .get_operations_created_by(address)
+    }
```
adrien-zinger, line 75: It's unnecessary to use pool to get that information. It can be retrieved by any storage.

AurelienFT: Yes changed.

massa-models/src/block.rs
```diff
@@ -126,7 +125,7 @@ pub struct Block {
     /// signed header
     pub header: WrappedHeader,
     /// operations
-    pub operations: Vec<WrappedOperation>,
+    pub operations: OperationIds,
```
damip, line 128: Critical: operation order in blocks matters, but OperationIds is a Set.

* operation order is chosen by block producer
* operations are executed in the order they appear in the block
* the operation global hash in the header is sensitive to operation order
* it is therefore consensus-critical for the operations order to be the same for everyone for every block

```suggestion
    pub operations: Vec<OperationId>,
```

damip (changes requested):

massa-factory-worker/src/endorsement_factory.rs
```diff
@@ -0,0 +1,224 @@
...
+                    // the selected block producer is not managed locally => continue
+                    None => continue,
+                };
+            }
+        }
+
+        // quit if there is nothing to produce
+        if producers_indices.is_empty() {
+            return;
+        }
```
adrien-zinger, line 153: `producers_indices` will always be empty here. Maybe it should be like the following proposal? I would also prefer a foreign method like `get_producers_indices`.

```suggestion
        // get creators if they are managed by our wallet
        let producers_indices: Vec<(KeyPair, usize)> = Vec::new();
        {
            let wallet = self.wallet.read().expect("could not lock wallet");
            for (index, producer_addr) in producer_addrs.into_iter().enumerate() {
                // check if the block producer address is handled by the wallet
                producers_indices.push(match wallet.find_associated_keypair(&producer_addr) {
                    // the selected block producer is managed locally => continue to attempt endorsement production
                    Some(kp) => (kp, index),
                    // the selected block producer is not managed locally => continue
                    None => continue,
                });
            }
        }

        // quit if there is nothing to produce
        if producers_indices.is_empty() {
            return;
        }
```

damip:
@massalabs/core-team  OperationIds seems to be used everywhere, even in places where operation order should matter (see https://github.com/massalabs/massa/pull/2862#discussion_r946848973)

Given how little benefit and how huge of a problem this type redefinition brings, I think it would be smart to completely remove it and write explicitly `Set<OperationId>` or `Vec<OperationId>` depending on the situation. It wasn't really that much shorter anyways.

adrien-zinger:
> @massalabs/core-team OperationIds seems to be used everywhere, even in places where operation order should matter (see [#2862 (comment)](https://github.com/massalabs/massa/pull/2862#discussion_r946848973) but also in Protocol etc...)
> 
> Given how little benefit and how huge of a problem this type redefinition brings, I think it would be smart to completely remove it and write explicitly `Set<OperationId>` or `Vec<OperationId>` depending on the situation. It wasn't really that much shorter anyways.

We can replace the `type OperationIds = Set...` with `type OperationIds = Vec...`. :slightly_smiling_face: it has been declared for that.

damip:
> > @massalabs/core-team OperationIds seems to be used everywhere, even in places where operation order should matter (see [#2862 (comment)](https://github.com/massalabs/massa/pull/2862#discussion_r946848973) but also in Protocol etc...)
> > Given how little benefit and how huge of a problem this type redefinition brings, I think it would be smart to completely remove it and write explicitly `Set<OperationId>` or `Vec<OperationId>` depending on the situation. It wasn't really that much shorter anyways.
> 
> We can replace the `type OperationIds = Set...` with `type OperationIds = Vec...`. slightly_smiling_face it has been declared for that. We can also declare an implementation for `insert`, that would avoid a refactoring over all the code base.

Basically, there are places where we want to quickly lookup the presence of an operation ID, in which case we want a Set. In other places we just want to preserver order and iterate in order, in which case we want a Vec. There is a requirement for both at different places of the codebase. That means that something named "OperationIds" is misleading, and critical consequences of that are currently seen in the codebase.

It should therefore split in two (OrderedOperationIds and UnorderedOperationIds), but I feel `Vec<OperationId>` and `Set<OperationId>` is just shorter and more explicit

massa-protocol-worker/src/protocol_network.rs
```diff
@@ -0,0 +1,426 @@
...
+            let operations_ids = match self.storage.retrieve_block(hash) {
+                Some(wrapped_block) => wrapped_block.read().content.operations.clone(),
+                None => {
+                    // let the node know we don't have the block.
+                    all_blocks_info.push((*hash, ReplyForBlocksInfo::NotFound));
+                    continue;
+                }
+            };
+            let block_info = match info_wanted {
+                AskForBlocksInfo::Info => ReplyForBlocksInfo::Info(operations_ids),
```
damip, line 179: operations_ids needs to be ordered ( not a Set )

massa-protocol-worker/src/protocol_network.rs
```diff
@@ -0,0 +1,426 @@
...
+                AskForBlocksInfo::Operations(op_ids) => {
```
damip, line 180: op_ids needs to be ordered ( not a Set )

massa-protocol-worker/src/protocol_network.rs
```diff
@@ -0,0 +1,426 @@
...
+                        &[*hash],
+                        true,
+                        Instant::now(),
+                        self.config.max_node_known_blocks_size,
+                    );
+
+                    // Send only the missing operations.
+                    let needed_ops = operations_ids
+                        .into_iter()
+                        .filter(|id| op_ids.contains(id))
```
damip, line 192: * no need to perform this `filter`
* first make sure that when the `AskForBlocksInfo::Operations` message is deserialized, we fail if there are too many ops being asked (constant config parameter)
* before we do `send_block_info` on response to `AskForBlocksInfo::Operations` we should first gather all the asked operations from storage, and if any is absent, send NotFound
* if all are found, send all the WrappedOperations throught send_block_info (and not just their IDs)

massa-protocol-worker/src/protocol_network.rs
```diff
@@ -0,0 +1,426 @@
...
+    /// - `operation_ids`: IDs of the operations contained by the block.
+    ///
+    /// # Result
+    /// return an error if stopping asking block failed. The error should be forwarded at the
+    /// root. todo: check if if make panic.
+    async fn on_block_operation_list_received(
+        &mut self,
+        from_node_id: NodeId,
+        block_id: BlockId,
+        operation_ids: OperationIds,
```
damip, line 227: operation_ids should be ordered

massa-protocol-worker/src/protocol_network.rs
```diff
@@ -0,0 +1,426 @@
...
+            }
+        };
+        let mut total_hash: Vec<u8> = vec![];
+        operation_ids.iter().for_each(|op_id| {
+            let op_hash = op_id.hash().into_bytes();
+            total_hash.extend(op_hash);
+        });
+
+        // Check operation_list against expected operations hash from header.
+        if info.header.content.operation_merkle_root == Hash::compute_from(&total_hash) {
```
damip, line 247: let's rename operation_merkle_root to operation_global_hash

damip: to avoid 1 level of indentation, use:

```
if info.header.content.operation_merkle_root != Hash::compute_from(&total_hash) {
    let _ = self.ban_node(&from_node_id).await;
    return Ok(());
}
```

massa-network-exports/src/commands.rs
```diff
@@ -138,20 +135,41 @@ pub enum NodeEventType {
...
+    Operations(OperationIds),
+}
+
+/// Reply with the info about a block.
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub enum ReplyForBlocksInfo {
+    /// The info about the block is required(list of operations ids).
+    Info(OperationIds),
+    /// The actual operations required.
+    Operations(OperationIds),
```
damip, line 154: I don't understand where the actual operations are sent.
Shouldn't this be `Operations(Vec<WrappedOperation>),` ?

damip (changes requested):

massa-network-exports/src/commands.rs
damip: @gterzian  see this comment and the ones above

massa-models/src/block.rs
AurelienFT: Done.

AurelienFT:
[followup list](https://github.com/massalabs/massa/pull/2862#issuecomment-1210462093)

massa-protocol-worker/src/protocol_network.rs
damip: @gterzian

massa-protocol-worker/src/protocol_network.rs
damip: @gterzian

damip:
[followup list](https://github.com/massalabs/massa/pull/2862#issuecomment-1210462093)

massa-network-exports/src/commands.rs
gterzian: @damip The actual operations are obtained from storage at https://github.com/massalabs/massa/pull/2862/files#diff-2350ca2a3fef681c7029e38fa91036024b30b38322f5807357e98ec037030873R200

gterzian:
updated list: https://github.com/massalabs/massa/pull/2862#issuecomment-1210462093

massa-network-exports/src/commands.rs
gterzian: Actually I guess we can change this to use the new storage system?Will file an issue.

massa-pos-worker/src/controller.rs
```diff
@@ -61,22 +62,23 @@ impl SelectorController for SelectorControllerImpl {
...
-            cache_lock.read().as_ref().map_err(|err| err.clone())?;
-        }
-        {
-            // send command
-            let (cvar, lock) = &*self.input_data;
-            let mut data = lock.lock();
-            data.push_back(Command::DrawInput {
+        // check status
+        let (_cache_cv, cache_lock) = &*self.cache;
+        cache_lock.read().as_ref().map_err(|err| err.clone())?;
```
gterzian, line 67: It looks like the lock is still being held when sending on the channel, which could result in deadlock if:
1. the channel is full(blocking send)
2. the worker is blocked trying to acquire a write lock(see `process_draws_result`)

@damip

damip: good catch ! solved

massa-protocol-worker/src/protocol_network.rs
damip: discussed in https://github.com/massalabs/massa/issues/2884

massa-protocol-worker/src/protocol_network.rs
damip: @gterzian was this solved ?

damip:
Merging !

Issue #2854, opened 2022-08-02
Add ARM build in CI for release.


AurelienFT:
Hello the Massa team,

It would be great if, when you release a new version, a package with ARM architecture compatible binaries is also generated.

This is not a blocking issue because ARM processor users can build the binaries directly from the sources but it would be more convenient if the binaries would be included in the release.

_Originally posted by @fsidhoum in https://github.com/massalabs/massa/discussions/2812_

fsidhoum:
I am working in this issue.

Should we generate an ARM64 package for windows and macOS too, or only for linux ?

AurelienFT:
> I am working in this issue.
> 
> Should we generate an ARM64 package for windows and macOS too, or only for linux ?

The more it's the best IMO

qdrn:
> I am working in this issue.
> 
> Should we generate an ARM64 package for windows and macOS too, or only for linux ?

It is likely that you will need to update the docker image that is currently used by the CI/CD. If that's the case then ping me so that I can update it.

fsidhoum:
> > I am working in this issue.
> > Should we generate an ARM64 package for windows and macOS too, or only for linux ?
> 
> It is likely that you will need to update the docker image that is currently used by the CI/CD. If that's the case then ping me so that I can update it.

I managed to build an ARM package for linux. To do so, I didn't change the Docker image used by Github Actions. I used the argument "use-cross" in the "actions-rs/cargo@v1" step.

Changing the "ubuntu-18.04" image will not have any affect while the Github Actions runner use an x86_64 architecture. And there is not arm64 public runner available.

I have done some tests with the arm64 binairies : this works great.
I have to do some more for x86_64 binairies, then I will be ready to do a pull request.

qdrn:
> > > I am working in this issue.
> > > Should we generate an ARM64 package for windows and macOS too, or only for linux ?
> > 
> > 
> > It is likely that you will need to update the docker image that is currently used by the CI/CD. If that's the case then ping me so that I can update it.
> 
> I managed to build an ARM package for linux. To do so, I didn't change the Docker image used by Github Actions. I used the argument "use-cross" in the "actions-rs/cargo@v1" step.
> 
> Changing the "ubuntu-18.04" image will not have any affect while the Github Actions runner use an x86_64 architecture. And there is not arm64 public runner available.
> 
> I have done some tests with the arm64 binairies : this works great. I have to do some more for x86_64 binairies, then I will be ready to do a pull request.

Great!

[...]

PR #2864, opened 2022-08-09
Add the ARM64 build target in cd.yml

fsidhoum:
This pull request is linked to the issue #2854.

It adds, when releasing, the generation of a package for linux with arm64 processor architecture.

To build the arm64 binaries on a Github Actions runner with x86_64 processor architecture, it uses [cross](https://github.com/cross-rs/cross).

To enable cross, it passes the value `true` to the argument `use-cross` of the Github Action `actions-rs/cargo@v1`.

When compiling, `cross` creates a new Docker container and build the sources inside it. But, by default, the container created doesn't provide the openssl library. To tell `cross` to install some dependancies before compiling, I have created the config file `Cross.toml`. In this file, for each target architectures, we can define the docker image used by `cross` and a command executed before building.

The more targets we want, the more complex the config will be. That's why, I kept only the arm64 linux target. I think that the others arm64 architectures will not be so expected by users. Moreover, it uses `cross` only for arm64 architectures to not have to add some configs.

Links :
- [Release](https://github.com/fsidhoum/massa/releases/tag/TEST.13.0)
- [CI Workflow](https://github.com/fsidhoum/massa/actions/runs/2820017800)
- [CD Workflow](https://github.com/fsidhoum/massa/actions/runs/2820019853)

### DIFF
--- a/.github/workflows/cd.yml
+++ b/.github/workflows/cd.yml
@@ -14,16 +14,29 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-18.04, macOS-latest, windows-latest]
+        platform: [x86_64, arm64]
         include:
           - os: ubuntu-18.04
-            target: x86_64-unknown-linux-musl
+            platform: x86_64
+            target: x86_64-unknown-linux-gnu
             name: release_linux.tar.gz
+          - os: ubuntu-18.04
+            platform: arm64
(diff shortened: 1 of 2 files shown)


qdrn (approved):

qdrn:
bors merge

PR #2877, opened 2022-08-17
Fix public API adding storage and pool function

adrien-zinger:


massa-storage/src/lib.rs
```diff
@@ -539,6 +536,47 @@ impl Storage {
     pub fn get_block_indexes(&self) -> &Arc<RwLock<BlockIndexes>> {
         &self.blocks
     }
+
+    /// Get all operations created by an address
```
AurelienFT, line 540: I think the documentation is wrong here.

massa-storage/src/lib.rs
```diff
@@ -539,6 +536,47 @@ impl Storage {
...
+    pub fn get_operations(&self) -> Operations {
```
AurelienFT, line 541: Maybe specify that is will get you the locals only

massa-api/src/public.rs
```diff
@@ -361,30 +357,24 @@ impl Endpoints for API<Public> {
...
-                //TOxDO: https://github.com/massalabs/massa/issues/2866
-                //pool_command_sender.get_operations(operation_ids.clone()),
-                async { Ok((Map::default(), Storage::default())) },
-                consensus_command_sender.get_operations(operation_ids)
-            );
-            let (pool_res, consensus_res) = (pool_res?, consensus_res?);
+            let consensus_res = consensus_command_sender
+                .get_operations(operation_ids)
+                .await?;
+            let pool_res = pool_command_sender.get_operations();
```
AurelienFT, line 363: The get operations is made with specific ids. Then, there is no other usage of `get_operations()` and so I think we shouldn't create this function because it's very heavy to run and don't think there is use case for this.

AurelienFT (commented):
General note : I have removed Operations and OperationsIds in #2878 so if you can avoid using them it will be easier to merge both.

massa-api/src/public.rs
adrien-zinger: I'll take by id :+1: but I need the operations bellow. I don't get your second point.

damip:
@adrien-zinger can you solve the conflicts so we can merge :) ?

damip:
bors merge

PR #2878, opened 2022-08-17
Remove shortened operations name and maintain operation order in blocks.

### LINKED ISSUE #2874: Remvoe OperationIds type redefinition and correct ordering
MAKE A PR TOWARDS BRANCH [release_testnet_14](https://github.com/massalabs/massa/tree/release_testnet_14)

Operation order in blocks matters:

* operation order is chosen by the block producer
* operations are executed in the order they appear in the block
* the operation global hash in the header is sensitive to operation order
* it is therefore consensus-critical for the operations order to be the same for everyone for every block

However, OperationIds seems to be used almost everywhere, even in places where operation order should matter (see https://github.com/massalabs/massa/pull/2862#discussion_r946848973 but also in Protocol etc...)

Given how little benefit and how huge of a problem this type redefinition brings, I think it would be smart to completely remove it and write explicitly `Set<OperationId>` or `Vec<OperationId>` depending on the situation. It wasn't really that much shorter anyways.

AurelienFT:
Fix #2874

### DIFF
--- a/massa-protocol-worker/src/checked_operations.rs
+++ b/massa-protocol-worker/src/checked_operations.rs
@@ -21,7 +21,7 @@ impl CheckedOperations {
         self.0.insert(prefix, *id).is_none()
     }
 
-    /// Get a set of [OperationIds] matching with the givec `prefix`.
+    /// Get a set of [Set<OperationId>] matching with the givec `prefix`.
     pub fn get(&self, prefix: &OperationPrefixId) -> Option<&OperationId> {
         self.0.get(prefix)
     }
--- a/massa-storage/src/endorsement_indexes.rs
+++ b/massa-storage/src/endorsement_indexes.rs
@@ -16,6 +16,9 @@ pub struct EndorsementIndexes {
 }
 
 impl EndorsementIndexes {
+    /// Insert a batch of endorsements and populate the indexes.
+    /// Arguments:
+    /// - endorsements: the endorsements to insert
     pub(crate) fn batch_insert(&mut self, endorsements: Vec<WrappedEndorsement>) {
         for endorsement in endorsements {
             let id = endorsement.id;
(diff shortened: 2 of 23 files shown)


adrien-zinger (approved):

massa-protocol-worker/src/checked_operations.rs
```diff
@@ -21,7 +21,7 @@ impl CheckedOperations {
         self.0.insert(prefix, *id).is_none()
     }
 
-    /// Get a set of [OperationIds] matching with the givec `prefix`.
+    /// Get a set of [Set<OperationId>] matching with the givec `prefix`.
```
damip, line 24: To me it looks more like the function returns an Option with just a single operation ID

massa-storage/src/endorsement_indexes.rs
```diff
@@ -40,28 +46,47 @@ impl EndorsementIndexes {
...
+    /// - endorsement_ids: the endorsements to link to the block
     pub(crate) fn link_endorsements_with_block(
         &mut self,
         block_id: &BlockId,
-        operations: &Vec<EndorsementId>,
+        endorsement_ids: &Vec<EndorsementId>,
     ) {
         self.index_by_block
             .entry(*block_id)
             .or_default()
```
damip, line 60: This won't insert the element if it is missing.

Use `.or_insert_with(|| Default::default())`

massa-storage/src/operation_indexes.rs
```diff
@@ -7,12 +10,15 @@ pub struct OperationIndexes {
     /// Operation structure container
     pub(crate) operations: Map<OperationId, WrappedOperation>,
     /// Structure mapping creators with the created operations
-    index_by_creator: Map<Address, OperationIds>,
+    index_by_creator: Map<Address, Set<OperationId>>,
     /// Structure mapping block ids with the operations
-    index_by_block: Map<BlockId, OperationIds>,
+    index_by_block: Map<BlockId, Vec<OperationId>>,
```
damip, line 15:
```suggestion
    index_by_block: Map<BlockId, Set<OperationId>>,
```

AurelienFT:
Done in https://github.com/massalabs/massa/pull/2862/commits/30d79ddb16024dd6576145fb337b7371b9ea7f85

PR #2885, opened 2022-08-18
Add two indexes for block.

### LINKED ISSUE #2881: Index to find to which blocks a given operation/endorsement is included
Add an index to Storage to easily find into which blocks a given operation/endorsement is included.

This should probably go into BlockIndices as `index_by_op: Map<OperationId, Set<BlockId>>` and  `index_by_endorsement: Map<EndorsementId, Set<BlockId>>`

AurelienFT:
Fix #2881


AurelienFT:
Close in favor of #2890

PR #2888, opened 2022-08-18
Add protocol storage management

adrien-zinger:
Storage in protocol + remove old protocol-pool events.

I also comment some unit tests, I need first to rebase on latest release 14...


massa-protocol-worker/src/protocol_worker.rs
```diff
@@ -903,15 +877,17 @@ impl ProtocolWorker {
...
+
+            // Store operation, claim locally
+            let mut ops = self.storage.clone_without_refs();
+            ops.store_operations(new_operations.into_values().collect());
+            self.storage
+                .claim_operation_refs(ops.get_local_operation_ids());
+
+            // Add to pool, propagate when received outside of a header.
+            self.pool_controller.add_operations(ops);
         }
```
adrien-zinger, line 891: :warning: I changed the place of the prune (I think it's better to prune first)... But maybe we should also prune only the oldest right? In a follow up

gterzian (commented):
I'm wondering if we don't still require an argument to `note_operations_from_node` that explicitly restricts propagation when the operation was received "within a block", because in the current logic it seems that if we see an operation for the first time(as per `checked_operations`), then it will be propagated even though if it came from within a block. 

Or perhaps that is what we want to do from now on?

gterzian:
In the process of reviewing the whole thing... @adrien-zinger

massa-consensus-worker/src/consensus_worker.rs
```diff
@@ -681,6 +682,7 @@ impl ConsensusWorker {
 
                 // Store block in shared storage.
                 self.block_db.storage.store_block(block);
+                self.block_db.storage.extend(storage);
```
AurelienFT, line 685: The block should be in the storage also and so we avoid passing it in the event we just pass the block id ?

massa-pool-worker/src/tests/mod.rs
```diff
@@ -1,7 +1,7 @@
 // Copyright (c) 2022 MASSA LABS <info@massa.net>
 
-mod config;
-mod mock_protocol_controller;
-mod operation_pool_tests;
-mod scenario;
-pub(crate) mod tools;
+//mod config;
```
AurelienFT, line 3: I don't get this.

massa-protocol-worker/src/protocol_network.rs
```diff
@@ -365,19 +358,28 @@ impl ProtocolWorker {
...
+            .clone()
+            .unwrap()
+            .into_iter()
+            .enumerate()
+            .map(|(idx, op_id)| (op_id, idx))
+            .collect();
+
+        // Let the consensus store the block, operations and endorsements are already stored
+        // with [note_operations_from_node] and [note_endorsements_from_node]
+        let mut storage = self.storage.clone_without_refs();
```
AurelienFT, line 372: Related to my previous comment we can claim here to have in the storage

massa-protocol-worker/src/protocol_worker.rs
```diff
@@ -903,15 +877,17 @@ impl ProtocolWorker {
...
             self.prune_checked_operations();
+
+            // Store operation, claim locally
+            let mut ops = self.storage.clone_without_refs();
+            ops.store_operations(new_operations.into_values().collect());
+            self.storage
+                .claim_operation_refs(ops.get_local_operation_ids());
+
+            // Add to pool, propagate when received outside of a header.
+            self.pool_controller.add_operations(ops);
```
gterzian, line 890: Since this is a sync call, do you need to pass a storage to it? It would seem that you can just pass the operations directly, and pool will store them before the call returns?

massa-protocol-worker/src/protocol_worker.rs
```diff
@@ -903,15 +877,17 @@ impl ProtocolWorker {
...
-            .await;
-
             // prune checked operations cache
             self.prune_checked_operations();
+
+            // Store operation, claim locally
+            let mut ops = self.storage.clone_without_refs();
+            ops.store_operations(new_operations.into_values().collect());
+            self.storage
+                .claim_operation_refs(ops.get_local_operation_ids());
```
gterzian, line 887: Why do you need to claim them locally? Could you not simply have `note_operations_..` return a storage which owns the newly processed operations, and send this on a message to consensus?

gterzian (commented):
Looks good overall, and I have a few questions about the use of storage. 

Can we write down something about how the "ownership" of objects is managed between protocol, pool, and consensus? I would have assumed that protocol wouldn't need to "own" references directly, just have access to storage and let pool and consensus manage ownership.

AurelienFT:
If you can check this comment @adrien-zinger also : https://github.com/massalabs/massa/pull/2897#discussion_r952418170 :) Thank you !

massa-protocol-worker/src/protocol_worker.rs
damip: eventually, pool `add_operations` should not be sync anymore, it should be a fire-and-forget mpsc towards a separate thread to avoid blocking protocol when performing parsing, ordering and pruning on the pool side

massa-protocol-worker/src/protocol_worker.rs
damip: sounds good for now

massa-consensus-worker/src/consensus_worker.rs
damip: can you recheck this with the latest conflict resolution ?

massa-pool-worker/src/tests/mod.rs
damip: can you recheck this with the latest conflict resolution ?

massa-protocol-worker/src/protocol_network.rs
damip: can you recheck this with the latest conflict resolution ?

massa-protocol-worker/src/protocol_worker.rs
damip: @gterzian here is a doc: https://github.com/massalabs/massa/pull/2901

massa-protocol-exports/src/protocol_controller.rs
```diff
@@ -30,7 +25,7 @@ pub enum ProtocolEvent {
         block_id: BlockId,
         /// block slot
         slot: Slot,
-        /// storage instance containing the block and its dependencies (except the parents)
+        /// storage containing endorsements and operations
```
damip, line 28:
```suggestion
         /// storage instance containing the block and its dependencies (except the parents)
```

massa-pool-worker/src/tests/mod.rs
```diff
@@ -1,7 +1,7 @@
...
+//mod mock_protocol_controller;
+//mod operation_pool_tests;
+//mod scenario;
+//pub(crate) mod tools;
```
damip, line 7: why is this commented in this PR ?

massa-protocol-exports/src/tests/tools.rs
```diff
@@ -278,26 +275,26 @@ where
...
+//    loop {
+//        tokio::select! {
+//            evt_opt = protocol_event_receiver.wait_event() => match evt_opt {
+//                Ok(orig_evt) => if let Some(res_evt) = filter_map(orig_evt) { return Some(res_evt); },
+//                _ => return None
+//            },
+//            _ = &mut timer => return None
+//        }
+//    }
+//}
```
damip, line 297: there are no more protocol pool events, this should be removed

massa-protocol-worker/src/tests/mod.rs
```diff
@@ -1,10 +1,10 @@
...
-mod scenarios;
-mod tools;
+//mod ask_block_scenarios;
+//mod ban_nodes_scenarios;
+//mod cache_scenarios;
+//mod endorsements_scenarios;
+//mod in_block_operations_scenarios;
+//mod operations_scenarios;
+//mod scenarios;
+//mod tools;
```
damip, line 10: let's uncomment this

damip (commented):
Some missing things:
* manage storage in `ProtocolCommand::IntegratedBlock { block_id, storage }`
* use storage when communicating with pool (see doc)

damip:
> still require an argument to `note_operations_from_node` that explicitly restricts

Indeed, for now we should not propagate operations that are propagating within a block

damip:
> I'm wondering if we don't still require an argument to `note_operations_from_node` that explicitly restricts propagation when the operation was received "within a block", because in the current logic it seems that if we see an operation for the first time(as per `checked_operations`), then it will be propagated even though if it came from within a block.
> 
> Or perhaps that is what we want to do from now on?

> still require an argument to `note_operations_from_node` that explicitly restricts

@gterzian  This is not necessary anymore.
We already use the node's knowledge to send the ops only to the nodes that haven't seen it before.
On the other hand, when a block is announced to a node, the node only asks for the operations that it doesn't already have.

PR #2889, opened 2022-08-18
Update api.rst

rykcod:
Suite au constat de @GGCom sur discord


AurelienFT:
Hello,

We have move the doc to an openrpc.json that is on the same folder. It's not up to date right now and we are updating it for the next testnet (14) in this PR #2862 . The modification is already there and much more !

However thanks for your contribution ! I will close the PR.

Issue #2883, opened 2022-08-18
Ensure that Storage and Pool are used properly in the Protocol module


gterzian:
Part of https://github.com/massalabs/massa/pull/2862

gterzian:
Note that the interface to pool has changed, actually in the [new interface](https://github.com/massalabs/massa/pull/2862/files#diff-fdd55872f57360f9a4ae73ab6a3988a750d5d35c9786f104f883f99936e353c4R16) I don't see a mechanism currently that matches the previous "propagate these operations" as was available in https://github.com/massalabs/massa/blob/d315aadbdc25a91b4bb160952c56d7026748fcef/massa-pool/src/pool_worker.rs#L326

Should we add such a mechanism? It could be used to avoid propagating operations received inside the new "block info" workflow(see https://github.com/massalabs/massa/pull/2862/files#diff-cd40aa0563a3ce2721769e8d6fb242a0e813ab40a0c119977fc9b144505ba7cdR58).  cc @massalabs/core-team

damip:
> Note that the interface to pool has changed, actually in the [new interface](https://github.com/massalabs/massa/pull/2862/files#diff-fdd55872f57360f9a4ae73ab6a3988a750d5d35c9786f104f883f99936e353c4R16) I don't see a mechanism that matches the previous "propagate these operations" as was available in
> 
> https://github.com/massalabs/massa/blob/d315aadbdc25a91b4bb160952c56d7026748fcef/massa-pool/src/pool_worker.rs#L326
> 
> Should we add such a mechanism? It could be used to avoid propagating operations received inside the new "block info" workflow(see https://github.com/massalabs/massa/pull/2862/files#diff-cd40aa0563a3ce2721769e8d6fb242a0e813ab40a0c119977fc9b144505ba7cdR58). cc @massalabs/core-team

We need something more advanced, see followup: https://github.com/massalabs/massa/issues/2904

PR #2890, opened 2022-08-18
Improve storage, pos, consensus, API

damip:
I spent some time reading the whole refactoring PR and improving it during the week end to minimize collisions. Please take a look.

Things that I did NOT touch here to not collide too much with other work in progress:
* PoS/roll bootstrap (@Eitu33)
* Proper use of Storage in Protocol (@adrien-zinger, @gterzian)
* Protocol needs to notify Pool of new endorsements/operations (@gterzian)
* Do not use constants for config (@AurelienFT)
* Tests ( @adrien-zinger , @gterzian )
* Do a pass on the API documentation (TODO)
* Better PoS error management: re-bootstrap on cycles unavailable instead of panicking (TODO)
* Do not panic if the state sent to us by the bootstrap srv is inconsistent in any way, just re-bootstrap (TODO)

I did however:
* improve the API further
* various performance improvements
  * most notably avoiding linear lookups in histories when it can be done in constant time
* improve Storage
  * simplify its interface
  * drew inspiration from https://github.com/massalabs/massa/pull/2885 to implement op/endorsement indices per block (@AurelienFT I basically used your code with some slight performance improvements on deletion)
* remove unused things in Consensus and simplify things there
* simplify and correct the PoS system (@Eitu33 , @adrien-zinger):
  * it wasn't properly following the new PoS algorithm, most notably it was using the same cycle as source for both the seed and the distribution
  * now the selector is fed the cycle to draw, directly together with lookback seed and rolls 
  * the draw algorithm was heavily simplified and its performance improved: used WeightedAliasIndex (https://docs.rs/rand_distr/0.4.3/rand_distr/weighted_alias/struct.WeightedAliasIndex.html) for `O(1)` sampling instead of `O(log(N))`
  * now, the selector is fed existing complete cycles (eg. init or bootstrap) when give_selector_controller is called. Blocks until the cycles that could be drawn are drawn on init.
  * now reading draws when the selector failed in any way will return the proper error
  * the controller can now wait for a given draw to be available
* remove warnings


docs/technical-doc/api.rst
```diff
@@ -58,8 +58,6 @@ slot, timestamp), clique count, connected nodes count.
         "clique_count": Number,
         "end_timespan": Number,// stats time interval, millis since 1970-01-01
         "final_block_count": Number,
-        "final_operation_count": Number,
```
AurelienFT, line 61: Modify the openrpc instead. This has been removed on main.

massa-storage/src/endorsement_indexes.rs
```diff
@@ -8,89 +10,108 @@ use massa_models::{
...
-    pub(crate) fn unlink_endorsements_from_block(&mut self, block_id: &BlockId) {
-        self.index_by_block.remove(block_id);
+    /// - endorsement_id: the endorsement id to remove
+    pub(crate) fn remove(&mut self, endorsement_id: &EndorsementId) -> Option<WrappedEndorsement> {
+        // update to-block index
+        self.index_to_block.remove(endorsement_id);
+
+        if let Some(e) = self.endorsements.remove(endorsement_id) {
+            // update creator index
+            self.index_by_creator
```
AurelienFT, line 76: Maybe delete index by creator entry if empty at the end (applies to all the others (block, operation))

massa-graph/src/bootstrapable_graph.rs
```diff
@@ -22,35 +22,23 @@ use std::ops::Bound::Included;
...
-    /// Best parents hashes in each thread.
-    pub best_parents: Vec<(BlockId, u64)>,
-    /// Latest final period and block hash in each thread.
-    pub latest_final_blocks_periods: Vec<(BlockId, u64)>,
-    /// Head of the incompatibility graph.
-    pub gi_head: Map<BlockId, Set<BlockId>>,
-    /// List of maximal cliques of compatible blocks.
-    pub max_cliques: Vec<Clique>,
+    /// list of final blocks
+    pub final_blocks: Vec<ExportActiveBlock>,
```
AurelienFT, line 26: Can you explain why we don't need the rest ?

AurelienFT (commented):
Nice refactor of the refactor 😆

docs/technical-doc/api.rst
damip: reverted

massa-storage/src/endorsement_indexes.rs
damip: done

massa-graph/src/bootstrapable_graph.rs
damip: basically we won't be transferring non-final blocks (so the cliques we send are empty)

damip:
thanks for the review @AurelienFT  !

massa-protocol-worker/src/protocol_network.rs
```diff
@@ -247,23 +247,25 @@ impl ProtocolWorker {
...
-                            .expect("unexpected checked operation not in storage");
-                        info.operations_size =
-                            info.operations_size.saturating_add(op.serialized_size());
-                        return false;
-                    }
-                    true
-                })
-                .collect();
+            let missing_operations = {
+                let ops = self.storage.read_operations();
```
gterzian, line 251: I prefer the previous `retrieve_operation(operation_id)` API of storage, because here you are holding a (read)lock over all operations in storage while looping over the ids and processing each one, which could result in loss of parallelism.

On the other hand, acquiring a lock for an individual operation at each iteration should be ok, especially since `parking_lot` is optimized for the uncontended case.

> Uncontended lock acquisition and release is done through fast inline paths which only require a single atomic operation (source: https://docs.rs/crate/parking_lot/latest).

This comment sort of applies also to the entire structure of indexes inside storage, since previously we had map containing other arcs/locks, which allowed for quick retrieval of a reference, followed by th acquisition of a lock around a single object. 

Now we have a single lock around for example a `BlockIndexes`, and the coarser lock makes it harder to achieve parallelism in principle. 

Parking lot is designed around the idea of have a lot of very fine grain locks, whereas it now seems we are rather going for a coarser grained(and more complex because of the manual reference counting) design, and I'm not sure why actually. Maybe you can explain the rationale behind this?

massa-protocol-worker/src/protocol_worker.rs
```diff
@@ -336,11 +337,9 @@ impl ProtocolWorker {
         cmd: ProtocolCommand,
         timer: &mut std::pin::Pin<&mut Sleep>,
     ) -> Result<(), ProtocolError> {
-        massa_trace!("protocol.protocol_worker.process_command.begin", {
-            "cmd": cmd
-        });
         match cmd {
-            ProtocolCommand::IntegratedBlock { block_id } => {
+            ProtocolCommand::IntegratedBlock { block_id, storage } => {
+                // TODO properly manage storage
```
gterzian, line 342: You could refer to https://github.com/massalabs/massa/issues/2883

gterzian (approved):
LGTM, although I'm wondering why storage keeps getting more complicated. Also one nit.

massa-protocol-worker/src/protocol_network.rs
damip: Thanks for the comments !

One problem with `retrieve_operation` is that things might have changed within a loop (ie the block containing the list of operations we were reading was dropped). Now we can achieve what we used to do with retrieve_operation easily, but we can also keep a finer locking for this kind of situations.

As for the performance impact, it's a real question, and is hard to know a priori. Here is a little playground to test the perf impact of parking_lot mutexes: https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=2b7c64dfaab0c5d4aefc006a1dab83b4
Looks like we get a x4 performance gain by keeping the same lock.
Even if it uses a single atomic operation, it is a big CPU-wide sync impact, and a common cause of cache misses.

Regarding  per-element locks, we used to have them only around blocks (not ops/endorsements) because we wanted blocks to be subject to interior changes, but this is not the case anymore as the only thing we do is add/remove elements from the global indexes so we need this global lock on the container. Note that we use RwLocks to make the very frequent concurrent reads more efficient.

damip: @gterzian  interesting read: https://spcl.inf.ethz.ch/Publications/.pdf/atomic-bench.pdf

Main take-away: 

`Atomics prevent any instruction level parallelism, significantly limiting the bandwidth (up to 30x in comparison with simple writes), even if there are no dependencies between the successive operation`

massa-protocol-worker/src/protocol_worker.rs
damip: I voluntarily didn't touch this to not collide with https://github.com/massalabs/massa/issues/2883 and https://github.com/massalabs/massa/pull/2888

damip:
Solved all conflicts. merging

AurelienFT (commented):
Good !

massa-protocol-worker/src/protocol_network.rs
gterzian: > Here is a little playground to test the perf impact of parking_lot mutexes

Yes taking the lock, and then looping, is going to be faster than taking the lock at each iteration, from the perspective of a single thread. 

However from the perspective of a multi-threaded system, the one "fast" thread could be causing every other thread to wait for the lock for the duration of the loop. 

So it's not about performance of a single thread, it's about opportunity for parallelism. By the way, if the system is designed around opportunities for parallelism, it is usually quite easy to optimize parts, like some kind of hot loop. However, if the system is not designed for parallelism, it is very hard to introduce it later...

> Basically, any atomic operation might temporarily impact all cores, and might cause cache misses

Interesting article, although I've just scanned the conclusion. 

It seems to apply mostly to parallel data structures, like a parallel map, that do a lot of atomic operations, and less-so to simply having a lock do a single operation(in the fast uncontended case as is the case with parking lot), followed by simply using the underlying locked structure without further atomic operations. 

Basically, it's a good argument against using complicated parallel data structures, vs simple locks. Thanks for sharing it. 

@damip

Issue #2884, opened 2022-08-18
Fix response to AskForBlocksInfo::Operations


gterzian:
* no need to perform this `filter`
* first make sure that when the `AskForBlocksInfo::Operations` message is deserialized, we fail if there are too many ops being asked (constant config parameter)
* before we do `send_block_info` on response to `AskForBlocksInfo::Operations` we should first gather all the asked operations from storage, and if any is absent, send NotFound
* if all are found, send all the WrappedOperations throught send_block_info (and not just their IDs)
* make sure the order of operations in the block are preserved

_Originally posted by @damip in https://github.com/massalabs/massa/pull/2862#discussion_r946954656_

gterzian:
@damip I have a few questions:

> send all the WrappedOperations throught send_block_info (and not just their IDs)

Currently, the idea is that protocol sends a message to network saying "send those operations", using the IDs, and network will retrieve the serialized form of the op from storage when sending the message. 

Note that network doesn't need the wrapped op, just the serialized form. 

> make sure the order of operations in the block are preserved

The current idea is that the operations and the block are separated, so the "order" in which operations are received from the network seems irrelevant. The only thing that matter, I think, is:

1. We store a list of operation for the block(in order). 
2. We have all those operations present in storage. 

So passing around a `Vec<OperationId>` in the (network) message, instead of a set, does not seem to contributed to the above goals. 

> we should first gather all the asked operations from storage, and if any is absent, send NotFound

If a node replies to another node with info on the block, it has the block, so it should also have all the operations as well.  

> no need to perform this filter

Actually I think we should strengthen it and make sure we ops are part of the block, to avoid having a node ask for nonsense and reply with potentially non-existing ops(which would fail when network would try to retrieve them from storage). 


Re all of the above, we can change the current workflow and we should clarify exactly what we want to do here.

gterzian:
It should work something like this:

1. Node A has a block(integrated it). Therefore it also has the ordered list of ops.
2. Node A propagates header.
3. Node B receives header.
4. Node B adds block to wishlist, asks Node A for info.
5. Node A sends `BlockInfoReply::Info(list)`, where `list` is the ordered list of ops. 
6. Node B asks for a set of missing ops.
7. Node A sends the set of missing ops. 

So basically, `BlockInfoReply::Info(list)` should contain the ordered list, while the rest of the workflow to get the actual missing ops does not need an ordered list.

damip:
@gterzian 

> @damip I have a few questions:
[...]

PR #2891, opened 2022-08-19
Block info msg fix

gterzian:
Serialization part of https://github.com/massalabs/massa/issues/2884


gterzian:
@damip Ready for review. This one basically just chunks outgoing "block info" messages to avoid sending to many at once. When receiving the messages we already check the length of it inside the deserializer...

gterzian:
@AurelienFT Do you want to review?

AurelienFT (commented):
I don't have the full scope of this change but it's complicated to follow here. I think we might consider splitting the code in another function.

massa-network-worker/src/node_worker.rs
```diff
@@ -160,64 +193,80 @@ impl NodeWorker {
...
-                                            for operation_id in &operation_ids {
-                                                match op_store.get(operation_id) {
-                                                    Some(operation) => {
-                                                        wrapped_operation_serializer
-                                                            .serialize(operation, &mut res)?;
-                                                    }
-                                                    None => {
-                                                        return Err(NetworkError::MissingOperation)
+                                for replies in
+                                    reply_list.chunks(MAX_ASK_BLOCKS_PER_MESSAGE as usize)
```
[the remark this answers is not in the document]
damip: I think we should put all test-default and runtime constants in a separate crate and only include it in main and the tests that need them. That way we will solve the problem of people using them everywhere. I opened a followup: https://github.com/massalabs/massa/issues/2898

gterzian:
Ok thanks for the review, I have replaced the constant with the use of the config... @AurelienFT

AurelienFT:
> Ok thanks for the review, I have replaced the constant with the use of the config... Re-splitting it into another function, I don't really see how it would make it clearer, actually personally I think it might make it less so, and in any case it appears like a minor issue, so maybe lets just leave it like that for now? @AurelienFT

Ok ! no problem

AurelienFT (approved):

massa-network-worker/src/node_worker.rs
```diff
@@ -7,7 +7,7 @@ use super::{
 use itertools::Itertools;
 use massa_logging::massa_trace;
 use massa_models::{
-    constants::{MAX_ASK_BLOCKS_PER_MESSAGE, MAX_ENDORSEMENTS_PER_MESSAGE, NODE_SEND_CHANNEL_SIZE},
+    constants::{MAX_ENDORSEMENTS_PER_MESSAGE, NODE_SEND_CHANNEL_SIZE},
```
damip, line 10: the code still uses constants, let's get rid of all of them

damip: opened a followup

PR #2920, opened 2022-08-25
Fix node basic features in sandbox.

### LINKED ISSUE #2917: Repair factory tests
Make the tests of Factory compile and pass

AurelienFT:
Fix included here : 

- Stop the node where stuck in protocol stop
- Endorsement pool crash at launch because try to fetch endorsements that are not created at first period.
- Initial rolls not stored in `cycle_history`
- Get block endpoint API was override
- Fix #2917 

Working on : 
- No endorsements in block
- No rewards for block creator
- Client formatting

More to come in this PR...

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2019,6 +2019,7 @@ dependencies = [
  "tempfile",
  "thiserror",
  "tokio",
+ "tracing",
 ]
 
 [[package]]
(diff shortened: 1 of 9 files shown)


damip:
Very nice !
Let's make separate PRs for each correction, to let people building tests use the corrections we make here without colliding

PR #2921, opened 2022-08-25
Enhancements for storage in protocol

adrien-zinger:
- Keep a storage while getting block information to ensure that we don't drop the operations.
- Replace Set<OperationId> with a storage in Propagate operation.

### DIFF
--- a/massa-protocol-worker/src/checked_operations.rs
+++ b/massa-protocol-worker/src/checked_operations.rs
@@ -22,7 +22,9 @@ impl CheckedOperations {
     }
 
     pub fn extend(&mut self, ids: &Set<OperationId>) {
-        let _ = ids.iter().map(|id| self.insert(id));
+        ids.iter().for_each(|id| {
+            self.insert(id);
+        });
     }
 
     /// Get a operation id matching with the givec `prefix` or None if there is none.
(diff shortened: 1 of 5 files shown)


massa-protocol-worker/src/checked_operations.rs
```diff
@@ -21,6 +21,10 @@ impl CheckedOperations {
         self.0.insert(prefix, *id).is_none()
     }
 
+    pub fn extend(&mut self, ids: &Set<OperationId>) {
+        let _ = ids.iter().map(|id| self.insert(id));
```
damip, line 25:
```suggestion
        ids.iter().for_each(|id| self.insert(id));
```

damip (approved):
Very good !

massa-protocol-worker/src/protocol_network.rs
```diff
@@ -199,6 +200,18 @@ impl ProtocolWorker {
             })
     }
 
+    /// Return the sum of all operation's serialized sizes in the Set<Id>
+    fn get_total_operations_size(storage: &Storage, operation_ids: Set<OperationId>) -> usize {
```
damip, line 204:
```suggestion
    fn get_total_operations_size(storage: &Storage, operation_ids: &Set<OperationId>) -> usize {
```

PR #2924, opened 2022-08-26
Do not create initial cycle when the node did not bootstrap

Eitu33:


Eitu33:
@damip @AurelienFT behaviour has been tested

damip:
I think it's "sequential", not "consequential" :)

damip:
Out of curiosity: why remove it AFTER bootstrap, and not clear beforehand if we know that we will be bootstrapping ? I feel like it would more logical and much less error prone than keeping this artifact for so long

Eitu33:
The initial cycle is now not created if the bootstrap happened, rather than creating it all the time and deleting it afterward.

Putting this in draft for now as the bootstrap test if failing with the new behaviour.

Eitu33:
Done.

AurelienFT (approved):

PR #2926, opened 2022-08-26
Fix block rewards not given.

AurelienFT:
Not all rewards are given.


massa-consensus-worker/src/consensus_worker.rs
```diff
@@ -572,7 +572,7 @@ impl ConsensusWorker {
             .clone()
             .into_iter()
             .filter_map(|b_id| match self.block_db.get_active_block(&b_id) {
-                Some((a_b, storage)) if a_b.is_final => {
+                Some((a_b, storage)) => {
                     Some((a_b.slot, (a_b.block_id, storage.clone_without_refs())))
```
damip, line 576: this `clone_without_refs` as well as the one on line 566 should be clone()

massa-consensus-worker/src/consensus_worker.rs
```diff
@@ -572,7 +572,7 @@ impl ConsensusWorker {
             .clone()
```
damip, line 572: remove this clone.

Also, line562, the clone should be removed as well

damip (changes requested):

damip (approved):

PR #2930, opened 2022-08-27
Regorganize config

damip:
In this PR, I did the following:
* simplified the management of configs
* reorganized massa-models with the goal of being able to split the crate in the future
* removed bad implementations of thread selection related to sharding (eg. wrapped_block.thread was set to the block creator's thread)
* removed many usages of config constants in the code
* simplified some config passing

Note on broken tests:
* pool => everything ok except "test_pool" that does not respond to its mock mpsc (not implemented)
* factory tests => problem: endorsement creation in parallel to the expected block creation interferes with expected messages


damip:
@massalabs/core-team FYI, please rebase your PRs on this

PR #2934, opened 2022-08-29
Critical execution tests.

Eitu33:


massa-execution-worker/src/speculative_roll_state.rs
```diff
@@ -156,16 +155,19 @@ impl SpeculativeRollState {
...
+            .entry(*creator)
+            .or_default();
+        if let Some(id) = block_id {
+            production_stats.block_success_count =
+                production_stats.block_success_count.saturating_add(1);
+            self.added_changes.seed_bits.push(id.get_first_bit());
+        } else {
+            production_stats.block_failure_count =
+                production_stats.block_failure_count.saturating_add(1);
+            self.added_changes.seed_bits.push(slot.get_first_bit());
```
damip, line 170: Can you explain this change ?

Eitu33: In the previous behaviour, if you were to update the production stats of an address that creates or misses its first assigned block, nothing will happen.

AurelienFT (approved):

PR #2916, opened 2022-08-25
Protocol tests: block propagation

gterzian:
FIX https://github.com/massalabs/massa/issues/2907

PR #2937, opened 2022-08-29
Clean up massa-network

AurelienFT:
This PR clean the code of massa-network : 

- Remove `ToSend` enum that was introduced for shared storage and old serialization format that was incompatible. Now it's useless
- So now we send directly Message to binders so that the serialization/deserialization is only in binders.
- Remove the chunks of block informations has it's already chunk by the parent function and also was bugged because the `send_data` call was in a too much inner loop.
- Pass full object to network and therefore remove usage of storage and fetch elements directly in protocol.
- So the two enum for blocks infos was regrouped.

This also fix a bug of nodes that disconnect because the serializing was bugged because the `send_data` was in the wrong loop and so it sent only the first element in the first message that serialize the full size. of the list.

This PR has been tested in simulator.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2035,7 +2035,6 @@ dependencies = [
  "massa_network_exports",
  "massa_serialization",
  "massa_signature",
- "massa_storage",
  "massa_time",
  "nom 7.1.1",
  "num_enum",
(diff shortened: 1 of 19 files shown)


damip:
@AurelienFT  Good catch for the bug, and very welcome code improvements !

@gterzian FYI

gterzian (approved):
LGTM, I actually need the fix in https://github.com/massalabs/massa/pull/2916

gterzian:
bors merge

AurelienFT:
Can't use bors CI don't pass.

AurelienFT:
bors cancel

PR #2940, opened 2022-08-30
Fix ops propagation

AurelienFT:
Remove duplicate constat that had a bad value causing the node that receive operations to ban the node.


damip:
Bravo !

PR #2941, opened 2022-08-30
Change storage default.

### LINKED ISSUE #2939: Remove Storage::default() and call it Storage::create_root()
Default is error prone as it incites creating new Storage objects instead of deriving them from the root one

AurelienFT:
Close #2939


damip:
perfect !

PR #2943, opened 2022-08-30
Add back desync.

AurelienFT:
I just uncommented. But I don't think it's enough if you have more precisions @damip .


massa-consensus-worker/src/consensus_worker.rs
```diff
@@ -34,8 +41,10 @@ pub struct ConsensusWorker {
     latest_final_periods: Vec<u64>,
     /// clock compensation
     clock_compensation: i64,
-    /// Final block stats `(time, creator)`
-    final_block_stats: VecDeque<(MassaTime, Address)>,
+    /// Final block stats `(time, creator, is_from_protocol)`
+    final_block_stats: VecDeque<(MassaTime, Address, bool)>,
+    /// Blocks that come from protocol used for stats and ids are removed when inserted in `final_block_stats`
```
damip, line 46: What if the block never becomes final because it ends up in a stale clique and is discarded for example ? Does it stay in this container ?
Maybe a good way to prevent that would be to avoid this HashSet and have an "origin" property for blocks in graph, alongside the `storage` property that follows blocks through their lifecycle in graph

massa-consensus-worker/src/consensus_worker.rs
```diff
@@ -502,7 +514,7 @@ impl ConsensusWorker {
     fn prune_stats(&mut self) -> Result<()> {
         let start_time =
             MassaTime::now(self.clock_compensation)?.saturating_sub(self.stats_history_timespan);
-        self.final_block_stats.retain(|(t, _)| t >= &start_time);
+        self.final_block_stats.retain(|(t, _, _)| t >= &start_time);
```
damip, line 517: This is rather inefficient, maybe:
```rust
while let Some((t, _, _)) = self.final_block_stats.front() {
    if t < start_time { self.final_block_stats.pop_front() } else { break }
}
```

damip (commented):
Nice to have ! A few comments tho

massa-consensus-worker/src/consensus_worker.rs
AurelienFT: It can be a big overload just for desynchronization to have one more parameter with origin everywhere maybe we could just add a prune for this hashset

damip (approved):

PR #2948, opened 2022-08-31
correct storage usage in protocol

damip:


AurelienFT:
Tested on simulator works well for basic features (block and ops propagation)

Issue #2935, opened 2022-08-29
Protocol problems


damip:
* objects must not be sent to Storage before their signatures are verified. Signature verif happens in Protocol (with cache control). Therefore, only full objects must be transferred between Protocol and Network.
* Network uses an overcomplicated and redundant sending/receiving system that was created at the beginnings of Storage. It should be refactored back to simply sending/receiving serializable `Message`s. All links with Storage should be removed in Network.
* Many memory leaks in Storage management within Protocol. For example, full blocks and their dependencies are added to self.storage when they need to be propagated, but only operations removed from the signature verif cache are ever removed from self.storage
* Signature verification caches (eg. checked_headers) are misused for block retrieval status management and could be cleared anytime

gterzian:
Thanks for noting.

gterzian:
> objects must not be sent to Storage before their signatures are verified. Signature verif happens in Protocol (with cache control). Therefore, only full objects must be transferred between Protocol and Network.

> 
Network uses an overcomplicated and redundant sending/receiving system that was created at the beginnings of Storage. It should be refactored back to simply sending/receiving serializable Messages. All links with Storage should be removed in Network.

I think these two were done already by @AurelienFT in https://github.com/massalabs/massa/pull/2937

gterzian:
> Many memory leaks in Storage management within Protocol. For example, full blocks and their dependencies are added to self.storage when they need to be propagated, but only operations removed from the signature verif cache are ever removed from self.storage

I think it's a good example of a problem introduced by the current "per component" approach to manage the lifetime of objects in storage. 

It's quite hard to know when a block can be "dropped" locally by protocol, for example: Should it be done when all active nodes have been noted as knowing about it? What about if a new node connects, and the block is actually still relevant? 

On the other hand, consensus can know when a block is so old that no node should ever need it anymore. 

And protocol can just "check storage" when a node asks about something, and if it's gone, just answer "not found". 

So protocol uses storage in a "dumb" way, without knowing when objects are to be dropped. 

On the other hand, something like execution can keep an `Arc` to a block it still needs to execute, so that even if consensus removes it from storage it can still be executed(if execution is very busy or something, although the other problem then could be running out of memory). 

@damip

damip:
> > Many memory leaks in Storage management within Protocol. For example, full blocks and their dependencies are added to self.storage when they need to be propagated, but only operations removed from the signature verif cache are ever removed from self.storage
> 
[...]

PR #2949, opened 2022-08-31
Fix pruning of checked headers if block is in wishlist

gterzian:
Addressing part of https://github.com/massalabs/massa/issues/2935

> Signature verification caches (eg. checked_headers) are misused for block retrieval status management and could be cleared anytime


massa-protocol-worker/src/protocol_worker.rs
```diff
@@ -860,7 +860,8 @@ impl ProtocolWorker {
     /// Prune `checked_headers` if it is too large
     fn prune_checked_headers(&mut self) {
         if self.checked_headers.len() > self.config.max_known_blocks_size {
-            self.checked_headers.clear();
+            self.checked_headers
+                .drain_filter(|id, _| !self.block_wishlist.contains_key(id));
```
damip, line 864: I don't understand, I think this basically prevents cache cleanup but doesn't actually retain the object

AurelienFT:
This has been fixed with this PR : https://github.com/massalabs/massa/pull/2956

PR #2966, opened 2022-09-05
Debug memory leak

AurelienFT:


damip:
Perfect :)

damip:
Very good work.

Just one thing: I would avoid ever calling `self.storage.store_*` or `self.storage.claim_*`.
Why do we need those ?

damip (approved):

damip:
Amazing work !

Issue #2961, opened 2022-09-05
Make tests compile and pass


damip:
* [x] consensus tests don't compile @gterzian 
* [x] factory tests timeout @Eitu33 
* [x] final-state tests don't compile @Eitu33 
* [x] massa-ledger tests don't compile
* [x] network tests don't compile @gterzian 
* [x] pool tests don't pass
* [x] PoS tests don't compile

Eitu33:
FI I'm taking factory final state and ledger for now.

damip:
@gterzian can you take care of consensus and network ?

gterzian:
@damip I'll do network first, I haven't been much involved with the consensus changes so perhaps the relevant person can do it? Otherwise I'll give it a try after network...

Eitu33:
For anyone that will work on it since I will be moving on to something else, the branch I worked on for the Factory, Final state and Ledger tests is `fix-factory-ledger-state-tests`. Final state and Ledger tests are ok but Factory tests still need some changes.

PR #2967, opened 2022-09-06
Fix network tests

gterzian:
Part of https://github.com/massalabs/massa/issues/2961

### DIFF
--- a/massa-network-worker/src/binders.rs
+++ b/massa-network-worker/src/binders.rs
@@ -14,7 +14,6 @@ use massa_serialization::Serializer;
 use massa_serialization::{DeserializeError, Deserializer};
 use std::convert::TryInto;
 use tokio::io::{AsyncReadExt, AsyncWriteExt};
-use tracing::log::warn;
 
 /// Used to serialize and send data.
 pub struct WriteBinder {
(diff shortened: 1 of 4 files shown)


gterzian:
@AurelienFT Ready for review. Note that it seems we're not checking the length of the outgoing `Message::AskForBlocks`.

AurelienFT:
> AskForBlocks

i don't understand what we are not checking ? The number of informations asked ? In serialization/deserialization ?

massa-network-worker/src/binders.rs
```diff
@@ -174,7 +173,6 @@ impl ReadBinder {
             .message_deserializer
             .deserialize::<DeserializeError>(&self.buf)
             .map_err(|err| {
-                warn!("error deserializing message: {:?}", err);
```
AurelienFT, line 177: I think we shouldn't remove this warn. The node should never receive bad message and also it can help us debug during testnet.

gterzian: Makes sense, re-added it.

gterzian:
@AureliaDolo Thanks for the review, comments addressed.

AurelienFT (approved):

gterzian:
> i don't understand what we are not checking ? The number of informations asked ? In serialization/deserialization

Yes it's the number of asks, in the serialization. Compare how the de-serializer uses `max_ask_block` in the `U32VarIntDeserializer`.  @AurelienFT

AurelienFT:
> > i don't understand what we are not checking ? The number of informations asked ? In serialization/deserialization
> 
> Yes it's the number of asks, in the serialization. Compare how the de-serializer uses `max_ask_block` in the `U32VarIntDeserializer`. @AurelienFT

We never have limits on the serializers because they are used with internal data.

PR #2972, opened 2022-09-07
Fix consensus tests

gterzian:
Part of https://github.com/massalabs/massa/issues/2961


damip:
Still needs some work to allow Consensus tests to compile and at least 1 or 2 to pass

gterzian:
@damip Ready for review. This makes them compile, and there is one test passing. Seems like there are some infrastructural issues that need to be dealt with for all tests, for example: https://github.com/massalabs/massa/issues/2978

Note that I left some warnings in the code, because I think some of them will go away as we continue working on making the tests pass.

PR #2977, opened 2022-09-07
Fix memory leak bootstrap

AurelienFT:
The leak come from the data_graph. I think this variable and the others of the bootstrap sessions are not cleared until the end of the select. I have tried a bunch of things without success :
- Removing the move in bootstrap_sessions and dont clone the variables
- Removing while that clean bootstrap sessions at the end
- Trying to clear/override bootstrap_sessions in the next branch of the select

I don't have much other ideas


AurelienFT:
The node grow a bit at start and then it stabilize. So I think we can release testnet in those conditions

massa-bootstrap/src/server.rs
```diff
@@ -240,9 +259,6 @@ impl BootstrapServer {
             }
         }
 
-        // wait for bootstrap sessions to finish
-        while bootstrap_sessions.next().await.is_some() {}
```
damip, line 244: let's put this back

massa-consensus-exports/src/consensus_controller.rs
```diff
@@ -104,7 +104,7 @@ impl ConsensusCommandSender {
 
     /// get bootstrap snapshot
     pub async fn get_bootstrap_state(&self) -> Result<BootstrapableGraph, ConsensusError> {
-        let (response_tx, response_rx) = oneshot::channel::<BootstrapableGraph>();
+        let (response_tx, mut response_rx) = mpsc::channel::<Box<BootstrapableGraph>>(10);
```
damip, line 107: let's set capacity to 1 instead of 10

PR #2997, opened 2022-09-12
Fix bootstrap of changes

damip:


massa-bootstrap/src/client.rs
```diff
@@ -84,7 +84,7 @@ async fn stream_final_state(
                     write_final_state
                         .async_pool
                         .apply_changes_unchecked(&final_state_changes.async_pool_changes);
-                    if !final_state_changes.roll_state_changes.is_empty() {
+                    if !final_state_changes.roll_state_changes.is_empty() {  // TODO is this correct ???
```
damip, line 87: `pos_state.apply_changes` also updates the cycle/slot etc... even in cases where there are no changes. 
Are we sure this condition makes sense ?

Eitu33: This is outdated after the refactoring, resolving.

Eitu33:
bors merge

AurelienFT:
bors merge

PR #3000, opened 2022-09-12
Op executesc add datastore

sydhds:


massa-models/src/operation.rs
```diff
@@ -695,14 +722,18 @@ impl Deserializer<OperationType> for OperationTypeDeserializer {
                         context("Failed data deserialization", |input| {
                             self.data_deserializer.deserialize(input)
                         }),
+                        context("Failed data deserialization", |input| {
```
AurelienFT, line 725: The message here is the same as above. It should be an other to be able to find easily the source of the error message.

AurelienFT (commented):
Very good usage of serialization/deserialization.

massa-models/src/operation.rs
sydhds: Just fixed the error message :)

damip (commented):
Very good work !

Issue #2976, opened 2022-09-07
Change endorsements


qdrn:
Endorsements should satisfy the following constraints:
- Endorsements should be created and propagated to the network by endorsers at half-slot time.
- Endorsements should endorse the best parent in the given thread as seen by the endorser at this timestamp.
- Endorsements created at a half-period can only be included in the block of the next period.

Side note: currently when we include the endorsements in a block we verify that they endorse the parent of the created block in this thread.

cc @sebastien-forestier @leoloco

damip:
## Current implementation

Currently, we PoS-select endorsement creators targeting the blockclique block at slot `(period, thread)` if any.
The endorsements selected at `(period, thread)` are created/emitted at time `genesis_timestamp + period * t0 + thread * (t0/thread_count) + t0/2`.

## Implementation change

We need to PoS-select endorsement creators that are allowed to be included in blocks at slot `(period, thread)`.
The endorsements selected at `(period, thread)` are created/emitted at time `genesis_timestamp + period * t0 + thread * (t0/thread_count) - t0/2` and endorse the best_parent of their thread.

sebastien-forestier:
One additionnal change to be discussed: take into account the compatible endorsements in the pool to compute the cliques fitness

qdrn:
> One additionnal change to be discussed: take into account the compatible endorsements in the pool to compute the cliques fitness

What cases do you have in mind ? If you create a block you create the one that gives the clique with the highest fitness (I guess ?). If you create endorsements then you probably don't have the other endorsements.

damip:
Part 1 is done here: https://github.com/massalabs/massa/pull/3004
Not merging yet (breaking)

AurelienFT:
Waiting for feeedback from research team.

qdrn:
I think we can close this one. Right ?

AurelienFT:
> take into account the compatible endorsements in the pool to compute the cliques fitness

This hasn't been discuss but could be moved to a discussion

qdrn:
> 

Yeah, I don't think it belongs to an issue.

PR #3004, opened 2022-09-13
Change endorsements

adrien-zinger:
#2976

On processing a slot, select the latest block on the same thread as endorsed block.
Target the current slot to check the endorsements for the best parent.

### DIFF
--- a/massa-consensus-exports/src/consensus_controller.rs
+++ b/massa-consensus-exports/src/consensus_controller.rs
@@ -161,7 +161,7 @@ impl ConsensusCommandSender {
     pub fn get_latest_blockclique_block_at_slot(
         &self,
         slot: Slot,
-    ) -> Result<Option<BlockId>, ConsensusError> {
+    ) -> Result<BlockId, ConsensusError> {
         let (response_tx, response_rx) = oneshot::channel();
         self.0
             .blocking_send(ConsensusCommand::GetLatestBlockcliqueBlockAtSlot { slot, response_tx })
--- a/massa-consensus-exports/src/consensus_controller.rs
+++ b/massa-consensus-exports/src/consensus_controller.rs
@@ -157,6 +157,26 @@ impl ConsensusCommandSender {
         })
     }
 
+    /// get latest block id of a slot in a blockclique
+    pub fn get_latest_blockclique_block_at_slot(
+        &self,
+        slot: Slot,
+    ) -> Result<Option<BlockId>, ConsensusError> {
+        let (response_tx, response_rx) = oneshot::channel();
+        self.0
+            .blocking_send(ConsensusCommand::GetLatestBlockcliqueBlockAtSlot { slot, response_tx })
+            .map_err(|_| {
+                ConsensusError::SendChannelError(
+                    "send error consensus command get_blockclique_block_at_slot".into(),
+                )
+            })?;
+        response_rx.blocking_recv().map_err(|_| {
+            ConsensusError::ReceiveChannelError(
+                "consensus command get_blockclique_block_at_slot response read error".to_string(),
+            )
+        })
+    }
+
     /// get current consensus stats
     pub async fn get_stats(&self) -> Result<ConsensusStats, ConsensusError> {
         let (response_tx, response_rx) = oneshot::channel();
--- a/massa-factory-worker/src/endorsement_factory.rs
+++ b/massa-factory-worker/src/endorsement_factory.rs
@@ -62,7 +62,7 @@ impl EndorsementFactoryWorker {
         // get delayed time
         let shifted_now = MassaTime::now(self.cfg.clock_compensation_millis)
             .expect("could not get current time")
-            .saturating_sub(self.half_t0);
+            .saturating_sub(self.cfg.t0);
 
         // if it's the first computed slot, add a time shift to prevent double-production on node restart with clock skew
         let base_time = if previous_slot.is_none() {
@@ -74,7 +74,7 @@ impl EndorsementFactoryWorker {
         // get closest slot according to the current absolute time
         let mut next_slot = get_closest_slot_to_timestamp(
             self.cfg.thread_count,
-            self.cfg.t0,
+            self.half_t0,
             self.cfg.genesis_timestamp,
             base_time,
         );
@@ -96,7 +96,7 @@ impl EndorsementFactoryWorker {
             next_slot,
         )
         .expect("could not get block slot timestamp")
-        .saturating_add(self.half_t0)
+        .saturating_sub(self.half_t0)
         .estimate_instant(self.cfg.clock_compensation_millis)
         .expect("could not estimate block slot instant");
 
--- a/massa-factory-worker/src/endorsement_factory.rs
+++ b/massa-factory-worker/src/endorsement_factory.rs
@@ -164,17 +164,14 @@ impl EndorsementFactoryWorker {
             // error getting block ID at target slot
             Err(_) => {
                 warn!(
-                    "could not get blockclique block to create endorsement targeting slot {}",
+                    "could not get latest blockclique block to create endorsement targeting slot {}",
                     slot
                 );
                 return;
             }
 
-            // the target slot is a miss: ignore
-            Ok(None) => return,
-
-            // there is a block a the target slot
-            Ok(Some(b_id)) => b_id,
+            // latest block found
+            Ok(b_id) => b_id,
         };
 
         // produce endorsements
--- a/massa-factory-worker/src/endorsement_factory.rs
+++ b/massa-factory-worker/src/endorsement_factory.rs
@@ -164,7 +164,7 @@ impl EndorsementFactoryWorker {
             // error getting block ID at target slot
             Err(_) => {
                 warn!(
-                    "could not get latest blockclique block to create endorsement targeting slot {}",
+                    "could not get latest blockclique block to create endorsement to be included at slot {}",
                     slot
                 );
                 return;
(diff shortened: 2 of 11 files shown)


massa-consensus-exports/src/consensus_controller.rs
```diff
@@ -157,6 +157,26 @@ impl ConsensusCommandSender {
         })
     }
 
+    /// get latest block id of a slot in a blockclique
+    pub fn get_latest_blockclique_block_at_slot(
+        &self,
+        slot: Slot,
+    ) -> Result<Option<BlockId>, ConsensusError> {
```
damip, line 164: same here, no need for an Option

massa-graph/src/block_graph.rs
```diff
@@ -2678,6 +2675,59 @@ impl BlockGraph {
             })
     }
 
+    /// get the latest blockclique (or final) block ID at a given slot, if any
+    pub fn get_latest_blockclique_block_at_slot(&self, slot: &Slot) -> Option<BlockId> {
```
damip, line 2679: this should always return a block ID, no need for an Option

massa-graph/src/block_graph.rs
```diff
@@ -2678,6 +2675,59 @@ impl BlockGraph {
...
+            .for_each(|id| {
+                let period = *block_periods.get(id).unwrap();
+                if period > latest_final_block_period && period < slot.period {
+                    latest_final_block_period = period;
+                    latest_final_block_id = *id;
+                }
+            });
+
+        Some(latest_final_block_id)
+    }
```
damip, line 2729: A more efficient way (that doesn't require adding an index to storage and iterating over all the blocks of that thread, nor locking storage at all) would be:
* iterate over every block B in the blockclique:
  * get `self.block_statuses.get(B)`
    * if the block is not Active, panic
    * if the block is final, panic
    * if the block is not in the target slot's thread, continue
    * if the block's period is equal or higher than the target slot's period, continue
    * if the block's period is strictly higher than latest_final_block_period:
      * set latest_final_block_period = B's period
      * set latest_final_block_id = B

adrien-zinger: great!

massa-factory-worker/src/endorsement_factory.rs
```diff
@@ -156,23 +156,26 @@ impl EndorsementFactoryWorker {
...
-            };
+        let endorsed_block: BlockId = match self
+            .channels
+            .consensus
+            .get_latest_blockclique_block_at_slot(slot)
+        {
+            // error getting block ID at target slot
+            Err(_) => {
+                warn!(
+                    "could not get blockclique block to create endorsement targeting slot {}",
```
damip, line 167:
```suggestion
                    "could not get latest blockclique block to create endorsement to be included at slot {}",
```

massa-factory-worker/src/endorsement_factory.rs
```diff
@@ -156,23 +156,26 @@ impl EndorsementFactoryWorker {
...
+                    slot
+                );
+                return;
+            }
+
+            // the target slot is a miss: ignore
+            Ok(None) => return,
```
damip, line 174: should never happen

massa-graph/src/block_graph.rs
```diff
@@ -1570,13 +1570,10 @@ impl BlockGraph {
     fn check_endorsements(
         &self,
         header: &WrappedHeader,
-        parent_in_own_thread: &ActiveBlock,
+        slot: Slot,
```
damip, line 1573: no need for the "slot" parameter, it can be found in header.content.slot

massa-graph/src/block_graph.rs
```diff
@@ -1592,10 +1589,10 @@ impl BlockGraph {
                 )));
             }
             // check that the endorsement slot matches the endorsed block
-            if endorsement.content.slot != parent_in_own_thread.slot {
+            if endorsement.content.slot != slot {
                 return Ok(EndorsementsCheckOutcome::Discard(DiscardReason::Invalid(
-                    format!("endorsement targets a block with wrong slot. Block's parent: {}, endorsement: {}",
-                            parent_in_own_thread.slot, endorsement.content.slot),
+                    format!("endorsement targets a block with wrong slot. Current slot: {}, endorsement: {}",
```
damip, line 1594: endorsement slot does not match the slot of the block that contains it

massa-pool-worker/src/endorsement_pool.rs
```diff
@@ -148,17 +148,17 @@ impl EndorsementPool {
     /// get endorsements for block creation
     pub fn get_block_endorsements(
         &self,
-        target_slot: &Slot,
+        slot: &Slot, // slot of the block
```
damip, line 151:
```suggestion
        slot: &Slot, // slot of the block that will contain the endorsement
```

massa-storage/src/block_indexes.rs
```diff
@@ -23,6 +23,8 @@ pub struct BlockIndexes {
     index_by_op: PreHashMap<OperationId, PreHashSet<BlockId>>,
     /// Structure mapping endorsement id with ids of blocks they are contained in
     index_by_endorsement: PreHashMap<EndorsementId, PreHashSet<BlockId>>,
+    /// Structure mapping threads with their Block id
+    index_by_thread: HashMap<u8, PreHashSet<BlockId>>,
```
damip, line 27: not needed

damip:
Some things to check:
* [ ] in the endorsement pool, when we prune the pool after receiving a finality notification from consensus, now we should remove all endorsements whose slot <= latest_final_slot[endorsement.thread]
* [ ] in the endorsement factory, when creating an endorsement that will be included in slot S, we should produce the endorsement at time(S) - t0/2

massa-graph/src/block_graph.rs
```diff
@@ -2678,6 +2675,46 @@ impl BlockGraph {
...
+    /// get the latest blockclique (or final) block ID at a given slot, if any
+    pub fn get_latest_blockclique_block_at_slot(&self, slot: &Slot) -> BlockId {
+        let (mut latest_final_block_id, mut latest_final_block_period) = self
+            .latest_final_blocks_periods
+            .get(slot.thread as usize)
+            .unwrap_or_else(|| panic!("unexpected not found latest final block period"));
+
+        self.max_cliques
+            .iter()
+            .filter(|c| c.is_blockclique)
```
damip, line 2687: Note that this will iterate on all cliques to filter them.
I recommend doing `let blockclique = &self.max_cliques.find(|c| c.is_blockclique).expect("no blockclique found").block_ids;` which is short-circuiting (will not scan the whole container but stop at the first found element)

damip: See https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.find

massa-graph/src/block_graph.rs
```diff
@@ -2678,6 +2675,46 @@ impl BlockGraph {
...
+                            if a_block.is_final {
+                                panic!(
+                                    "unexpected final block on getting latest blockclique at slot"
+                                );
+                            }
+                            if a_block.slot.thread == slot.thread
+                                && a_block.slot.period <= slot.period
+                                && a_block.slot.period > latest_final_block_period
+                            {
+                                latest_final_block_period = a_block.slot.period;
```
damip, line 2706: the naming "latest_final_block_period / latest_final_block_id" is kinda wrong because it will almost never be a final block, I would recommend best_block_period / best_block_id

massa-graph/src/block_graph.rs
```diff
@@ -2678,6 +2675,46 @@ impl BlockGraph {
...
+                            a_block,
+                            storage: _,
+                        }) => {
+                            if a_block.is_final {
+                                panic!(
+                                    "unexpected final block on getting latest blockclique at slot"
+                                );
+                            }
+                            if a_block.slot.thread == slot.thread
+                                && a_block.slot.period <= slot.period
```
damip, line 2703: this should be a strict inferiority

massa-graph/src/block_graph.rs
```diff
@@ -2678,6 +2675,46 @@ impl BlockGraph {
...
+                    .block_ids
+                    .iter()
+                    .for_each(|id| match self.block_statuses.get(id) {
+                        Some(BlockStatus::Active {
+                            a_block,
+                            storage: _,
+                        }) => {
+                            if a_block.is_final {
+                                panic!(
+                                    "unexpected final block on getting latest blockclique at slot"
```
damip, line 2699: latest blockclique *block* at slot

massa-factory-worker/src/endorsement_factory.rs
```diff
@@ -62,7 +62,7 @@ impl EndorsementFactoryWorker {
         // get delayed time
         let shifted_now = MassaTime::now(self.cfg.clock_compensation_millis)
             .expect("could not get current time")
-            .saturating_sub(self.half_t0);
+            .saturating_sub(self.cfg.t0);
 
         // if it's the first computed slot, add a time shift to prevent double-production on node restart with clock skew
         let base_time = if previous_slot.is_none() {
```
adrien-zinger, line 68: I'm very unsure about that changes :roll_eyes: I miss the logic of that timing, if we can add a documentation and an example it would be great!

adrien-zinger: NB. that's not the unique change in that file obviously

damip: I made a proposal here: https://github.com/massalabs/massa/pull/3004/commits/0d6d08178019fcddbc6c16adb147de101bef3e21

damip (approved):

damip (changes requested):
OK, let's not merge yet tho (breaking)

damip (approved):

Issue #3013, opened 2022-09-15
Add POS mock for consensus tests


gterzian:
Currently some tests([example](https://github.com/massalabs/massa/blob/a56bc624fc21464a1137b650065dd6dbd9f8cb0c/massa-consensus-worker/src/tests/inter_cycle_batch_finalization.rs#L48)) are failing at https://github.com/massalabs/massa/blob/a56bc624fc21464a1137b650065dd6dbd9f8cb0c/massa-graph/src/block_graph.rs#L1308

I guess we need to add a mock of the POS to be able to control this from the tests. 

I'll try using the mock exported by `massa-pos-exports`, which seems to suffer from the same problem as https://github.com/massalabs/massa/issues/2955

PR #2984, opened 2022-09-09
Consensus tests: use exported mock pool controller

gterzian:
Fix https://github.com/massalabs/massa/issues/2978

PR #3014, opened 2022-09-15
Use POS mock in consensus tests

gterzian:
Fix https://github.com/massalabs/massa/issues/3013

### DIFF
--- a/massa-pos-exports/src/controller_traits.rs
+++ b/massa-pos-exports/src/controller_traits.rs
@@ -11,8 +11,6 @@ use massa_hash::Hash;
 use massa_models::address::Address;
 use massa_models::api::IndexedSlot;
 use massa_models::slot::Slot;
-
-#[cfg(feature = "testing")]
 use std::collections::{HashMap, VecDeque};
 
 /// interface that communicates with the selector worker thread
@@ -62,7 +60,6 @@ pub trait SelectorController: Send + Sync {
     /// Get every [Selection]
     ///
     /// Only used in tests for post-bootstrap selection matching.
-    #[cfg(feature = "testing")]
     fn get_entire_selection(&self) -> VecDeque<(u64, HashMap<Slot, Selection>)>;
 }
 
--- a/massa-pos-exports/src/controller_traits.rs
+++ b/massa-pos-exports/src/controller_traits.rs
@@ -11,6 +11,8 @@ use massa_hash::Hash;
 use massa_models::address::Address;
 use massa_models::api::IndexedSlot;
 use massa_models::slot::Slot;
+
+#[cfg(feature = "testing")]
 use std::collections::{HashMap, VecDeque};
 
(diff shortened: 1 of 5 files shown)


gterzian:
Note, this is branched off of https://github.com/massalabs/massa/pull/2984

gterzian:
@AurelienFT ready for review.

Note the related issue: https://github.com/massalabs/massa/issues/3016

massa-pos-worker/src/controller.rs
```diff
@@ -157,16 +155,8 @@ impl SelectorController for SelectorControllerImpl {
...
     fn get_entire_selection(&self) -> VecDeque<(u64, HashMap<Slot, Selection>)> {
-        let (_, lock) = &*self.cache;
-        let cache_guard = lock.read();
-        let cache = cache_guard.as_ref().map_err(|err| err.clone()).unwrap();
-        cache
-            .0
-            .iter()
-            .map(|cycle_draws| (cycle_draws.cycle, cycle_draws.draws.clone()))
-            .collect()
+        panic!("Method should only be used by a mock of the selector controller.");
```
AurelienFT, line 159: We are using it in the tests of the bootstrap and we don't wanna use a mock right there

gterzian: Thanks for the info, I've re-added the method body, leaving the feature flag out. By the way it seems bootstrap doesn't use that feature, and the compilation problem only occurred when it is used. We can fix it with the other issue...

massa-pos-exports/src/controller_traits.rs
```diff
@@ -62,7 +60,6 @@ pub trait SelectorController: Send + Sync {
     /// Get every [Selection]
     ///
     /// Only used in tests for post-bootstrap selection matching.
-    #[cfg(feature = "testing")]
```
AurelienFT, line 65: Why did you removed the constraint ?

gterzian: Because of https://github.com/massalabs/massa/issues/3016(to be fixed in a follow-up)

AurelienFT: Did you tried with the comment I made in the issue ? I think it solves the problem.

gterzian: I haven't tried, and I think it's best done in a follow-up, so that this PR is not blocked by something unrelated.

gterzian:
So it seems the problem in https://github.com/massalabs/massa/issues/3016 only happens if you run the tests for a single package, for example `cargo test -p massa_consensus_worker`...

AurelienFT (approved):

AurelienFT:
bors merge

AurelienFT:
bors merge

AurelienFT:
bors cancel

AurelienFT:
bors merge

PR #3023, opened 2022-09-16
Remove banning in edge case and more print

AurelienFT:
Case : If we already received info about this step or the next one
* don't ban
* ignore
* mark as not knowing block
* print

Case : If we haven’t received the infos from the precedent step
Can happen if consensus removes the block from wishlist during the process
* ignore 
* warn
*  mark as not knowing block

Case : If block is bigger than max size
that is a problem.
* Warn 
* ban the sender

Case : If not all operations has been received when asking
* gather the ops that are good AND asked
* mark the node as not knowing the block
* print a warning

Case : If we receive a bad header asked by protocol, not block propagation
* print warning
* ban

Solutions : Add print in all this cases and remove banning as we are on testnet and so we can have data, except for “bad header asked by protocol, not block propagation” that we should keep the ban

### DIFF
--- a/massa-pos-worker/src/draw.rs
+++ b/massa-pos-worker/src/draw.rs
@@ -57,6 +57,7 @@ pub(crate) fn perform_draws(
         ),
     };
 
+    let mut five_first_slots: Vec<(Slot, Selection)> = Vec::new();
     loop {
         // draw block creator
         let producer = if cur_slot.period > 0 {
@@ -71,14 +72,13 @@ pub(crate) fn perform_draws(
             .map(|_index| addresses[dist.sample(&mut rng)])
             .collect();
 
+        let selection = Selection {
+            producer,
+            endorsements,
+        };
+        five_first_slots.push((cur_slot, selection.clone()));
         // add to draws
-        cycle_draws.draws.insert(
-            cur_slot,
-            Selection {
-                producer,
-                endorsements,
-            },
-        );
+        cycle_draws.draws.insert(cur_slot, selection);
 
         if cur_slot == last_slot {
             break;
@@ -92,11 +92,7 @@ pub(crate) fn perform_draws(
         "Draws for cycle {} complete. Look_back seed was {:#?}. Five first selections is : {:#?}",
         cycle,
         lookback_seed.to_bytes(),
-        cycle_draws
-            .draws
-            .iter()
-            .take(5)
-            .collect::<Vec<(&Slot, &Selection)>>()
+        five_first_slots
     );
 
     Ok(cycle_draws)
--- a/massa-pos-worker/src/draw.rs
+++ b/massa-pos-worker/src/draw.rs
@@ -58,6 +58,7 @@ pub(crate) fn perform_draws(
     };
 
     let mut five_first_slots: Vec<(Slot, Selection)> = Vec::new();
+    let mut count = 0;
     loop {
         // draw block creator
         let producer = if cur_slot.period > 0 {
@@ -76,7 +77,10 @@ pub(crate) fn perform_draws(
             producer,
             endorsements,
         };
-        five_first_slots.push((cur_slot, selection.clone()));
+        if count < 5 {
+            five_first_slots.push((cur_slot, selection.clone()));
+            count += 1;
+        }
         // add to draws
         cycle_draws.draws.insert(cur_slot, selection);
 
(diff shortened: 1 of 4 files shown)


massa-pos-worker/src/draw.rs
```diff
@@ -71,14 +72,13 @@ pub(crate) fn perform_draws(
             .map(|_index| addresses[dist.sample(&mut rng)])
             .collect();
 
+        let selection = Selection {
+            producer,
+            endorsements,
+        };
+        five_first_slots.push((cur_slot, selection.clone()));
```
damip, line 79: this will not only add the first 5 but all of them

AurelienFT: Fixed.

massa-protocol-worker/src/protocol_network.rs
```diff
@@ -242,7 +242,15 @@ impl ProtocolWorker {
...
-                let _ = self.ban_node(&from_node_id).await;
+                warn!(
+                    "Node {} sent us header for block id {} but we already received it.",
+                    from_node_id, block_id
+                );
+                if let Some(node) = self.active_nodes.get_mut(&from_node_id) && node.asked_blocks.contains_key(&block_id) {
+                    node.asked_blocks.remove(&block_id);
+                    node.insert_known_blocks(&[block_id], false, Instant::now(), self.config.max_node_known_blocks_size);
+                }
+
```
damip, line 253: shall we run the ask_for_block here ?

massa-protocol-worker/src/protocol_network.rs
```diff
@@ -251,6 +259,12 @@ impl ProtocolWorker {
             .await
             .is_err()
```
damip, line 260: let's get the actual error and add it inside the message

massa-protocol-worker/src/protocol_network.rs
```diff
@@ -301,19 +315,37 @@ impl ProtocolWorker {
...
         } else {
-            let _ = self.ban_node(&from_node_id).await;
+            warn!(
+                "Node {} sent us an operation list but we don't have block id {} in our wishlist.",
+                from_node_id, block_id
+            );
+            if let Some(node) = self.active_nodes.get_mut(&from_node_id) && node.asked_blocks.contains_key(&block_id) {
+                node.asked_blocks.remove(&block_id);
+                node.insert_known_blocks(&[block_id], false, Instant::now(), self.config.max_node_known_blocks_size);
+            }
```
damip, line 325: shall we run the ask_for_block here ?

massa-protocol-worker/src/protocol_network.rs
```diff
@@ -301,19 +315,37 @@ impl ProtocolWorker {
...
 
         let header = if let Some(header) = &info.header {
             header
         } else {
-            let _ = self.ban_node(&from_node_id).await;
+            warn!("Node {} sent us an operation list but we don't have receive the header of block id {} yet.", from_node_id, block_id);
+            if let Some(node) = self.active_nodes.get_mut(&from_node_id) && node.asked_blocks.contains_key(&block_id) {
+                node.asked_blocks.remove(&block_id);
+                node.insert_known_blocks(&[block_id], false, Instant::now(), self.config.max_node_known_blocks_size);
+            }
```
damip, line 336: shall we run the ask_for_block here ?

massa-protocol-worker/src/protocol_network.rs
```diff
@@ -301,19 +315,37 @@ impl ProtocolWorker {
...
         if info.operation_ids.is_some() {
-            let _ = self.ban_node(&from_node_id).await;
+            warn!(
+                "Node {} sent us an operation list for block id {} but we already received it.",
+                from_node_id, block_id
+            );
+            if let Some(node) = self.active_nodes.get_mut(&from_node_id) && node.asked_blocks.contains_key(&block_id) {
+                node.asked_blocks.remove(&block_id);
+                node.insert_known_blocks(&[block_id], false, Instant::now(), self.config.max_node_known_blocks_size);
+            }
```
damip, line 348: shall we run the ask_for_block here ?

massa-protocol-worker/src/protocol_network.rs
AurelienFT: It's running automatically at short intervals and if received need connection/block header etc so I think we don't need to add a run here

massa-protocol-worker/src/protocol_network.rs
```diff
@@ -379,6 +413,10 @@ impl ProtocolWorker {
             .note_operations_from_node(operations.clone(), &from_node_id)
             .is_err()
```
damip, line 414: let's keep the actual error and add it to the warn msg

massa-protocol-worker/src/protocol_network.rs
damip: okay

damip (approved):

AurelienFT:
bors merge

Issue #3033, opened 2022-09-19
Operation propagation(outside of block)


gterzian:
A. When we send a new operation via the API:

1. It is [propagated](https://github.com/massalabs/massa/blob/d87f9f11ad623153a55524e922c245431d19b0d9/massa-api/src/public.rs#L857) to all nodes [active at that time](https://github.com/massalabs/massa/blob/d87f9f11ad623153a55524e922c245431d19b0d9/massa-protocol-worker/src/protocol_worker.rs#L410).

B. When we [receive a full operation](https://github.com/massalabs/massa/blob/d87f9f11ad623153a55524e922c245431d19b0d9/massa-protocol-worker/src/protocol_network.rs#L116) from the network:
1. We [note the operation](https://github.com/massalabs/massa/blob/d87f9f11ad623153a55524e922c245431d19b0d9/massa-protocol-worker/src/worker_operations_impl.rs#L140) from the node, which includes
2. Sending the operation [to pool](https://github.com/massalabs/massa/blob/d87f9f11ad623153a55524e922c245431d19b0d9/massa-protocol-worker/src/protocol_worker.rs#L932).

Previously, there was a [`propagate` flag](https://github.com/massalabs/massa/blob/5010bcaea3e1e3772ae8756f580f731c3130c9df/massa-protocol-worker/src/protocol_worker.rs#L1109) to the call adding the operation to pool, and if true, pool would then [call back into protocol](https://github.com/massalabs/massa/blob/5010bcaea3e1e3772ae8756f580f731c3130c9df/massa-pool/src/pool_worker.rs#L326) to propagate the operation.

I think this last part might be missing from the current workflow, and that "it works" when people keep trying to add the operation via the API, because that is like doing A above in a loop, meaning it eventually reaches all nodes, or perhaps it happens to eventually reach a block producer.

Perhaps there is another mechanism by which we announce newly received operations to the network, I can't see it however. The only call to `protocol.propagate_operations` seems to be the one made by `api` referred to above, and protocol only calls `network.send_operations_batch` in response to receiving that command. 

The fact that an operation, when received from the network, is never propagated, could actually be tested with a unit-test(or perhaps rather a failing test asserting that it is propagated).

gterzian:
cc @massalabs/core-team

damip:
Good catch !

I've made a little PR for correcting it: https://github.com/massalabs/massa/pull/3035

damip:
[...]

PR #3034, opened 2022-09-19
Protocol: add test for the propagation of operations 

gterzian:
Fix https://github.com/massalabs/massa/issues/3033

PR #3035, opened 2022-09-19
correct op and endorsements propagation

damip:
Should correct the problem discovered by @gterzian  in https://github.com/massalabs/massa/issues/3033

### DIFF
--- a/massa-protocol-worker/src/protocol_worker.rs
+++ b/massa-protocol-worker/src/protocol_worker.rs
@@ -792,7 +792,7 @@ impl ProtocolWorker {
         }
 
         let (_endorsement_ids, endorsements_reused) = match self
-            .note_endorsements_from_node(header.content.endorsements.clone(), source_node_id, false)
+            .note_endorsements_from_node(header.content.endorsements.clone(), source_node_id)
             .await
         {
             Err(_) => {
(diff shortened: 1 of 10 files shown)


massa-protocol-worker/src/protocol_worker.rs
```diff
@@ -929,7 +937,10 @@ impl ProtocolWorker {
             ops.store_operations(new_operations.into_values().collect());
 
             // Add to pool, propagate when received outside of a header.
-            self.pool_controller.add_operations(ops);
+            self.pool_controller.add_operations(ops.clone());
+
+            // Request propagation
+            self.propagate_ops(ops).await;
```
gterzian: I think we need to add a `propagate` boolean, to `note_operations_from_node`, and set it to `true` when calling it from `on_block_full_operations_received`, and `false` otherwise.

gterzian (changes requested):
Looks good, should just add the boolean, you can use the test added at https://github.com/massalabs/massa/pull/3034

massa-protocol-worker/src/protocol_worker.rs
damip: Since separate op propagation, we don't need to do this anymore. We always propagate ops to all the nodes that don't know them (whether they come from blocks or naked). And if a node is missing some ops to fill up a block, they can ask them explicitly.

AurelienFT:
Should also correct #3036

damip:
All tests passed. merging

PR #3064, opened 2022-09-22
Executed ops hash and debug logs

Eitu33:


massa-final-state/src/executed_ops.rs
```diff
@@ -22,39 +25,66 @@ use std::ops::Bound::{Excluded, Included};
 
 /// A structure to list and prune previously executed operations
 #[derive(Debug, Default, Clone, PartialEq, Eq)]
-pub struct ExecutedOps(PreHashMap<OperationId, Slot>);
+pub struct ExecutedOps {
+    /// Map of the executed operations
+    ops: PreHashMap<OperationId, Slot>,
+    /// Cumulated hash of the executed operations
+    pub hash: Option<Hash>,
```
AurelienFT, line 32: Why having an option if better then having a default hash from empty data for example or filled of 0 ?

massa-final-state/src/executed_ops.rs
```diff
@@ -22,39 +25,66 @@ use std::ops::Bound::{Excluded, Included};
...
     pub fn extend(&mut self, other: ExecutedOps) {
-        self.0.extend(other.0);
+        if let Some(other_hash) = other.hash {
+            if let Some(current_hash) = self.hash.as_mut() {
+                *current_hash ^= other_hash;
+            } else {
+                self.hash = Some(other_hash);
+            }
+        }
+        self.ops.extend(other.ops);
```
AurelienFT, line 55: I think you should put this in scrope of the if `if let Some(other_hash) = other.hash` so that ip ensure you don't add executed ops without adding hash. Even if right now it's not possible and you know it because you made the code if someone else make modifications in future it could lead to errors.

massa-final-state/src/executed_ops.rs
```diff
@@ -22,39 +25,66 @@ use std::ops::Bound::{Excluded, Included};
...
-        self.0
-            .retain(|_id, last_valid_slot| *last_valid_slot >= max_slot);
+        let (kept, removed): (PreHashMap<OperationId, Slot>, PreHashMap<OperationId, Slot>) = self
+            .ops
+            .iter()
+            .partition(|(_, &last_valid_slot)| last_valid_slot >= max_slot);
+        let hash = self
+            .hash
+            .as_mut()
+            .expect("critical: an ExecutedOps object with ops must also contain a hash");
```
AurelienFT, line 83: This will fail if prune is called and no operations as been inserted yet ?

massa-final-state/src/executed_ops.rs
```diff
@@ -219,17 +249,24 @@ impl ExecutedOpsSerializer {
...
         self.u64_serializer.serialize(&entry_count, buffer)?;
 
         // encode entries
-        for (op_id, slot) in &value.0 {
+        for (op_id, slot) in &value.ops {
             buffer.extend(op_id.to_bytes());
             self.slot_serializer.serialize(slot, buffer)?;
         }
 
+        // note: mimic the behaviour of OptionSerializer because we have not hash serializer
```
AurelienFT, line 263: maybe open an issue to make a hash serializer.

massa-final-state/src/executed_ops.rs
```diff
@@ -262,17 +301,23 @@ impl Deserializer<ExecutedOps> for ExecutedOpsDeserializer {
...
-                    |input| self.slot_deserializer.deserialize(input),
-                )),
-            ),
+            tuple((
+                length_count(
+                    context("Failed length deserialization", |input| {
+                        self.u64_deserializer.deserialize(input)
+                    }),
+                    tuple((
+                        |input| self.operation_id_deserializer.deserialize(input),
```
AurelienFT, line 310: I think it should be cool to have some context for errors here

AurelienFT (commented):
Good one ! Just few interrogations and also I think one or two unit tests could be a great enhancement

massa-final-state/src/executed_ops.rs
Eitu33: Because `ExecutedOps` is created multiple times (in comparison with `Ledger` that is created only once), I'm suspect that having an initial hash in this case would result in different hashes between two nodes with the same executed ops list.

massa-final-state/src/executed_ops.rs
Eitu33: Added the following check right above:
```rust
if removed.is_empty() {
    return;
}
```

massa-final-state/src/executed_ops.rs
Eitu33: Implemented the hash serializer in the end.

AurelienFT (approved):

AurelienFT (approved):
Thank you for the support you give me ! I can count on you and that's insane.

Eitu33:
@AurelienFT no problem ser me n u 2 the moon together 🚀 🤝

AurelienFT (approved):

PR #3065, opened 2022-09-22
Remove useless transfer balance.

AurelienFT:
This transfer was useful when 2 balances were in place to transfer from sequential to parallel before sending to SCs. Now it perform only a ping pong so we can remove it.


Eitu33 (approved):

Eitu33:
No 🏓  here

damip:
In that case, the "coins" parameter for executesc should be removed completely

Eitu33 (approved):
No 🏓 and no 🪙  either

PR #3068, opened 2022-09-23
Fix spelling and force it in CI.

AurelienFT:


Eitu33 (approved):
Aurelien is a man of many words 🥇

PR #3075, opened 2022-09-23
Fix ExecutedOps extend XOR computing.

Eitu33:


AurelienFT (approved):
Without option it's easier to read. Well played

PR #3077, opened 2022-09-26
Improve storage perf in Execution

damip:
This PR aims at massively improving the performance of Execution.

* [x] try in sandbox/simulation/labnet
  * [x] make tests compile
  * [x] make tests pass 
* [x] add comments and docs


massa-execution-worker/src/controller.rs
```diff
@@ -87,21 +95,34 @@ pub struct ExecutionControllerImpl {
...
+
+        // extend block info
+        input_data.block_storage.extend(block_storage);
+
+        // extend finalized blocks
+        input_data.finalized_blocks.extend(finalized_blocks);
+
+        // update blockclique
+        if new_blockclique.is_some() {
+            input_data.new_blockclique = new_blockclique;
```
AurelienFT, line 121: I don't understand in the code where do we skip the process of the `new_blockclique` field if he is the same as before. Can you point me to this code @damip ?

massa-execution-worker/src/execution.rs
```diff
@@ -960,15 +885,120 @@ impl ExecutionState {
...
+        } else {
+            // cache entry absent
+            info!(
+                "speculative execution cache empty, executing final slot={}/block={:?}",
+                slot, target_id
+            );
+        }
+
+        // truncate the whole execution queue
+        self.active_history.write().0.clear();
```
AurelienFT, line 989: Why do you clear whole speculative cache when a block became final ?

AurelienFT (commented):
I don't understand in details the slot sequencer so I don't know if there is errors. For the rest it seems ok.

massa-execution-worker/src/controller.rs
damip: In consensus, we set new_blockclique to None if it's the same as before: https://github.com/massalabs/massa/blob/c0865fe4d79bb0f0e63c26332608b023a7697ebf/massa-consensus-worker/src/consensus_worker.rs#L608-L611

massa-execution-worker/src/execution.rs
damip: In most cases, the slot that is becoming final has already been executed as speculative in the past and is therefore present at the front of the active history.

So when a slot becomes final, we first check first element of the active history:
* if it is present and its input matches the one of the slot being finalized (same slot, same block ID or miss) then we can just remove that element from the front of the active history and apply its already-computed output to the final state. That's the normal situation
* if it is absent (history empty) it means that we just bootstrapped or that the execution is lagging a lot => we need to execute the slot now and apply its output to the final state
* if it is present but mismatches the slot being finalized, it means that the whole blockclique has changed suddenly (deep fork). This is theoretically not impossible but the probability of it happening is at the very limit of our security conditions. Hence the warning. In that case we need to discard the whole history and execute the slot being finialized, then apply its result to the final state.

In the code, we return early in the 1st case here https://github.com/massalabs/massa/blob/c0865fe4d79bb0f0e63c26332608b023a7697ebf/massa-execution-worker/src/execution.rs#L969-L972 and the clear doesn't happen.

In the 2 other cases the function continues and we do the history clear

damip:
> I don't understand in details the slot sequencer so I don't know if there is errors. For the rest it seems ok.

That means that I need to write the document explaining the algorithm. I'm on it

massa-execution-worker/src/controller.rs
AurelienFT: Yes but then if it's None in consensus it will not change the blockclique here but the field `new_blockclique` still old the previous state ? And if yes what prevent it to re-compute it ?

massa-execution-worker/src/execution.rs
AurelienFT: > and is therefore present at the front of the active history.

This is what my question is about why the final slot is at the front ? Because they become final one by one with the condition that the precedent one is final also ?

massa-execution-worker/src/controller.rs
damip: Exactly. We notify Execution only if there is a new blockclique. And when Execution takes the execution input data, it will always get the latest blockclique (or None if there is no new one since the last take). That way if Execution is slow, there might be multiple blockclique changes in-between takes and it won't have to process all blockcliques, just the latest one.

For final blocks it's different because they all need to be executed (new final blocks don't cancel previous ones). So we accumulate the final blocks in the execution input data so that when Execution takes them, it gets their whole accumulation to process them all.

There is also another optimization in the input data: the block_storage. Before this PR, whenever a block was finalized, we would send a full copy of its storage to Execution. Also, when the blockclique changed, we used to send a full copy of all the storages of all the blocks of the new blockclique. That was a lot of storage clones/drops. Now only send block_storage from consensus to execution for blocks that weren't sent to Execution before.

massa-execution-worker/src/execution.rs
damip: Exactly, SCE finality is slot-sequential, so the next slot to finalize is almost certainly the earliest current speculative slot.

damip:
> I don't understand in details the slot sequencer so I don't know if there is errors. For the rest it seems ok.

Should be clearer now

AurelienFT (approved):
Slot sequencer is very well documented and explicit even if it's big. If it has been tested in real conditions it's ok for me.

damip:
bors merge

damip:
bors merge

PR #3081, opened 2022-09-26
Rework consensus and refactor graph

AurelienFT:
This PR aims to remove consensus as it's now only a module that call block graph.

All the code from consensus will be removed or moved to graph.

The async from consensus will be remove also 

Changes in algorithm: 
- get_all_final_blocks will now return storage to avoid two fetching


AurelienFT:
@sydhds @gterzian  Can you review this PR that is a draft for now ? 

The architecture of consensus is ok to me I designed it in a separate crate. There is one/two little TODO I will address a bit more doc.

I tried to split the existing behavior to avoid having unreadable files and functions. I think I can still split a bit more will work on it.

I also added a lot of documentation even if some functions don't have yet.

To give a bit overview of the architecture. Graph will run in a separate thread. To interact with him we will use a controller that share a state with the thread. Pretty much all the code is in the state.  The thread write in the state and the controller read it.
Graph will make calculation each slots + whenever he receive a block/header from the controller.

massa-graph-2-worker/src/controller.rs
```diff
@@ -0,0 +1,136 @@
...
+use massa_storage::Storage;
+use parking_lot::RwLock;
+use std::sync::{mpsc::SyncSender, Arc};
+
+use crate::{commands::GraphCommand, state::GraphState};
+
+#[derive(Clone)]
+pub struct GraphControllerImpl {
+    command_sender: SyncSender<GraphCommand>,
+    shared_state: Arc<RwLock<GraphState>>,
```
gterzian, line 27: I think you should avoid mixing channels and locks, because the two will not synchronize and it could easily lead to confusing behavior. 

For example, if you send a `GraphCommand::RegisterBlock(id)`, and then do `get_block_statuses(vec![id])`, even if it's from the same thread, you will sometimes still get `NotFound` back. 

The solution is to either do everything with channels(using oneshots within commands to get replies when needed), or do everything with a lock/condvar combo such as `Arc<(Condvar, RwLock<(GraphState, Option<GraphCommand>)>)>`

gterzian (changes requested):
Looks good, just one comment on the overall architecture for now...

massa-graph-2-worker/src/controller.rs
damip: This is done mostly to be able to:
* send commands through the channel without waiting for them to be processed from the point of view of the sending thread, and channels are very much optimal for that (much faster than locks)
* still be able to read the current state of the graph as processed so far (for this we need a shared state)

Note that sending commands and reading the state is done from different, mutually-asynchronous tasks.

But I agree that some documentation warning that the two are not in sync is important.

AurelienFT:
Close in favor of #3162

PR #3084, opened 2022-09-27
Pool protection

adrien-zinger:
Response to a problem of accumulation of useless operations in the pool and avoid possible attacks.

Start a thread which clean periodically the pool.


damip:
In the end we took another direction on this one. Closing for now but keeping the branch.

Issue #3054, opened 2022-09-21
Robustify endorsement/operation/block propagation for testnet 15


damip:
* [ ] document the block/endo/op propagation processes
* [ ] document all functions and their behavior
* [ ] try in sandbox/simulation/labnet:
  * operation/endorsement/block propagation under heavy load
  * locate performance bottlenecks causing a slowdown of finality
* [ ] continue adding critical unit tests

PR #3086, opened 2022-09-27
Robustify operations propagation

gterzian:
FIX https://github.com/massalabs/massa/issues/3054

### DIFF
--- a/massa-node/base_config/config.toml
+++ b/massa-node/base_config/config.toml
@@ -81,6 +81,8 @@
     # Maximum number of batches in the memory buffer.
     # Dismiss the new batches if overflow
     operation_batch_buffer_capacity = 10024
+    # Immediately announce ops if overflow
+    operation_announcement_buffer_capacity = 10024
     # Start processing batches in the buffer each `operation_batch_proc_period` in millisecond
     operation_batch_proc_period = 500
     # All operations asked are prune each `operation_asked_pruning_period` millisecond
--- a/massa-node/base_config/config.toml
+++ b/massa-node/base_config/config.toml
@@ -85,6 +85,8 @@
     operation_batch_proc_period = 500
     # All operations asked are prune each `operation_asked_pruning_period` millisecond
     asked_operations_pruning_period = 100000
+    # Interval at which operations are announced in batches.
+    operation_announcement_interval = 500
     # Max number of operation per message, same as network param but can be smaller
     max_operations_per_message = 1024
     # Time threshold after which operation are not propagated
--- a/massa-node/base_config/config.toml
+++ b/massa-node/base_config/config.toml
@@ -86,7 +86,7 @@
     # All operations asked are prune each `operation_asked_pruning_period` millisecond
     asked_operations_pruning_period = 100000
     # Interval at which operations are announced in batches.
-    operation_announcement_interval = 500
+    operation_announcement_interval = 150
     # Max number of operation per message, same as network param but can be smaller
     max_operations_per_message = 1024
     # Time threshold after which operation are not propagated
--- a/massa-protocol-exports/src/tests/tools.rs
+++ b/massa-protocol-exports/src/tests/tools.rs
@@ -274,7 +274,7 @@ pub fn create_protocol_config() -> ProtocolConfig {
         operation_batch_buffer_capacity: 1000,
         operation_batch_proc_period: 200.into(),
         asked_operations_pruning_period: 500.into(),
-        operation_announcement_interval: 500.into(),
+        operation_announcement_interval: 150.into(),
         max_operations_per_message: 1024,
         thread_count: 32,
         max_serialized_operations_size_per_block: 1024,
--- a/massa-protocol-exports/src/tests/tools.rs
+++ b/massa-protocol-exports/src/tests/tools.rs
@@ -272,6 +272,7 @@ pub fn create_protocol_config() -> ProtocolConfig {
         max_known_endorsements_size: 1000,
         max_node_known_endorsements_size: 1000,
         operation_batch_buffer_capacity: 1000,
+        operation_announcement_buffer_capacity: 1000,
         operation_batch_proc_period: 200.into(),
         asked_operations_pruning_period: 500.into(),
         operation_announcement_interval: 150.into(),
--- a/massa-protocol-exports/src/tests/tools.rs
+++ b/massa-protocol-exports/src/tests/tools.rs
@@ -274,6 +274,7 @@ pub fn create_protocol_config() -> ProtocolConfig {
         operation_batch_buffer_capacity: 1000,
         operation_batch_proc_period: 200.into(),
         asked_operations_pruning_period: 500.into(),
+        operation_announcement_interval: 500.into(),
         max_operations_per_message: 1024,
         thread_count: 32,
         max_serialized_operations_size_per_block: 1024,
--- a/massa-protocol-worker/src/protocol_worker.rs
+++ b/massa-protocol-worker/src/protocol_worker.rs
@@ -1,7 +1,7 @@
 //! Copyright (c) 2022 MASSA LABS <info@massa.net>
 
 use crate::checked_operations::CheckedOperations;
-use crate::sig_verifier::verify_sigs;
+use crate::sig_verifier::verify_sigs_batch;
 use crate::{node_info::NodeInfo, worker_operations_impl::OperationBatchBuffer};
 
 use massa_logging::massa_trace;
@@ -929,7 +929,7 @@ impl ProtocolWorker {
         }
 
         // optimized signature verification
-        verify_sigs(
+        verify_sigs_batch(
             &new_operations
                 .iter()
                 .map(|(op_id, op)| (*op_id.get_hash(), op.signature, op.creator_public_key))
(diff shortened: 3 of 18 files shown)


gterzian:
@AurelienFT Could you please run the `charge.py` command with the simulator on this branch? I'm having some troubles installing it locally(see https://github.com/massalabs/massa-network-simulator/issues/35). Thanks!

gterzian:
@massalabs/core-team Ready for review.

Haven't been able to run the simulator yet(see the various [issues](https://github.com/massalabs/massa-network-simulator/issues)I filed), so if someone who has could run this on through it, it would be appreciated.

massa-protocol-worker/src/protocol_worker.rs
```diff
@@ -474,7 +490,22 @@ impl ProtocolWorker {
                 );
             }
             ProtocolCommand::PropagateOperations(storage) => {
-                self.propagate_ops(&storage).await;
+                // Note: should we claim refs locally?
```
gterzian, line 493: Wondering about this.

massa-protocol-worker/src/protocol_worker.rs
```diff
@@ -213,6 +216,7 @@ impl ProtocolWorker {
                 config.operation_batch_buffer_capacity,
             ),
             storage,
+            operations_to_announce: Default::default(),
```
damip, line 219: you can preallocate it with the max op batch announcement size

massa-protocol-worker/src/protocol_worker.rs
```diff
@@ -292,9 +296,14 @@ impl ProtocolWorker {
                     self.update_ask_block(&mut block_ask_timer).await?;
                 }
 
-                // operation ask timer
+                // operation ask, and announce, timer
                 _ = &mut operation_batch_proc_period_timer => {
```
damip, line 300: It should be a different timer

massa-protocol-worker/src/protocol_worker.rs
```diff
@@ -344,6 +346,20 @@ impl ProtocolWorker {
...
+    /// Add an list of operations to a buffer for announcement at the next interval,
+    /// or immediately if the buffer is full.
+    async fn note_operations_to_announce(&mut self, operations: &[OperationId]) {
+        // Add the operations to a list for announcement at the next interval.
+        self.operations_to_announce.extend_from_slice(operations);
+
+        // If the buffer is full,
+        // announce operations immediately,
+        // clearing the data at the same time.
+        if self.operations_to_announce.len() > self.config.max_known_ops_size {
```
damip, line 358: the buffer size should not be max_known_ops_size but another specific config parameter (eg. op_announcement_buffer_size)

massa-protocol-worker/src/protocol_worker.rs
```diff
@@ -344,6 +346,20 @@ impl ProtocolWorker {
         }
     }
```
damip, line 347: at the end of announce_ops, reset the announcement timer

massa-protocol-worker/src/protocol_worker.rs
```diff
@@ -309,33 +318,26 @@ impl ProtocolWorker {
...
 
-    /// request the propagation of a set of operations
-    async fn propagate_ops(&mut self, storage: &Storage) {
-        let operation_ids = storage.get_op_refs();
+    /// Announce a set of operations to active nodes who do not know about it yet.
+    /// Side effects:
+    /// - notes nodes as knowing about those operations from now on.
+    /// - empties the buffer of operations to announce.
+    async fn announce_ops(&mut self) {
```
damip, line 325: * reset the announcement timer
* if the buffer is empty, quit to avoid iterating on nodes

massa-protocol-worker/src/protocol_worker.rs
damip: That's an important question.

For now I think we can just ignore it because we will end up sending only the ops that are kept in the pool.
In the future (eg. when the pool gets separated out) we should somehow keep the refs to those ops for the duration of the propagation.

damip:
@gterzian

Very good work ! (with a few caveats)

Some missing things to make it really effective:
* can you integrate https://github.com/massalabs/massa/pull/3095 in this PR now that we have batches ?
* once it's done, can you also try again with the simulator that was corrected by @sydhds  and @AurelienFT ?
* then finish the robustify checklist in https://github.com/massalabs/massa/issues/3054

massa-protocol-worker/src/protocol_worker.rs
gterzian: Why? It seems to make sense to announce, and ask, at the same interval.

massa-protocol-worker/src/protocol_worker.rs
gterzian: How about we re-use `operation_batch_buffer_capacity`?

massa-protocol-worker/src/protocol_worker.rs
gterzian: Ah ok I see that the ask is not done at a fixed interval.

massa-protocol-worker/src/protocol_worker.rs
gterzian: > reset the announcement timer

I think I'll just use a tokio interval.

gterzian:
@damip Thanks for the review, comments addressed. I suggest we merge this one first, and then I can integrate the parallelization via a rebase of [that branch](https://github.com/massalabs/massa/pull/3095).

massa-protocol-worker/src/protocol_worker.rs
damip: The operation asking system and the announcement batching system are separate, independent mechanisms. Let's keep the configs separate for more flexibility

massa-protocol-worker/src/protocol_worker.rs
damip: The operation asking system and the announcement batching system are separate, independent mechanisms. Let's keep the configs separate for more flexibility

massa-node/base_config/config.toml
```diff
@@ -85,6 +85,8 @@
     operation_batch_proc_period = 500
     # All operations asked are prune each `operation_asked_pruning_period` millisecond
     asked_operations_pruning_period = 100000
+    # Interval at which operations are announced in batches.
+    operation_announcement_interval = 500
```
damip, line 89: 500ms is very high. Let's set it to 150ms

massa-protocol-exports/src/tests/tools.rs
```diff
@@ -274,6 +274,7 @@ pub fn create_protocol_config() -> ProtocolConfig {
         operation_batch_buffer_capacity: 1000,
         operation_batch_proc_period: 200.into(),
         asked_operations_pruning_period: 500.into(),
+        operation_announcement_interval: 500.into(),
```
damip, line 277: 150 ms

massa-protocol-worker/src/protocol_worker.rs
damip: @gterzian  the 1st point (resetting the timer) was not addressed in the case where the buffer is full and we send immediately

gterzian: @damip Does it matter? We could also just remove that, or do it first and leave the newly added ops in the buffer for the next tick. We can tweak it later based on how the batching actually turns out I think.

massa-protocol-worker/src/protocol_worker.rs
gterzian: Ok, done.

massa-protocol-worker/src/protocol_worker.rs
```diff
@@ -292,9 +303,24 @@ impl ProtocolWorker {
...
+
+                    // Reset timer.
+                    let now = Instant::now();
+                    let next_tick = now
+                        .checked_add(self.config.operation_announcement_interval.into())
+                        .ok_or(TimeError::TimeOverflowError)?;
+                    operation_announcement_interval.set(sleep_until(next_tick));
+                }
+
+                // operation ask, and announce, timer
```
AurelienFT, line 319: The documentation seems old here

gterzian: yes, removed.

gterzian:
@AurelienFT @damip Thanks for the reviews, I think I have addressed all your comments. The protocol worker test suite passes, however I still haven't been able to run it on the simulator, and it doesn't look like I'll be able to for today.

massa-protocol-worker/src/protocol_worker.rs
damip: * if we only send on timer, sometimes we will get huge batches and extra latency for no reason
* if we send on buffer full but only reset the timer when it ticks, sometimes we will send tiny batches for no reason except that we sent a full buffer just before
* if we reset the timer every time we send, we get the best compromise: low latency when the buffer is full, otherwise maximal buffer size at a given capped latency

damip:
Ok, @sydhds  can you take over for us to finish in time ?

damip:
@gterzian  It's crucial that you figure out how to run the simulator, we can't go without it anymore. Let's make sure you have it for tomorrow, and open issues if anything goes wrong to get help until it works.

PR #3107, opened 2022-10-04
Change storage constants

AurelienFT:


docs/technical-doc/storage-costs.rst
```diff
@@ -5,29 +5,29 @@ Storage Costs
 Explanations
 ------------
 
-In Massa the ledger is shared across all nodes of the network. We need to set a size limit to be able to run a node without
+In Massaa the ledger is shared across all nodes of the network. We need to set a size limit to be able to run a node without
```
qdrn, line 8: Typo in Massa :)

qdrn:
Side note: it's probably obvious to us and people familiar with blockchains that addresses are 32 bytes but not necessarily for newcomers. It could be more explicit I think.

PR #3122, opened 2022-10-05
issues 2721

MerryPoppins92:


qdrn:
Hello @MerryPoppins92 !
Thanks for your contribution! Is your PR ready to be merged ?

MerryPoppins92:
Hello Adrien,

Thank you for reaching me and hope you are doing well!

I receive an email from github saying that the merge is not possible due to
some issues. I would like to do a code review with someone if it's possible

excellente journee

Merry

massa-node/src/main.rs
```diff
@@ -25,22 +26,21 @@ use massa_ledger_worker::FinalLedger;
...
-    VERSION,
+    MAX_BOOTSTRAP_FINAL_STATE_PARTS_SIZE, MAX_BOOTSTRAP_MESSAGE_SIZE, MAX_DATASTORE_ENTRY_COUNT,
+    MAX_DATASTORE_KEY_LENGTH, MAX_DATASTORE_VALUE_LENGTH, MAX_DATA_ASYNC_MESSAGE,
+    MAX_ENDORSEMENTS_PER_MESSAGE, MAX_FUNCTION_NAME_LENGTH, MAX_GAS_PER_BLOCK,
+    MAX_LEDGER_CHANGES_COUNT, MAX_MESSAGE_SIZE, MAX_OPERATIONS_PER_BLOCK, MAX_PARAMETERS_SIZE,
+    NETWORK_CONTROLLER_CHANNEL_SIZE, NETWORK_EVENT_CHANNEL_SIZE, OPERATION_VALIDITY_PERIODS,
+    PERIODS_PER_CYCLE, ROLL_PRICE, T0, THREAD_COUNT, VERSION,
+};
+use massa_models::config::{
+    ASYNC_POOL_PART_SIZE_MESSAGE_BYTES, CHANNEL_SIZE, DELTA_F0, NETWORK_NODE_COMMAND_CHANNEL_SIZE,
```
AurelienFT, line 41: Import all constants in the same one as above.

massa-node/src/main.rs
```diff
@@ -80,6 +89,7 @@ async fn launch(
     StopHandle,
     StopHandle,
 ) {
+    // info!("{}", boot);
```
AurelienFT, line 92: It's a debug print that should be removed before merging (same for all above)

massa-node/src/main.rs
```diff
@@ -601,6 +607,8 @@ struct Args {
     /// Wallet password
     #[structopt(short = "p", long = "pwd")]
     password: Option<String>,
+    #[structopt(short = "b", long = "boot")]
```
AurelienFT, line 610: A documentation is needed for file that will be used by users.

massa-node/src/settings.rs
```diff
@@ -197,6 +192,60 @@ pub struct ProtocolSettings {
...
+    let file = "./src/whiteliste.yaml";
+    let mut file = File::open(file).expect("Unable to open file");
+    let mut contents = String::new();
+
+    file.read_to_string(&mut contents)
+        .expect("Unable to read file");
+
+    let docs = YamlLoader::load_from_str(&contents).unwrap();
+    let mut bootstrap_list: Vec<SocketAddr> = vec![];
+    for s in docs[0]["whitelist"].as_vec().unwrap() {
```
AurelienFT, line 236: You can use a map here

massa-node/src/settings.rs
```diff
@@ -180,15 +180,10 @@ pub struct ProtocolSettings {
     /// Maximum number of batches in the memory buffer.
     /// Dismiss the new batches if overflow
     pub operation_batch_buffer_capacity: usize,
-    /// Maximum number of operations in the announcement buffer.
-    /// Immediately announce if overflow.
-    pub operation_announcement_buffer_capacity: usize,
```
AurelienFT, line 185: This shouldn't be deleted

massa-node/src/main.rs
```diff
@@ -601,6 +607,8 @@ struct Args {
...
+    boot: Option<String>,
```
AurelienFT, line 611: Passing an argument to use a config file is complex for user. The most easy way for the user to do it it's like our config.toml. We create a file by default in your case whitelist: [] and then the user can edit it.
Sadly this doesn't match the requirements of the issue that ask this list to be live editable. I think a way to do it live executable is to use the API.

massa-node/src/main.rs
```diff
@@ -649,6 +657,17 @@ async fn main(args: Args) -> anyhow::Result<()> {
         .with(tracing_layer)
         .init();
 
+    let x = match args.boot {
+        None => {
```
AurelienFT, line 661: You can use is_some : https://doc.rust-lang.org/std/option/enum.Option.html#method.is_some to perform this conversion

qdrn:
I think you need to merge the main branch (at least).

massa-node/src/main.rs
```diff
@@ -149,8 +159,25 @@ async fn launch(
     let stop_signal = signal::ctrl_c();
     tokio::pin!(stop_signal);
 
+    // info!("{:?}", SETTINGS.bootstrap.bootstrap_list);
+
+    let bootstrap_list = match boot {
+        true =>  {
+            let x = load_file2();
```
AurelienFT, line 166: You are not using the return of the function that read the file you are just using the number of entries in it and you are generating new public key instead of using the ones in the file.

massa-node/src/whiteliste.yaml
```diff
@@ -0,0 +1,8 @@
+whitelist:
```
AurelienFT, line 1: We use toml for our config file I think we should keep toml for consistency.

AurelienFT (commented):
I don’t see any modification on the behavior of the bootstrap or anythink. The issue ask that the bootstrap server use a whitelist to allow or not people to bootstrap from this server. Your code just modify the list that is used to bootstrap as a client. 
Also the code have a lot of debug prints and dead code that should be in a PR ready to review. The code doesn’t compile because not all modifications is pushed. The code isn’t formatted also and not clippy compatible

damip (changes requested):
PR rejected (see comments)

PR #3123, opened 2022-10-05
Add Denoucement op + ser/deser + Hasher trait

sydhds:


massa-models/src/denunciation.rs
```diff
@@ -0,0 +1,628 @@
...
+    ///   endorsed_block: BlockId(Hash::compute_from("test".as_bytes()))
+    /// };
+    /// let mut buffer = Vec::new();
+    /// EndorsementSerializerLW::new().serialize(&endorsement, &mut buffer).unwrap();
+    /// ```
+    fn serialize(&self, value: &Denunciation, buffer: &mut Vec<u8>) -> Result<(), SerializeError> {
+        self.slot_serializer.serialize(&value.slot, buffer)?;
+        buffer.extend(value.pub_key.to_bytes());
+        let denunciation_kind = value.is_for_block() as u8;
+        buffer.extend([denunciation_kind]);
```
AurelienFT, line 242: You can use push

massa-models/src/denunciation.rs
```diff
@@ -0,0 +1,628 @@
...
+                context("Failed pub_key deserialization", |input| {
+                    self.pub_key_deserializer.deserialize(input)
+                }),
+                context("Failed slot deserialization", |input| take(1usize)(input)),
+            )),
+        )
+        .parse(buffer)?;
+
+        // TODO: rework this
+        let is_for_block_ = matches!(is_for_block, [1]);
```
AurelienFT, line 312: You could use the num_enum crate that we use in other modules like bootstrap and network

massa-models/src/denunciation.rs
```diff
@@ -0,0 +1,628 @@
...
+
+        let (rem2, proof): (_, DenunciationProof) = match is_for_block_ {
+            true => context(
+                "Failed Block Denunciation deser",
+                tuple((
```
AurelienFT, line 317: You could create a parser for this 4 fields that are commun with endorsements and use it to avoid code duplication

massa-models/src/denunciation_interest.rs
```diff
@@ -0,0 +1,18 @@
+// Copyright (c) 2022 MASSA LABS <info@massa.net>
+
+use crate::endorsement::WrappedEndorsement;
+use crate::operation::WrappedOperation;
+use crate::block::WrappedHeader;
+
+/// What can be sent to Denunciation factory (in order to create Denunciation)
+#[derive(Debug)]
+pub enum DenunciationInterest {
```
AurelienFT, line 9: Maybe `DenunciationEvent` is a naming that match more the rest of the project

massa-models/src/operation.rs
```diff
@@ -631,6 +652,10 @@ impl OperationTypeDeserializer {
                 max_op_datastore_key_length,
                 max_op_datastore_value_length,
             ),
+            denunciation_deserializer: DenunciationDeserializer::new(
+                // FIXME
+                32, 16,
```
AurelienFT, line 657: To not forget

massa-pool-worker/src/endorsement_pool.rs
```diff
@@ -72,6 +80,12 @@ impl EndorsementPool {
             }
         }
         self.storage.drop_endorsement_refs(&removed);
+
+        // Send to Denunciation factory
+        let de_interest = DenunciationInterest::Final(self.last_cs_final_periods.clone());
+        self.de_items_tx
+            .send(de_interest)
```
AurelienFT, line 87: I think we can use try_send to avoid blocking here

massa-protocol-worker/src/protocol_worker.rs
```diff
@@ -934,6 +951,14 @@ impl ProtocolWorker {
 
             // Check operation signature only if not already checked.
             if !self.checked_operations.contains_id(&operation_id) {
+
+                // Skip invalid Denunciation op
+                if !match &operation.content.op {
```
AurelienFT, line 956: I don't understand well this condition I think it need more doc. But as far as I understand what does happens if others operations are also invalid ? We will say pass but we should fail

massa-protocol-exports/src/protocol_controller.rs
```diff
@@ -157,6 +157,20 @@ impl ProtocolCommandSender {
...
 
+    /// Propagate a batch of operation ids (from pool).
+    ///
+    /// note: Full `OperationId` is replaced by a `OperationPrefixId` later by the worker.
+    pub fn propagate_operations_sync(&mut self, operations: Storage) -> Result<(), ProtocolError> {
+        massa_trace!("protocol.command_sender.propagate_operations", {
+            "operations": operations.get_op_refs()
+        });
+        self.0
+            .blocking_send(ProtocolCommand::PropagateOperations(operations))
```
AurelienFT, line 168: I think try_send here could be better

massa-pool-worker/src/endorsement_pool.rs
```diff
@@ -115,6 +129,12 @@ impl EndorsementPool {
                         panic!("endorsement is expected to be absent from endorsements_sorted at this point");
                     }
                     added.insert(endo.id);
+
+                    // Send endorsements to Denunciation Factory
+                    let de_interest = DenunciationInterest::WrappedEndorsement(endo.clone());
```
AurelienFT, line 134: The operations and header are sent to denunciation_factory from protocol why endorsement is sent from the pool ?

massa-protocol-worker/src/protocol_worker.rs
```diff
@@ -909,6 +919,13 @@ impl ProtocolWorker {
             massa_trace!("protocol.protocol_worker.note_header_from_node.ok", { "node": source_node_id,"block_id":block_id, "header": header});
             return Ok(Some((block_id, true)));
         }
+
+
+        if let Err(e) = self.de_factory_sender.send(
```
AurelienFT, line 924: Use try send here maybe

massa-protocol-worker/src/protocol_worker.rs
```diff
@@ -996,6 +1021,22 @@ impl ProtocolWorker {
...
+            let de_interest_ops = new_operations
+                .iter()
+                .filter(|(_, wrapped_op)| {
+                    matches!(wrapped_op.content.op, OperationType::Denunciation { .. })
+                })
+                .map(|(_, wrapped_op)| wrapped_op.clone())
+                .collect::<Vec<WrappedOperation>>();
+
+            if !de_interest_ops.is_empty() {
+                if let Err(e) = self.de_factory_sender.send(
```
AurelienFT, line 1034: Use try send here maybe

AurelienFT (commented):
Good work ! The factory is well ! A lot of variables names have dash at the end is there any reasons ?

massa-models/src/denunciation_interest.rs
Eitu33, line 9: As said this morning I believe this should be in `massa-factory-exports` but it's really no big deal if it stays here tbh

massa-models/src/operation.rs
```diff
@@ -631,6 +652,10 @@ impl OperationTypeDeserializer {
                 max_op_datastore_key_length,
                 max_op_datastore_value_length,
             ),
+            denunciation_deserializer: DenunciationDeserializer::new(
+                // FIXME
```
Eitu33, line 656: Do not forget this

massa-pos-exports/src/deferred_credits.rs
```diff
@@ -53,6 +54,27 @@ impl DeferredCredits {
...
+
+        if self.0.contains_key(slot) {
+            self.0
+                .entry(*slot)
+                .and_modify(|current_credits| {
+                    current_credits
+                        .entry(*address)
+                        .and_modify(|e| {
+                            let sub_amount = e.saturating_sub(*amount);
+                            result = Some(min(*e, *amount));
```
Eitu33, line 70: I'm not sure I understand the behaviour of this return value, min of the address' credit amount at the given slot and the amount to be removed? In what case is that needed?

massa-protocol-worker/src/protocol_worker.rs
```diff
@@ -996,6 +1021,22 @@ impl ProtocolWorker {
             self.note_operations_to_announce(&to_announce, op_timer)
                 .await;
 
+            // send Operation(Denunciation) to DenunciationFactory
+            let de_interest_ops = new_operations
+                .iter()
+                .filter(|(_, wrapped_op)| {
+                    matches!(wrapped_op.content.op, OperationType::Denunciation { .. })
+                })
+                .map(|(_, wrapped_op)| wrapped_op.clone())
```
Eitu33, line 1030: You can use a `filter_map()` here instead of `map` then `filter`

Eitu33 (commented):
Nice work 😄 Just a few questions / suggestions

massa-execution-worker/src/execution.rs
```diff
@@ -616,6 +620,64 @@ impl ExecutionState {
...
     }
 
+    pub fn execute_denunciation_op(
+        &self,
+        operation: &OperationType,
+        _sender_addr: Address,
+        block_credits: &mut Amount,
+    ) -> Result<(), ExecutionError> {
+
+        const ROLL_COUNT: u64 = 1;
```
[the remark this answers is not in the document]
sydhds: Yes only 1 roll (or min(100, available) coins from deferred credits

massa-protocol-worker/src/protocol_worker.rs
sydhds: Here we are only checking if Denunciation operation are invalid, others operations will always be inserted in new operations no?

massa-protocol-exports/src/protocol_controller.rs
sydhds: propagate_endorsements use blocking_send as well as the tokio mpsc is bounded. Why do you think 'try_send' is better and what do you think should be done if try_send fails?

massa-pool-worker/src/endorsement_pool.rs
sydhds: If I am not mistaken, the endorsements from both the network and the endorsement factory are stored in the endorsement pool and thus we can feed the denunciation factory from only one place. What do you think?

massa-execution-worker/src/execution.rs
```diff
@@ -616,6 +620,64 @@ impl ExecutionState {
...
+                .checked_mul_u64(ROLL_COUNT)
+                .ok_or_else(||
+                    ExecutionError::RuntimeError(format!("Unable to multiply roll price by {}", ROLL_COUNT))
+                )?;
+
+            // Set call stack
+            // This needs to be defined before anything can fail, so that the emitted event contains the right stack
+            context.stack = vec![ExecutionStackElement {
+                address: addr_denounced,
+                coins: amount,
```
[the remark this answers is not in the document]
sydhds: done

massa-execution-worker/src/execution.rs
```diff
@@ -305,6 +305,9 @@ impl ExecutionState {
             OperationType::Transaction { .. } => {
                 self.execute_transaction_op(&operation.content.op, sender_addr)
             }
+            OperationType::Denunciation { .. } => {
```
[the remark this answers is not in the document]
sydhds: Good question... What do you think @damip ?

massa-execution-worker/src/speculative_roll_state.rs
```diff
@@ -476,4 +507,55 @@ impl SpeculativeRollState {
...
+
+        // First try to rm in added_changes
+        let res = self.added_changes.deferred_credits.sub_amount(slot, addr, &amount_to_rm);
+        if let Some(a) = res {
+            amount_removed = amount_removed.checked_add(a)?;
+            amount_to_rm = amount_to_rm.checked_sub(amount_removed)?;
+        }
+
+        // If not found, try to rm from active history
+        let mut hist = self.active_history.write();
```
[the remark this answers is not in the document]
sydhds: done

massa-protocol-worker/src/protocol_worker.rs
sydhds: good catch! fixed!

massa-protocol-worker/src/protocol_worker.rs
sydhds: Also fixed!

sydhds:
> Good work ! The factory is well ! A lot of variables names have dash at the end is there any reasons ?

Most of the time, I put a dash at the end of a var if it's a Result or an Option and I extract the 'real' var after like:

```
let amount_: Option<Amount> = ...
if let Some(amount) = amount_ {
  todo!()
}
```

Does it make sense? What would you write on your side?

massa-execution-worker/src/context.rs
```diff
@@ -635,6 +635,59 @@ impl ExecutionContext {
...
+                .checked_mul_u64(roll_count)
+                .ok_or_else(||
+                    ExecutionError::RuntimeError(format!("Cannot mult roll price by {}", roll_count))
+                )?;
+
+            // Remove deferred credits starting at self.slot until
+            let mut amount_slashed = Amount::zero();
+            let mut amount_to_rm = amount_to_slash;
+            for (slot, _amount) in credits.iter() {
+                if amount_slashed >= amount_to_slash {
```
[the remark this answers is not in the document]
sydhds: I only break if amount_slashed (== 2 in your example) >= amount_to_slash (== 4 in your example) or did I miss something here?
As remove_deferred_credits returns the amount removed (that can be < to the amount given as a parameter) I need to use amount_rm no?

massa-execution-worker/src/context.rs
```diff
@@ -635,6 +635,59 @@ impl ExecutionContext {
...
+    {
+        let slashed_coins;
+
+        // First try to slash roll if any
+        if self.speculative_roll_state.try_burn_rolls(
+            denounced_addr,
+            roll_count,
+        ).is_err() {
+
+            // if no roll were slashed, that can likely be done from deferred credit
```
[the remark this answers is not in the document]
sydhds: If selected, a participant has at least 1 roll. If he tries to sell it in the mean time, then he has 0 roll but some deferred credits.
I you read the code for the function 'is_expired_for_denunciation', you will see that we reject Denunciation based on a constant DENUNCIATION_EXPIRE_CYCLE_DELTA_EXPIRE_COUNT (== 3 cycles) that should always guarantee some deferred credits are there (4 cycles to sell a roll). What do you think of this?

massa-execution-worker/src/execution.rs
```diff
@@ -616,6 +620,64 @@ impl ExecutionState {
...
+                    let amount = slashed_amount
+                        .checked_div_u64(2)
+                        .ok_or_else(||
+                            ExecutionError::RuntimeError(
+                                format!("Unable to divide slashed amount: {} by 2", slashed_amount))
+                        )?;
+                    *block_credits = block_credits.saturating_add(amount);
+                },
+                Err(e) => warn!("Unable to slash rolls or deferred credits: {}", e),
+            }
```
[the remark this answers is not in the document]
sydhds: This is an incentive for the Block creator to include this Denunciation. And remember that the block reward is also for the endors'ed and the endorsement creator...

massa-execution-worker/src/speculative_roll_state.rs
```diff
@@ -476,4 +507,55 @@ impl SpeculativeRollState {
...
+                .final_state
+                .write()
+                .pos_state
+                .rm_deferred_credits(slot, addr, amount);
+
+            if let Some(a) = res {
+                amount_removed = amount_removed.checked_add(a)?;
+                // No need to update amount_to_tm here
+            }
+        }
```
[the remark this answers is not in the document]
sydhds: In my unit tests, the deferred credits are in added_changes but I dunno if it's relevant from the 'real' cases. What do you think @damip ?

massa-execution-worker/src/tests/scenarios_mandatories.rs
```diff
@@ -850,6 +854,244 @@ fn events_from_switching_blockclique() {
     manager.stop();
 }
 
+#[test]
+#[serial]
+pub fn slash_roll() {
```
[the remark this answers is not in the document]
sydhds: +100. I've added some more documentation. Let me know if it now looks ok...

massa-protocol-worker/src/protocol_worker.rs
sydhds: done

massa-models/src/denunciation_interest.rs
sydhds: I seems like it's gonna add quite some useless dependencies. Let's talk about that...

massa-factory-worker/src/denunciation_factory.rs
```diff
@@ -0,0 +1,375 @@
...
+
+    fn process_new_endorsement(&mut self, wrapped_endorsement: WrappedEndorsement) {
+
+        let key = (wrapped_endorsement.content.slot, wrapped_endorsement.content.index);
+        if self.seen_endorsement_denunciation.contains(&key) {
+            return;
+        }
+
+        if self.seen_endorsement_denunciation.len() > ENDORSEMENT_DENUNCIATION_CACHE_MAX_SIZE ||
+            self.endorsements_by_slot_index.len() > ENDORSEMENT_BY_CACHE_MAX_SIZE {
```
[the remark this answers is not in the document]
sydhds: True for the other constants but these one are only temporary limits so maybe we can keep them in this file?

massa-execution-worker/src/speculative_roll_state.rs
```diff
@@ -476,4 +507,55 @@ impl SpeculativeRollState {
...
+        let res = self.added_changes.deferred_credits.sub_amount(slot, addr, &amount_to_rm);
+        if let Some(a) = res {
+            amount_removed = amount_removed.checked_add(a)?;
+            amount_to_rm = amount_to_rm.checked_sub(amount_removed)?;
+        }
+
+        // If not found, try to rm from active history
+        let mut hist = self.active_history.write();
+        while amount_removed < *amount {
+            match hist.0.iter_mut().rev().next() {
```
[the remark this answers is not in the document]
sydhds: As it's in a while loop, it seems I cannot use back_mut() here no?

massa-factory-worker/src/denunciation_factory.rs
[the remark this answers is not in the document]
sydhds: And done for the others

massa-factory-worker/src/block_factory.rs
```diff
@@ -192,7 +193,41 @@ impl BlockFactoryWorker {
...
+                    OperationType::Denunciation { data: de } => {
+                        let op = Operation {
+                            fee: Default::default(),
+                            expire_period: 0,
+                            op: OperationType::Denunciation { data: de.clone() },
+                        };
+                        // safe to unwrap as this should never fail
+                        let wrapped_op: WrappedOperation = Operation::new_wrapped(op,
+                                               OperationSerializer::new(),
+                                               block_producer_keypair).unwrap();
```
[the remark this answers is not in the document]
sydhds: Stealing is implemented because nothing can prevent a block producer to do so and it 'should' do it to maximize its profits.

massa-factory-worker/src/block_factory.rs
```diff
@@ -234,6 +271,15 @@ impl BlockFactoryWorker {
             block_id, slot, block_producer_addr
         );
 
+        // Send new operations (stolen denunciations) to pool
+        self.channels.storage.store_operations(new_ops.clone());
```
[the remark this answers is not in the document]
sydhds: Done

massa-factory-worker/src/denunciation_factory.rs
```diff
@@ -0,0 +1,375 @@
...
+use massa_models::timeslots::get_closest_slot_to_timestamp;
+
+const DENUNCIATION_EXPIRE_CYCLE_DELTA_EXPIRE_COUNT: u64 = 3;
+const ENDORSEMENT_DENUNCIATION_CACHE_MAX_SIZE: usize = 4096;
+const ENDORSEMENT_BY_CACHE_MAX_SIZE: usize = 4096;
+const BLOCK_HEADER_DENUNCIATION_CACHE_MAX_SIZE: usize = 4096;
+const BLOCK_HEADER_BY_CACHE_MAX_SIZE: usize = 4096;
+
+/// Structure gathering all elements needed by the factory thread
+pub(crate) struct DenunciationFactoryWorker {
```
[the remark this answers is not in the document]
sydhds: done

massa-protocol-exports/src/protocol_controller.rs
AurelienFT: Because I don't think you need to block here you are not waiting any data and I don't know why propagate_endorsement is blocking. It could be non-blocking also

massa-pool-worker/src/endorsement_pool.rs
AurelienFT: It's the same for block header that can be produce from block factory they are both sent to consensus. Consensus receive them all I think it should be better to make this send from the consensus.

AurelienFT:
> > Good work ! The factory is well ! A lot of variables names have dash at the end is there any reasons ?
> 
> Most of the time, I put a dash at the end of a var if it's a Result or an Option and I extract the 'real' var name after, like:
> 
> ```
> let amount_: Option<Amount> = ...
> if let Some(amount) = amount_ {
>   todo!()
> }
> ```
> 
> Does it make sense? What would you write on your side?

Ok I understand. It's ok for me nice convention

massa-pos-exports/src/deferred_credits.rs
sydhds: This can happen if we try to remove more than what is available. 

Let say we want to remove 43 credits when 42 are available; at the end we return 42.

massa-factory-worker/src/denunciation_factory.rs
```diff
@@ -0,0 +1,375 @@
...
+use massa_models::endorsement::WrappedEndorsement;
+use itertools::Itertools;
+use massa_models::operation::{Operation, OperationSerializer, OperationType, WrappedOperation};
+use crossbeam_channel::{select, Receiver};
+use massa_models::block::WrappedHeader;
+use massa_models::denunciation::DenunciationProof;
+use massa_models::denunciation_interest::DenunciationInterest;
+use massa_models::timeslots::get_closest_slot_to_timestamp;
+
+const DENUNCIATION_EXPIRE_CYCLE_DELTA_EXPIRE_COUNT: u64 = 3;
```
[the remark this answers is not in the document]
sydhds: done

massa-models/src/operation.rs
sydhds: Fixed

massa-execution-worker/src/context.rs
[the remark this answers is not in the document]
Eitu33: Oh yeah ok that makes sense

massa-execution-worker/src/execution.rs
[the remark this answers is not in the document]
Eitu33: > This is an incentive for the Block creator to include this Denunciation

Okay but then I don't see any code that rewards the producer of a denunciation, I just don't see for what reason they would do it, what do they gain from it? This is also linked to the stealing system we talk about below, I get that the block producer gets its way but what incentive is there for the denunciator?

> remember that the block reward is also for the endors'ed and the endorsement creator

That's what I meant by "selected addresses"

massa-execution-worker/src/speculative_roll_state.rs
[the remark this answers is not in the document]
Eitu33: Ah yeah my bad I missed it

massa-execution-worker/src/tests/scenarios_mandatories.rs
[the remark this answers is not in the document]
Eitu33: Yep that's great 👍

massa-factory-worker/src/block_factory.rs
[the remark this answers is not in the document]
Eitu33: Ok I see, but then my next question would be https://github.com/massalabs/massa/pull/3123#discussion_r1005705008

massa-models/src/denunciation_interest.rs
Eitu33: Yeah after giving it some more thought I agree with you, since it would require to include the factory exports in modules that only need this struct

massa-execution-worker/src/speculative_roll_state.rs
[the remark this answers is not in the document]
sydhds: Fixed

massa-factory-worker/src/block_factory.rs
```diff
@@ -192,7 +193,41 @@ impl BlockFactoryWorker {
...
 
         // gather operations and compute global operations hash
-        let (op_ids, op_storage) = self.channels.pool.get_block_operations(&slot);
+        let (op_ids_, op_storage) = self.channels.pool.get_block_operations(&slot);
+
+        // 'Steal' op denunciation
+        // TODO: add doc about why we stole those
+        let ops = op_storage.read_operations();
+        let mut new_ops: Vec<WrappedOperation> = Vec::new();
+        let new_ops_ref = &mut new_ops;
```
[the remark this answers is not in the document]
sydhds: Old and useless code. Removed!

massa-execution-worker/src/execution.rs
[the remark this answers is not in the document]
sydhds: From what I remember, the producer (of a Denunciation) wants to avoid bad actors on the graph (in order to protect its coins and rolls).

massa-factory-worker/src/denunciation_factory.rs
```diff
@@ -0,0 +1,375 @@
...
+            self.endorsements_by_slot_index.len() > ENDORSEMENT_BY_CACHE_MAX_SIZE {
+            // TODO: Remove this for Testnet 17
+            //       Debug feature for Testnet 16.x only
+            debug!("[De Factory] new endorsements: cache full: {}, {}",
+                self.seen_endorsement_denunciation.len(),
+                self.endorsements_by_slot_index.len());
+            return;
+        }
+
+        let mut denunciations: Vec<Denunciation> = Vec::with_capacity(1);
```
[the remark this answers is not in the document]
sydhds: Because it's easier to collect from an iterator to a Vec<Denunciation> and it's easy to deal with to create WrappedOperation  (with an iterator)...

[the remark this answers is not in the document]
AurelienFT: I think you don't need to collect. And for WrappedOperation creation I think it's easier to create one directly than mapping over a vector with 1 element

massa-execution-worker/src/speculative_roll_state.rs
```diff
@@ -476,4 +507,55 @@ impl SpeculativeRollState {
 
         credits
     }
+
+    /// Remove a given amount from deferred credits and return the removed amount
+    pub fn remove_deferred_credits(&mut self, slot: &Slot, addr: &Address, amount: &Amount) -> Option<Amount> {
+
+        let mut amount_to_rm = *amount;
+        let mut amount_removed = Amount::zero();
```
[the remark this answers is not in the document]
sydhds: Fixed

massa-execution-worker/src/context.rs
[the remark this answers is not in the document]
sydhds: Simplified!

massa-factory-worker/src/block_factory.rs
```diff
@@ -234,6 +271,15 @@ impl BlockFactoryWorker {
             block_id, slot, block_producer_addr
         );
 
+        // Send new operations (stolen denunciations) to pool
+        self.channels.storage.store_operations(new_ops.clone());
+        // And now propage them
+        let mut new_ops_storage = self.channels.storage.clone_without_refs();
+        new_ops_storage.store_operations(new_ops);
+        if let Err(err) = self.channels.protocol.propagate_operations_sync(new_ops_storage) {
```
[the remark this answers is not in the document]
sydhds: I need to propagate them because they are 'stolen' denunciation operations and thus the others nodes do not know about them yet...

massa-factory-worker/src/denunciation_factory.rs
```diff
@@ -0,0 +1,375 @@
...
+                        }
+                    }
+                }
+            })
+    }
+
+
+    fn cleanup_cache(&mut self) {
+
+        let next_slot = self.get_next_slot();
```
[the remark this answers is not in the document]
sydhds: Denunciation operation if included in a block would be deleted later one. I think the best is to clean Denunciation op after a steal in block factory?

massa-factory-worker/src/denunciation_factory.rs
```diff
@@ -0,0 +1,375 @@
...
+                if wrapped_endos.len() == 1 {
+
+                    wrapped_endos.push(wrapped_endorsement);
+
+                    denunciations.extend(
+                        wrapped_endos.iter()
+                            .take(2)
+                            .tuples()
+                            .map(|(we1, we2)| {
+                                Denunciation::from((we1, we2))
```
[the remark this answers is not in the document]
sydhds: We already know that the Denunciations are both at the same slot & index. However, the denunciation factory can still be tricked if we provide the same endorsement twice. So better to fix this...

massa-models/src/denunciation.rs

massa-factory-worker/src/denunciation_factory.rs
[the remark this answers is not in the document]
sydhds: Fixed for endorsement & block denunciation

massa-factory-worker/src/denunciation_factory.rs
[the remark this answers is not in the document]
AurelienFT: You have to clean all references to an object in the storage to delete it. If denunciation factory still have a reference it will not be deleted

massa-execution-worker/src/speculative_roll_state.rs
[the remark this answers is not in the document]
Eitu33: > We still slash in the same order (earliest-to-be-credited first)

But we don't here do we? We slash in changes > active > final. So we're taking away the last that would be reimbursed since we go through the more recent credits and it's always 3 cycles for reimbursement.

> And we can't only look into the final state of deferred credits when creating a candidate slashing, otherwise we might end up attempting to slash multiple times something that was already slashed in the active history but not in the final state yet.

Not sure I understand this, if we keep the current behaviour meaning we go through the credits and stop iterating once we reach the amount that was to be slashed that won't happen no?

> We can't slash the final state from a candidate slot or we will run into consensus problems.

I was speaking of setting the slashing as execution output apply it to final_state if the times comes. But as a matter of fact I just noticed that we currently write on final state in the `speculative_roll_state` which is never supposed happen because the eventual changes can't be undone. final_state in the "speculative" versions is only meant to be read, I think we should consider using a read only lock to avoid this in the future.

massa-models/src/denunciation.rs
sydhds: Did not know about this one, ty! Fixed.

massa-models/src/denunciation_interest.rs
sydhds: I think DenunciationEvent is kinda misleading as here we regroup things that can later make a Denunciation.

massa-pool-worker/src/endorsement_pool.rs
sydhds: Fixed

massa-protocol-exports/src/protocol_controller.rs
sydhds: I'm not against using try_send here but I still don't know what to do if the try_send fails when the queue is full. With blocking_send, we ensure that we will propagate ops...

massa-models/src/denunciation.rs
sydhds: Done

massa-pool-worker/src/endorsement_pool.rs
sydhds: I think it makes more sense that it comes from the pools but your point stands too. What do you think @damip ?

massa-factory-worker/src/denunciation_factory.rs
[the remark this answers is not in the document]
sydhds: Done

massa-factory-worker/src/denunciation_factory.rs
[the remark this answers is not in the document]
sydhds: Done

massa-protocol-exports/src/protocol_controller.rs
AurelienFT: Yes but we wait a lot. I think that propagating denunciations isn't a mandatory thing and so the network can run without it so if we can't send because the channel is full we just don't do anything IMO.

massa-factory-worker/src/block_factory.rs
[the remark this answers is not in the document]
AurelienFT: This behaviour is a bit unclear for me. I trust you and I will check it again when I will make the final review

massa-hash/src/hash.rs
```diff
@@ -317,6 +317,18 @@ impl FromStr for Hash {
...
+///
+pub trait Hasher<T> {
+    /// Serialize a value `T` into a buffer of `u8`.
+    ///
+    /// ## Parameters
+    /// * value: the value to be serialized.
+    ///
+    /// ## Returns
+    /// A Result with the serialized data.
+    fn hash(&self, value: &T, public_key: &[u8], content: &[u8]) -> Result<Hash, SerializeError>;
```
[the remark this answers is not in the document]
sydhds: I used a custom Hasher trait with a default impl that should do the job (no need to define new_wrapped_with_hasher anymore). What do you think of it?

AurelienFT (commented):
Nothing more to add. Just need to clear the lasts comments from my previous review. However, I'm not a big fan of the stolen procedure I think it need more documentation

massa-execution-worker/src/context.rs
```diff
@@ -635,6 +635,54 @@ impl ExecutionContext {
         )
     }
 
+    /// Try to slash 1 roll from the denounced address.
```
damip, line 638:
```suggestion
    /// Try to slash a given number of rolls from the denounced address. If roll slashing fails, try to slash the equivalent amount of deferred credits
```

massa-execution-worker/src/context.rs
```diff
@@ -635,6 +635,54 @@ impl ExecutionContext {
...
+    ///
+    /// # Arguments
+    /// * `denounced_addr`: address to slash the roll or deferred credit from
+    pub fn try_slash_roll(
+        &mut self,
+        denounced_addr: &Address,
+        roll_count: u64,
```
damip, line 645: maybe let's focus on the "roll_count = 1" case

damip: otherwise, "try_burn_rolls" misbehaves:  if it cannot slash ALL the rolls but just a part of them, it will still completely fail and not even try to slash "what it can"

damip: otherwise, "try_burn_rolls" misbehaves:  if it cannot slash ALL the rolls but just a part of them, it will still completely fail and not even try to slash "what it can"

massa-execution-worker/src/context.rs
```diff
@@ -635,6 +635,54 @@ impl ExecutionContext {
...
+        let slashed_coins;
+
+        // First try to slash roll if any
+        if self.speculative_roll_state.try_burn_rolls(
+            denounced_addr,
+            roll_count,
+        ).is_err() {
+
+            // if no roll were slashed, that can likely be done from deferred credit
+            let credits = self.speculative_roll_state.get_address_deferred_credits(denounced_addr, self.slot);
```
damip, line 657: This function returns all deferred credits after a given slot, included.

But deferred credits are credited at the beginning of the slot execution (before the block is executed) by calling context::execute_deferred_credits.
This means that if there is a deferred credit for the current slot, we cannot slash it because it was already credited.

In summary: credits targeting exactly "self.slot" should not be taken into account

massa-execution-worker/src/context.rs
```diff
@@ -635,6 +635,54 @@ impl ExecutionContext {
...
+            let amount_to_slash = self.config.roll_price
+                .checked_mul_u64(roll_count)
+                .ok_or_else(||
+                    ExecutionError::RuntimeError(format!("Cannot mult roll price by {}", roll_count))
+                )?;
+
+            // Remove deferred credits starting at self.slot until
+            let mut amount_to_rm = amount_to_slash;
+            for (slot, _amount) in credits.iter() {
```
damip, line 666: skip the slot == self.slot iteration, as explained above

massa-execution-worker/src/context.rs
```diff
@@ -635,6 +635,54 @@ impl ExecutionContext {
...
+                if amount_to_rm == Amount::zero() {
+                    break;
+                }
+                let amount_slashed_at_slot = self.speculative_roll_state
+                    .remove_deferred_credits(slot, denounced_addr, &amount_to_rm);
```
damip, line 671: pass amount_to_rm by value, not reference

massa-execution-worker/src/speculative_roll_state.rs
```diff
@@ -476,4 +507,56 @@ impl SpeculativeRollState {
 
         credits
     }
+
+    /// Remove a given amount from deferred credits and return the removed amount
```
damip, line 511: Explain that if it cannot slash everytging, it will still try to slash as much as it can

massa-execution-worker/src/context.rs
```diff
@@ -635,6 +635,54 @@ impl ExecutionContext {
...
+                amount_to_rm = amount_to_rm.saturating_sub(amount_slashed_at_slot)
```
damip, line 672: I feel like the whole loop iterating on credits is not useful. I would do it like this:
1- try to slash the rolls, list the number of coins that we failed to slash
2- just call remove_deferred_credits (rename it slash_deferred_credits) on the remainder, if any

massa-execution-worker/src/speculative_roll_state.rs
```diff
@@ -476,4 +507,56 @@ impl SpeculativeRollState {
...
+    pub fn remove_deferred_credits(&mut self, slot: &Slot, addr: &Address, amount: &Amount) -> Amount {
+
+        let amount_zero = Amount::zero();
+        let mut amount_to_rm = *amount;
+
+        // First try to rm in added_changes
+        let res = self.added_changes.deferred_credits.sub_amount(slot, addr, &amount_to_rm);
```
damip, line 518: sub_amount => slash_amount

massa-execution-worker/src/speculative_roll_state.rs
```diff
@@ -476,4 +507,56 @@ impl SpeculativeRollState {
...
+        if let Some(a) = res {
+            amount_to_rm = amount_to_rm.saturating_sub(a);
+        }
+
+        // If not found, try to rm from active history
+        // Early test to avoid locking active history
+        if amount_to_rm > amount_zero {
+            let mut hist = self.active_history.write();
+            let mut hist_iter = hist.0.iter_mut().rev();
+            while amount_to_rm > amount_zero {
```
damip, line 528: This does not behave properly: we might "slash" some history elements that are overwritten later in history or through added_changes

massa-execution-worker/src/execution.rs
```diff
@@ -305,6 +305,9 @@ impl ExecutionState {
             OperationType::Transaction { .. } => {
                 self.execute_transaction_op(&operation.content.op, sender_addr)
             }
+            OperationType::Denunciation { .. } => {
+                self.execute_denunciation_op(&operation.content.op, sender_addr, block_credits)
```
damip, line 309: To be clean, I would pass a mutable reference to new_block_credits, and move the "*block_credits = new_block_credits;" line after the call to execution

massa-execution-worker/src/execution.rs
```diff
@@ -616,6 +620,59 @@ impl ExecutionState {
...
     }
 
+    pub fn execute_denunciation_op(
+        &self,
+        operation: &OperationType,
+        _sender_addr: Address,
+        block_credits: &mut Amount,
+    ) -> Result<(), ExecutionError> {
+
+        const ROLL_COUNT: u64 = 1;
```
damip, line 630: maybe SLASH_ROLL_COUNT would be a more explicit name

massa-execution-worker/src/execution.rs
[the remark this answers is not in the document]
damip: A denunciation can be "stolen" easily (the block producer just sets his own signature on it to benefit from the denunciation), so there is no point in crediting the denunciation producer.

The incentive for emitting a denunciation is : 
* if you're the block producer => you benefit from the rewarded block coins
* if you're the producer of the block being endorsed => you benefit from the rewarded block coins of your descendant
* if you're the producer of an endorsement => you benefit from the rewarded block coins of the block that includes your endorsement
* if you're none of those => you indirectly benefit from helping the network

massa-execution-worker/src/speculative_roll_state.rs
```diff
@@ -146,6 +146,37 @@ impl SpeculativeRollState {
...
         credit.insert(*seller_addr, roll_price.saturating_mul_u64(roll_count));
+        Ok(())
+    }
+
+    /// Try to burn `roll_count` rolls from the seller address.
+    ///
+    /// # Arguments
+    /// * `addr`: address to sell the rolls from
+    /// * `roll_count`: number of rolls to sell
+    pub fn try_burn_rolls(
```
damip, line 157: try_slash_rolls

massa-execution-worker/src/execution.rs
```diff
@@ -616,6 +620,59 @@ impl ExecutionState {
...
+
+        // process denunciation operations only
+        let denunciation = match operation {
+            OperationType::Denunciation { data } => data,
+            _ => panic!("unexpected operation type"),
+        };
+
+        if denunciation.is_valid() {
```
damip, line 638: the denunciation validity should be checked in the context of the block: if it refers to an event that is "too old" it is invalid

massa-execution-worker/src/speculative_roll_state.rs
```diff
@@ -146,6 +146,37 @@ impl SpeculativeRollState {
...
+        &mut self,
+        addr: &Address,
+        roll_count: u64,
+    ) -> Result<(), ExecutionError> {
```
damip, line 161: this function should not fail if it can slash only part of the rolls, but return the number of rolls effectively slashed

massa-factory-worker/src/block_factory.rs
```diff
@@ -192,7 +193,42 @@ impl BlockFactoryWorker {
...
         // gather operations and compute global operations hash
-        let (op_ids, op_storage) = self.channels.pool.get_block_operations(&slot);
+        let (op_ids_, op_storage) = self.channels.pool.get_block_operations(&slot);
+
+        // 'Steal' op denunciation
+        // TODO: add doc about why we stole those
+        let ops = op_storage.read_operations();
+        let mut new_ops: Vec<WrappedOperation> = Vec::new();
+        let mut ops_to_clean: PreHashSet<OperationId> = Default::default();
+        let op_ids: Vec<OperationId> = op_ids_
```
damip, line 203: there is a problem with theft: it changes the operation ID so the list of previously executed op IDs is desync

damip: we can discuss this

massa-execution-worker/src/execution.rs
```diff
@@ -616,6 +620,59 @@ impl ExecutionState {
         Ok(())
     }
 
+    pub fn execute_denunciation_op(
```
damip, line 623: it feels strange to charge the operation fee for the denunciation. We can discuss this

massa-execution-worker/src/context.rs
sydhds: I'll go with multi-roll slashing :)

massa-execution-worker/src/context.rs
sydhds: done

massa-execution-worker/src/context.rs
sydhds: done

massa-execution-worker/src/speculative_roll_state.rs
sydhds: done

massa-execution-worker/src/speculative_roll_state.rs
sydhds: done

massa-execution-worker/src/context.rs
sydhds: done

massa-execution-worker/src/speculative_roll_state.rs
sydhds: done

massa-execution-worker/src/context.rs
sydhds: As we discuss it, remove_deferred_credits slash a given amount at a given slot so we need the loop here ;)

massa-execution-worker/src/speculative_roll_state.rs
sydhds: done

massa-execution-worker/src/execution.rs
sydhds: Renamed and moved to ExecutionConfig!

massa-execution-worker/src/execution.rs
sydhds: done

AurelienFT:
Closing this in favor of #3666

PR #3136, opened 2022-10-09
improve op propag

damip:
In this PR, we make sure that when someone announces us an operation, we mark the sender node as knowing the op.


damip:
@massalabs/core-team FYI

PR #3150, opened 2022-10-11
Feature/ext unit test 1

sydhds:
  * [ ] make tests compile
  * [ ] make tests pass 

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -210,3 +210,14 @@ jobs:
         with:
           branch: gh-pages
           folder: target/doc
+
+  unit_tests:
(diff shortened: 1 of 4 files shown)


sydhds:
wondering if we should remove src files from main repo? @damip

damip:
> wondering if we should remove src files from main repo? @damip

I guess yes since otherwise it would be a duplicate (right ?)

sydhds:
> > wondering if we should remove src files from main repo? @damip
> 
> I guess yes since otherwise it would be a duplicate (right ?)

Yes that would be a duplicate. I'll make a new PR to remove the sources then :)

AurelienFT (approved):

PR #3161, opened 2022-10-18
Feature/node worker full duplex 2

sydhds:
  * [x] make tests compile
  * [x] make tests pass 


massa-network-worker/src/node_worker.rs
```diff
@@ -162,14 +120,34 @@ impl NodeWorker {
                     }
                 };
             }
+
+            if let Err(e) = socket_writer.write_half.shutdown().await {
```
damip, line 124: note: this does not shutdown the reader

massa-network-worker/src/node_worker.rs
```diff
@@ -207,74 +185,19 @@ impl NodeWorker {
...
-                            }
-                            Message::Endorsements(endorsements) => {
-                                massa_trace!("node_worker.run_loop. receive Message::Endorsement", {"node": self.node_id, "endorsements": endorsements});
-                                self.send_node_event(NodeEvent(self.node_id, NodeEventType::ReceivedEndorsements(endorsements))).await;
-                            }
-                            _ => {
-                                // TODO: Write a more user-friendly warning/logout after several consecutive fails? see #1082
-                                massa_trace!("node_worker.run_loop.self.socket_reader.next(). Unexpected message Warning", {});
-                            },
+                res = &mut node_reader_handle => {
```
damip, line 188: a select branch should also be added for node_writer_handle because if the writer exits but not the reader we won't capture it as it is

massa-network-worker/src/node_worker.rs
```diff
@@ -207,74 +185,19 @@ impl NodeWorker {
...
-                        exit_reason = ConnectionClosureReason::Failed;
-                        break;
-                    },
-                },
-
+                    };
+                    break;
+                }
                 // node command
                 cmd = self.node_command_rx.recv() => {
```
damip, line 202: To reduce contention, I recommend not managing this here but directly providing writer_command_tx upstream so that the network main loop communicates directly with the node's writer task

massa-network-worker/src/node_worker.rs
```diff
@@ -410,6 +333,168 @@ impl NodeWorker {
...
+        // 3- Stop node_reader_handle
+        if !reader_joined {
+            // Abort the task otherwise socket_reader.next() is stuck, waiting for some data to read
+            debug!(
+                "node_worker.run_loop.cleanup.node_reader_handle.abort, node_id: {}",
+                self.node_id
+            );
+            node_reader_handle.abort();
+        }
+
```
damip, line 345: also stop the writer task

massa-network-worker/src/node_worker.rs
```diff
@@ -410,6 +333,168 @@ impl NodeWorker {
...
         Ok(exit_reason)
     }
 }
+
+async fn node_sender_handle(
```
damip, line 350: node_reader_handler

TODO: document this function

massa-network-worker/src/node_worker.rs
```diff
@@ -410,6 +333,168 @@ impl NodeWorker {
...
+                exit_reason = ConnectionClosureReason::Failed;
+                break;
+            }
+        }
+    }
+
+    exit_reason
+}
+
+async fn send_node_event(
```
damip, line 456: document function

massa-network-worker/src/node_worker.rs
```diff
@@ -410,6 +333,168 @@ impl NodeWorker {
...
+    node_event_tx: &mut Sender<NodeEvent>,
+    event: NodeEvent,
+    _max_send_wait: MassaTime,
+) {
+    let result = node_event_tx
+        // .send_timeout(event, max_send_wait.to_duration())
+        .send_timeout(event, tokio::time::Duration::from_millis(10000))
```
damip, line 463: this timeout should not be hardcoded

massa-network-worker/src/node_worker.rs
```diff
@@ -410,6 +333,168 @@ impl NodeWorker {
...
+                event
+            );
+        }
+        Err(SendTimeoutError::Timeout(event)) => {
+            warn!("Failed to send NodeEvent due to timeout: {:?}.", event);
+        }
+    }
+}
+
+pub fn try_send_to_node(
```
damip, line 479: document this function

massa-network-worker/src/node_worker.rs
sydhds: This line has not been added with the intention to close the reader but to close the writer in a proper way... 

@damip Would you llike me to remove it?

damip: oh ok, then:
* if graceful shutdown takes program time, we should remove it and let the kernel handle things in the background 
* if graceful shutdown has no performance impact, we can leave it

massa-network-worker/src/node_worker.rs
sydhds: At this point, the node_writer_handle is already stopped but to (at line 320):

`match node_writer_handle.await {
`
 Trying to abort the node_writer_handle result in a compilation error:

>    --> massa-network-worker/src/node_worker.rs:344:13
>     |
> 131 |         tokio::pin!(node_writer_handle);
>     |         ------------------------------- move occurs because `node_writer_handle` has type `Pin<&mut tokio::task::JoinHandle<Result<(), NetworkError>>>`, which does not implement the `Copy` trait
> ...
> 320 |             match node_writer_handle.await {
>     |                                     ------ `node_writer_handle` moved due to this method call
> ...
> 344 |             node_writer_handle.abort();
>     |             ^^^^^^^^^^^^^^^^^^^^^^^^^^ value borrowed here after move
>     |
>

massa-network-worker/src/node_worker.rs
sydhds: done

massa-network-worker/src/node_worker.rs
sydhds: done

massa-network-worker/src/node_worker.rs
sydhds: done

massa-network-worker/src/node_worker.rs
```diff
@@ -162,14 +120,34 @@ impl NodeWorker {
...
+                massa_trace!("node_worker.run_loop.loop.writer_shutdown error: {}", err);
+            }
+
             Ok(())
         });
         tokio::pin!(node_writer_handle);
         let mut writer_joined = false;
 
+        let node_reader_handle = tokio::spawn(async move {
+            node_sender_handle(
```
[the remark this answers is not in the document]
sydhds: Renamed

massa-network-worker/src/node_worker.rs
damip: my bad

massa-network-worker/src/node_worker.rs
damip: what's different compared to node_reader_handle ? I think we should handle their closing the same way

massa-network-worker/src/node_worker.rs
damip: I think we can just drop abruptly :)

massa-network-worker/src/node_worker.rs
```diff
@@ -410,6 +333,170 @@ impl NodeWorker {
...
+}
+
+/// Send a node event via node_event_tx queue - used by node_reader_handle
+async fn send_node_event(
+    node_event_tx: &mut Sender<NodeEvent>,
+    event: NodeEvent,
+    max_send_wait: MassaTime,
+) {
+    let result = node_event_tx
+        .send_timeout(event, max_send_wait.to_duration())
```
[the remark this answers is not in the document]
sydhds: From what I see, it's set to 0 in base_config/config.toml. Are you sure you'd like a 5000 ms (5s) wait for each node event?

massa-network-worker/src/node_worker.rs
sydhds: node_writer_handle is mostly waiting writer_command_rx and it will exit when writer_command_tx is dropped (which is done at line 311).

On the other end, node_reader_handle is mostly waiting for socket_reader; the only 'easy' to close this task is to abort it (done at line 338). If you prefer we can investigate, a more elegant solution (like using tokio Notify + select in node_reader_handle)?

Note that the code is always closing first the writer (by first checking writer_joined at line 314) then is always closing the reader (by first checking reader_joined).

massa-network-worker/src/node_worker.rs
sydhds: Done

massa-node/base_config/config.toml
```diff
@@ -133,8 +133,10 @@
     max_operations_per_message = 1024
     # max number of endorsements per message
     max_endorsements_per_message = 1024
-    # max milliseconds to wait while sending an event before dropping it
-    max_send_wait = 0
+    # max milliseconds to wait while sending a node event before dropping it
+    max_send_wait_node_event = 5_000
```
damip, line 137: does toml support "_" ?

damip, line 137: I think toml doesn't suppoer "_" separators

```suggestion
    max_send_wait_node_event = 5000
```

massa-network-worker/src/node_worker.rs
```diff
@@ -65,111 +62,52 @@ impl NodeWorker {
...
-                    None => {
-                        massa_trace!("node_worker.run_loop.loop.writer_command_rx.recv. None", {});
-                        break;
-                    }
-                };
-            }
-            Ok(())
+            node_writer_handle(
+                &mut socket_writer,
+                &mut self.node_command_rx,
```
damip, line 82: Is it legal to spawn a task with a reference to self.node_command_rx ?!
To make things clean, I would use an `Option<Receiver>` for self.node_command_rx and take() the receiver just before the spawn. Then it would be moved to the task which would be its sole owner

sydhds: It compiles and seems to work fine but that would be interesting to have a deeper look and understand this...

damip (approved):

damip (changes requested):
this is pushing docs

Issue #3124, opened 2022-10-06
Synchronize OpenRPC specification


aoudiamoncef:
The current state of OpenRPC specification does't match latest input/output. We have to check every method IO. 

- [ ] get_block
- [ ] get_blockclique_block_by_slot
- [ ] get_cliques
- [ ] get_datastore_entries
- [ ] get_endorsements
- [ ] get_filtered_sc_output_event
- [ ] get_graph_interval
- [ ] get_operations
- [ ] get_stakers
- [ ] get_status
- [ ] node_add_staking_private_keys
- [ ] node_ban_by_id
- [ ] node_ban_by_ip
- [ ] node_get_staking_addresses
- [ ] node_remove_from_whitelist
- [ ] node_remove_staking_addresses
- [ ] node_sign_message
- [ ] node_stop
- [ ] node_unban_by_id
- [ ] node_unban_by_ip
- [ ] node_whitelist
- [ ] send_operations

PR #3180, opened 2022-10-21
Return gas cost when calling execute_read_only_bytecode

sydhds:
  * [x] make tests compile
  * [x] make tests pass 


sydhds:
API spec not updated, as @aoudiamoncef will take care of it here:

https://github.com/massalabs/massa/issues/3124

massa-execution-worker/src/execution.rs
```diff
@@ -1054,19 +1056,23 @@ impl ExecutionState {
...
+
+                remaining_gas
             }
-        }
+        };
 
         // return the execution output
-        Ok(context_guard!(self).settle_slot())
+        let mut execution_output = context_guard!(self).settle_slot();
+        execution_output.gas_cost = Some(req.max_gas - remaining_gas);
```
damip, line 1074:
```suggestion
        execution_output.gas_cost = Some(req.max_gas.saturating_sub(remaining_gas));
```

massa-execution-worker/src/execution.rs
```diff
@@ -1054,19 +1056,23 @@ impl ExecutionState {
...
+        Ok(execution_output)
```
damip, line 1075: Here is an improvement:

```
{
        context.register_used_gas(req.max_gas.saturating_sub(remaining_gas));
        Ok(context.settle_slot())
}
```

Where `register_used_gas` increments the context's `slot_used_gas` with the provided amount of gas. Then `settle_slot` sets the `execution_output.used_gas`.

This should also be used for the normal, non-readonly execute_slot.

Why ? because:
* it avoids carrying around a sense-less used_gas property in the execution outputs that is nonzero only for read-only calls, and zero for normal slot execution, even if the execution costs gas
* in the future it will allow us to know how much gas is left at the end of a block execution, which can be filled-up with extra autonomous SCs

massa-execution-worker/src/context.rs
```diff
@@ -713,6 +713,7 @@ impl ExecutionContext {
             block_id: std::mem::take(&mut self.opt_block_id),
             state_changes,
             events: std::mem::take(&mut self.events),
+            gas_cost: None
```
damip, line 716: it is a bit surprising that this is used only in the case of readonly executions, and kept to none when it's a normal execution

massa-execution-worker/src/execution.rs
damip: another option is to not add gas_cost to ExecutionOutput at all, but to allow execute_readonly_request to return a pair (execution_output, used_gas)

massa-execution-exports/src/types.rs
```diff
@@ -47,6 +47,8 @@ pub struct ExecutionOutput {
     pub state_changes: StateChanges,
     /// events emitted by the execution step
     pub events: EventStore,
+    /// optional gas cost (Only when: read only execute bytecode)
+    pub gas_cost: Option<u64>
```
damip, line 51: remove gas_cost from the ExecutionOutput structure altogether

AurelienFT (approved):

PR #3183, opened 2022-10-25
fix get_stakers sort result

aoudiamoncef:
  * [x] make tests compile
  * [x] make tests pass 


AurelienFT:
Did you tested it on a local node @aoudiamoncef ?

aoudiamoncef:
Short answer Yes.

this code was tested well the last time before been merged

I tested it in Rust playground

I tested it in sandbox

I wasn't able to bootstrap to testnet...

AurelienFT (approved):

aoudiamoncef:
I don't understand the sanity issue. All is green in my workstation. 

Bonus: I tested in testnet and it works as expected

PR #3185, opened 2022-10-26
Add bootstrap whitelist and fix error messages.

### LINKED ISSUE #2721: Bootstrap IP whitelist
Add an optional whitelist and blacklist of IP addresses in the config file, filtering who is allowed to bootstrap from our node.

When the whitelist is absent, all IPs that are not in the blacklist are allowed.
When the whitelist is present, only Ips that are in the whitelist but not the blacklist are allowed.
An absent blacklist is equivalent to an empty blacklist.

By default, a whitelist containing the routability test bot IPs (ipv4 and ipv6) should be present, and an empty blacklist. On the testnet, all massa labs servers should allow all IPs to bootstrap (no whitelist).

The whitelist/blacklist filtering should happen as the first thing on a TCPListener. Whitelisting should not bypass per-ip and simultaneous connections limitations.

When obtaining an IP address either from the white/blacklist or from an incoming connection attempt on a TCPListener, always normalize it before use:
```rust
fn normalize_ip(ip: IpAddr) -> IpAddr {
   match ip {
       IpAddr::V4(ip) => ip.to_ipv6_mapped(),
       IpAddr::V6(ip) => ip,
   }.to_canonical()
}
```
Why not just to_canonical ? Because the case in which the incoming ip is ipv4 but was mapped to ipv6 by the os, it would fail the comparison with a canonicalized ipv4 from the config (eg. Ipv4 is not converted to ipv6 by canonicalize)

AurelienFT:
  * [x] make tests compile
  * [x] make tests pass 

Fix #2721 and #3122


Eitu33 (approved):
Only good boys will bootstrap now 👼

Issue #3184, opened 2022-10-26
Create rpc.discover method


aoudiamoncef:
**Is your feature request related to a problem? Please describe.**
OpenRPC Playground sends request to introspect the OpenRPC spec from the server by calling `rpc.discover` method which isn't implemented yet. 

**Describe the solution you'd like**
We'll implement the method which returns `openrpc.json` file content.

We could use [JsonSchema](https://docs.rs/jsonschema/latest/jsonschema/) to parse and validate the file content. But it's a nice to have.

PR #3186, opened 2022-10-27
Add rpc.discover method to Public API

aoudiamoncef:
  * [ ] make tests compile
  * [ ] make tests pass 


aoudiamoncef:
Related to https://github.com/massalabs/massa/issues/3184

aoudiamoncef:
@AurelienFT is it useful to add this method to `massa-client` CLI ?

AurelienFT:
> @AurelienFT is it useful to add this method to `massa-client` CLI ?

No.

massa-api/src/lib.rs
```diff
@@ -5,7 +5,7 @@
 #![warn(unused_crate_dependencies)]
 use crate::error::ApiError::WrongAPI;
 use error::ApiError;
-use jsonrpc_core::{BoxFuture, IoHandler, Value};
+use jsonrpc_core::{BoxFuture, IoHandler, serde_json, Value};
```
AurelienFT, line 8: I don't think usage related to this import

AurelienFT (commented):
I think it also need a bit of formatting. Did you try the playground with your endpoint ? I think you can't because you don't have a domain name. Let's wait the end of the tests of consensus in labnet (ETA 2-3hours) and launch this in labnet to test it.

aoudiamoncef:
You're right, I tested only via Postman. Sorry for formatting issues.

AurelienFT:
Tested in labnet works well !

PR #3187, opened 2022-10-27
Add function to regroup param in double quotes.

### LINKED ISSUE #2570: Parameters with quotes in CLI don't work
When there is quotes in the command line of the CLI it don't parse the arguments correctly. The following isn't parsed correctly :
``` 
call_smart_contract <your address> 2R4zRvGc5GcX4eCWrM5zsboFKodCUuWa7X8biiDBQMoLohwH4N setResolver '{"name": "<name_of_your_website>", "address": "<your_address>"}' 1000000000 0 0 0
```

I except that the single quotes will be matched and consider as one argument but it's not the case. The current workaround is to not use the client : 
```
./massa-client call_smart_contract <your address> 2R4zRvGc5GcX4eCWrM5zsboFKodCUuWa7X8biiDBQMoLohwH4N setResolver '{"name": "<name_of_your_website>", "address": "<your_address>"}' 1000000000 0 0 0
```

AurelienFT:
  * [x] make tests compile
  * [x] make tests pass 

Fix #2570 also avoid fetching the same command twice with two up arrows hit.

### DIFF
--- a/massa-client/src/repl.rs
+++ b/massa-client/src/repl.rs
@@ -173,8 +173,15 @@ impl<T: ToString> History<T> for CommandHistory {
             self.history.pop_back();
         }
         let string_value = val.to_string();
-        CommandHistory::write_to_saved_history(&string_value);
-        self.history.push_front(string_value);
+        if let Some(last_command) = self.history.iter().next() {
+            if last_command != &string_value {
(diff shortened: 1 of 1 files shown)


Eitu33 (approved):
You could've probably use nom for this hahaha but I'll agree to it 💀 💀

PR #3188, opened 2022-10-27
Remove SC unit test sources (already moved to massa-unit-tests-src repo)

sydhds:
  * [ ] make tests compile
  * [ ] make tests pass 


sydhds:
Follow up of: 

https://github.com/massalabs/massa/pull/3150

AurelienFT (approved):

PR #3189, opened 2022-10-28
Do not use format! for Operation && OperationType comparison

sydhds:
  * [x] make tests compile
  * [x] make tests pass 


Eitu33 (approved):
Absolutely breathtaking 😮

Eitu33:
Jokes aside that's actually nice

AurelienFT (approved):

PR #3191, opened 2022-10-28
Limit operation max size in protocol

### LINKED ISSUE #3090: Max operation size in protocol
If an operation is strictly bigger than MAX_BLOCK_SIZE, it is considered invalid => ignore and ban sender

aoudiamoncef:
close #3090

  * [x] make tests compile
  * [x] make tests pass 

### DIFF
--- a/massa-protocol-worker/src/protocol_network.rs
+++ b/massa-protocol-worker/src/protocol_network.rs
@@ -114,19 +114,8 @@ impl ProtocolWorker {
             }
             NetworkEvent::ReceivedOperations { node, operations } => {
                 massa_trace!(OPS, { "node": node, "operations": operations});
-                if operations.iter().any(|operation| {
-                    operation.serialized_size()
-                        > self.config.max_serialized_operations_size_per_block
-                }) {
-                    warn!(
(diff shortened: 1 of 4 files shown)


massa-protocol-worker/src/protocol_network.rs
```diff
@@ -117,8 +117,19 @@ impl ProtocolWorker {
             }
             NetworkEvent::ReceivedOperations { node, operations } => {
                 massa_trace!(OPS, { "node": node, "operations": operations});
-                self.on_operations_received(node, operations, op_timer)
-                    .await;
+                if operations.iter().any(|operation| {
```
AurelienFT, line 120: I think this should be better to be in `note_operations_from_node` here : https://github.com/massalabs/massa/blob/main/massa-protocol-worker/src/protocol_worker.rs#L932

Because is the function that is triggered each time we receive operation even if it's not in the propagation process. This will be more flexible and will be checked all time.

You don't ban in `note_operations_from_node` but you return an error and `on_operations_received` will ban

AurelienFT (approved):
Good job ! You can squash and merge it :)

PR #3192, opened 2022-10-28
Consensus bootstrap streaming

Eitu33:
* [x] try in simulation & labnet
* [x] unit tests on the added & changed features
  * [x] make tests compile
  * [x] make tests pass 


massa-consensus-worker/src/controller.rs
```diff
@@ -87,41 +91,62 @@ impl ConsensusController for ConsensusControllerImpl {
...
+
+        for b_id in &difference {
+            if let Some(BlockStatus::Active { a_block, storage }) =
                 read_shared_state.block_statuses.get(b_id)
             {
+                if final_blocks.len() as u64 >= self.bootstrap_part_size {
+                    break;
+                }
+                if let StreamingStep::Finished(Some(slot)) = execution_cursor {
+                    if a_block.slot > slot {
```
AurelienFT, line 128: Why this check is only if execution cursor is finished ?

massa-consensus-worker/src/controller.rs
```diff
@@ -87,41 +91,62 @@ impl ConsensusController for ConsensusControllerImpl {
...
+                        continue;
+                    }
+                }
                 if a_block.is_final {
-                    // filter only final actives
-                    return true;
+                    let mut export = ExportActiveBlock::from_active_block(a_block, storage);
```
AurelienFT, line 133: The heaviest computation on this function is to fetch the ops and you set them to `Vec::new()` I think you should make the object yourself or a version of the function that do not search the ops

massa-consensus-worker/src/state/prune.rs
```diff
@@ -12,7 +17,8 @@ impl ConsensusState {
     /// prune active blocks and return final blocks, return discarded final blocks
     fn prune_active(&mut self) -> Result<PreHashMap<BlockId, ActiveBlock>, ConsensusError> {
         // list required active blocks
-        let mut retain_active = self.list_required_active_blocks()?;
+        let mut retain_active: PreHashSet<BlockId> =
+            self.list_required_active_blocks()?.into_iter().collect();
```
AurelienFT, line 21: I think this `into_iter().collect()` is useless

AurelienFT (commented):
Very good PR. Really clap Thomas

massa-bootstrap/src/client.rs
```diff
@@ -100,39 +102,52 @@ async fn stream_final_state(
...
+                    *next_bootstrap_message = BootstrapClientMessage::AskBootstrapPart {
                         last_slot: Some(slot),
                         last_ledger_step,
                         last_pool_step,
                         last_cycle_step,
                         last_credits_step,
                         last_ops_step,
+                        last_consensus_step,
                     };
-                    debug!(
```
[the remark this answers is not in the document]
Eitu33: I removed them because they do not bring any useful information in case of a bootstrap behaviour problem, the logs we need are those are in [`server.rs`](https://github.com/massalabs/massa/blob/consensus-blocks-streaming/massa-bootstrap/src/server.rs#L418)

massa-consensus-worker/src/controller.rs
Eitu33: Because we want no blocks after the execution cursor once the final state bootstrap is done

Eitu33: Ah do you mean that we could potentially get blocks that are in the future of the upcoming last execution slot here while execution streaming is ongoing? If that is what you meant you have a point that is not what we want

massa-consensus-worker/src/controller.rs
Eitu33: Yeah I agree will do so on Monday

massa-consensus-worker/src/state/prune.rs
Eitu33: Indeed, leftovers of a previous version

Eitu33: Done

massa-consensus-worker/src/controller.rs
Eitu33: Done

AurelienFT (approved):

Eitu33:
Last possible issues have been investigated and validated by @damip :

- [x]  Block operations request after a recent consensus bootstrap (Not an issue)
- [x] API behaviour on blocks without operations (Not an issue)
- [x] Pruning of the consensus bootstrap cursor (Done)
- [x] Init of the execution controller without complete operations after consensus bootstrap (Not an issue)

Eitu33:
Merging

PR #3205, opened 2022-11-04
update lib config 0.11 to 0.13

modship:
  * [X] make tests compile
  * [X] make tests pass 


AurelienFT (approved):
Thanks for your contribution. Very good PR. ❤️

AurelienFT:
Tested in sandbox and bootstrap with a second node.

PR #3210, opened 2022-11-09
Final state hash

Eitu33:
### Context

This PR covers:

1. Async pool hash
2. PoS deferred credits hash
3. PoS cycle history hash
4. Final state global hash

The rest of final state hash computations have already been implemented

### TODOS

* [ ] unit tests on the added / changed features
  * [ ] make tests compile
  * [ ] make tests pass 


massa-async-pool/src/pool.rs
```diff
@@ -52,13 +59,16 @@ impl AsyncPool {
...
                 // delete a message from the pool
-                Change::Delete(msg_id) => {
-                    self.messages.remove(msg_id);
+                Change::Delete(message_id) => {
+                    if let Some(removed_message) = self.messages.remove(message_id) {
+                        self.hash ^= removed_message.hash;
+                    }
                 }
             }
         }
```
AurelienFT, line 74: In the function `settle_slot` just after there is `self.messages` that is extended and popped and there is no modifications of the hash.

massa-async-pool/src/pool.rs
```diff
@@ -123,13 +133,13 @@ impl AsyncPool {
         // gather all selected items and remove them from self.messages
         // iterate in decreasing priority order
         self.messages
-            .drain_filter(|_, msg| {
+            .drain_filter(|_, message| {
```
AurelienFT, line 136: Drain filter remove elements from `self.messages` maybe we should XOR their hash also ?

massa-final-state/src/final_state.rs
```diff
@@ -117,15 +118,37 @@ impl FinalState {
...
+        // 5. executed operations hash
+        hash_concat.extend(self.executed_ops.hash.to_bytes());
         debug!(
-            "executed_ops hash at slot {}: {:?}",
+            "executed_ops hash at slot {}: {}",
             slot, self.executed_ops.hash
         );
+        // 6. final state hash
+        let final_state_hash = Hash::compute_from(&hash_concat);
+        debug!("final_state hash at slot {}: {}", slot, final_state_hash);
```
AurelienFT, line 151: Can you make this one as "info" ? It will help us a lot to debug the peoples and our nodes that often run in info.

massa-pos-exports/src/deferred_credits.rs
```diff
@@ -70,7 +167,7 @@ impl DeferredCredits {
 
     /// Insert/overwrite a deferred credit
     pub fn insert(&mut self, addr: Address, slot: Slot, amount: Amount) {
-        let entry = self.0.entry(slot).or_default();
+        let entry = self.credits.entry(slot).or_default();
         entry.insert(addr, amount);
```
AurelienFT, line 171: This insertion shouldn't have an extend ?

massa-async-pool/src/pool.rs
Eitu33: take_batch_to_execute is used only speculatively, hash computations are done on finality

massa-pos-exports/src/deferred_credits.rs
Eitu33: Extend? You mean hash update? If yes this is also only used speculatively, no need to do so

massa-pos-exports/src/cycle_info.rs
```diff
@@ -33,6 +99,100 @@ pub struct CycleInfo {
...
+        roll_counts: BTreeMap<Address, u64>,
+        rng_seed: BitVec<u8>,
+        production_stats: PreHashMap<Address, ProductionStats>,
+    ) -> Self {
+        let hash_computer = CycleInfoHashComputer::new();
+        let mut hash = Hash::from_bytes(CYCLE_INFO_HASH_INITIAL_BYTES);
+
+        // compute the cycle hash
+        hash ^= hash_computer.compute_cycle_hash(cycle);
+        hash ^= hash_computer.compute_complete_hash(complete);
```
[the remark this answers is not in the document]
Eitu33: Because you can't just compute one hash from the serialized cycle on new, it would be different than a cycle that was partially updated multiple times on a running node. You are right about the collide though, added an identifier to avoid that.

massa-async-pool/src/pool.rs
```diff
@@ -52,13 +59,16 @@ impl AsyncPool {
         for change in changes.0.iter() {
             match change {
                 // add a new message to the pool
-                Change::Add(msg_id, msg) => {
-                    self.messages.insert(*msg_id, msg.clone());
+                Change::Add(message_id, message) => {
+                    self.messages.insert(*message_id, message.clone());
```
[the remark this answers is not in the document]
Eitu33: Done

massa-final-state/src/final_state.rs
Eitu33: Done

massa-async-pool/src/pool.rs
Eitu33: settle_slot is used only speculatively, hash computations are done on finality

massa-pos-exports/src/cycle_info.rs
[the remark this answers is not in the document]
AurelienFT: Shouldn't you add identifier to all of them ? Collision can't happens with seed or rolls ?

AurelienFT:
@Eitu33 As a general it would be nice to have a comment on the function is they are speculative or final so that we know it on the future and so need to change the hash or not

Eitu33:
Only apply_changes functions are usually final-only so I don't believe it is that hard to track, but I agree that it could be more explicit

massa-pos-exports/src/cycle_info.rs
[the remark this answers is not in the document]
Eitu33: There is a very slight chance that bitvec might collide with something indeed, will add an ident for it too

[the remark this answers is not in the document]
Eitu33: Collision on this is low enough for us to not care about it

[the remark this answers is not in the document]
Eitu33: I did some other updates on the CycleInfo hash computation, would be great if you can take a look again at `cycle_info.rs`

AurelienFT (approved):

PR #3213, opened 2022-11-09
Adds binaries for Mac M1

### LINKED ISSUE #3212: Add ARM64 binaries for MacOS
For now, MacOS binaries provided in [releases](https://github.com/massalabs/massa/releases) are only available for `x86_64` processors. Most new Macs come with an `arm64` processor.

Add `arm64` binaries for MacOS in [releases](https://github.com/massalabs/massa/releases).

- Let `arm64` Mac users compile binaries themselves. But that's not really user friendly and requires way more time and disk space.
- ...

**Additional context**
- Seeing how the [CD workflow](https://github.com/massalabs/massa/blob/99b5d6dab5acc31eb30da11b3ee6ff4da9cee6cc/.github/workflows/cd.yml) is made, it should be as easy as removing [MacOS `arm64`](https://github.com/massalabs/massa/blob/99b5d6dab5acc31eb30da11b3ee6ff4da9cee6cc/.github/workflows/cd.yml#L38) from the matrix [`exclude`](https://github.com/massalabs/massa/blob/99b5d6dab5acc31eb30da11b3ee6ff4da9cee6cc/.github/workflows/cd.yml#L35) and add it to the [`include`](https://github.com/massalabs/massa/blob/99b5d6dab5acc31eb30da11b3ee6ff4da9cee6cc/.github/workflows/cd.yml#L18).

qdrn:
Closes #3212

### DIFF
--- a/.github/workflows/cd.yml
+++ b/.github/workflows/cd.yml
@@ -33,13 +33,11 @@ jobs:
             target: x86_64-apple-darwin
             name: release_macos.tar.gz
           - os: macOS-latest
-            platform: x86_64
+            platform: arm64
             target: aarch64-apple-darwin
             name: release_macos_aarch64.tar.gz
         exclude:
           - os: windows-latest
-            platform: arm64
-          - os: macOS-latest
             platform: arm64
     runs-on: ${{ matrix.os }}
     steps:


qdrn:
Can you test the binaries generated for Mac M1 here @thomas-senechal https://github.com/qdrn/massa/releases/tag/TEST.13.0 and they work accept and merge this PR ?

qdrn:
Can also work with @Eitu33 :)

peterjah (approved):
FYI, Cross does not support aarch64-apple-darwin ( at least not like this) But The build should work because it will use cargo as fallback!

AurelienFT (approved):
I have tested with my M1

AurelienFT:
bors merge

qdrn:
> FYI, Cross does not support aarch64-apple-darwin ( at least not like this) But the build should work because it will use cargo as fallback!

Thanks for the info. Maybe we can find something that is cleaner in the future.

PR #3219, opened 2022-11-16
migrate API to jsonrpsee

damip:
  * [x] make tests compile
  * [x] make tests pass 

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1839,7 +1839,6 @@ dependencies = [
  "async-speed-limit",
  "bitvec",
  "displaydoc",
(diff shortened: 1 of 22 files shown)


AurelienFT (changes requested):
Tried to launch massa-client and had this error :
 
```
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: missing field `enable_http`', /Users/aurelienfoucault/massa/massa-models/src/config/massa_settings.rs:63:25
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

AurelienFT (approved):

Issue #3085, opened 2022-09-27
Inconsistency in size of sub SC


AurelienFT:
In this test : https://github.com/massalabs/massa/blob/e73a532f8ecf0c3333b42e3e353c13492827a9f2/massa-execution-worker/src/tests/scenarios_mandatories.rs#L121 we use the size of a sub smart contract to calculate the storage costs.

The size we hard-coded is the size we get in `set_bytecode` of `speculative_ledger.rs`.
I tried to get the size of the sub smart contract built (his .wasm) and the size of his base64 none of them correspond to the size hard-coded.

damip:
is this still relevant ?

AurelienFT:
Yes

PR #3221, opened 2022-11-19
Byte array update

Eitu33:
  * [x] make tests compile
  * [x] make tests pass


tools/setup_test.rs
```diff
@@ -63,35 +63,35 @@ fn dl_file(url: &str, to_file: &str, max_size: u64) -> Result<(), DlFileError> {
 
 fn main() -> Result<(), std::io::Error> {
 
-    println!("Using tag: {} for release of massa-unit-tests-src repo...", TAG);
```
sydhds, line 66: Those println! are in my opinion quite important to avoid mistakes so maybe we can keep them?

tools/setup_test.rs
```diff
@@ -63,35 +63,35 @@ fn dl_file(url: &str, to_file: &str, max_size: u64) -> Result<(), DlFileError> {
...
 
     if Path::new(&extract_folder).exists() {
         println!("Please remove the folder: {} before runnning this script", extract_folder);
         std::process::exit(1);
     }
 
-    println!("Downloading: {}...", url);
-    let res = dl_file(&url, &path, ARCHIVE_MAX_SIZE);
+    // println!("Downloading: {}...", url);
+    // let res = dl_file(&url, &path, ARCHIVE_MAX_SIZE);
```
sydhds, line 79: The script should by default download from github. the 'local' option is only for developers for quick iteration so better to add an command line option to active it

massa-execution-worker/Cargo.toml
```diff
@@ -21,7 +21,7 @@ massa_execution_exports = { path = "../massa-execution-exports" }
 massa_models = { path = "../massa-models" }
 massa_storage = { path = "../massa-storage" }
 massa_hash = { path = "../massa-hash" }
-massa-sc-runtime = { git = "https://github.com/massalabs/massa-sc-runtime", tag = "v0.8.2" }
+massa-sc-runtime = { path = "../../massa-sc-runtime" }
```
sydhds, line 24: Just a reminder here to switch to a released version of massa-sc-runtime

massa-execution-worker/src/tests/scenarios_mandatories.rs
```diff
@@ -160,7 +163,7 @@ fn test_nested_call_gas_usage() {
     std::thread::sleep(Duration::from_millis(10));
 
     // length of the sub contract test.wasm
-    let bytecode_sub_contract_len = 3715;
+    let bytecode_sub_contract_len = 4374;
```
sydhds, line 166: @AurelienFT If I remember you created an issue for this hardcoded value. Maybe we can add this issue here as a comment?

massa-execution-worker/src/tests/scenarios_mandatories.rs
```diff
@@ -789,7 +807,7 @@ fn datastore_manipulations() {
     // you can check the source code of the following wasm file in massa-sc-examples
```
sydhds, line 807: outdated

massa-execution-worker/src/tests/scenarios_mandatories.rs
```diff
@@ -807,7 +825,7 @@ fn datastore_manipulations() {
     );
 
     // Length of the value left in the datastore. See sources for more context.
-    let value_len = 10;
+    let value_len = 4;
```
sydhds, line 828: Cannot the SC generate an event, retrieve it in order to avoid this hardcoding?

massa-execution-worker/src/tests/scenarios_mandatories.rs
```diff
@@ -922,13 +940,18 @@ fn events_from_switching_blockclique() {
 /// Return a result that should be unwrapped in the root `#[test]` routine.
 fn create_execute_sc_operation(
```
sydhds, line 941: It seems the function create_execute_sc_operation_with_datastore was already avail. so maybe we can have only one function here?

sydhds (changes requested):

massa-execution-worker/src/tests/scenarios_mandatories.rs
AurelienFT: https://github.com/massalabs/massa/issues/3085

AurelienFT (commented):
Nothing more to say than the review above.

tools/setup_test.rs
Eitu33: Yes my bad this was not supposed to be in the final PR, just some tmp changes that I forgot to revert

tools/setup_test.rs
Eitu33: Same as above

massa-execution-worker/Cargo.toml
Eitu33: Yep, just waiting for every corresponding PR to be ready

massa-execution-worker/src/tests/scenarios_mandatories.rs
Eitu33: That's a good idea to avoid hardcoding in this case, but I believe it is not the place to make that change

massa-execution-worker/src/tests/scenarios_mandatories.rs
Eitu33: Done

massa-execution-worker/src/tests/scenarios_mandatories.rs
sydhds: Would you mind adding an issue in github and add the link in a comment then?

Eitu33: https://github.com/massalabs/massa/issues/3236

sydhds (approved):

PR #3229, opened 2022-11-21
Fix deferred credits

Eitu33:
  * [x] make tests compile
  * [x] make tests pass

### DIFF
--- a/massa-execution-worker/src/context.rs
+++ b/massa-execution-worker/src/context.rs
@@ -658,12 +658,21 @@ impl ExecutionContext {
     /// * `slot`: associated slot of the deferred credits to be executed
     /// * `credits`: deferred to be executed
     pub fn execute_deferred_credits(&mut self, slot: &Slot) {
-        let credits = self.speculative_roll_state.get_deferred_credits(slot);
-        for (addr, amount) in credits {
-            if let Err(e) = self.transfer_coins(None, Some(addr), amount, false) {
+        let executed_credits = self.speculative_roll_state.get_deferred_credits(slot);
+        for (address, amount) in executed_credits {
+            self.speculative_roll_state
+                .added_changes
+                .deferred_credits
+                .0
+                .entry(*slot)
+                .or_default()
+                .entry(address)
+                .and_modify(|credit_amount| *credit_amount = Amount::from_raw(0))
+                .or_insert(amount);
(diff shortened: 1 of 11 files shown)


AurelienFT:
@Eitu33 Can you add some description on the changes and why ?

Eitu33:
@AurelienFT @sydhds What this PR aims to fix:

1. Deferred credits changes were build to be absolute changes, meaning that the apply of speculative changes to the final deferred credits must override the values for any given existing entry. This is not what was done, the previous behaviour was accumulating the credits what would result in credits slashes never being applying because the change value was set to 0 and that 0 was later summed to the final value instead of replacing it.
2. The execution of deferred credits only credited the coins equivalent to the given address without setting the credit values to 0, what is even worst is that the only pruning operation we perform on credits removes only the credits that reached 0. The ram was increasing indefinitely because of that.
3. The previous deferred credits bootstrap didn't take into consideration that credits preceding the execution slot could change, with the denunciations coming they can, so the credits had to parallelly stream its changes in case some happened during the main credits bootstrap process.

massa-execution-worker/src/context.rs
```diff
@@ -658,12 +658,27 @@ impl ExecutionContext {
...
-            if let Err(e) = self.transfer_coins(None, Some(addr), amount, false) {
+        let executed_credits = self.speculative_roll_state.get_deferred_credits(slot);
+
+        for (address, amount) in executed_credits {
+            // REVIEW NOTE: not sure which one is the best option here:
+            // * option 1: current one, requires to find the deferred_credits slot entry on every iteration even though the slot is always the same one
+            // * option 2: find the entry once but iterate twice because of the mutable reference incompatibility coming later from transfer_coins
+            // * option 3: get a copy of the slot credits first, modify the copy and replace the value in the deferred_credits after the iteration, but if transfer_coins fails no updates will be applied
+
+            self.speculative_roll_state
```
AurelienFT, line 669: Fetching in an hashmap is O(1) so there is not much problems fetching each slot I think.

AurelienFT (approved):

massa-execution-worker/src/context.rs
Eitu33: Removing the review comment and merging

PR #3232, opened 2022-11-22
Upgrade versions

AurelienFT:
  * [x] make tests compile
  * [x] make tests pass 

Need to fix compilation

### DIFF
--- a/massa-signature/Cargo.toml
+++ b/massa-signature/Cargo.toml
@@ -13,7 +13,7 @@ ed25519-dalek = { version = "1.0", features = ["batch"] }
 serde = { version = "1.0", features = ["derive"] }
 thiserror = "1.0"
 nom = "7.1"
-rand = "0.7"
+rand = "0.8"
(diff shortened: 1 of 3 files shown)


massa-signature/Cargo.toml
```diff
@@ -13,7 +13,7 @@ ed25519-dalek = { version = "1.0", features = ["batch"] }
 serde = { version = "1.0", features = ["derive"] }
 thiserror = "1.0"
 nom = "7.1"
-rand = "0.7"
+rand = "0.8"
```
damip, line 16: I'm afraid this one can't be updated for now

AurelienFT:
Rand can't be update for now but should be able soon as the update has been done on `ed25519_dalek` but in a RC version. https://github.com/dalek-cryptography/ed25519-dalek/issues/162#issuecomment-1321563329

Eitu33 (approved):
Astonishing

PR #3240, opened 2022-11-23
Keep old testnet final ledger.

AurelienFT:
  * [ ] make tests compile
  * [ ] make tests pass 

Need to be tested, don't think it's a good thing to keep it for testnet 17.

### DIFF
--- a/massa-node/src/main.rs
+++ b/massa-node/src/main.rs
@@ -135,9 +135,12 @@ async fn launch(
 
     // Remove current disk ledger if there is one
     // NOTE: this is temporary, since we cannot currently handle bootstrap from remaining ledger
-    if SETTINGS.ledger.disk_ledger_path.exists() {
+    let now = MassaTime::now(0).expect("could not get now time");
+    if *GENESIS_TIMESTAMP > now && SETTINGS.ledger.disk_ledger_path.exists() {
         std::fs::remove_dir_all(SETTINGS.ledger.disk_ledger_path.clone())
             .expect("disk ledger delete failed");
+    } else {
+        info!("Loading old ledger for next episode");
     }
 
     // Create final ledger
--- a/massa-bootstrap/src/client.rs
+++ b/massa-bootstrap/src/client.rs
@@ -426,13 +426,6 @@ pub async fn get_state(
         // init final state
         {
             let mut final_state_guard = final_state.write();
-            // load ledger from initial ledger file
-            final_state_guard
-                .ledger
-                .load_initial_ledger()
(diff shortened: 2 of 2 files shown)


damip:
That being said,
most of the SCs from 16 won't work on 17

AurelienFT:
> That being said, most of the SCs from 16 won't work on 17

Yes I think we will implement it in testnet 18.

damip:
note: this has impacts on the deploy scripts for labnet/testnet

AurelienFT:
> note: this has impacts on the deploy scripts for labnet/testnet

yes we need to remove the `storage` folder.

AurelienFT:
This PR is included in a more massive work about version and I think we should have some brainstorming before doing things like this. Moving it to TEST.19

Leo-Besancon:
This PR is tightly linked to the https://github.com/massalabs/massa/issues/3611 issue, so I think we should coordinate based on what we want to / can deliver for testnet_21, and what we want in the future.

In PR https://github.com/massalabs/massa/pull/3624, I made the assumption that we only keep the ledger if we want to restart from a snapshot, which is the case for only 1 node. (The other nodes will bootstrap the ledger from it as usual).

**If we include the network restart capabilities in testnet_21**, I would argue to keep the same approach: only have one node load the ledger with a CLI flag, and the other bootstrap from it. It's more expensive for the servers, but prevents syncing issues between them.

**If we don't include the network restart capabilities**, I think the current approach is fine for testnet_21. Maybe just add a `--remove-ledger` flag in sandbox mode to avoid having to delete it manually on every launch. (maybe even have it as the default?)

massa-node/src/main.rs
```diff
@@ -142,9 +142,12 @@ async fn launch(
 
     // Remove current disk ledger if there is one
     // NOTE: this is temporary, since we cannot currently handle bootstrap from remaining ledger
-    if SETTINGS.ledger.disk_ledger_path.exists() {
+    let now = MassaTime::now(0).expect("could not get now time");
+    if *GENESIS_TIMESTAMP > now && SETTINGS.ledger.disk_ledger_path.exists() {
```
Leo-Besancon, line 146: Why do you not use the keep_ledger flag here? If we are in sandbox, this will always remove the ledger, which we might not want.

Leo-Besancon (changes requested):
I think you still need to read the flag from the config.toml file (or from CLI).

AurelienFT:
Sorry @Leo-Besancon I have forgotten to commit a file...

PR #3243, opened 2022-11-24
Initial code for get_keys SC abi

sydhds:
  * [x] make tests compile
  * [x] make tests pass 


damip:
@sydhds  FYI https://github.com/massalabs/massa/blob/main/massa-execution-worker/src/execution.rs#L1127

sydhds:
Associated PR:
* https://github.com/massalabs/massa-unit-tests-src/pull/11
* https://github.com/massalabs/massa-sc-runtime/pull/173
* https://github.com/massalabs/massa-as-sdk/pull/78

massa-execution-worker/Cargo.toml
```diff
@@ -21,7 +21,8 @@ massa_execution_exports = { path = "../massa-execution-exports" }
 massa_models = { path = "../massa-models" }
 massa_storage = { path = "../massa-storage" }
 massa_hash = { path = "../massa-hash" }
-massa-sc-runtime = { git = "https://github.com/massalabs/massa-sc-runtime", branch = "testnet_17" }
+# massa-sc-runtime = { git = "https://github.com/massalabs/massa-sc-runtime", branch = "testnet_17" }
+massa-sc-runtime = { path = "../../massa-sc-runtime" }
```
AurelienFT, line 25: Don't merge before changing this.
(This comment is to not forget)

massa-execution-worker/src/speculative_ledger.rs
```diff
@@ -360,6 +361,29 @@ impl SpeculativeLedger {
...
+    pub fn get_keys(&self, addr: &Address) -> Option<Vec<Vec<u8>>> {
+
+        let mut added_keys = self.added_changes.get_keys(addr);
+        let (mut final_keys, mut candidate_keys) = get_final_and_candidate_datastore_keys(
+            &self.final_state.read().ledger,
+            &self.active_history.read().0,
+            addr);
+
+        final_keys.append(&mut candidate_keys);
+        added_keys.append(&mut final_keys);
```
AurelienFT, line 380: It's possible to have multiple times the same key even if only one exist here ? For example if it exists in `final_state` and has been modified in `added_changes` ?

sydhds: Nop as final_keys is a BTreeSet ;)

massa-execution-worker/src/speculative_ledger.rs
```diff
@@ -360,6 +361,29 @@ impl SpeculativeLedger {
...
+    ///
+    /// # Arguments
+    /// * `addr`: address to query
+    ///
+    /// # Returns
+    /// `Some(Vec<Vec<u8>>)` for found keys, `None` if the address does not exist.
+    pub fn get_keys(&self, addr: &Address) -> Option<Vec<Vec<u8>>> {
+
+        let mut added_keys = self.added_changes.get_keys(addr);
+        let (mut final_keys, mut candidate_keys) = get_final_and_candidate_datastore_keys(
```
damip, line 374: Do not use `get_final_and_candidate_datastore_keys` here because it performs a copy of the final state of the keys, that we don't need here. It is only needed in the API.

massa-execution-worker/src/speculative_ledger.rs
```diff
@@ -360,6 +361,29 @@ impl SpeculativeLedger {
...
+    /// Gets a copy of a datastore keys for a given address
+    ///
+    /// # Arguments
+    /// * `addr`: address to query
+    ///
+    /// # Returns
+    /// `Some(Vec<Vec<u8>>)` for found keys, `None` if the address does not exist.
+    pub fn get_keys(&self, addr: &Address) -> Option<Vec<Vec<u8>>> {
+
+        let mut added_keys = self.added_changes.get_keys(addr);
```
damip, line 373: Error: if a key was deleted in added_changes, that deletion won't be taken into account

massa-ledger-exports/src/ledger_changes.rs
```diff
@@ -750,6 +750,44 @@ impl LedgerChanges {
         }
     }
 
+    pub fn get_keys(
```
damip, line 753: this duplicates code and also it won't work as expected: information on key deletion is lost.
I recommend removing this function.

massa-execution-worker/src/speculative_ledger.rs
```diff
@@ -360,6 +361,29 @@ impl SpeculativeLedger {
...
+            &self.active_history.read().0,
+            addr);
+
+        final_keys.append(&mut candidate_keys);
+        added_keys.append(&mut final_keys);
+        Some(added_keys
+            .iter()
+            .cloned()
+            .collect())
+    }
```
damip, line 385: What I would recommend:
```rust
pub fn get_datastore_keys(&self, addr: &Address) -> Option<BTreeSet<Vec<u8>>> {  // note that I renamed the function to keep consistency. I also changed the return format to use a btreeset to keep determinism in data ordering.
    // here, get the final keys from the final ledger
    let mut keys: Option<BTreeSet<Vec<u8>>> = self.final_state.read().get_datastore_keys(addr);  // TODO change this function so that it returns an Option, especially None if the ledger entry is absent !!!  Also make sure it returns a btreeset for ordering determinism

    // here, traverse the history from oldest to newest with added_changes at the end, applying additions and deletions
    let active_history = active_history.read();
    let changes_iterator = active_history.0.iter().map(|item| item.state_changes.ledger_changes).chain(std::iter::once(&self.added_changes)); 
    for ledger_changes in changes_iterator {
        match ledger_changes.get(addr) {
            // address absent from the changes
            None => (),

            // address ledger entry being reset to an absolute new list of keys
            Some(SetUpdateOrDelete::Set(new_ledger_entry)) => {
                keys = Some(new_ledger_entry.datastore.keys().cloned().collect());
            }

            // address ledger entry being updated
            Some(SetUpdateOrDelete::Update(entry_updates)) => {
                let ref_keys = keys.get_or_insert_default();
                for (ds_key, ds_update) in &entry_updates.datastore {
                    match ds_update {
                        SetOrDelete::Set(_) => ref_keys.insert(ds_key.clone()),
                        SetOrDelete::Delete => ref_keys.remove(ds_key),
                    };
                }
            }

            // address ledger entry being deleted
            Some(SetUpdateOrDelete::Delete) => {
                keys = None;
            }
        }
    }
    
    keys
}
```

damip (changes requested):

massa-execution-worker/src/common.rs
```diff
@@ -0,0 +1,48 @@
+use massa_ledger_exports::{LedgerController, SetOrDelete, SetUpdateOrDelete};
+use massa_models::address::Address;
+
+use std::collections::{BTreeSet, VecDeque};
+use massa_execution_exports::ExecutionOutput;
+
+/// Get every final and active datastore key of the given address
+pub fn get_final_and_candidate_datastore_keys(
```
damip, line 8: I recommend removing the "common" file as this function is only useful for the API (we don't need a fully copy of the final state in any other case)

massa-execution-worker/src/execution.rs
```diff
@@ -1124,39 +1124,12 @@ impl ExecutionState {
         &self,
         addr: &Address,
     ) -> (BTreeSet<Vec<u8>>, BTreeSet<Vec<u8>>) {
-        // here, get the final keys from the final ledger, and make a copy of it for the candidate list
```
damip, line 1127: I recommend putting this back here instead of a separate "common" file

massa-execution-worker/src/speculative_ledger.rs
sydhds: done

massa-execution-worker/src/common.rs
sydhds: done

massa-execution-worker/src/execution.rs
sydhds: done

massa-ledger-exports/src/ledger_changes.rs
damip: I think it is still there :)

massa-ledger-worker/src/ledger_db.rs
```diff
@@ -330,13 +330,13 @@ impl LedgerDB {
...
     /// A `BTreeSet` of the datastore keys
-    pub fn get_datastore_keys(&self, addr: &Address) -> BTreeSet<Vec<u8>> {
+    pub fn get_datastore_keys(&self, addr: &Address) -> Option<BTreeSet<Vec<u8>>> {
         let handle = self.db.cf_handle(LEDGER_CF).expect(CF_ERROR);
 
         let mut opt = ReadOptions::default();
         opt.set_iterate_upper_bound(end_prefix(data_prefix!(addr)).unwrap());
 
-        self.db
+        Some(self.db
```
damip, line 339: I think this will still return Some even if the address does not exist in the ledger (?)

damip: We might need your help here @Eitu33

Eitu33: > I think this will still return Some even if the address does not exist in the ledger (?)

Yes it will return `Some([])` in that case

Eitu33: If the required behaviour is `None` when the addr isn't in the ledger you should check for its balance or bytecode entry before, can't use the datastore size as indicator as it will have the same result for non-existant addr and addr with no datatstore

damip: Ok then let's look for the balance

Eitu33: Yeah we could set the rockdb iterator lower bound to the address balance prefix identifier which would include the balance and bytecode in the retrieved keys, and then handle it as you wish

massa-ledger-exports/src/ledger_changes.rs
sydhds: My bad. I just removed it!

massa-ledger-worker/src/ledger_db.rs
damip: Nice ! Can you sync on that together ?

sydhds: Seen with Thomas so I've impl a 'quick fix' so we can merge this for TESTNET 17 and I'd like to create an issue to impl a better version (with rockdb iterator). Is this ok for you @damip ?

damip: sounds good !

massa-execution-worker/src/speculative_ledger.rs
```diff
@@ -369,11 +369,15 @@ impl SpeculativeLedger {
...
     pub fn get_keys(&self, addr: &Address) -> Option<BTreeSet<Vec<u8>>> {
-        // here, get the final keys from the final ledger
-        let mut keys: Option<BTreeSet<Vec<u8>>> = self.final_state
-            .read()
-            .ledger
-            .get_datastore_keys(addr);
+
+        let mut keys: Option<BTreeSet<Vec<u8>>> = match self.final_state.read().ledger.entry_exists(addr) {
+            // here, get the final keys from the final ledger
+            true => self.final_state
```
damip, line 375: problem: this is going to deadlock (final_state is locked in the scope above)

damip (changes requested):

massa-execution-worker/src/speculative_ledger.rs
sydhds: Does it apply also for parking_lot::RwLock (we are using this one and not the std::sync::RwLock impl)?

sydhds: Code updated anyway!

damip: > Does it apply also for parking_lot::RwLock (we are using this one and not the std::sync::RwLock impl)?

In parking_lot, it does not panic, but deadlocks: `Note that attempts to recursively acquire a read lock on a RwLock when the current thread already holds one may result in a deadlock.`

In any case, it is good practice to minimize locks, and avoid recursive locks in the same thread. When it is absolutely necessary,  re-entrant locks should be used: https://docs.rs/parking_lot/latest/parking_lot/type.ReentrantMutex.html

AurelienFT:
@sydhds the PR has been merged on runtime you can change back the path in dependencies

AurelienFT (approved):

massa-execution-worker/src/speculative_ledger.rs
```diff
@@ -360,6 +361,55 @@ impl SpeculativeLedger {
...
+    /// # Arguments
+    /// * `addr`: address to query
+    ///
+    /// # Returns
+    /// `Some(Vec<Vec<u8>>)` for found keys, `None` if the address does not exist.
+    pub fn get_keys(&self, addr: &Address) -> Option<BTreeSet<Vec<u8>>> {
+
+        let read_guard = self.final_state.read();
+        let mut keys: Option<BTreeSet<Vec<u8>>> = match read_guard.ledger.entry_exists(addr) {
+            true => read_guard.ledger.get_datastore_keys(addr),
```
damip, line 375: I think a more logical way would be to have the final state `get_datastore_keys` do the entry_exists check and return None if the entry doesn't exist, instead of always returning Some

aoudiamoncef: @sydhds is related to this https://github.com/massalabs/massa/issues/3275 ?

damip: https://github.com/massalabs/massa/issues/3275 is a followup to improve the performance of get_datastore_keys, here it is about the function actually doing what it is supposed to

damip (approved):

Issue #3225, opened 2022-11-21
Client CLI broken


damip:
Client CLI input breaks down:
* on Windows
* when the console is not wide enough (line break)

PR #3249, opened 2022-11-24
Change dialoguer for rustyline.

AurelienFT:
  * [x] make tests compile
  * [x] make tests pass 

Remove dialoguer (except for password) and use rustyline instead.

Fix https://github.com/massalabs/massa/issues/3225


massa-client/src/repl.rs
```diff
@@ -94,161 +86,107 @@ fn group_parameters(parameters: Vec<String>) -> Vec<String> {
...
+        completer: MassaCompleter::new(),
+        validator: MatchingBracketValidator::new(),
+    };
+    let config = Config::builder()
+        .auto_add_history(true)
+        .completion_prompt_limit(100)
+        .completion_type(CompletionType::List)
+        .max_history_size(10000)
+        .build();
+    let mut rl: Editor<MyHelper> = Editor::with_config(config).unwrap();
```
aoudiamoncef, line 114: I'm not a big fan of unwrap(). Could be replaced with `?`or handle the error explicitly

aoudiamoncef:
After locally testing the client, It seems good for me.

AurelienFT:
> After locally testing the client, It seems good for me.

Can you approve then please :) ? So that we can merge it.

aoudiamoncef (approved):

Issue #3263, opened 2022-11-27
Remove floating part from time display in bootstrap


aoudiamoncef:
Actual:
![image](https://user-images.githubusercontent.com/22281426/204162615-f013fd15-fbf9-4526-8a2e-1c75c58251d5.png)

Expected: 
Display time in secondes without floating part. 20s instead of 20.345675334s

modship:
Hi Moncef, 
#3265

aoudiamoncef:
@modship thank you for the PR ❤️

PR #3265, opened 2022-11-28
#3263 Remove floating part from time display in bootstrap

modship:
  * [ ] make tests compile
  * [ ] make tests pass 


massa-bootstrap/src/server.rs
```diff
@@ -355,23 +355,27 @@ pub async fn stream_bootstrap_information(
                 .get_executed_ops_part(last_ops_step);
             exec_ops_part = ops_data;
 
-            if let Some(slot) = last_slot && slot != final_state_read.slot {
-                if slot > final_state_read.slot {
-                    return Err(BootstrapError::GeneralError(
-                        "Bootstrap cursor set to future slot".to_string(),
-                    ));
+           final_state_changes = if let Some(slot) = last_slot  {
+                if slot != final_state_read.slot {
```
AurelienFT, line 359: Why not leaving this in the upper line with a `&&` and so having one `else` as before ?

AurelienFT (commented):
Thank you for the PR. Nice PR.

massa-bootstrap/src/server.rs
modship: Intellij show error on this line : 
image

and 

image

i think it just interpretor error.
I splited the pull request in two commit, i can drop the second or you can cherry pick the first. Like you want

AurelienFT: I think it's better is you drop the second one. Maybe your intellij doesn't support condition after if let.

modship: well it's done :)

AurelienFT (approved):

Issue #2627, opened 2022-05-18
Use bs58_check functions exclusively for debug purposes


Eitu33:
## Context

Address human-readable format is now handled through ToString and FromStr traits implementations. There is no purpose (debug excepted) for bs58_check functions anymore.

## TODO

- [ ] Set all the Adresses bs58_check functions as private
- [ ] Make sure they are no external occurrences of them

setar:
Hi, i have this problem on last branch of testnet ( fa6bdbf84e642f004b34c1103e1f2f0cbe2a2d06 ) 
```
massa@massa-node:~/massa/massa-node$ cargo run --release
warning: /home/massa/massa/massa-node/Cargo.toml: unused manifest key: build
    Finished release [optimized] target(s) in 4.04s
     Running `/home/massa/massa/target/release/massa-node`
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: parsing error : public key bs58_check parsing error: ... 
--- skipped ---
```

Any idea to solve it ?

Eitu33:
@setar Public keys format has changed, if you are using a PubKey, PrivKey, or Address from the previous episode your node won't work

setar:
Thanks, resolved via regenerate new keys

adrien-zinger:
@Eitu33 it looks like a good first issue, right?

AurelienFT:
Done in #3266

PR #3266, opened 2022-11-28
Use bs58_check functions exclusively for debug purposes

thomas-senechal:
resolves https://github.com/massalabs/massa/issues/2627

  * [x] make tests compile
  * [x] make tests pass 

I noticed that the Address' `from_bs58_check` and `to_bs58_check` were not used anymore so I deleted them from the code. 
I updated the examples that were using them and moved the Address' `bs58_check` examples as an example for the Address' `from_str` function.


AurelienFT (approved):
Thank you for the PR and welcome to the repository :)

PR #3301, opened 2022-12-05
Change latest_final_blocks_periods data type & its value update

Eitu33:
* [ ] try in labnet
  * [ ] make tests compile
  * [ ] make tests pass 

### DIFF
--- a/massa-consensus-worker/src/state/mod.rs
+++ b/massa-consensus-worker/src/state/mod.rs
@@ -1,4 +1,4 @@
-use std::collections::{HashMap, VecDeque};
+use std::collections::{HashMap, VecDeque, BTreeMap};
 
 use massa_consensus_exports::{
     block_graph_export::BlockGraphExport,
@@ -45,8 +45,9 @@ pub struct ConsensusState {
     pub active_index: PreHashSet<BlockId>,
     /// Save of latest periods
     pub save_final_periods: Vec<u64>,
-    /// One (block id, period) per thread
-    pub latest_final_blocks_periods: Vec<(BlockId, u64)>,
+    /// One block id per thread for the last 3 periods
+    /// IMPORTANT TODO: prune this
+    pub latest_final_blocks_periods: BTreeMap<u64, Vec<BlockId>>,
     /// One `(block id, period)` per thread TODO not sure I understand the difference with `latest_final_blocks_periods`
     pub best_parents: Vec<(BlockId, u64)>,
     /// Every block we know about
--- a/massa-consensus-worker/src/state/graph.rs
+++ b/massa-consensus-worker/src/state/graph.rs
@@ -5,7 +5,12 @@ use massa_consensus_exports::{
     error::ConsensusError,
 };
 use massa_logging::massa_trace;
-use massa_models::{block::BlockId, clique::Clique, prehash::PreHashSet, slot::Slot};
+use massa_models::{
+    block::{Block, BlockId},
+    clique::Clique,
+    prehash::PreHashSet,
+    slot::Slot,
+};
 
 use super::ConsensusState;
 
@@ -345,12 +350,11 @@ impl ConsensusState {
                     }];
                 }
                 // update latest final blocks
-                if final_block.slot.period
-                    > self.latest_final_blocks_periods[final_block.slot.thread as usize].1
-                {
-                    self.latest_final_blocks_periods[final_block.slot.thread as usize] =
-                        (block_id, final_block.slot.period);
-                }
+                let thread_list: &mut Vec<BlockId> = self
+                    .latest_final_blocks_periods
+                    .entry(final_block.slot.period)
+                    .or_insert_with(|| Vec::with_capacity(self.config.thread_count as usize));
+                thread_list[final_block.slot.thread as usize] = final_block.block_id;
                 // update new final blocks list
                 self.new_final_blocks.insert(block_id);
             } else {


massa-consensus-worker/src/state/mod.rs
```diff
@@ -96,6 +96,18 @@ impl ConsensusState {
         }
     }
 
+    /// Get a full active block
+    ///
+    /// Returns an error if it was not found
+    pub fn get_full_active_block_with_err(
```
damip, line 102: the proper formulation would be try_get_active_block or just get_active_block

massa-consensus-worker/src/state/mod.rs
```diff
@@ -199,125 +211,153 @@ impl ConsensusState {
...
+    fn list_latest_final_blocks_at(
+        &self,
+        slot: Slot,
+    ) -> Result<Vec<(BlockId, u64)>, ConsensusError> {
+        let mut latest: Vec<(BlockId, u64)> = Vec::with_capacity(self.config.thread_count as usize);
+        for id in self.active_index.iter() {
+            if let Some((block, _storage)) = self.get_full_active_block(id) {
+                if let Some((_id, period)) = latest.get(block.slot.thread as usize) {
+                    if block.is_final && block.slot <= slot && block.slot.period > *period {
+                        latest[block.slot.thread as usize] = (*id, block.slot.period)
```
damip, line 226: this will panic because "latest" is not initialized

massa-consensus-worker/src/state/mod.rs
```diff
@@ -199,125 +211,153 @@ impl ConsensusState {
...
+                    }
+                } else {
+                    latest[block.slot.thread as usize] = (*id, block.slot.period)
```
damip, line 229: this will panic as well

massa-consensus-worker/src/state/mod.rs
```diff
@@ -199,125 +211,153 @@ impl ConsensusState {
...
+    /// list the latest final blocks at the given slot
+    ///
+    /// exclusively used by `list_required_active_blocks`
+    fn list_latest_final_blocks_at(
+        &self,
+        slot: Slot,
+    ) -> Result<Vec<(BlockId, u64)>, ConsensusError> {
+        let mut latest: Vec<(BlockId, u64)> = Vec::with_capacity(self.config.thread_count as usize);
+        for id in self.active_index.iter() {
+            if let Some((block, _storage)) = self.get_full_active_block(id) {
```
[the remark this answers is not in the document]
Eitu33: Okay but are we sure we want to have an error in this case if the block is missing?

Eitu33:
@damip Edges cases are being properly handled, except for one, the client side post-bootstrap `list_required_active_blocks` check and re-launch

A call to `list_required_active_blocks` is made shortly after the bootstrap has succeeded, but currently terminates the client node if it fails. The feature allowing to re-launch a bootstrap procedure in the same process is beyond the scope of this issue I believe, as the initialisation order of the `massa-node` modules involves that some substantial work must be done in order to handle this.

We can have a call about this issue if you like

massa-consensus-worker/src/state/mod.rs
```diff
@@ -199,125 +215,150 @@ impl ConsensusState {
...
         }
+        for index in 0..(self.config.thread_count as usize) {
+            if latest[index].is_none() {
+                return Err(ConsensusError::ContainerInconsistency(format!(
+                    "could not find latest block for thread {}",
+                    index
+                )));
+            }
+        }
+        Ok(latest.into_iter().map(|opt| opt.unwrap()).collect())
```
damip, line 243: This whole part can be done with a single iteration and without unwraps:

```
latest.into_iter().enumerate().map(|(thread, opt)| opt.ok_or_else(|| {
    ConsensusError::ContainerInconsistency(format!(
       "could not find latest block for thread {}",
       thread
   )
}.collect()
```

massa-consensus-worker/src/state/mod.rs
```diff
@@ -199,125 +215,150 @@ impl ConsensusState {
...
+        for index in 0..(self.config.thread_count as usize) {
+            if earliest[index].is_none() {
+                return Err(ConsensusError::ContainerInconsistency(format!(
+                    "could not find earliest block for thread {}",
+                    index
+                )));
+            }
+        }
+        Ok(earliest.into_iter().map(|opt| opt.unwrap()).collect())
+    }
```
damip, line 275: same remark as above to make this more compact, safe, readable and efficient

massa-consensus-worker/src/state/mod.rs
```diff
@@ -199,125 +215,150 @@ impl ConsensusState {
...
-                    earliest_retained_periods[retain_slot.thread as usize],
-                    retain_slot.period,
-                );
+        // add all the active blocks that are after the effective_latest_finals of their thread
+        self.add_active_blocks_after(&mut kept_blocks, &effective_latest_finals, end_slot);
+
+        // do the following 3 times
+        for _ in 0..3 {
+            // extend kept_blocks with the parents of the current kept_blocks
+            let kept_blocks_clone = kept_blocks.clone();
```
damip, line 337: kept_blocks is potentially large. Instead of cloning it, it might be better to iterate over the current kept_blocks, accumulate the new items to be added in a separate container, and then add them to kept_blocks at the end of the outer loop iteration

massa-consensus-worker/src/state/mod.rs
```diff
@@ -199,125 +215,150 @@ impl ConsensusState {
...
-                    .0.slot;
-                earliest_retained_periods[retain_slot.thread as usize] = std::cmp::min(
-                    earliest_retained_periods[retain_slot.thread as usize],
-                    retain_slot.period,
-                );
+        // add all the active blocks that are after the effective_latest_finals of their thread
+        self.add_active_blocks_after(&mut kept_blocks, &effective_latest_finals, end_slot);
+
+        // do the following 3 times
+        for _ in 0..3 {
```
damip, line 335: sorry, it's not 0..3 but 0..2 (twice)

massa-consensus-worker/src/state/mod.rs
damip: looks like clippy agrees :)

Eitu33:
Corresponding issue opened here https://github.com/massalabs/massa/issues/3308

massa-consensus-worker/src/state/mod.rs
```diff
@@ -331,19 +337,20 @@ impl ConsensusState {
...
         self.add_active_blocks_after(&mut kept_blocks, &effective_latest_finals, end_slot);
 
-        // do the following 3 times
-        for _ in 0..3 {
+        // do the following 2 times
+        for _ in 0..2 {
             // extend kept_blocks with the parents of the current kept_blocks
-            let kept_blocks_clone = kept_blocks.clone();
-            for id in kept_blocks_clone.iter() {
+            let mut cumulated_parents: Vec<BlockId> = Vec::new();
```
damip, line 343: This might get very big because of heavy redundancy.
We should use a prehashset

Eitu33: Ah yes true mb

Eitu33: Done

damip (approved):

PR #3312, opened 2022-12-07
Use final hash in pos selections

Eitu33:
* [ ] try in labnet
* [ ] make tests compile
* [ ] make tests pass 

### DIFF
--- a/massa-final-state/src/final_state.rs
+++ b/massa-final-state/src/final_state.rs
@@ -136,10 +136,10 @@ impl FinalState {
         let n = (self.pos_state.cycle_history.len() == self.config.pos_config.cycle_history_length)
             as usize;
         for cycle_info in self.pos_state.cycle_history.iter().skip(n) {
-            hash_concat.extend(cycle_info.global_hash.to_bytes());
+            hash_concat.extend(cycle_info.cycle_global_hash.to_bytes());
             debug!(
                 "cycle ({}) hash at slot {}: {}",
-                cycle_info.cycle, slot, cycle_info.global_hash
+                cycle_info.cycle, slot, cycle_info.cycle_global_hash
             );
         }
         // 5. executed operations hash
--- a/massa-final-state/src/final_state.rs
+++ b/massa-final-state/src/final_state.rs
@@ -106,19 +106,9 @@ impl FinalState {
         self.pos_state
             .apply_changes(changes.pos_changes.clone(), self.slot, true)
             .expect("could not settle slot in final state proof-of-stake");
+        // TODO do not panic above: it might just mean that the lookback cycle is not available
         self.executed_ops
             .apply_changes(changes.executed_ops_changes.clone(), self.slot);
-
-        // cycle history debug log
-        debug!(
-            "After slot {} PoS cycle list is {:?}",
-            slot,
-            self.pos_state
-                .cycle_history
-                .iter()
-                .map(|c| (c.cycle, c.is_complete()))
-                .collect::<Vec<(u64, bool)>>()
-        );
 
         // push history element and limit history size
         if self.config.final_history_length > 0 {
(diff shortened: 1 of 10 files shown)


Eitu33:
Sorry for the base branch updates it's to remove diffs after merging `main` changes in `testnet_18`

massa-final-state/src/final_state.rs
```diff
@@ -151,6 +151,10 @@ impl FinalState {
         // 6. final state hash
         let final_state_hash = Hash::compute_from(&hash_concat);
         info!("final_state hash at slot {}: {}", slot, final_state_hash);
+        // 7. save final state hash in the latest cycle
+        if let Some(cycle) = self.pos_state.cycle_history.back_mut() {
```
damip, line 155: a more robust and less error-prone way would be to create a feed_cycle_state_hash function in the pos state, and have it internally use get_cycle_index to make sure it grabs the right state

massa-pos-exports/src/pos_final_state.rs
```diff
@@ -242,10 +242,16 @@ impl PoSFinalState {
...
+                // it will later be combined with rng_seed from cycle - 2 to determine the selection seed
+                // do this here to avoid a potential attacker manipulating the selections
+                let state_hash = cycle_info
+                    .final_state_hash_snapshot
+                    .expect("critical: a complete cycle must contain a final state hash snapshot");
+                (cycle_info.roll_counts.clone(), Some(state_hash))
             }
             // looking back to negative cycles
-            None => self.initial_rolls.clone(),
+            None => (self.initial_rolls.clone(), None),
```
damip, line 254: instead of None, could we use the hash of the initial state here ?

massa-pos-exports/src/pos_final_state.rs
```diff
@@ -259,7 +265,11 @@ impl PoSFinalState {
                 if !cycle_info.complete {
                     return Err(PosError::CycleUnfinished(c));
                 }
-                Hash::compute_from(&cycle_info.rng_seed.clone().into_vec())
+                let mut seed = cycle_info.rng_seed.clone().into_vec();
```
damip, line 268: I think we should also prepend the serialized current cycle number to the seed in order to make it really unique

massa-final-state/src/final_state.rs
```diff
@@ -98,15 +146,15 @@ impl FinalState {
...
+        // unwrap is justified because every error in PoS `apply_changes` is critical
         self.ledger
             .apply_changes(changes.ledger_changes.clone(), self.slot);
         self.async_pool
             .apply_changes_unchecked(&changes.async_pool_changes);
         self.pos_state
             .apply_changes(changes.pos_changes.clone(), self.slot, true)
-            .expect("could not settle slot in final state proof-of-stake");
-        // TODO do not panic above: it might just mean that the lookback cycle is not available
+            .unwrap();
```
damip, line 157: I don't understand this change (from expect to unwrap)

damip: The reason of the expect was initially that an unavailable PoS cycle should cause a new bootstrap attempt instead of a crash, to be implemented in the future

damip (approved):
Very nice !
Maybe it is worth testing it on the labnet before we deliver it

massa-final-state/src/final_state.rs
```diff
@@ -32,8 +32,12 @@ pub struct FinalState {
     /// history of recent final state changes, useful for streaming bootstrap
     /// `front = oldest`, `back = newest`
     pub changes_history: VecDeque<(Slot, StateChanges)>,
+    /// hash of the final state, it is computed on finality
+    pub final_state_hash: Hash,
 }
 
+const FINAL_STATE_HASH_INITIAL_BYTES: &[u8; 32] = &[0; HASH_SIZE_BYTES];
```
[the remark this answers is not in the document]
Eitu33: this was supposed to be `Hash::from_bytes(FINAL_STATE_HASH_INITIAL_BYTES)`, good catch I updated it

Eitu33:
Yeah this def needs some extensive testing, will do once all the finished work is merged in `testnet_18` 👍

Issue #3076, opened 2022-09-24
API refactoring


damip:
Refactor the API:
1 - specify the business needs with Core, the Innovation team, direct users (eg. Dusa) end the community Builders. Try to split endpoints as much as possible. Use snake_case everywhere (https://github.com/massalabs/massa/pull/2823). Decide if we fuse the private/public APIs and allow finer access control through other means.
2 - write a new API that uses https://github.com/paritytech/jsonrpsee crate and:
  * enable more advanced access control
  * enable HTTPS encryption
  * enable call batching to read/write time-consistent information
  * allow websocket streaming
  * enable rate limiting

damip:
@aoudiamoncef  (no need to start yet, I'm just preparing the project so that you can start efficiently when you arrive :) )

aoudiamoncef:
related to https://github.com/massalabs/massa/issues/2799

PR #3314, opened 2022-12-08
Add WebSockets support + ApiV2

aoudiamoncef:
  * [x] make tests compile
  * [x] make tests pass 

### DIFF
--- a/massa-consensus-worker/src/controller.rs
+++ b/massa-consensus-worker/src/controller.rs
@@ -227,25 +227,27 @@ impl ConsensusController for ConsensusControllerImpl {
     }
 
     fn register_block(&self, block_id: BlockId, slot: Slot, block_storage: Storage, created: bool) {
-        block_storage.read_blocks().get(&block_id).map(|value| {
-            let operations: Vec<(OperationId, Option<Wrapped<Operation, OperationId>>)> = value
-                .content
-                .operations
-                .iter()
-                .map(|operation_id| {
-                    match block_storage.read_operations().get(operation_id).cloned() {
-                        Some(wrapped_operation) => (*operation_id, Some(wrapped_operation)),
-                        None => (*operation_id, None),
-                    }
-                })
-                .collect();
+        if self.ws_config.enabled {
+            block_storage.read_blocks().get(&block_id).map(|value| {
+                let operations: Vec<(OperationId, Option<Wrapped<Operation, OperationId>>)> = value
+                    .content
+                    .operations
+                    .iter()
+                    .map(|operation_id| {
+                        match block_storage.read_operations().get(operation_id).cloned() {
+                            Some(wrapped_operation) => (*operation_id, Some(wrapped_operation)),
+                            None => (*operation_id, None),
+                        }
+                    })
+                    .collect();
 
-            let _ = self.ws_config.block_sender.send(value.content.clone());
-            self.ws_config.filled_block_sender.send(FilledBlock {
-                header: value.content.header.clone(),
-                operations,
-            })
-        });
+                let _ = self.ws_config.block_sender.send(value.content.clone());
+                self.ws_config.filled_block_sender.send(FilledBlock {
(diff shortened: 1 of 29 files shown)


massa-consensus-worker/src/worker/mod.rs
```diff
@@ -35,6 +36,18 @@ pub struct ConsensusWorker {
...
+pub struct WsConfig {
+    /// Whether WebSockets are enabled
+    pub enabled: bool,
+    /// Broadcast sender(channel) for blocks headers
+    pub block_header_sender: Sender<BlockHeader>,
+    /// Broadcast sender(channel) for blocks
+    pub block_sender: Sender<Block>,
+    /// Broadcast sender(channel) for filled blocks
+    pub filled_block_sender: Sender<FilledBlock>,
+}
```
Eitu33, line 49: Was there any discussion I'm unaware of where it was decided that we'd use tokio channels for this? Why not use crossbeam instead?

massa-consensus-worker/src/controller.rs
```diff
@@ -219,6 +227,28 @@ impl ConsensusController for ConsensusControllerImpl {
...
+                        }
+                    })
+                    .collect();
+
+                let _ = self.ws_config.block_sender.send(value.content.clone());
+                self.ws_config.filled_block_sender.send(FilledBlock {
+                    header: value.content.header.clone(),
+                    operations,
+                })
+            });
```
Eitu33, line 249: You are using `map` to produce side effects, `map` should only be used when it is needed to modify the value and wield its result, please use `if let Some(value)` here

massa-consensus-worker/src/worker/mod.rs
aoudiamoncef: IMHO the new library is built on top of Tokio so it seems to be a safe choice.

this is the [example](https://github.com/paritytech/jsonrpsee/blob/master/examples/examples/ws_pubsub_broadcast.rs) which I followed to dev this feature.

https://github.com/massalabs/massa/issues/3076

massa-models/src/config/constants.rs
```diff
@@ -64,7 +64,7 @@ lazy_static::lazy_static! {
         if cfg!(feature = "sandbox") {
             "SAND.0.0"
         } else {
-            "TEST.18.0"
+            "TEST.17.2"
```
AurelienFT, line 67: Shouldn't be push I think.

massa-node/base_config/config.toml
```diff
@@ -11,6 +11,8 @@
     bind_private = "127.0.0.1:33034"
     # port on which the node API listens for public requests. Can be exposed to the Internet
     bind_public = "0.0.0.0:33035"
+    # port on which the node API listens for HTTP requests and WebSOckets subscriptions. Can be exposed to the Internet
+    bind_api = "0.0.0.0:33036"
```
AurelienFT, line 15: Maybe rename it to v2 to be more clear for users.

aoudiamoncef: Just add it in the comment could be fine, I don't went to add v2 everywhere

AurelienFT: Adding it in config file that users reads doesn't seems too much to me but you can just put a comment ok

massa-models/src/config/constants.rs
aoudiamoncef: Good catch

massa-node/base_config/config.toml
aoudiamoncef: Done

massa-models/src/config/constants.rs
aoudiamoncef: Done

aoudiamoncef:
Please don't merge this until all the PR checkboxes are filled. We should write OpenRPC doc

AurelienFT (approved):

PR #3315, opened 2022-12-08
Keep events after SC execution failure

Eitu33:
* [ ] try in sandbox & labnet
* [x] unit tests on the added & changed features
  * [x] make tests compile
  * [x] make tests pass 


Eitu33:
`execution_error.wasm` corresponding code can be found in https://github.com/massalabs/massa-unit-tests-src/pull/16

massa-client/src/cmds.rs
```diff
@@ -576,13 +576,14 @@ impl Command {
             }
 
             Command::get_filtered_sc_output_event => {
-                let p_list: [&str; 6] = [
+                let p_list: [&str; 7] = [
```
AurelienFT, line 579: Need to update the help at the top to match this new behavior

massa-execution-exports/src/event_store.rs
```diff
@@ -87,6 +87,11 @@ impl EventStore {
                     (Some(_), None) => return false,
                     _ => (),
                 }
+                match filter.is_error {
+                    Some(is_error) if x.context.is_error != is_error => return false,
+                    None if x.context.is_error => return false,
```
AurelienFT, line 92: So by default we don't show the events on a operation with an execution that failed ? That's a different behavior then the others ones. If there is a good reason we can keep it but we need to put it on the help in the description of the CLI and in the API docs

AurelienFT (commented):
Need to update the `openrpc.json` for the api that has been changed. Also I think it's better to merge the PR on unit tests before to not break CI.

massa-execution-exports/src/event_store.rs
Eitu33: This was Damir's insight, thought process was that we don't want the current behaviour to change for users that are building stuff on massa. Current implementation is not breaking on input nor on output values

massa-client/src/cmds.rs
Eitu33: Done

massa-execution-exports/src/event_store.rs
AurelienFT: I'm just saying that" not specifying a filter means "it filter" " : that is a different behavior than the other filters on the CLI so this needs to be specify.

As a side note you said "we don't want the current behaviour to change for users" but we are changing it here because the error message that is shown when an operation fail will now not be displayed when no filter. This is a change.

Eitu33: Yeah true, tbh I don't have a strong opinion about what should be adopted here, both behaviours seem justified to me

AurelienFT: That's my opinion too

massa-execution-exports/src/event_store.rs
```diff
@@ -9,7 +9,7 @@ use std::collections::VecDeque;
 
 /// Store for events emitted by smart contracts
 #[derive(Default, Debug, Clone)]
-pub struct EventStore(VecDeque<SCOutputEvent>);
+pub struct EventStore(pub VecDeque<SCOutputEvent>);
```
AurelienFT, line 12: Why you changed this to public ?

AurelienFT (approved):

massa-execution-exports/src/event_store.rs
Eitu33: To avoid reimplementing iterator functions for `EventStore`, need them to perform a diff with the snapshot in order to deduce events coming from failed executions

PR #3317, opened 2022-12-09
Test/cbindgen test 1

sydhds:
  * [ ] make tests compile
  * [ ] make tests pass 


damip:
@sydhds I think we can close this in favor of GRPC ?

PR #3318, opened 2022-12-09
Gas limit for read-only / execute / call SC execution

Eitu33:
  * [x] make tests compile
  * [x] make tests pass 


massa-execution-worker/src/execution.rs
```diff
@@ -208,8 +208,14 @@ impl ExecutionState {
...
 
-        // check remaining block gas
+        // check operation gas
         let op_gas = operation.get_gas_usage();
+        if op_gas > self.config.max_gas_per_execution {
+            return Err(ExecutionError::ExecutionGasOverThreshold(format!(
+                "execution gas for operation {} is {} which is above the maximum allowed {}",
+                operation.id, op_gas, self.config.max_gas_per_execution
+            )));
+        }
```
damip, line 218: this part is not necessary

massa-execution-worker/src/execution.rs
```diff
@@ -1011,6 +1017,13 @@ impl ExecutionState {
         // TODO ensure that speculative things are reset after every execution ends (incl. on error and readonly)
         // otherwise, on prod stats accumulation etc... from the API we might be counting the remainder of this speculative execution
 
+        if req.max_gas > self.config.max_gas_per_execution {
+            return Err(ExecutionError::ExecutionGasOverThreshold(format!(
+                "execution gas for read-only call is {} which is above the maximum allowed {}",
+                req.max_gas, self.config.max_gas_per_execution
+            )));
+        }
```
damip, line 1025: this is the only useful one. All the other ones are redundant

massa-api/src/public.rs
```diff
@@ -215,6 +223,14 @@ impl MassaRpcServer for API<Public> {
...
         } in reqs
         {
+            if max_gas > self.0.api_settings.max_gas_per_execution {
+                return Err(ApiError::BadRequest(format!(
+                    "one of the provided read-only call requests has a gas amount ({}) above the allowed maximum ({})",
+                    max_gas, self.0.api_settings.max_gas_per_execution
+                ))
+                .into());
+            }
+
```
damip, line 233: not necessary because the check happens in execution anyways

massa-api/src/public.rs
```diff
@@ -879,6 +895,14 @@ impl MassaRpcServer for API<Public> {
...
                         ApiError::ModelsError(ModelsError::DeserializeError(err.to_string()))
                     })?;
+                let op_gas = op.get_gas_usage();
+                if op_gas > self.0.api_settings.max_gas_per_execution {
+                    return Err(ApiError::BadRequest(format!(
+                        "operation {} has a gas amount ({}) above the allowed maximum ({})",
+                        op.id, op_gas, self.0.api_settings.max_gas_per_execution
+                    ))
+                    .into());
+                }
```
damip, line 905: not necessary: op won't be executed if its gas is above max_block_gas

massa-execution-worker/src/execution.rs
damip: Also, let's rename `self.config.max_gas_per_execution` into `self.config.max_read_only_gas`

massa-models/src/config/constants.rs
```diff
@@ -196,6 +196,8 @@ pub const POOL_CONTROLLER_CHANNEL_SIZE: usize = 1024;
 
 /// Maximum of GAS allowed for a block
 pub const MAX_GAS_PER_BLOCK: u64 = 1_000_000_000;
+/// Maximum of GAS allowed for a single execution
+pub const MAX_GAS_PER_EXECUTION: u64 = 100_000_000;
```
damip, line 200: it shouldn't be a constant but a config file entry

damip (approved):

PR #3320, opened 2022-12-10
Add bootsrap IPs lists management

### LINKED ISSUE #3316: Add endpoint to edit bootstrap white/black lists
Whitelist/Blacklists allow people give rights to an IP to connect to their node.

Currently they are only editable via a JSON file.

TODO: Add API endpoint and client command to manage those configs dynamically.

aoudiamoncef:
  * [x] make tests compile
  * [x] make tests pass 

closes #3316


aoudiamoncef:
image

aoudiamoncef:
```
command > node_bootsrap_whitelist
Error: check if your node is running: RPC call failed: ErrorObject { code: ServerError(-32001), message: "Internal server error: failed to read bootsrap whitelist configuration file: No such file or directory (os error 2)", data: None }
```

I don't like the fact that we just throw the error back to the client... may be we should do something in the RPC method. Actually, we throw an error when bootsrap whitelist file doesn't exist. 

Proposal: 

- Check if the file exists, if not throw an error with an explicite message telling the client that everyone could bootsrap. 

- Return null instead of throwing an error.  null =/= [] (seems to be a good idea)

aoudiamoncef:
> image

@damip this is the actual workflow

massa-client/src/cmds.rs
```diff
@@ -251,6 +261,30 @@ pub enum Command {
     when_moon,
 }
 
+#[derive(Debug, Display, EnumString, EnumIter)]
+#[strum(serialize_all = "snake_case")]
+pub enum CLIOperation {
```
Eitu33, line 266: Let's rename this to `ListOperations` since those are the only commands that make use of that

massa-client/src/cmds.rs
```diff
@@ -989,29 +1023,179 @@ impl Command {
...
+                                Ok(()) => {
+                                    if !json {
+                                        println!("Request of remove from bootsrap blacklist successfully sent!")
+                                    }
+                                    Ok(Box::new(()))
+                                }
+                                Err(e) => rpc_error!(e),
+                            }
+                        }
+                        _ => bail!("Not implemented"),
```
Eitu33, line 1073: Let's match `AllowAll` explicitly and improve this error message, also in `node_peers_whitelist`

massa-api/src/private.rs
```diff
@@ -82,9 +86,9 @@ impl MassaRpcServer for API<Private> {
     async fn node_sign_message(&self, message: Vec<u8>) -> RpcResult<PubkeySig> {
         let network_command_sender = self.0.network_command_sender.clone();
         match network_command_sender.node_sign_message(message).await {
-            Ok(public_key_signature) => return Ok(public_key_signature),
-            Err(e) => return Err(ApiError::from(e).into()),
-        };
+            Ok(public_key_signature) => Ok(public_key_signature),
+            Err(e) => Err(ApiError::from(e).into()),
+        }
```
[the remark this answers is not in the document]
aoudiamoncef: I tried to find a better way to do it but I stick with this solution. I'm really interested if y' there is a better way to achieve this purpose.

massa-client/src/cmds.rs
aoudiamoncef: It was my first implementation. But I changed in case we add a new Operation to ListOperation we don't have ton propagate to match arms.

massa-api/src/private.rs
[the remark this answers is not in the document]
aoudiamoncef: It was like this before: 

```rust
    fn node_sign_message(&self, message: Vec<u8>) -> BoxFuture<Result<PubkeySig, ApiError>> {
        let network_command_sender = self.0.network_command_sender.clone();
        let closure = async move || Ok(network_command_sender.node_sign_message(message).await?);
        Box::pin(closure())
    }
`
``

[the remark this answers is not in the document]
Comment (Eitu33) [reply massa-api/src/private.rs:91]: `impl From<use massa_network_exports::NetworkError> for JsonRpseeError` directly, we use `ApiError` everywhere as in intermediated. Basically implemented all the conversions to a type we never return and that will be converted itself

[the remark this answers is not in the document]
Comment (aoudiamoncef) [reply massa-api/src/private.rs:91]: Proposal: 
```rust
    async fn node_sign_message(&self, message: Vec<u8>) -> RpcResult<PubkeySig> {
        let network_command_sender = self.0.network_command_sender.clone();
        network_command_sender.node_sign_message(message).await
        .map_err(|e| ApiError::NetworkError(e).into())
    }
```

massa-client/src/cmds.rs
Eitu33: We'd want to propagate anyway, if someone misuses the command I expect it to retrieve a proper error not juste a "Not implemented"

massa-api/src/private.rs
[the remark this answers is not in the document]
Eitu33: This is just replacing the `match` with a `map_err` whole point of my comment was to removing explicit conversions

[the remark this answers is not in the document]
Eitu33: We just need to implement the conversion traits for the error type that we decide to use and not use intermediate types (`ApiError` in this implementation)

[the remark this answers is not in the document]
Eitu33: But let's discuss about it some other time this isn't in the scope of this PR

massa-api/src/private.rs
```diff
@@ -118,48 +122,47 @@ impl MassaRpcServer for API<Private> {
...
+        w_wallet
+            .remove_addresses(&addresses)
+            .map_err(|e| ApiError::WalletError(e).into())
     }
 
     async fn get_staking_addresses(&self) -> RpcResult<PreHashSet<Address>> {
         let node_wallet = self.0.node_wallet.clone();
-        let addresses_set = node_wallet.write().get_wallet_address_list();
-        Ok(addresses_set)
+        let w_wallet = node_wallet.read();
```
aoudiamoncef, line 132: Why this was `write()`?

sydhds:
> ```
> command > node_bootsrap_whitelist
> Error: check if your node is running: RPC call failed: ErrorObject { code: ServerError(-32001), message: "Internal server error: failed to read bootsrap whitelist configuration file: No such file or directory (os error 2)", data: None }
> ```
> 
> I don't like the fact that we just throw the error back to the client... may be we should do something in the RPC method. Actually, we throw an error when bootsrap whitelist file doesn't exist.
> 
> Proposal:
> 
> * Check if the file exists, if not throw an error with an explicite message telling the client that everyone could bootsrap.
> * Return null instead of throwing an error.  null =/= [] (seems to be a good idea)

massa-api/src/private.rs
AurelienFT: No idea. Probably a bad copy/paste

massa-api/src/private.rs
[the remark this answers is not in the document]
AurelienFT: Can you open an issue @aoudiamoncef so that we don't loose this and close this comment

massa-client/src/cmds.rs
sydhds: Checked and it's now fixed; closing it

massa-client/src/cmds.rs
sydhds: Checked and it's fixed. Closing it...

massa-api/src/private.rs
```diff
@@ -213,23 +217,225 @@ impl MassaRpcServer for API<Private> {
...
+            ScrudOperation::Delete,
+        )
     }
 
     async fn get_openrpc_spec(&self) -> RpcResult<Value> {
         crate::wrong_api::<Value>()
     }
 }
+
+fn bootsrap_list_call(
```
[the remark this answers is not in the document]
sydhds: Renamed + doc added. Closing it...

massa-api/src/private.rs
```diff
@@ -213,23 +217,225 @@ impl MassaRpcServer for API<Private> {
...
+    ips: Vec<IpAddr>,
+    list_type: ListType,
+    scrud_operation: ScrudOperation,
+) -> RpcResult<()> {
+    match scrud_operation {
+        ScrudOperation::Create => {
+            get_file_with_length(bootstrap_list_file.clone(), &list_type, true).and_then(|tuple| {
+                if tuple.1 == 0 {
+                    write_to_jsonfile(bootstrap_list_file, BTreeSet::from_iter(ips), &list_type)
+                } else {
```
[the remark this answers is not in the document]
sydhds: Checked and looks ok. Closing it...

massa-node/base_config/config.toml
```diff
@@ -14,7 +14,7 @@
     # max number of arguments per RPC call
     max_arguments = 128
     # path to the openrpc specification file used in `rpc.discover` method
-    openrpc_spec_path = "base_config/openrpc.json"
+    openrpc_spec_file = "base_config/openrpc.json"
```
sydhds, line 17: Not very important but here we specify a filepath so I would rename it to filepath to be really explicit?

massa-models/src/api.rs
```diff
@@ -15,6 +15,7 @@ use massa_time::MassaTime;
 use serde::{Deserialize, Serialize};
 use std::collections::BTreeMap;
 use std::net::IpAddr;
+use strum::Display;
```
sydhds, line 18: Why do we need strum here? Cannot we implement Display on our own here?

massa-client/src/cmds.rs
```diff
@@ -1069,8 +1257,16 @@ async fn send_operation(
 
 /// TODO: ugly utilities functions
 /// takes a slice of string and makes it into a `Vec<T>`
-pub fn parse_vec<T: std::str::FromStr>(args: &[String]) -> anyhow::Result<Vec<T>, T::Err> {
-    args.iter().map(|x| x.parse::<T>()).collect()
+pub fn parse_vec<T: std::str::FromStr>(args: &[String]) -> anyhow::Result<Vec<T>, Error>
+where
+    T::Err: Display,
```
sydhds, line 1262: Not sure why we need to add a constraint here. Cannot we just format with {:?} ?

AurelienFT:
@aoudiamoncef Can you fix the conflict so that we can final review and merge ?

sydhds (changes requested):
Good PR! 

Can we plan to split massa-client/cmds.rs because it's huge and quite hard to read easily (maybe create an issue for that)?

massa-models/src/api.rs
aoudiamoncef: Of course we could implement it.

massa-client/src/cmds.rs
aoudiamoncef: `parse_vec` was flagged as UGLY method. Any improvements are welcome. I tried to make it better

massa-node/base_config/config.toml
AurelienFT: I don't think "file" is better than "path" because it's a "path". "path" and "filepath" are both ok but as we are using "path" everywhere (and a lot of projects does) I don't see any reason to rename "filepath" everywhere

aoudiamoncef: In massa `_path` wins compared to `_file`

massa-client/src/cmds.rs
sydhds: Not sure to understand... Who flagged this function as ugly? and why?

aoudiamoncef: @AurelienFT should knows the history of this method

aoudiamoncef:
```rust
/// TODO: ugly utilities functions
/// takes a slice of string and makes it into a `Vec<T>`
pub fn parse_vec<T: std::str::FromStr>(args: &[String]) -> anyhow::Result<Vec<T>, Error>
where
    T::Err: Display,
{
    args.iter()
        .map(|x| {
            x.parse::<T>()
                .map_err(|e| anyhow!("failed to parse \"{}\" due to: {}", x, e))
        })
        .collect()
}
```

I need to display which value is causing the error to the user. Feel free if you have a better idea to implement this

sydhds (approved):

PR #3323, opened 2022-12-12
Local execution

Eitu33:


massa-execution-worker/src/tests/scenarios_mandatories.rs
```diff
@@ -321,6 +321,98 @@ fn send_and_receive_async_message() {
     manager.stop();
 }
 
+#[test]
```
AurelienFT, line 324: I think it's cool if we always have the walkthrough like in the others tests !

AurelienFT (commented):
Nice one.

massa-execution-worker/src/tests/scenarios_mandatories.rs
Eitu33: Done

AurelienFT (approved):
"Mergeable" when the dependencies are merged

PR #3333, opened 2022-12-15
Init sc update

Eitu33:


AurelienFT (approved):

massa-execution-worker/src/interface_impl.rs
```diff
@@ -343,6 +343,27 @@ impl Interface for InterfaceImpl {
...
     }
 
+    /// Check whether or not the caller has write access in the current context
+    ///
+    /// # Returns
+    /// true if the caller has write access
+    fn caller_has_write_access(&self) -> Result<bool> {
+        let context = context_guard!(self);
+        let mut call_stack_iter = context.get_call_stack().into_iter().rev();
+        let caller = if let Some(last) = call_stack_iter.next() {
```
AurelienFT, line 353: Something that I find strange is that you seems to fetch the right element in the stack here that is the caller address but after you are iterating again in `get_owned_addresses_for ` to find it again. Couldn't be possible to factorize them?

Eitu33: True, I improved it

AurelienFT (approved):

PR #3342, opened 2022-12-22
Add `get_blocks` method to JSONRPC API

aoudiamoncef:
  * [x] make tests compile
  * [x] make tests pass 

### DIFF
--- a/massa-api/src/public.rs
+++ b/massa-api/src/public.rs
@@ -588,11 +588,10 @@ impl MassaRpcServer for API<Public> {
         let blocks = ids
             .into_iter()
             .filter_map(|id| {
-                let block = match storage.read_blocks().get(&id).cloned() {
-                    Some(b) => b.content,
-                    None => {
-                        return None;
-                    }
+                let block = if let Some(wrapped_block) = storage.read_blocks().get(&id) {
+                    wrapped_block.content.clone()
+                } else {
+                    return None;
                 };
                 if let Some(graph_status) = consensus_controller
                     .get_block_statuses(&[id])
--- a/massa-api/src/public.rs
+++ b/massa-api/src/public.rs
@@ -580,40 +580,48 @@ impl MassaRpcServer for API<Public> {
         Ok(res)
     }
 
-    /// gets a block. Returns None if not found
+    /// gets a block(s). Returns nothing if not found
     /// only active blocks are returned
-    async fn get_block(&self, id: BlockId) -> RpcResult<BlockInfo> {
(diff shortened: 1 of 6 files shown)


massa-node/base_config/openrpc.json
```diff
@@ -267,7 +270,7 @@
             "params": [
                 {
                     "name": "endorsementId",
-                    "description": "Must be an endorsement id",
+                    "description": "Must be an endorsement(s) id(s)",
```
sydhds, line 273: Maybe better: "Need to provide at least one valid endorsement id"

massa-client/src/cmds.rs
```diff
@@ -548,13 +548,13 @@ impl Command {
                 }
             }
 
-            Command::get_block => {
+            Command::get_blocks => {
                 if parameters.len() != 1 {
-                    bail!("wrong param numbers")
+                    bail!("wrong param numbers, [IpAddr] shouldn't be empty")
```
sydhds, line 553: Error message could be better: Expecting at least one IpAddr ?

massa-api/src/public.rs
```diff
@@ -580,40 +580,47 @@ impl MassaRpcServer for API<Public> {
...
-            content: Some(BlockInfoContent {
-                is_final,
-                is_in_blockclique,
-                is_candidate,
-                is_discarded,
-                block,
-            }),
-        })
+            .filter_map(|id| {
+                let block = if let Some(wrapped_block) = storage.read_blocks().get(&id) {
```
sydhds, line 591: I would suggest to rewrite it like that (so it will be easier to read):

```

let block = if let Some(...) = ... {
     ....
     if let Some(graph_status) = consensus_controller... {
         ...
        return Some(BlockInfoContent { ... });
    }
}
return None;



```

sydhds (changes requested):

massa-client/src/cmds.rs
aoudiamoncef: Errors messages... you know it's HARD, but having some convention would be cool as this one is repetitive.

massa-api/src/public.rs
aoudiamoncef: Done

massa-client/src/cmds.rs
aoudiamoncef: Done

massa-node/base_config/openrpc.json
aoudiamoncef: done

sydhds (approved):

PR #3347, opened 2022-12-27
Add broadcast operations via WebSocket

aoudiamoncef:
  * [x] make tests compile
  * [x] make tests pass 


massa-node/base_config/config.toml
```diff
@@ -132,6 +132,8 @@
     max_operations_propagation_time = 32000
     # time threshold after which endorsement are not propagated
     max_endorsements_propagation_time = 48000
+    # operations sender(channel) capacity
+    ws_operations_capacity = 128
```
AurelienFT, line 136: That's a really low number as we have multiple thousands of op in 1 second and you can receive batch that are way more huge then 128. I think you can put 5_000 at least

massa-ws/src/lib.rs
```diff
@@ -0,0 +1,37 @@
...
+
+use jsonrpsee::{core::error::SubscriptionClosed, SubscriptionSink};
+use serde::Serialize;
+use tokio::sync::broadcast::Sender;
+use tokio_stream::wrappers::BroadcastStream;
+
+/// Re-exports libraries to not require any additional
+/// dependencies to be explicitly added on the module(s) side.
+#[doc(hidden)]
+pub mod __reexports {
```
AurelienFT, line 15: Never saw this way of re-exporting. The idiomatic way (I believe) is to `pub use` the part you want to re-export. I have made this in my previous PR : https://github.com/massalabs/massa/blob/testnet_18/massa-execution-exports/src/lib.rs#L56 (I re-export some massa-sc-runtime stuff)

massa-protocol-worker/src/protocol_worker.rs
```diff
@@ -1080,6 +1100,11 @@ impl ProtocolWorker {
 
         Ok(())
     }
+
+    /// New produced operations.
+    pub(crate) fn subscribe_new_operations(&self, sink: SubscriptionSink) {
+        broadcast_via_ws(self.ws_config.operation_sender.clone(), sink);
```
AurelienFT, line 1106: I don't think making a function for a line of code it's worth it.

massa-consensus-worker/src/controller.rs
```diff
@@ -301,29 +298,14 @@ impl ConsensusController for ConsensusControllerImpl {
     }
 
     fn subscribe_new_blocks_headers(&self, sink: SubscriptionSink) {
```
AurelienFT, line 300: While you are refactoring I don't really understand what this functions are doing. Is not suppose to be the API that start the thread when it receive someone subscribing ? And the modules only broadcast ? 

So in your module you just need to know about a tokio sender and the ws related stuff is contained in API.

massa-node/base_config/openrpc.json
```diff
@@ -1125,7 +1125,33 @@
...
+                }
+            ],
+            "params": [],
+            "result": {
+                "schema": {
+                    "$ref": "#/components/schemas/Operation"
+                },
+                "name": "Operation"
+            },
+            "name": "subscribe_new_filled_blocks",
```
AurelienFT, line 1152: I think the name isn't good here.

AurelienFT (commented):
Good job just a few concern about code splitting.

massa-node/base_config/config.toml
aoudiamoncef: Yes, I knew that you'll give us a good one

massa-ws/src/lib.rs
aoudiamoncef: IMHO, this way is used to share an entire dependency. But in our case may be a `pub use` could be used.

AurelienFT: I don't think is necessary to share a whole module yes.

massa-protocol-worker/src/protocol_worker.rs
aoudiamoncef: this call will be replaced by: 

```rust
/// Brodcast the stream(sender) content via a WebSocket
pub fn broadcast_via_ws<T: Serialize + Send + Clone + 'static>(
    sender: Sender<T>,
    mut sink: SubscriptionSink,
) {
    let rx = BroadcastStream::new(sender.subscribe());
    tokio::spawn(async move {
        match sink.pipe_from_try_stream(rx).await {
            SubscriptionClosed::Success => {
                sink.close(SubscriptionClosed::Success);
            }
            SubscriptionClosed::RemotePeerAborted => (),
            SubscriptionClosed::Failed(err) => {
                sink.close(err);
            }
        };
    });
}
```

aoudiamoncef: If we do this, the `massa-ws` module will be removed.

massa-consensus-worker/src/controller.rs
aoudiamoncef: May be the module `massa-ws` could be removed. see the comment above please

massa-node/base_config/openrpc.json
aoudiamoncef: done

massa-protocol-worker/src/protocol_worker.rs
AurelienFT: I'm saying that `broadcast_via_ws` is already a function and making the function `subscribe_new_operations` that only have one line of code seems to be an overhead.

massa-consensus-worker/src/controller.rs
AurelienFT: It's not related to the comment above. It's about using the websocket things only in API and only have sender in the other modules.

massa-node/base_config/config.toml
aoudiamoncef: done

massa-ws/src/lib.rs
aoudiamoncef: done

massa-protocol-worker/src/protocol_worker.rs
aoudiamoncef: done

massa-consensus-worker/src/controller.rs
aoudiamoncef: done

massa-node/base_config/openrpc.json
```diff
@@ -1125,7 +1125,33 @@
...
+            ],
+            "params": [],
+            "result": {
+                "schema": {
+                    "$ref": "#/components/schemas/Operation"
+                },
+                "name": "Operation"
+            },
+            "name": "subscribe_new_operations",
+            "summary": "New produced operations",
```
AurelienFT, line 1153:
```suggestion
            "summary": "Subscribe to new received operations",
```

massa-node/base_config/openrpc.json
```diff
@@ -1234,6 +1260,42 @@
...
+            ],
+            "result": {
+                "schema": {
+                    "type": "boolean"
+                },
+                "name": "unsubscribe result",
+                "description": "unsubscribe success message"
+            },
+            "name": "unsubscribe_new_operations",
+            "summary": "Unsubscribe from new produced operations",
```
AurelienFT, line 1297:
```suggestion
            "summary": "Unsubscribe from new received operations",
```

massa-node/base_config/openrpc.json
```diff
@@ -1234,6 +1260,42 @@
...
+            "description": "Unsubscribe from new produced operations."
```
AurelienFT, line 1298:
```suggestion
            "description": "Unsubscribe from new received operations."
```

massa-node/base_config/openrpc.json
```diff
@@ -1125,7 +1125,33 @@
...
+            "params": [],
+            "result": {
+                "schema": {
+                    "$ref": "#/components/schemas/Operation"
+                },
+                "name": "Operation"
+            },
+            "name": "subscribe_new_operations",
+            "summary": "New produced operations",
+            "description": "New produced operations."
```
AurelienFT, line 1154:
```suggestion
            "description": "New received operations."
```

massa-api/src/api_trait.rs
```diff
@@ -34,4 +34,12 @@ pub trait MassaApi {
 		item = FilledBlock
 	)]
     fn subscribe_new_filled_blocks(&self);
+
+    /// New produced operations.
```
AurelienFT, line 38:
```suggestion
    /// New received operations.
```

AurelienFT (commented):
Much better separation and organisation.

Do we also want to announce the operations that are send in the API of this node ? Because currently we only take the ones that come from the network but we don't take the one sent by user through the API.

I think this could be in a separate issue because it implies more concerns, can you just specify in the description of openrpc that we don't broadcast op send through the node API on this node. And open an issue to do it later on.

Thank you

AurelienFT:
The tests are not compiling.

aoudiamoncef:
We could add existing

> Much better separation and organisation.
>
> Do we also want to announce the operations that are send in the API of this node ? Because currently we only take the ones that come from the network but we don't take the one sent by user through the API.

 IMHO, we could do It rapidly, we have to push new RAW operations but without any control...

> I think this could be in a separate issue because it implies more concerns, can you just specify in the description of openrpc that we don't broadcast op send through the node API on this node. And open an issue to do it later on.

W'll explicitly do it.

AurelienFT:
> IMHO, we could do It rapidly, we have to push new RAW operations but without any control...

That's the problem on my mind we don't have verifications and the node operator already have them if he catch them in the API. Let's see in the near future to merge this for now.

aoudiamoncef:
@AurelienFT , tests seems to be green. We could continue

AurelienFT (approved):

PR #3363, opened 2023-01-02
Add API V2 Http/WebSocket clients to Massa SDK

### LINKED ISSUE #3361: Missing Massa Http/WebSocket clients in Massa SDK
We have WS client [example](https://github.com/paritytech/jsonrpsee/blob/master/examples/examples/core_client.rs) in `jsonrpsee` repository.

aoudiamoncef:
  * [x] make tests compile
  * [x] make tests pass 

closes #3361

### DIFF
--- a/massa-client/base_config/config.toml
+++ b/massa-client/base_config/config.toml
@@ -6,8 +6,8 @@ timeout = 1000
 ip = "127.0.0.1"
 private_port = 33034
 public_port = 33035
-api_port = 33036
-[http]
+
+[client]
     # maximum size in bytes of a request
     max_request_body_size = 52428800
     # request timeout
@@ -22,3 +22,7 @@ api_port = 33036
     max_log_length = 4096
     # custom headers passed to the server with every request (default is empty).
     headers = []
+
+    [client.http]
+        # whether to enable HTTP.
+        enabled = true
--- a/massa-client/src/settings.rs
+++ b/massa-client/src/settings.rs
@@ -24,6 +24,7 @@ pub struct DefaultNode {
     pub ip: IpAddr,
     pub private_port: u16,
     pub public_port: u16,
+    pub api_port: u16,
 }
 
 /// Http Client settings.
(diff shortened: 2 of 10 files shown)


massa-client/base_config/config.toml
```diff
@@ -6,8 +6,8 @@ timeout = 1000
 ip = "127.0.0.1"
 private_port = 33034
 public_port = 33035
-api_port = 33036
```
AurelienFT, line 9: Where is defined the port for the net API now ?

massa-client/src/settings.rs
```diff
@@ -26,17 +26,26 @@ pub struct DefaultNode {
...
     pub request_timeout: MassaTime,
     pub max_concurrent_requests: usize,
     pub certificate_store: String,
     pub id_kind: String,
     pub max_log_length: u32,
     pub headers: Vec<(String, String)>,
+    pub http: HttpSettings,
+}
+
+///TODO add WebSocket to CLI
```
AurelienFT, line 43: I don't think we want to do this. There is no use case to have a WS in a CLI and there is no CLI format for that

massa-consensus-exports/src/events.rs
```diff
@@ -3,4 +3,6 @@
 pub enum ConsensusEvent {
     /// probable desynchronization detected, need re-synchronization
     NeedSync,
+    /// Network is ended should be send after `end_timestamp`
+    Stop,
```
AurelienFT, line 7: I have updated testnet_19, if you merge it on your branch we shouldn't see this changes anymore.

massa-client/base_config/config.toml
aoudiamoncef: We don't integrate the new API in `massa-client` yet. 

https://github.com/massalabs/massa/blob/3751e18022f7c29f7ce6b292e4d146921ecd563d/massa-node/base_config/config.toml#L15

massa-client/src/settings.rs
aoudiamoncef: May be in headless CLI(non interactive) it could be integrated, but It's not a common usecase so, it's not urgent at all

massa-consensus-exports/src/events.rs
aoudiamoncef: done

AurelienFT (approved):

AurelienFT:
Nice job ! 🎉

PR #3364, opened 2023-01-02
Add slot.period/parents.is_empty() checks to BlockHeader deser

Ben-PH:
  * [ ] make tests compile
  * [ ] make tests pass 


massa-models/src/block.rs
```diff
@@ -933,7 +947,7 @@ mod test {
         // create block header
         let orig_header = BlockHeader::new_wrapped(
             BlockHeader {
-                slot: Slot::new(1, 1),
+                slot: Slot::new(0, 1),
```
Ben-PH, line 950: Does this change correct an error? my current understanding is that a genesis block (i.e. the first block in a thread) has period of 0

AurelienFT: This makes the test correct. Before the test wasn't testing the right behavior.

Ben-PH:
Not exactly sure how to go about "check that the endorsements all target the parent block in the same thread (if not genesis)"

a block header has a vector of block ids, but not sure how to identify:

a) which one is on the same thread
b) check if it's actually that blocks parents.

my initial naive attempt was this (`self` is a `BlockHeader`):
```rust
            let parent_id = self.parents[self.slot.thread as usize];
            for endo in self.endorsements.iter() {
                if endo.content.endorsed_block != parent_id {
                    return Err("Invariant broken: endorsement doesn't match parent".into());
                }
            }
```

AurelienFT:
> Not exactly sure how to go about "check that the endorsements all target the parent block in the same thread (if not genesis)"
> 
> a block header has a vector of block ids, but not sure how to identify:
> 
> a) which one is on the same thread b) check if it's actually that blocks parents.
> 
> my initial naive attempt was this (`self` is a `BlockHeader`):
> 
> ```rust
>             let parent_id = self.parents[self.slot.thread as usize];
>             for endo in self.endorsements.iter() {
>                 if endo.content.endorsed_block != parent_id {
>                     return Err("Invariant broken: endorsement doesn't match parent".into());
>                 }
>             }
> ```

The same thread you can check by using the slot thread as an index as you made it's great.

They are parents because they are in the parent field. If it appears in the future that this blocks have parents that are invalid (too old block, incompatibles etc..) this will be check by consensus after.

massa-models/src/block.rs
```diff
@@ -759,28 +835,54 @@ impl Deserializer<BlockHeader> for BlockHeaderDeserializer {
...
+
+                    if endo.verify_signature().is_err() {
+                        Err(nom::Err::Failure(ContextError::add_context(
+                            rest,
+                            "Endorsement signature invalid",
+                            ParseError::from_error_kind(rest, nom::error::ErrorKind::Fail),
+                        )))
+                    } else {
+                        Ok((rest, endo))
+                    }
```
Ben-PH, line 856: Would we prefer to add signature checking to the `deserialize_with` implementation?

massa-models/src/block.rs
```diff
@@ -759,28 +835,54 @@ impl Deserializer<BlockHeader> for BlockHeaderDeserializer {
...
+            slot,
+            parents,
+            operation_merkle_root,
+            endorsements,
+        };
+
+        {
+            #[cfg(any(test, feature = "testing"))]
+            assert!(header.assert_invariants().is_ok());
+        }
```
Ben-PH, line 883: Thinking behind this, is that asserting invariants within this function makes for much simpler code, and is less prone to errors/bugs. By making this available in any environment run with the testing flag, it tests more thoroughly for errors in the deserialiser. 

There are the obvious drawbacks of code that is somewhat redundant.

shall we keep it around?

massa-models/src/block.rs
```diff
@@ -759,28 +835,54 @@ impl Deserializer<BlockHeader> for BlockHeaderDeserializer {
...
             length_count::<&[u8], Wrapped<Endorsement, EndorsementId>, u32, E, _, _>(
                 context("Failed length deserialization", |input| {
                     self.length_endorsements_deserializer.deserialize(input)
                 }),
                 context("Failed endorsement deserialization", |input| {
-                    endorsement_deserializer.deserialize_with(&self.endorsement_serializer, input)
+                    let (rest, endo) = endorsement_deserializer
+                        .deserialize_with(&self.endorsement_serializer, input)?;
+
+                    if endo.verify_signature().is_err() {
```
AurelienFT, line 848: After a discussion with @damip we remembered that we don't want to verify the signatures on the objects when in deserialization because we can receive objects multiple times sent by different peers and so we want to avoid calculating signatures for objects that we will drop few seconds after.

We prefer calculate the signatures after verifying uniqueness  [example with block header](https://github.com/massalabs/massa/blob/main/massa-protocol-worker/src/protocol_worker.rs#L839).

Ben-PH: so we don't want this in the block-header deserealization?

```rust
                    if endo.verify_signature().is_err() {
                        Err(nom::Err::Failure(ContextError::add_context(
                            rest,
                            "Endorsement signature invalid",
                            ParseError::from_error_kind(rest, nom::error::ErrorKind::Fail),
                        )))
                    } else {
                        Ok((rest, endo))
                    }
```

just return the the result of `deserialize_with`?

AurelienFT: We don't want to verify any signatures in the deserializers. We want to do it in others parts of the code to avoid calculate signatures of possible duplicate objects that we receive in the network

Ben-PH: Where should I check for signatures in that case?

massa-models/src/block.rs
```diff
@@ -504,6 +505,59 @@ pub struct BlockHeader {
...
+                if endo.content.index >= ENDORSEMENT_COUNT {
+                    return Err::<u32, Box<dyn std::error::Error>>(
+                        "Invariant broken: endorsement index out of bounds".into(),
+                    );
+                }
+                Ok(endo.content.index)
+            })
+            .try_collect::<std::collections::HashSet<_>>()?
+            .len()
+            == self.endorsements.len();
```
Ben-PH, line 555: There has to be a better way, such as an `try_into_set` or a `try_map`, but nothing jumped out at me at a quick glance

AurelienFT (commented):
Nice one. It really add more safety to the block header deserializer. Well done. You mentioned me a test that but the CI is running well, no ?

massa-models/src/block.rs
```diff
@@ -715,77 +784,130 @@ impl Deserializer<BlockHeader> for BlockHeaderDeserializer {
...
-                                tag(&[1]),
-                                count(
-                                    context("Failed block_id deserialization", |input| {
-                                        self.hash_deserializer
-                                            .deserialize(input)
-                                            .map(|(rest, hash)| (rest, BlockId(hash)))
-                                    }),
-                                    self.thread_count as usize,
-                                ),
+            context("Failed BlockHeader deserialization", |input| {
```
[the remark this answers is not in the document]
Ben-PH: Perhaps if I went into this already competent/confidant with noms patterns, I would have kept the tuple around. In the process of understanding it, I found this deconstruction a bit more straight forward, particularly given that we are validating the first two elements of the tuple before completing it in the merkle deser.

Personally, I find this a bit more readable, but I don't feel strongly about it.

massa-models/src/block.rs
```diff
@@ -504,6 +506,65 @@ pub struct BlockHeader {
...
+                endo.verify_signature()?;
+                // ..and that they are within range
+                if endo.content.index >= ENDORSEMENT_COUNT {
+                    return Err::<u32, Box<dyn std::error::Error>>(
+                        "Invariant broken: endorsement index out of bounds".into(),
+                    );
+                }
+                Ok(endo.content.index)
+            })
+            .try_collect::<std::collections::HashSet<_>>();
```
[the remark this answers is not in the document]
Ben-PH: > Nice one. It really add more safety to the block header deserializer. Well done. You mentioned me a test that but the CI is running well, no ?

It's the doc-tests that are failing.

There is a "bug" that I mention [here](https://github.com/massalabs/massa/pull/3364#issuecomment-1373799739) but i forgot to add the patch that I was talking about.

My analysis is that the deserializer will always assume there are `THREAD_COUNT` parents to deserialize, even if the actual count is different. Everything still fails, just for the wrong reasons.

Ben-PH:
> Nice one. It really add more safety to the block header deserializer. Well done. You mentioned me a test that but the CI is running well, no ?

It's the doc-tests that are failing.

There is a "bug" that I mention [here](https://github.com/massalabs/massa/pull/3364#issuecomment-1373799739) but i forgot to add the patch that I was talking about.

My analysis is that the deserializer will always assume there are `THREAD_COUNT` parents to deserialize, even if the actual count is different. Everything still fails, just for the wrong reasons.

massa-models/src/block.rs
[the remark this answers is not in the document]
Ben-PH: replaced with

```rust
        // assert that the endorsement indexes are all unique...
        let mut set = HashSet::new();
        for endo in self.endorsements.iter() {
            // ...and check signatures + invariants while at it
            endo.check_invariants()?;

            if !set.insert(endo.content.index) {
                return Err("Endorsement duplicate index found".into());
            }
        }
        Ok(())

```

massa-models/src/block.rs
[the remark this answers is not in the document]
AurelienFT: I was speaking about tuple with the two first elements. I think about the readability it's slightly more readable yes but the `tuple` combinator link the two parser of nom and allow nom to make some memory consumption optimizations etc.. It's certainly more optimized and provide more safety.

I believe it's the goal of nom https://github.com/rust-bakery/nom#streaming-formats

But it's clearly debatable I invoke @Eitu33 and @sydhds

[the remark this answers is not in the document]
Ben-PH: Hadn't considered the optimisations aspect.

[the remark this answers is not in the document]
Ben-PH: I've updated to use `tuple`. If we decide to revert, we can rebase and discord commit f414a23

Ben-PH:
I think Ive gotten to the bottom of why I'm getting doc test failings, and why not on CI

First, it only fails with the `testing` feature enabled. This introduces a signature verification of endorsements when deserialising them, which is the source of the error.

The cause of verification failing appears to be the choice of endorsement deserializer when making an endorsement verifiable, putting that endorsement into a block header, then making that block header verifiable.

```rust
/// Deserializer for `BlockHeader`
pub struct BlockHeaderDeserializer {
    slot_deserializer: SlotDeserializer,
    endorsement_serializer: EndorsementSerializer,
    length_endorsements_deserializer: U32VarIntDeserializer,
    hash_deserializer: HashDeserializer,
    thread_count: u8,
    endorsement_count: u32,
}
```

So the signed endorsement using a LW serializer later gets deserialized using the non-LW deserializer, it would seem.

If the endorsements are verifiable with a serializer that is consistent with the serializer for the block-header, there are no issues.

Ben-PH:
commit [a445c49](https://github.com/massalabs/massa/pull/3364/commits/a445c49a10bf6f370f7d6253767fc9de4b41e066) fixes the doc-tests.

The error seems to be inconsistency in the deserializers used (block header deser doesn't use the LW version.)

It only gets triggered when it's tested with the "testing" feature, as it adds signature verification to the deser process. I tried adding `doctest` to the `cfg(any(...`, but that didn't activate the checks, it would seem.

AurelienFT (approved):
Nice one ! Good job 🎉

Issue #3166, opened 2022-10-18
Client: avoid showing/using secret key and wallet password if we don't need to


sebastien-forestier:
Currently, when launching the client, it asks for the wallet password whereas there are plenty of use cases where you don't need the wallet (`get_status`) or the password (`wallet_info` to see my balance).

We should:
- not ask the wallet password until we want to sign something
- not show the secret key unless we really need it (for instance we could ask it through a get_secret_key(address)), and do not show it in wallet_info, 
- not show the public key either (use get_public_key(address) if you need it) as it can be confused with address
- `node_add_staking_secret_keys` secret key handling should be done automatically, and DO NOT MAKE THE USER COPY/PASTE THE SECRET KEY, for instance with `node_start_staking(address)`.

PR #3365, opened 2023-01-03
Added new commands (e.g. node_start_staking with address)

Leo-Besancon:
General checklist:

  * [x] make tests compile
  * [x] make tests pass 

Issue #3166 : 

- [x] not ask the wallet password until we want to sign something
- [x] not show the secret key unless we really need it (for instance we could ask it through a get_secret_key(address)), and do not show it in wallet_info,
- [x] not show the public key either (use get_public_key(address) if you need it) as it can be confused with address
- [x] node_add_staking_secret_keys secret key handling should be done automatically, and DO NOT MAKE THE USER COPY/PASTE THE SECRET KEY, for instance with node_start_staking(address).

### DIFF
--- a/.github/workflows/cd.yml
+++ b/.github/workflows/cd.yml
@@ -41,7 +41,7 @@ jobs:
             platform: arm64
     runs-on: ${{ matrix.os }}
     steps:
-      - uses: actions/checkout@v2
+      - uses: actions/checkout@v3
         with:
           submodules: "recursive"
       - uses: actions-rs/toolchain@v1
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -153,7 +153,7 @@ jobs:
       - name: "Use the current branch as dependency for gas-calibration"
         run: > 
           cd gas-calibration && 
-          sed -i 's/main/${{ github.head_ref || github.ref_name }}/g' Cargo.toml && 
+          sed -i 's!main!${{ github.head_ref || github.ref_name }}!g' Cargo.toml && 
           sed -i 's!massa-sc-runtime = { git = "https://github.com/massalabs/massa-sc-runtime", features = \["gas_calibration"\] }!'"$(cat ../massa-execution-worker/Cargo.toml | grep massa-sc-runtime | sed 's!/!\\/!g' | sed 's!}!, features = ["gas_calibration"]}!g' )"'!g' Cargo.toml
       - name: "Launch gas-calibration with one SC per ABI in massa-as-sdk to see if there is an ABI missing"
         run: >
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -21,7 +21,7 @@ jobs:
   sanity:
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v2
+      - uses: actions/checkout@v3
         with:
           submodules: "recursive"
       - uses: actions-rs/toolchain@v1
@@ -33,7 +33,7 @@ jobs:
       - uses: Swatinem/rust-cache@v2
         with:
           shared-key: "sanity"
-          save-if: ${{ github.ref == 'refs/pull/3358/merge' }}
+          save-if: ${{ github.ref_name == 'testnet_19' }}
       - uses: actions-rs/cargo@v1
         with:
           command: fmt
@@ -45,7 +45,7 @@ jobs:
     runs-on: ubuntu-latest
     continue-on-error: true
     steps:
-      - uses: actions/checkout@v2
+      - uses: actions/checkout@v3
         with:
           submodules: "recursive"
       - uses: actions-rs/toolchain@v1
@@ -55,7 +55,7 @@ jobs:
       - uses: Swatinem/rust-cache@v2
         with:
           shared-key: "check"
-          save-if: ${{ github.ref == 'refs/pull/3358/merge' }}
+          save-if: ${{ github.ref_name == 'testnet_19' }}
       - uses: actions-rs/cargo@v1
         with:
           command: check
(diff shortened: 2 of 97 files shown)


aoudiamoncef:
Hi 

```
error: Found argument '-p' which wasn't expected, or isn't valid in this context
```
May be we should keep this, if the user would add the password by default as always

Leo-Besancon:
> May be we should keep this, if the user would add the password by default as always

@aoudiamoncef: @Ben-PH mentionned using the -p flag may leak your password (e.g. with htop i believe, showing what command line was used to launch a process).

I can add the flag back if we'd rather have this possibility.

aoudiamoncef:
> > May be we should keep this, if the user would add the password by default as always
> 
> @aoudiamoncef: @Ben-PH mentionned using the -p flag may leak your password (e.g. with htop i believe, showing what command line was used to launch a process).
> 
> I can add the flag back if we'd rather have this possibility.

In non interactive mode, we have ton inject the password, is it handled in the new implementation ?

Leo-Besancon:
> In non interactive mode, we have ton inject the password, is it handled in the new implementation ?

If you run a command that needs the password in non interactive mode, it will prompt it immediately. Which may be break things if some scripts rely on the previous behaviour.

Ben-PH:
> > May be we should keep this, if the user would add the password by default as always
> 
> @aoudiamoncef: @Ben-PH mentionned using the -p flag may leak your password (e.g. with htop i believe, showing what command line was used to launch a process).
> 
> I can add the flag back if we'd rather have this possibility.

Do you think the following requirements for providing password be reasonable?

a) it's manually entered by the user, or interfaces with a trusted password manager (I think that would be really difficult outside a web browser, no?)
c) other security measures such as re-writing the memory containing the plain-text password before that variable goes out of scope (should be provided by libraries that handle password input) .

Ben-PH:
...with that said, the security flag I raised, and the discussion around what would be reasonable, would probably best be addressed and driven by someone dedicated to the security role, both the engineering and threat analysis side of thing.

aoudiamoncef:
IMHO, we will do what others do, in non interactive mode, we could use env variables ?

Leo-Besancon:
> IMHO, we will do what others do, in non interactive mode, we could use env variables ? we should keep the non interactive mode working

Good idea, I'll try and implement these changes then.

AurelienFT:
> IMHO, we will do what others do, in non interactive mode, we could use env variables ? we should keep the non interactive mode working

Why we change from parameters to env variable ? If we problem is having the password in the shell, users can use bash script that return the password as argument instead of the password himself.

We need to have a strong reason to break compatibility on this. Do we have one ?

AurelienFT:
Ok to have both nice improvement 👍

Leo-Besancon:
@aoudiamoncef Updated:
- -p / --pwd is restored
- We check MASSA_CLIENT_PASSWORD env variable (where should we document this?)
- Output for get_public_key and get_secret_key updated.

aoudiamoncef:
> 



> @aoudiamoncef Updated:
> 
> * -p / --pwd is restored
Nice
> * We check MASSA_CLIENT_PASSWORD env variable (where should we document this?)
massalabs/massa-docs project:  https://docs.massa.net/en/latest/testnet/wallet.html
> * Output for get_public_key and get_secret_key updated.
Well done

Ben-PH:
> We could keep both, if no param provided, we try to read `MASSA_CLIENT_PASSWORD` for example. It's a nice to have.
> 
> The security of the password outside of the CLI is the responsibility of the user.

I agree, with the caveat that we should aim to provide an environment that makes handling that responsibility as easy as possible, and directing them that way as much as possible. Perhaps we should put work-flows that call for manual user input front-and-center of our documentation/tutorials, with foot-notes/margin-notes/etc. directing people to the options that have the password in the CLI/env-variable/config file/etc.?

take for example windows forcing restarts on users to keep their OS updated and secure. The better approach would be to have an OS that can keep itself secure without having to restart. Failing that, to have the default user experience be to not keep the OS up to date in exchange for the convenience of not having updates forced on them, then too many users would have insecure systems, and Windows would have a security reputation that is ~~terrible~~ much worse than what it already is. 

This might be a good topic to start a discussion with security team candidates to look at their "security oriented mindset". Could go down a rabbit-hole of threat analysis and response, trade-offs between security and other concerns, etc.

aoudiamoncef:
@Leo-Besancon, please could you sync this branch ?

Leo-Besancon:
@aoudiamoncef Conflict resolved!

aoudiamoncef:
We could improve warning messages to be more accurate:

actual: 
```rust
client_warning!("do not share your key")
````
proposal:
```rust
client_warning!("do not share your secret key")
````

aoudiamoncef:
- After testing the CLI and as `get_public_key` and `get_secret_key` comes from wallet not an RPC call. add the prefix `wallet_` could be better. 
- May be invert "password_not_need" to "require_password" or "protected" and avoid negation is better.

Leo-Besancon:
@aoudiamoncef 

> * After testing the CLI and as `get_public_key` and `get_secret_key` comes from wallet not an RPC call. add the prefix `wallet_` could be better.

I agree!

> * May be invert "password_not_need" to "require_password" or "protected" and avoid negation is better.

Ok, this is what I add at the begining.
The problem I had was that if the dev forgets to put the "require_password" tag, the user will have an error (because we did not get the password from him and want to access the wallet).
Right now, if the dev forgets the tag, then the user is just unnecessarily asked for their password.

Which do you think is better?

aoudiamoncef:
> @aoudiamoncef
> 
> > * After testing the CLI and as `get_public_key` and `get_secret_key` comes from wallet not an RPC call. add the prefix `wallet_` could be better.
> 
> I agree!
> 
> > * May be invert "password_not_need" to "require_password" or "protected" and avoid negation is better.
> 
> Ok, this is what I had at the begining. The problem I had was that if the dev forgets to put the "require_password" tag, the user will have an error (because we did not get the password from him and want to access the wallet). Right now, if the dev forgets the tag, then the user is just unnecessarily asked for their password.
> 
> Which do you think is better?

I see, it's good for me, good job 👍🏻

aoudiamoncef (approved):

PR #3378, opened 2023-01-06
Fix bootstrap test tools

Eitu33:


massa-bootstrap/src/tests/tools.rs
```diff
@@ -437,6 +439,11 @@ pub fn get_boot_state() -> BootstrapableGraph {
         .unwrap();
 
     assert_eq_bootstrap_graph(&bootstrapable_graph_deserialized, &boot_graph);
+    for blk in bootstrapable_graph_deserialized.final_blocks.iter() {
+        for endo in blk.block.content.header.content.endorsements.iter() {
+            endo.verify_signature().unwrap();
+        }
+    }
```
Ben-PH, line 446: This won't be needed I don't think, as I'll be adding it to the deserialiser that only runts under `#[cfg(any(test, feature = "testing"))]`

Eitu33: Removed 👍

PR #3383, opened 2023-01-06
Add execute from final state for read only SC calls

### LINKED ISSUE #2950: rpc call `execute_read_only_call` to be extended with an optional json payload for returning final state
Smart contracts might often have code like this:

```
export function getSomeData(address: string): string {
  const data = Storage.get("KEY");
  generateEvent(data);

  return data;
}
```

If this read-only method is being called using `execute_read_only_call`, as discussed with Damir, the state reading will be based on a simulated execution at the time of call giving us a transient state. 

Since transient states are subject to change, we need to have an rpc call `execute_read_only_call` targeting final states and thus returning finality state data. I suggest adding a new param to the json rpc payload to force the rpc call to return only final slot storage data, if not set, same as now, transient.

aoudiamoncef:
  * [x] make tests compile
  * [x] make tests pass 

closes #2950

### DIFF
--- a/massa-execution-exports/src/types.rs
+++ b/massa-execution-exports/src/types.rs
@@ -69,6 +69,12 @@ pub struct ReadOnlyExecutionRequest {
     pub call_stack: Vec<ExecutionStackElement>,
     /// Target of the request
     pub target: ReadOnlyExecutionTarget,
+    /// optional event status
+    ///
+    /// Some(true) means final
+    /// Some(false) means candidate
+    /// None means final _and_ candidate
+    pub is_final: Option<bool>,
 }
 
 /// structure describing different possible targets of a read-only execution request
@@ -101,6 +107,12 @@ pub struct ReadOnlyCallRequest {
     pub target_func: String,
     /// Parameter to pass to the target function
     pub parameter: String,
+    /// optional event status
+    ///
+    /// Some(true) means final
+    /// Some(false) means candidate
+    /// None means final _and_ candidate
+    pub is_final: Option<bool>,
 }
 
 /// Structure describing an element of the execution stack.
--- a/massa-api/src/public.rs
+++ b/massa-api/src/public.rs
@@ -127,6 +127,7 @@ impl MassaRpcServer for API<Public> {
             address,
             bytecode,
             operation_datastore,
+            is_final,
         } in reqs
         {
             let address = address.unwrap_or_else(|| {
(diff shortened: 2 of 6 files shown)


aoudiamoncef:
Hello @AurelienFT @Eitu33,

Could I test this in sandbox mode ? (I tried without success)

AurelienFT:
> Hello @AurelienFT @Eitu33,
> 
> Could I test this in sandbox mode ? (I tried without success)

Yes you should be able to test it sandbox, you don't even need to have a staker as it's read_only execution (if you want to test read_only call you need a staker to produce blocks that will save an SC that you can read call later)

What's the current issue you have ?

aoudiamoncef:
@AurelienFT I tired to find the easiest way to test this feature but I struggled with the mocks. 

- When I two `execute_readonly_request` with a thread sleep between the calls, the slot never change. 
- When I print `controller.get_stats()` I see that the slot is affected by the time and it moves. 

I don't know why `execute_readonly_request` doesn't returns the active or final slots ?

The good news is that the feature seems to work when I make real JsonRPC calls.

AurelienFT:
@aoudiamoncef I will takeover the testing part.


> Is it mandatory to add `pub struct StateChanges { `
to the response ? or the slot field is enough ?

Yes that's part of #2950. 

> or move it out the execution module ?

It's already in another module `massa-final-state` and returned by the execution function you can send directly this structure to the client in the API.

massa-execution-exports/src/types.rs
```diff
@@ -69,6 +69,12 @@ pub struct ReadOnlyExecutionRequest {
     pub call_stack: Vec<ExecutionStackElement>,
     /// Target of the request
     pub target: ReadOnlyExecutionTarget,
+    /// optional event status
+    ///
+    /// Some(true) means final
+    /// Some(false) means candidate
+    /// None means final _and_ candidate
```
AurelienFT, line 76: I don't think it's possible to make the execution for final and candidate, this could be only a boolean.

aoudiamoncef: IMHO, we should be more explicit about this parameter. 

```rust
    /// Some(true) means start execution from final state
    /// Some(false) or None means start execution from active state
```

AurelienFT: If it's only true or false we don't need an option.

aoudiamoncef: I totally agree, we don't have to add a third state here. 

done in 0e537fca9c63fdb9101fe9b69f5ece480fc86135 as it's a part 2 of this PR. 

Sorry for miss.

PR #3384, opened 2023-01-06
Testnet 19 bph wrapper rename

Ben-PH:
  * [x] make tests compile
  * [x] make tests pass 

### DIFF
--- a/massa-models/src/secure_share.rs
+++ b/massa-models/src/secure_share.rs
@@ -48,7 +48,7 @@ pub trait Id {
     fn get_hash(&self) -> &Hash;
 }
 
-/// Trait that define a structure that can be wrapped.
+/// Trait that define a structure that can be signed for secure sharing.
 pub trait SecureShareContent
 where
     Self: Sized + Display,
@@ -78,7 +78,7 @@ where
         })
     }
 
-    /// Serialize the wrapped structure
+    /// Serialize the secured structure
     fn serialize(
         signature: &Signature,
         creator_public_key: &PublicKey,
@@ -91,7 +91,7 @@ where
         Ok(())
     }
 
-    /// Deserialize the wrapped structure
+    /// Deserialize the secured structure
     fn deserialize<
         'a,
         E: ParseError<&'a [u8]> + ContextError<&'a [u8]>,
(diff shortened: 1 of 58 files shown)


massa-models/src/secure_share.rs
```diff
@@ -45,16 +49,17 @@ pub trait Id {
...
-pub trait WrappedContent
+pub trait SecureShareContent
 where
     Self: Sized + Display,
 {
-    /// Creates a wrapped version of the object
-    fn new_wrapped<SC: Serializer<Self>, U: Id>(
+    /// Using the provided key-pair, generates values the are used
+    /// by the trust-free data sharing protocols.
+    fn secure<Ser: Serializer<Self>, ID: Id>(
```
AurelienFT, line 58: Maybe rename it to `new_secure`

AurelienFT (commented):
Nice refactor. Just a little adjustment on my side and it would be perfect

massa-models/src/secure_share.rs
Ben-PH: I like it. I'll put that together on Monday.

Ben-PH:
@Leo-Besancon I thought you'd be a valuable review from the perspective that the new naming convention (and some doc-comment tweaks) relate closely to our lessons with damir about how signing and verification works.

Leo-Besancon:
The proposed change does make it clearer to understand the intent of Wrapped types.
Just to be sure, you changed 
```
creator_address: op.creator_address,
```
to: 
```
creator_address: op.content_creator_address,
```

in order to differentiate between the creator of the wrapped content (e.g. the Block producer) and the creator of the wrapper struct?

Ben-PH:
> in order to differentiate between the creator of the wrapped content (e.g. the Block producer) and the creator of the wrapper struct?

That wasn't the intent, and I can now see how this rename is a touch misleading. This value is determined by the keypair that is generated, which is Rng'd, and passed into the creation function. Perhaps `signature_origin_address`? or `signature_public_key_addr`?

I'm trying to go for something that gives an intuitive sense that this is a) a massa address, b) said address is the result of the same public key that is used as the input for generating the signature.

Leo-Besancon:
Understood, content_creator_address is fine then.

Leo-Besancon (approved):

massa-models/src/secure_share.rs
Ben-PH: I went with `new_verifiable`. I'm thinking "secure" is a touch redundant to the type name, and it's a chance to document the properties that make it `SecureShare`

Issue #2608, opened 2022-05-09
Need ipv6 bootstrap address


rykcod:
It's possible to create an offical ipv6 node? Because for ipv6 only user node, its impossible to bootstrap without add a community ipv6 node in the custom config.toml

damip:
We need to figure this out.

AurelienFT:
First step here is to test if it really don't work.

Leo-Besancon:
I did reproduce the issue. A client with IPv6 only interface can't connect to a bootstrap server using the IPv4 protocol (for example, see [this link](https://serverfault.com/questions/1102241/purpose-of-ipv4-mapped-ipv6-address).

This means that we should:
- [ ] Let our bootstrap servers expose both an IPv4 and IPv6 address
- [ ] Then, we can put both addresses in the bootstrap ip list
- [ ] Finally, filter the bootstrap ip list based on the available interfaces of the node

PR #3391, opened 2023-01-09
2608 need ipv6 bootstrap address

Leo-Besancon:
  * [x] make tests compile
  * [x] make tests pass 

From #2608:

* [x] Let our bootstrap servers expose both an IPv4 and IPv6 address
* [x] Then, we can put both addresses in the bootstrap ip list
* [x] Finally, filter the bootstrap ip list based on the available interfaces of the node

### DIFF
--- a/massa-bootstrap/src/client.rs
+++ b/massa-bootstrap/src/client.rs
@@ -441,6 +441,7 @@ pub async fn get_state(
     let mut filtered_bootstrap_list: Vec<(SocketAddr, PublicKey)> =
         bootstrap_config.bootstrap_list.clone();
 
+    // We filter the bootstrap list to keep only the ip addresses we are compatible with
     match &bootstrap_config.bootstrap_protocol {
         Some(s) if s.as_str() == "ipv4" => {
             filtered_bootstrap_list = filtered_bootstrap_list
@@ -456,7 +457,7 @@ pub async fn get_state(
                 .filter(|&(addr, _pubkey)| addr.is_ipv6())
                 .collect()
         }
-        _ => filtered_bootstrap_list = filtered_bootstrap_list.clone(),
+        _ => {}
     };
 
     // we are after genesis => bootstrap
@@ -467,19 +468,19 @@ pub async fn get_state(
         ));
     }
 
-    let mut shuffled_list = filtered_bootstrap_list.clone();
-    shuffled_list.shuffle(&mut StdRng::from_entropy());
+    filtered_bootstrap_list.shuffle(&mut StdRng::from_entropy());
 
-    // we only keep one IP for each Public key.
-    let shuffled_hashmap: HashMap<PublicKey, SocketAddr> = shuffled_list
+    // we remove the duplicated public keys (if a bootstrap server appears both with its IPv4 and IPv6 address)
+    let shuffled_hashmap: HashMap<PublicKey, SocketAddr> = filtered_bootstrap_list
         .clone()
         .into_iter()
         .map(|(addr, pub_key)| (pub_key, addr))
         .collect();
-    shuffled_list = shuffled_hashmap
+    let mut shuffled_list: Vec<(SocketAddr, PublicKey)> = shuffled_hashmap
         .iter()
         .map(|(&pub_key, &addr)| (addr, pub_key))
         .collect();
+
     shuffled_list.shuffle(&mut StdRng::from_entropy());
 
     let mut next_bootstrap_message: BootstrapClientMessage =
--- a/massa-bootstrap/src/client.rs
+++ b/massa-bootstrap/src/client.rs
@@ -1,5 +1,5 @@
 use humantime::format_duration;
-use std::{collections::HashMap, net::SocketAddr, sync::Arc, time::Duration};
+use std::{collections::HashSet, net::SocketAddr, sync::Arc, time::Duration};
 
 use massa_final_state::FinalState;
 use massa_logging::massa_trace;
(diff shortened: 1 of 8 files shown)


massa-bootstrap/src/settings.rs
```diff
@@ -10,6 +10,8 @@ use std::{net::SocketAddr, path::PathBuf};
 pub struct BootstrapConfig {
     /// Ip address of our bootstrap nodes and their public key.
     pub bootstrap_list: Vec<(SocketAddr, NodeId)>,
+    /// IP version filter for bootstrap list, targeting "ipv4", "ipv6" or "both".
```
aoudiamoncef, line 13: we should say that "both" is the default value

massa-node/src/tests/config.toml
```diff
@@ -70,8 +70,10 @@
         ["149.202.86.103:31245", "5GcSNukkKePWpNSjx9STyoEZniJAN4U4EUzdsQyqhuP3WYf6nj"],
         ["149.202.89.125:31245", "5wDwi2GYPniGLzpDfKjXJrmHV3p1rLRmm4bQ9TUWNVkpYmd4Zm"],
         ["158.69.120.215:31245", "5QbsTjSoKzYc8uBbwPCap392CoMQfZ2jviyq492LZPpijctb9c"],
-        ["158.69.23.120:31245", "8139kbee951YJdwK99odM7e6V3eW7XShCfX5E2ovG3b9qxqqrq"]
+        ["158.69.23.120:31245", "8139kbee951YJdwK99odM7e6V3eW7XShCfX5E2ovG3b9qxqqrq"],,
```
aoudiamoncef, line 73: small typo issue here

massa-bootstrap/src/client.rs
```diff
@@ -437,15 +437,71 @@ pub async fn get_state(
         }
         return Ok(GlobalBootstrapState::new(final_state));
     }
+
+    let mut filtered_bootstrap_list = bootstrap_config.bootstrap_list.clone();
+
+    // We filter the bootstrap list to keep only the ip addresses we are compatible with
+    match &bootstrap_config.bootstrap_protocol {
+        Some(s) if s.as_str() == "ipv4" => {
```
aoudiamoncef, line 445: s -> ip_type

massa-bootstrap/src/client.rs
```diff
@@ -437,15 +437,71 @@ pub async fn get_state(
         }
         return Ok(GlobalBootstrapState::new(final_state));
     }
+
+    let mut filtered_bootstrap_list = bootstrap_config.bootstrap_list.clone();
+
+    // We filter the bootstrap list to keep only the ip addresses we are compatible with
+    match &bootstrap_config.bootstrap_protocol {
```
aoudiamoncef, line 444: you could avoid some repetition by using a private function

massa-bootstrap/src/client.rs
```diff
@@ -437,15 +437,71 @@ pub async fn get_state(
...
+
+            let new_bootstrap_list_len = filtered_bootstrap_list.len();
+
+            debug!(
+                "Filtered out {} IPv4 bootstrap addresses out of a total of {} bootstrap servers.",
+                prev_bootstrap_list_len as i32 - new_bootstrap_list_len as i32,
+                prev_bootstrap_list_len
+            );
+        }
+        _ => {}
```
aoudiamoncef, line 479: May be, we should be explicit here. "ipv4", "ipv6", "both" or thow an error

Leo-Besancon: Should I create a specific enum for it? So that if it is invalid it produces the error when it tries to deserialize the config instead.

aoudiamoncef: It seems to fit in enum, so i would say yes

massa-node/base_config/config.toml
```diff
@@ -204,13 +204,19 @@
...
+        ["[2001:41d0:602:21e4::]:31245", "N13Ykon8Zo73PTKMruLViMMtE2rEG646JQ4sCcee2DnopmVM3P5"],
     ]
+    # force the bootstrap protocol to use: "IPv4", "IPv6", or "Both". Defaults to using both protocols.
+    bootstrap_protocol = "Both"
     # path to the bootstrap whitelist file. This whitelist define IPs that can bootstrap on your node.
     bootstrap_whitelist_path = "base_config/bootstrap_whitelist.json"
     # path to the bootstrap blacklist file. This whitelist define IPs that will not be able to bootstrap on your node. This list is optional.
     bootstrap_blacklist_path = "base_config/bootstrap_blacklist.json"
-    # [optionnal] port on which to listen for incoming bootstrap requests
+    # [optional] port on which to listen for incoming bootstrap requests
```
aoudiamoncef, line 219: [optional] -> (optional) we could use this convention to avoid confusion with arrays

massa-bootstrap/src/client.rs
```diff
@@ -437,15 +464,35 @@ pub async fn get_state(
...
     massa_trace!("bootstrap.lib.get_state.init_from_others", {});
-    if bootstrap_config.bootstrap_list.is_empty() {
+    if filtered_bootstrap_list.is_empty() {
         return Err(BootstrapError::GeneralError(
             "no bootstrap nodes found in list".into(),
         ));
     }
-    let mut shuffled_list = bootstrap_config.bootstrap_list.clone();
+
+    filtered_bootstrap_list.shuffle(&mut StdRng::from_entropy());
```
AurelienFT, line 482: You don't need to shuffle it now because you already shuffle it at line 494

AurelienFT (commented):
Nice one. Good job !

massa-bootstrap/src/client.rs
Leo-Besancon: I decided to shuffle two times because:

- If I only shuffle after removing duplicates, then the IP protocol used for each bootstrap server will always be the same (same order as the bootstrap list), even if we should bootstrap on "Both" protocols.
- If I only shuffle before, then the order depends on the HashMap iterator (which is arbitrary and maybe not truly random).

Do you see another way to do so? Or maybe having the same protocol is not so bad?

AurelienFT: If I follow you here, if you shuffle it before the dedup, it will remove sometimes ipv4 and sometimes ipv6 as it's not in the order of the bootstrap list anymore ? Or there is something I miss ?

Leo-Besancon: Exactly! I think it's better to keep sometimes Ipv4 and sometimes Ipv6 if the bootstrap server exposes both.

AurelienFT: Ok then I think you could something like that to dedup : https://stackoverflow.com/a/47648303/10419533 that will avoid you the re-ordering of the hashmap and then the second shuffle and should optimize this.

Leo-Besancon: Good idea, done in latest commit.

AurelienFT (approved):
Nice work well done !

PR #3415, opened 2023-01-13
Broadcast local operations

### LINKED ISSUE #3355: Add node produced operations to `subscribe_new_operations`
In testnet 18, we introduced experimental support for WebSockets where `subscribe_new_operations` broadcast only received operations from other nodes. Produced operation from the node are missing.

aoudiamoncef:
  * [x] make tests compile
  * [x] make tests pass 

closes #3355

### DIFF
--- a/massa-protocol-worker/src/protocol_worker.rs
+++ b/massa-protocol-worker/src/protocol_worker.rs
@@ -511,8 +511,17 @@ impl ProtocolWorker {
                 self.checked_operations
                     .extend(operation_ids.iter().copied());
 
-                // Announce operations to active nodes not knowing about it.
                 let to_announce: Vec<OperationId> = operation_ids.iter().copied().collect();
+                // Broadcast operations to active sender(channel) subscribers.
+                if self.config.broadcast_enabled {
+                    for op_id in to_announce.clone() {
+                        if let Some(s_operation) = storage.read_operations().get(&op_id) {
+                            let _ = self.operation_sender.send(s_operation.content.clone());
+                        };
+                    }
+                }
+
+                // Announce operations to active nodes not knowing about it.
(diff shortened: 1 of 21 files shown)


massa-protocol-worker/src/protocol_worker.rs
```diff
@@ -511,8 +511,17 @@ impl ProtocolWorker {
                 self.checked_operations
                     .extend(operation_ids.iter().copied());
 
-                // Announce operations to active nodes not knowing about it.
                 let to_announce: Vec<OperationId> = operation_ids.iter().copied().collect();
+                // Broadcast operations to active sender(channel) subscribers.
+                if self.config.broadcast_enabled {
```
AurelienFT, line 516: In this function we receive operationids from the others nodes to let us know them and we ask their content if we don't already have them later on on the code.
So here you will fetch in the storage only ops that we already have and so has already been sent to the WS when the content has been retrieved.

Now that you want to broadcast all the operations you can retrieve the information for the `operation_pool` `add_operations` function ([code](https://github.com/massalabs/massa/blob/main/massa-pool-worker/src/operation_pool.rs#L101)). It's called by the `node_operations_from_node` where you had your previous send and the API when someone use `send_operations`

AurelienFT (changes requested):

massa-protocol-worker/src/protocol_worker.rs
aoudiamoncef: If I understand, we have to options:

1. Move the broadcast to add_operations in massa-pool-worker(I prefer this)
2. Broadcast from Public API.

aoudiamoncef: Done

AurelienFT (approved):
Nice one

PR #3416, opened 2023-01-13
Module caching

Eitu33:
  * [x] make tests compile
  * [x] make tests pass 


massa-execution-worker/src/execution.rs
```diff
@@ -70,6 +70,8 @@ pub(crate) struct ExecutionState {
     execution_interface: Box<dyn Interface>,
     // execution statistics
     stats_counter: ExecutionStatsCounter,
+    // cache of pre compiled sc modules
+    module_cache: Arc<RwLock<ModuleCache>>,
```
AurelienFT, line 74: When we talked about generics in the call the goal was to not have a dependency to wasmer and so have a generic module that wrap the wasmer module and maybe others in the future but don't have a type `ModuleCache`. I think the module cache should stay in Massa and use a `Module` defined in `massa-sc-runtime` this will avoid the `RwLock` and to pass this object to the execution.

AurelienFT: This would also be useful to have modules with differents VM/versions behind.

Eitu33: > I think the module cache should stay in Massa and use a Module defined in massa-sc-runtime this will avoid the RwLock and to pass this object to the execution.

I was expecting this would cause a misunderstanding, that is not the thought process behind the lock, that is needed for reasons that have nothing to do with this issue (the runtime env needing to keep a reference to the cache in order to properly run subsequent modules). 

> When we talked about generics in the call the goal was to not have a dependency to wasmer and so have a generic module that wrap the wasmer module and maybe others in the future but don't have a type ModuleCache.

> This would also be useful to have modules with differents VM/versions behind.

As for the comments regarding `ModuleCache` being not generic enough I do not agree with it, what `ModuleCache` does can whatever we want, we just need to keep its information in execution between multiple runtime calls, the rest should not be handled in the node code.

Eitu33: Updates will be made accordingly to what was decided with @AurelienFT & @sydhds, thought process behind the changes to be referenced here in the near future.

Eitu33:
@AurelienFT @sydhds 

So quite a lot changed since the above comment, TLDR is:
* cache is now a LRU in `massa-execution-worker`
* cache also contains the previous instance creation cost
* a module will not be saved in the cache if its instantiation failed
* `massa-sc-runtime` only takes a module as input
* `massa-sc-runtime` contains nothing related with the cache, it interacts with it through an interface function

massa-execution-worker/src/module_cache.rs
```diff
@@ -0,0 +1,58 @@
+use massa_execution_exports::ExecutionError;
+use massa_sc_runtime::{GasCosts, RuntimeModule};
+use schnellru::{ByLength, LruMap};
+use tracing::log::warn;
+
+pub struct ModuleCache {
+    gas_costs: GasCosts,
+    cache: LruMap<Vec<u8>, (RuntimeModule, u64)>,
```
sydhds, line 8: I think you should add some documentation for:
* Vec<u8>: non compiled bytecode
* u64: initial gas cost?

massa-execution-worker/src/worker.rs
```diff
@@ -86,7 +86,7 @@ impl ExecutionThread {
             let (req, resp_tx) = req_resp.into_request_sender_pair();
 
             // Acquire read access to the execution state and execute the read-only request
-            let outcome = self.execution_state.read().execute_readonly_request(req);
+            let outcome = self.execution_state.write().execute_readonly_request(req);
```
sydhds, line 89: Any reason for this change? Doc should be updated...

massa-execution-exports/src/settings.rs
```diff
@@ -57,6 +57,8 @@ pub struct ExecutionConfig {
     pub max_bytecode_size: u64,
     /// Max datastore value size
     pub max_datastore_value_size: u64,
+    /// Max size of the module cache
```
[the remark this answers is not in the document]
sydhds: + report this in config.toml as well

massa-execution-worker/src/context.rs
```diff
@@ -126,6 +127,9 @@ pub struct ExecutionContext {
 
     /// operation id that originally caused this execution (if any)
     pub origin_operation_id: Option<OperationId>,
+
+    // cache of pre compiled sc modules
```
sydhds, line 131: I'm picky but this is not "pre compiled", sc are compiled into assembly (like x86_64) by wasmer

massa-execution-worker/src/module_cache.rs
```diff
@@ -0,0 +1,58 @@
+use massa_execution_exports::ExecutionError;
```
sydhds, line 1: Add some doc here to explain how the cache is working and maybe some quick refresher for what is a LRU cache?

sydhds (changes requested):

massa-execution-worker/Cargo.toml
```diff
@@ -10,20 +10,22 @@ edition = "2021"
...
-criterion = {version = "0.4", optional = true}
+criterion = { version = "0.4", optional = true }
 parking_lot = { version = "0.12", features = ["deadlock_detection"] }
 tracing = "0.1"
 serde_json = "1.0"
 num = { version = "0.4", features = ["serde"] }
-tempfile = { version = "3.3", optional = true }    # use with gas_calibration feature
+schnellru = "0.2.0"
+# use with gas_calibration feature
+tempfile = { version = "3.3", optional = true }
```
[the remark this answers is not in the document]
Eitu33: Its declared in `[dev-dependencies]` but that does not cover special features

massa-execution-worker/src/module_cache.rs
```diff
@@ -0,0 +1,58 @@
...
+    /// If the module is not contained in the cache:
+    /// * create the module
+    /// * retrieve it
+    pub fn get_module(
+        &mut self,
+        bytecode: &[u8],
+        limit: u64,
+    ) -> Result<RuntimeModule, ExecutionError> {
+        if let Some((cached_module, init_cost)) = self.cache.get(bytecode) {
+            warn!("(CACHE) found");
```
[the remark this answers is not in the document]
Eitu33: Debug logs, removed

massa-execution-worker/src/module_cache.rs
Eitu33: Done

massa-execution-worker/src/module_cache.rs
Eitu33: Done in the `ModuleCache` description

massa-execution-worker/src/context.rs
Eitu33: It was "pre compiled" as in previously compiled not literal pre compilation, I can understand what that was misleading I updated it

massa-execution-worker/src/execution.rs
Eitu33: This has been taken care of, resolving

sydhds (approved):
Please do not forget to update massa-sc-runtime dependency before merge!

massa-execution-exports/Cargo.toml
```diff
@@ -18,7 +18,7 @@ massa_storage = { path = "../massa-storage" }
 massa_final_state = { path = "../massa-final-state" }
 massa_ledger_exports = { path = "../massa-ledger-exports", optional = true }
 parking_lot = { version = "0.12", features = ["deadlock_detection"], optional = true }
-massa-sc-runtime = { git = "https://github.com/massalabs/massa-sc-runtime" }
+massa-sc-runtime = { path = "../../massa-sc-runtime" }
```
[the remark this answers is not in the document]
Eitu33: used `testnet_19` runtime version

sydhds (approved):

sydhds (approved):

Eitu33:
TO THE MOOOOOON 🚀 

bors merge

Eitu33:
Called bors by mistake

PR #3417, opened 2023-01-13
Test Address versionning

Leo-Besancon:
Address::AddressV1 and Address::AddressV2 variants, each made with a different Hash algorithm.

/!\ The code compiles, but there should be some errors in the way we handle both versions!

  * [ ] make tests compile
  * [ ] make tests pass 

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2547,6 +2547,7 @@ dependencies = [
  "serde",
  "serde_json",
  "serial_test 0.10.0",
+ "sha2 0.10.6",
  "thiserror",
 ]
 
--- a/massa-async-pool/src/pool.rs
+++ b/massa-async-pool/src/pool.rs
@@ -9,7 +9,7 @@ use crate::{
     AsyncMessageDeserializer, AsyncMessageIdDeserializer, AsyncMessageIdSerializer,
     AsyncMessageSerializer, AsyncMessageTrigger,
 };
-use massa_hash::{Hash, HASH_SIZE_BYTES};
+use massa_hash::{Hash, HASHV1_SIZE_BYTES};
 use massa_ledger_exports::LedgerChanges;
 use massa_models::{slot::Slot, streaming_step::StreamingStep};
 use massa_serialization::{
@@ -24,7 +24,7 @@ use nom::{
 use std::collections::BTreeMap;
 use std::ops::Bound::{Excluded, Included, Unbounded};
 
-const ASYNC_POOL_HASH_INITIAL_BYTES: &[u8; 32] = &[0; HASH_SIZE_BYTES];
+const ASYNC_POOL_HASH_INITIAL_BYTES: &[u8; 32] = &[0; HASHV1_SIZE_BYTES];
 
 /// Represents a pool of sorted messages in a deterministic way.
 /// The final asynchronous pool is attached to the output of the latest final slot within the context of massa-final-state.
@@ -342,7 +342,7 @@ impl Deserializer<BTreeMap<AsyncMessageId, AsyncMessage>> for AsyncPoolDeseriali
 
 #[test]
 fn test_take_batch() {
-    use massa_hash::Hash;
+    /*use massa_hash::HashV2;
     use massa_models::{address::Address, amount::Amount, slot::Slot};
     use std::str::FromStr;
 
@@ -353,7 +353,7 @@ fn test_take_batch() {
         bootstrap_part_size: 100,
     };
     let mut pool = AsyncPool::new(config);
-    let address = Address(Hash::compute_from(b"abc"));
+    let address = Address(HashV2::compute_from(b"abc"));
     for i in 1..10 {
         let message = AsyncMessage::new_with_hash(
             Slot::new(0, 0),
@@ -373,5 +373,5 @@ fn test_take_batch() {
     }
     assert_eq!(pool.messages.len(), 9);
     pool.take_batch_to_execute(Slot::new(2, 0), 19);
-    assert_eq!(pool.messages.len(), 4);
+    assert_eq!(pool.messages.len(), 4);*/
 }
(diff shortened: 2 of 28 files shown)


Leo-Besancon:
Some example of refactoring we would have to do:

In `massa\massa-ledger-exports\key.rs`:
```
 let limit = ADDRESS_MAX_SIZE_BYTES + 1;
        buffer.extend(&value[..limit]);
        if value[ADDRESS_MAX_SIZE_BYTES] == DATASTORE_IDENT {
```
The Address size is basically hardcoded in the way we store the ledger.

Leo-Besancon:
# Address versioning 
* How do we compute the Address::get_thread() method in a way that does not depend on the version?
* How can we handle different address versions in our smart-contracts?

## First thoughts: 

To simplify, we only worry about the transition between two address versions, AddressV0 and AddressV1.

Vocabulary:
* Legacy address = AddressV0.
* New address = AddressV1.
* Legacy contract = Deployed at a legacy address. Can handle AddressV0 only.
* New contract = Deployed at a new address. Can handle both address types.

## Thread handling

### Scenario

Alice has 5 coins in her wallet, and maintains two nodes (one using AddressV0 and one using AddressV1).
She signs at the same time two transactions (T0 and T1) from his two addresses (behind the same keypair).
Tx0: Addr(Alice,V0) -> Bob (value of 5 coins)
Tx1: Addr(Alice,V1) -> Charlie (value of 5 coins)

Tx0 and Tx1 are inserted in two different threads Thread 0 and Thread 1 (because the first few bits of each address is different). She is spending the same coins, but the two threads are not aware of that at the beginning.

Here, maybe the double spend would be caught eventually (through cross-thread syncs), but it would lead to a lot of inefficiencies (because it takes too much time to realize some blocks are incompatible and this hinders finality).

### Methods to protect against this problem

a. Keep the previous hash first bits in the newer addresses.
1.New addresses may not have the following structure:
`AddressV1: [VersionNumber1][HashV1]`
   Instead, they would have the following structure, with T the number of bits needed for threads
`AddressV1: [VersionNumber1][HashV0[0..T]][HashV1]`
2. Problems: 
- Ugly code (we have to use two different hashing algos, just to assign correctly each address...)

b. Use another discriminant for the thread repartition

* Computed from Public Key? Do we have access to the Public Key every time we need to access this info?
* 

`todo!()`

c. A node would only accept Txs made from the new version. (A lot of breaking changes)

## Smart contract handling

### Scenario

At T0, Alice publishes SC0 that only handles AddressV0.

At T1, Massa develops AddressV1 standard.

At T2, every node on the network has been updated and handles both addresses AddressV0 and AddressV1.

After T2, the smart contract SC0 will be affected if:
* It uses calculations on the address bytes directly.
* It stores info on a given address (e.g. "address Addr(Alice, V0) owns this NFT".

`todo!()`

Alice would need to access from version V0, or update it ?
After T2, the smart contract SC0 will NOT be affected if:
It only interacts with addresses through Massa SDK features.

### Argument handling of address types in smart contracts

#### In legacy contracts

a. If the legacy contract does not manipulate addresses at all
No problem.

b. If the legacy contract uses addresses for Storage.set() and Storage.get(), or other imported functions from massa-as-sdk
Example: A basic NFT contract. It stores the NFT owner as Storage.
An user should be able to prove that two addresses are the same? 

c. If the legacy contract handles addresses bytes directly.
A new version of this contract, handling both address versions, should be deployed.

#### In new contracts

a. The new contract should have logic that differenciate two of the same address?

### Smart contract addresses access

a. A legacy contract will always be accessible from its original address.

b. A new contract will be deployed at a new address format.

c.1. A legacy contract won't be able to call a new contract?
c.2. A new contract should be able to call a legacy contract.

The address of a smart contract is computed in massa/mass_execution_worker/src/context.rs, in 
`pub fn create_new_sc_address(&mut self, bytecode: Vec<u8>) -> Result<Address, ExecutionError>`

In particular, it is computed has a hash of deterministic data 
```
let mut data: Vec<u8> = self.slot.to_bytes_key().to_vec();
// add the index of the created address within this context to the seed
data.append(&mut self.created_addr_index.to_be_bytes().to_vec());
// add a flag on whether we are in read-only mode or not to the seed
// this prevents read-only contexts from shadowing existing addresses
if self.read_only {
    data.push(0u8);
} else {
    data.push(1u8);
}
let address = Address::AddressV2(AddressV2(massa_hash::HashV2::compute_from(&data)));
```

Changing the hash algorithm will produce new contract addresses easily.

### Implementation choices 

* Use a Translation map? Such as proposed in these notes: https://notes.ethereum.org/@ipsilon/address-space-extension-exploration#ASE-Address-Space-Extension-with-Translation-Map
+ Users can still use both address versions transparently
+ Additionnal features ( ... )
- Additionnal storage
- Not clean
- Adds support for just 1 additional version (i.e. to support N versions, we need N-1 translation maps).
* 

`todo!()`

## References

Ethereum related issue: https://ethereum-magicians.org/t/increasing-address-size-from-20-to-32-bytes/5485
Proposal notes: https://notes.ethereum.org/@ipsilon/address-space-extension-exploration
In particular, the following points are raised: https://notes.ethereum.org/@ipsilon/address-space-extension-test-cases#Notable-cases--questions

damip:
How about forcing simultaneous versioning of addresses + keys ? eg. even if only the hashing algo is updated, the version numbers of newly created keys is also updated. And when the signature algo is updated, also increment the version of the derived addresses.

In that case, the Address structure can be an enum (like https://doc.rust-lang.org/std/net/enum.IpAddr.html ) with the following methods, each specialized to a given address version:
* get_size(&self) -> usize
* get_thread(&self) -> u8
And with a specialized serializer.

The PublicKey structure can also be an enum, and expose a compute_address(&self) -> Address method that yields the right address version.

Advantage: the "version" is actually the crypto version in use

Disvantage: both addresses and keys get an extra enum element even if just one of them changed.

Leo-Besancon:
This seems like a neet idea! @AurelienFT is currently working on a set of macros helping the versioning our various structures (avoiding some code duplication).

I think +

AurelienFT:
Next steps in #3442 #3471

PR #3435, opened 2023-01-17
Add HTTP GET queries proxying 

### LINKED ISSUE #3376: Support HTTP GET queries proxying
It could be interesting to add `GET` request proxing to RPC methods. This feature could simplify frontend queries and allow URIs sharing.

GET /addresses/{address_id} or ?ids=[address_id, address_id] => will be binded to `get_addresses` method

[JSONRPSEE HTTP PROXY](https://github.com/paritytech/jsonrpsee/blob/master/examples/examples/http_proxy_middleware.rs)

related to #2799

modship:
  * [ ] make tests compile
  * [ ] make tests pass 

closes #3376 

Supported resources are:

- [x]  addresses
- [x]  blocks
- [x]  endorsements
- [x]  operations


AurelienFT:
Is this PR still up to date ? If yes can you fix the CI and assign to someone for review ?

modship:
> 

ok i'll do it, i'm waiting for some comments

massa-api/src/proxy.rs
```diff
@@ -0,0 +1,212 @@
...
+                let decoded_query = decode(query)
+                    .expect(&*format!("Error when decode query for GET {}", path))
+                    .to_string();
+
+                for param in decoded_query.split('&') {
+                    // example param for addresses:"ids=["Address1", "Address2"]"
+                    let kv: Vec<&str> = param.splitn(2, '=').collect();
+                    if kv.len() == 2 {
+                        let value: Value = serde_json::from_str(kv[1]).unwrap_or_else(|_| {
+                            panic!("Error when deserialize params query for GET {}", path)
```
aoudiamoncef, line 116: This should be a handled error without panic

massa-api/src/proxy.rs
```diff
@@ -0,0 +1,212 @@
...
+                        });
+                        params
+                            .insert(value)
+                            .expect("Error when insert into ArrayParams");
```
aoudiamoncef, line 120: If we could avoid `expect` and handle the error

massa-api/src/proxy.rs
```diff
@@ -0,0 +1,212 @@
...
+use jsonrpsee::core::params::ArrayParams;
+use jsonrpsee::core::traits::ToRpcParams;
+use jsonrpsee::types::{Id, RequestSer};
+use serde_json::Value;
+use std::future::Future;
+use std::pin::Pin;
+use std::task::{Context, Poll};
+use urlencoding::decode;
+
+pub const URI_ADDRESSES: &str = "/addresses";
```
aoudiamoncef, line 20: We should pay attention to URL encoding for example: 

`%2Faddresses` is the same as `/addresses`

massa-api/src/proxy.rs
```diff
@@ -0,0 +1,212 @@
...
+/// See [`MassaProxyGetRequest`] for more details.
+#[derive(Debug, Clone)]
+pub struct MassaProxyGetRequestLayer {}
+
+impl MassaProxyGetRequestLayer {
+    /// Creates a new [`ProxyGetRequestLayer`].
+    ///
+    /// See [`MassaProxyGetRequest`] for more details.
+    pub fn new() -> Result<Self, RpcError> {
+        Ok(Self {})
```
aoudiamoncef, line 37: Why this the `new`is empty ? may be we could add some checks here and throw an error(failfast)

massa-api/src/proxy.rs
```diff
@@ -0,0 +1,212 @@
...
+
+            req = req.map(|_| body);
+        }
+
+        // Call the inner service and get a future that resolves to the response.
+        let fut = self.inner.call(req);
+
+        // Adjust the response if needed.
+        let res_fut = async move {
+            let res = fut.await.map_err(|err| err.into())?;
```
aoudiamoncef, line 163: Wrap the error with a custom one, this will facilitate the debugging

massa-api/src/proxy.rs
```diff
@@ -0,0 +1,212 @@
...
+                            HeaderValue::from_static("application/json; charset=utf-8"),
+                        )
+                        .body(Body::from(payload.result.to_string()))
+                        .expect("Unable to parse response body for type conversion")
+                } else {
+                    let error = serde_json::to_string(&jsonrpsee::types::ErrorResponse::borrowed(
+                        jsonrpsee::types::error::ErrorCode::InternalError.into(),
+                        Id::Null,
+                    ))
+                    .expect("built from known-good data; qed");
```
aoudiamoncef, line 194: Same here, avoid except as possible

massa-api/src/proxy.rs
modship: for errors i can directly return an http response to the caller each time a fonction can fail or create a struct like ```MassaProxyError``` or ```MassaProxyGetError``` who do the job for better error management. 
It depends on the future evolution of the proxy, I don’t know if it’s excessive or no
what do you think @aoudiamoncef ?

AurelienFT:
I don't feel we will keep on this with API v2 @aoudiamoncef what do you think we should do with this PR ?

aoudiamoncef:
We'll have the same thing in gRPC side but It'll be autogenerated via annotations. This PR was a good exercice to discover the Massa API side. 

We could stop the effort on it.

AurelienFT:
Ok thanks for context let's close this then. Thanks for the work on it @lilyanB

lilyanB:
I thinks you wanted mention @modship

AurelienFT:
Oh yes sorry 😆

PR #3436, opened 2023-01-17
Fix deadlock in get_blocks

### LINKED ISSUE #3399: Testnet API server random deadlocks
Sometimes the server that serve `test.massa.net` become inactive, no more logs but the process is still running.

It symptomatic of a deadlock. For now I haven't find anything on it. Just that it happens only on the node that serve the API so it potentially come from API or Websocket.

AurelienFT:
This PR fix #3399 

A read reference in the storage was kept for too long because of : https://internals.rust-lang.org/t/lifetime-in-match-if-let-and-while-let/14304/6 which cause the `prune()` function of the consensus to never release his write lock to `shared_state` because was blocked on waiting the lock access to the block to delete. But this lock was never release because he lives until the end of the `if let` and in this we are calling `get_block_statuses` which requires a read lock to `shared_state` that he can achieve because of the write lock on the consensus thread.

Here is an schema to explain the text above : 
Screenshot 2023-01-17 at 15 57 45

I just made the `if` before the code that fetch in the `shared_state`.
I tested this on my node on the testnet the endpoint still work. Need to test on testnet0 to be sure it's fully fixed.

### DIFF
--- a/massa-api/src/public.rs
+++ b/massa-api/src/public.rs
@@ -588,30 +588,33 @@ impl MassaRpcServer for API<Public> {
         let blocks = ids
             .into_iter()
             .filter_map(|id| {
-                if let Some(wrapped_block) = storage.read_blocks().get(&id).cloned() {
-                    if let Some(graph_status) = consensus_controller
-                        .get_block_statuses(&[id])
-                        .into_iter()
-                        .next()
-                    {
-                        let is_final = graph_status == BlockGraphStatus::Final;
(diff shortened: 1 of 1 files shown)


aoudiamoncef (approved):

AurelienFT:
I have looked at all the if let/while let and didn't find any other possible problems like this @aoudiamoncef

Eitu33 (approved):

AurelienFT:
bors merge

PR #3441, opened 2023-01-18
ledger-db restructuring

Ben-PH:
  * [ ] make tests compile
  * [ ] make tests pass 


Ben-PH:
I've been looking into making the results of the getters return a byte-slice instead of a vec as well. The issue is that the library uses the `DBPinnableSlice` struct to manage the lifetime of the data-pinning. When it goes out of scope, a rocksdb defined ffi function is used to unpin (free up) the data. Fundamentally, the question seems to be this:

How to make the v-table containing the drop method available outside of `massa-ledger-worker` library, without making our other libraries also dependent on rocksdb (which leads to longer compile times. like, _really_ long)

Eitu33:
We usually employ "refactor" when dealing with major behavioral changes of a module, I changed the title to restructuring to avoid misleading anyone. Also I'm not a big fan of the `ledger_db` subdir, I don't feel there is a need for that as the module is entirely related to the the `ledger_db` file anyway, feels weird to have a file and a directory with the same name too, I'd rather we keep everything at the root. Otherwise lgtm.

Ben-PH:
I'll revert file separation, then merge on monday.

I'll also raise a discussion in github about that (I personally prefer not to have `mod.rs` files, but I'll raise it in the discussion)

massa-ledger-worker/src/ledger_db.rs
```diff
@@ -31,16 +31,16 @@ use std::{
...
+pub const LEDGER_CF: &str = "ledger";
+pub const METADATA_CF: &str = "metadata";
+pub const OPEN_ERROR: &str = "critical: rocksdb open operation failed";
+pub const CRUD_ERROR: &str = "critical: rocksdb crud operation failed";
+pub const CF_ERROR: &str = "critical: rocksdb column family operation failed";
+pub const LEDGER_HASH_ERROR: &str = "critical: saved ledger hash is corrupted";
+pub const KEY_LEN_SER_ERROR: &str = "critical: key length serialization failed";
+pub const SLOT_KEY: &[u8; 1] = b"s";
+pub const LEDGER_HASH_KEY: &[u8; 1] = b"h";
+pub const LEDGER_HASH_INITIAL_BYTES: &[u8; 32] = &[0; HASH_SIZE_BYTES];
```
Eitu33, line 43: remove pub here please

massa-ledger-worker/src/ledger_db.rs
```diff
@@ -210,6 +197,162 @@ impl LedgerDB {
...
+        opt.set_iterate_range(data_prefix!(addr).clone()..end_prefix(data_prefix!(addr)).unwrap());
+
+        let mut iter = self
+            .db
+            .iterator_cf_opt(handle, opt, IteratorMode::Start)
+            .flatten()
+            .map(|(key, _)| key.split_at(ADDRESS_SIZE_BYTES + 1).1.to_vec())
+            .peekable();
+
+        // Return None if emtpty
```
Eitu33, line 250:
```suggestion
        // Return None if empty
        // TODO: function should return None if complete entry does not exist
        // and Some([]) if it does but datastore is empty
```

Ben-PH: I've corrected my mistake of only changing the typo.

massa-ledger-worker/src/ledger_db.rs
Ben-PH: Corrected. Apologies, when reverting the file-split, i removed the `(crate)`, but for some reason didn't remove the `pub` also.

Thanks for the catch.

Eitu33 (approved):

PR #3445, opened 2023-01-19
Provide alloc-free byte-slyce with DBPinnableSlice as Deref

Ben-PH:
  * [ ] make tests compile
  * [ ] make tests pass 

This adds more refactors to the upstream refactor branch. It assumes that passing up the deref trait object implicitly attaches the appropriate `Drop` impls as well (need to check that).

The intended advantages is to reduce the allocations needed.

Possible disadvantages include the increased use of dynamic dispatch.

### DIFF
--- a/massa-execution-worker/src/execution.rs
+++ b/massa-execution-worker/src/execution.rs
@@ -1128,16 +1128,17 @@ impl ExecutionState {
         address: &Address,
         key: &[u8],
     ) -> (Option<Vec<u8>>, Option<Vec<u8>>) {
-        let final_entry = self.final_state.read().ledger.get_data_entry(address, key);
+        let binding = self.final_state.read();
+        let final_entry = binding.ledger.get_data_entry(address, key);
         let search_result = self
             .active_history
             .read()
             .fetch_active_history_data_entry(address, key);
         (
-            final_entry.clone(),
+            final_entry.as_ref().map(|res| res.deref().to_vec()),
             match search_result {
                 HistorySearchResult::Present(active_entry) => Some(active_entry),
-                HistorySearchResult::NoInfo => final_entry,
+                HistorySearchResult::NoInfo => final_entry.map(|res| res.deref().to_vec()),
                 HistorySearchResult::Absent => None,
             },
         )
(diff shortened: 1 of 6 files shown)


Ben-PH:
If this approach is a welcome one, I'll apply such changes elsewhere.

massa-execution-worker/src/execution.rs
```diff
@@ -1128,16 +1128,17 @@ impl ExecutionState {
...
-        let final_entry = self.final_state.read().ledger.get_data_entry(address, key);
+        let binding = self.final_state.read();
+        let final_entry = binding.ledger.get_data_entry(address, key);
         let search_result = self
             .active_history
             .read()
             .fetch_active_history_data_entry(address, key);
         (
-            final_entry.clone(),
+            final_entry.as_ref().map(|res| res.deref().to_vec()),
```
AurelienFT, line 1138: I don't understand how it reduce allocations. In my mind, `to_vec()` makes a new allocation for the vector. I think I'm missing something. Can you explain it to me ?

Ben-PH: At this point in the codebase, the refactor is redundant, true. I'm waiting on @Eitu33 to discuss if replacing (alloc-heavy) `Vec<u8>` with a `Box<dyn Deref<Target = [u8]>>` might actually be useful.  

Assuming that we are all for it, the plan is to put in the change wherever we are cloning out the bytes, when just having a pinned reference into the DB will do.

One other consideration I have in mind, is to establish the pattern that when data is cloned out, it is done so into its deserialized form. Idea being "when you don't own the data, you hold a reference to the bytes that exist in the DB. When you _do_ own the data, you own it in its a cannonical type (e.g. as a byte-vec in a SecureSend, or the struct itself)".

Ben-PH:
Parking this. Can bring this work back up when it becomes more of a priority.

massa-execution-worker/src/execution.rs
AurelienFT: That's interesting yes. Maybe be worth looking in stabilization phase

PR #3447, opened 2023-01-19
Move BlockId into block_id.rs & BlockHeader into block_header.rs

sydhds:
  * [x] make tests compile
  * [x] make tests pass 


AurelienFT (approved):
Will be useful for the futur ! Thank you @sydhds

PR #3455, opened 2023-01-21
2660 Add a rust representation of disk ledger keys

thomas-senechal:
  * [x] make tests compile
  * [x] make tests pass 

### DIFF
--- a/massa-ledger-exports/src/key.rs
+++ b/massa-ledger-exports/src/key.rs
@@ -29,12 +29,18 @@ pub enum KeyType {
 #[derive(Default, Clone)]
 pub struct KeyTypeSerializer {
     vec_u8_serializer: VecU8Serializer,
+    // Wether is deserialized with VecU8Deserializer or not.
+    // This allows us to store the datastore key length at the beginning of the key or not.
+    // The datastore key length is useful when transfering multiple keys, like in packets,
+    // but isn't when storing a datastore key in the ledger.
+    datastore_key_length: bool
 }
 
 impl KeyTypeSerializer {
-    pub fn new() -> Self {
+    pub fn new(datastore_key_length: bool) -> Self {
         Self {
             vec_u8_serializer: VecU8Serializer::new(),
+            datastore_key_length
         }
     }
 }
@@ -46,7 +52,11 @@ impl Serializer<KeyType> for KeyTypeSerializer {
             KeyType::BYTECODE => buffer.extend(&[u8::from(KeyTypeId::BYTECODE)]),
             KeyType::DATASTORE(data) => {
                 buffer.extend(&[u8::from(KeyTypeId::DATASTORE)]);
-                self.vec_u8_serializer.serialize(data, buffer)?;
+                if self.datastore_key_length {
+                    self.vec_u8_serializer.serialize(data, buffer)?;
+                } else {
+                    buffer.extend(data);
+                }
             }
         }
         Ok(())
@@ -56,15 +66,21 @@ impl Serializer<KeyType> for KeyTypeSerializer {
 #[derive(Clone)]
 pub struct KeyTypeDeserializer {
     vec_u8_deserializer: VecU8Deserializer,
+    // Wether is deserialized with VecU8Deserializer or not.
+    // This allows us to store the datastore key length at the beginning of the key or not.
+    // The datastore key length is useful when transfering multiple keys, like in packets,
+    // but isn't when storing a datastore key in the ledger.
+    datastore_key_length: bool
 }
 
 impl KeyTypeDeserializer {
-    pub fn new(max_datastore_key_length: u8) -> Self {
+    pub fn new(max_datastore_key_length: u8, datastore_key_length: bool) -> Self {
         Self {
             vec_u8_deserializer: VecU8Deserializer::new(
                 Included(u64::MIN),
                 Included(max_datastore_key_length as u64),
             ),
+            datastore_key_length
         }
     }
 }
@@ -79,8 +95,12 @@ impl Deserializer<KeyType> for KeyTypeDeserializer {
             Ok(KeyTypeId::BALANCE) => Ok((rest, KeyType::BALANCE)),
             Ok(KeyTypeId::BYTECODE) => Ok((rest, KeyType::BYTECODE)),
             Ok(KeyTypeId::DATASTORE) => {
-                let (rest, data) = self.vec_u8_deserializer.deserialize(rest)?;
-                Ok((rest, KeyType::DATASTORE(data)))
+                if self.datastore_key_length {
+                    let (rest, data) = self.vec_u8_deserializer.deserialize(rest)?;
+                    Ok((rest, KeyType::DATASTORE(data)))
+                } else {
+                    Ok((&[], KeyType::DATASTORE(rest.to_vec())))
+                }
             }
             Err(_) => Err(nom::Err::Error(E::from_error_kind(
                 rest,
@@ -124,10 +144,10 @@ pub struct KeySerializer {
 
 impl KeySerializer {
     /// Creates a new `KeySerializer`
-    pub fn new() -> Self {
+    pub fn new(datastore_key_length: bool) -> Self {
         Self {
             address_serializer: AddressSerializer::new(),
-            key_type_serializer: KeyTypeSerializer::new(),
+            key_type_serializer: KeyTypeSerializer::new(datastore_key_length),
         }
     }
 }
@@ -164,10 +184,10 @@ pub struct KeyDeserializer {
 
 impl KeyDeserializer {
     /// Creates a new `KeyDeserializer`
-    pub fn new(max_datastore_key_length: u8) -> Self {
+    pub fn new(max_datastore_key_length: u8, datastore_key_length: bool) -> Self {
         Self {
             address_deserializer: AddressDeserializer::new(),
-            key_type_deserializer: KeyTypeDeserializer::new(max_datastore_key_length),
+            key_type_deserializer: KeyTypeDeserializer::new(max_datastore_key_length, datastore_key_length),
         }
     }
 }
--- a/massa-ledger-exports/src/key.rs
+++ b/massa-ledger-exports/src/key.rs
@@ -45,23 +45,9 @@ pub struct Key {
 }
 
 impl Key {
-    pub fn new_balance(address: &Address) -> Self {
+    pub fn new(address: &Address, key_type: KeyType) -> Self {
         Self {
-            key_type: KeyType::BALANCE,
-            address: *address,
-        }
-    }
-
-    pub fn new_bytecode(address: &Address) -> Self {
-        Self {
-            key_type: KeyType::BYTECODE,
-            address: *address,
-        }
-    }
-
-    pub fn new_datastore(address: &Address, key: Option<Vec<u8>>) -> Self {
-        Self {
-            key_type: KeyType::DATASTORE(key),
+            key_type,
             address: *address,
         }
     }
@@ -100,7 +86,7 @@ impl Serializer<Key> for KeySerializer {
     /// let mut serialized = Vec::new();
     /// let address = Address::from_str("AU12dG5xP1RDEB5ocdHkymNVvvSJmUL9BgHwCksDowqmGWxfpm93x").unwrap();
     /// let store_key = Hash::compute_from(b"test");
-    /// let mut key = Key::new_datastore(&address, Some(store_key.into_bytes().to_vec()));
+    /// let mut key = Key::new(&address, KeyType::DATASTORE(Some(store_key.into_bytes().to_vec())));
     /// KeySerializer::new().serialize(&key, &mut serialized).unwrap();
     /// ```
     fn serialize(&self, value: &Key, buffer: &mut Vec<u8>) -> Result<(), SerializeError> {
@@ -155,14 +141,14 @@ impl Deserializer<Key> for KeyDeserializer {
     /// let address = Address::from_str("AU12dG5xP1RDEB5ocdHkymNVvvSJmUL9BgHwCksDowqmGWxfpm93x").unwrap();
     /// let store_key = Hash::compute_from(b"test");
     ///
-    /// let mut key = Key::new_datastore(&address, Some(store_key.into_bytes().to_vec()));
+    /// let mut key = Key::new(&address, KeyType::DATASTORE(Some(store_key.into_bytes().to_vec())));
     /// let mut serialized = Vec::new();
     /// KeySerializer::new().serialize(&key, &mut serialized).unwrap();
     /// let (rest, key_deser) = KeyDeserializer::new(255).deserialize::<DeserializeError>(&serialized).unwrap();
     /// assert!(rest.is_empty());
     /// assert_eq!(key_deser, key);
     ///
-    /// let mut key = Key::new_balance(&address);
+    /// let mut key = Key::new(&address, KeyType::BALANCE);
     /// let mut serialized = Vec::new();
     /// KeySerializer::new().serialize(&key, &mut serialized).unwrap();
     /// let (rest, key_deser) = KeyDeserializer::new(255).deserialize::<DeserializeError>(&serialized).unwrap();
--- a/massa-ledger-exports/src/key.rs
+++ b/massa-ledger-exports/src/key.rs
@@ -4,35 +4,88 @@ use massa_models::{
 };
 use massa_serialization::{Deserializer, SerializeError, Serializer};
 use nom::error::{ContextError, ParseError};
+use num_enum::{IntoPrimitive, TryFromPrimitive};
 use std::ops::Bound::Included;
 
 pub const BALANCE_IDENT: u8 = 0u8;
 pub const BYTECODE_IDENT: u8 = 1u8;
 pub const DATASTORE_IDENT: u8 = 2u8;
 
+#[derive(PartialEq, Eq, Clone, IntoPrimitive, TryFromPrimitive, Debug)]
 #[repr(u8)]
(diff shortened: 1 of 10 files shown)


Ben-PH:
The key.rs merge conflict is coming from changes that I introduced. Would it be helpful if we pair up to resolve? Maybe we can find out why the +2 is needed here while we are at it: https://github.com/massalabs/massa/pull/3455/files#diff-1219a2cf735a0b9e73b4372b43c8191d263af08fae67b6e13eb30e855f2eca61R651

thomas-senechal:
> The key.rs merge conflict is coming from changes that I introduced. Would it be helpful if we pair up to resolve? Maybe we can find out why the +2 is needed here while we are at it: https://github.com/massalabs/massa/pull/3455/files#diff-1219a2cf735a0b9e73b4372b43c8191d263af08fae67b6e13eb30e855f2eca61R651

Thank a lot for your help but I'm used to it now since I already did it multiple times for this PR ^^
Every conflicts are fixed now, but I still have an issue with the `gas-costs-check`. I would love to have some help about that ! 

And about the +2, I talked about it with @AurelienFT and we think it's related to the fact that we are now using an enum and that it takes 2 bytes instead of one. But it would be great if we could make sure that this is why it happens to make sure it wont become an issue later. 

Thanks again ! 😊

Ben-PH:
Most likely, but I thought I accounted for the extra byte. Have a couple of ideas as to what's going on.

AurelienFT:
@Eitu33 I assign you as a reviewer on that. If you don't want/have time assign me back

AurelienFT:
As thomas is absent this week and this become quit important @sydhds can you look at it ?

massa-ledger-exports/src/key.rs
```diff
@@ -1,58 +1,77 @@
...
-macro_rules! balance_key {
-    ($addr:expr) => {
-        [&$addr.prefixed_bytes()[..], &[BALANCE_IDENT]].concat()
-    };
+#[repr(u8)]
+#[derive(PartialEq, Eq, Clone, Debug)]
+pub enum KeyType {
+    BALANCE = 0,
+    BYTECODE = 1,
+    DATASTORE(Option<Vec<u8>>) = 2,
```
sydhds, line 18: Would be nice to have some doc explaining what is Option<Vec<u8>> here (I guess this is the datastore key?). Any reason why we have an Option here?

massa-ledger-exports/src/key.rs
```diff
@@ -70,35 +89,34 @@ impl KeySerializer {
...
-    fn serialize(&self, value: &Vec<u8>, buffer: &mut Vec<u8>) -> Result<(), SerializeError> {
-        let limit = ADDRESS_SIZE_BYTES + 1;
-        buffer.extend(&value[..limit]);
-        if value[limit - 1] == DATASTORE_IDENT {
-            if value.len() > limit {
-                self.vec_u8_serializer
-                    .serialize(&value[limit..].to_vec(), buffer)?;
-            } else {
+    fn serialize(&self, value: &Key, buffer: &mut Vec<u8>) -> Result<(), SerializeError> {
+        buffer.extend(&value.address.prefixed_bytes());
```
sydhds, line 107: Any reason why AddressSerializer is not used here?

massa-ledger-worker/src/ledger_db.rs
```diff
@@ -585,21 +633,22 @@ impl LedgerDB {
...
                 handle,
                 opt,
-                IteratorMode::From(data_prefix!(addr), Direction::Forward),
+                IteratorMode::From(&serialized_key, Direction::Forward),
             )
             .flatten()
             .map(|(key, data)| {
                 (
-                    key.split_at(ADDRESS_SIZE_BYTES + 1).1.to_vec(),
+                    key.split_at(ADDRESS_SIZE_BYTES + 2).1.to_vec(), // DIRTY FIX: It's supposed to be 1 but 2 works. I couldn't find why.
```
sydhds, line 651: I think this should be investigated, maybe check with @Eitu33

sydhds (changes requested):
Nice PR ! :) 
A few things to clarify !

massa-ledger-exports/src/key.rs
```diff
@@ -1,58 +1,77 @@
...
-///
-/// TODO: add a separator identifier if the need comes to have multiple datastore
-#[macro_export]
-macro_rules! data_key {
-    ($addr:expr, $key:expr) => {
-        [&$addr.prefixed_bytes()[..], &[DATASTORE_IDENT], &$key].concat()
-    };
+/// Disk ledger keys representation
+#[derive(PartialEq, Eq, Clone, Debug)]
+pub struct Key {
```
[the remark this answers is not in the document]
thomas-senechal: I don't think it's a good idea to rename it from `Key` to `LedgerKey`. We already are in a module related to the Ledger, if we need more context in another module, then we should import `Key` using `ledger_export::Key` to add context. 

We could also import it like so:
```
use ledger_export::Key as LedgerKey;
```
This would allow us to have a clear context. The issue I see with this method is that we will have to use this import in every files using `Key`.

massa-ledger-exports/src/key.rs
```diff
@@ -1,58 +1,77 @@
...
 
-/// Bytecode key formatting macro
-///
-/// NOTE: still handle separate bytecode for now to avoid too many refactor at once
-#[macro_export]
-macro_rules! bytecode_key {
-    ($addr:expr) => {
-        [&$addr.prefixed_bytes()[..], &[BYTECODE_IDENT]].concat()
-    };
+impl KeyType {
```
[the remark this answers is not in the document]
sydhds: https://github.com/massalabs/massa/blob/589ae01ddc6ec19234e6cb726a8feab51acd4a43/massa-models/src/operation.rs#L234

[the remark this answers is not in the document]
thomas-senechal: Oh, so I think what is interesting there isn't the `repr(u32)` but the use of `IntoPrimitive` and `TryFromPrimitive` which allows us to do serialization and deserialization in a cleaner way.

I will change it to match that way.

massa-ledger-exports/src/key.rs
thomas-senechal: Fixed with https://github.com/massalabs/massa/pull/3455/commits/2014ce50d37f6c73538ba67f52054767fad38589 (and https://github.com/massalabs/massa/pull/3455/commits/a7da4a7c68bdbe92ce65bb5ff2bd842ba87122c7)

massa-ledger-exports/src/key.rs
thomas-senechal: This wasn't used anymore. I removed it in [this commit](https://github.com/massalabs/massa/pull/3455/commits/a7da4a7c68bdbe92ce65bb5ff2bd842ba87122c7).

massa-ledger-exports/src/key.rs
```diff
@@ -1,58 +1,77 @@
...
 }
 
-/// Datastore entry prefix formatting macro
-#[macro_export]
-macro_rules! data_prefix {
-    ($addr:expr) => {
-        &[&$addr.prefixed_bytes()[..], &[DATASTORE_IDENT]].concat()
-    };
+impl Key {
+    pub fn new_balance(address: &Address) -> Self {
```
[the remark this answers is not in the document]
thomas-senechal: Fixed in [this commit](https://github.com/massalabs/massa/pull/3455/commits/dec646c71403baf9a8605a495dbdaec786c8b8b6).

massa-ledger-exports/src/key.rs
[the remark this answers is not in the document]
thomas-senechal: It's not possible to use `IntoPrimitive` and `TryFromPrimitive` on the `KeyType` enum because of the DATASTORE key.

I don't see a better way of doing it than the one I used. If you have one, could you explain me ? If needed, we can make a call to talk about it.

massa-ledger-exports/src/key.rs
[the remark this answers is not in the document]
sydhds: Alright so can you use this (use ledger_export::Key as LedgerKey;) in bootstrap (and any other module that are not massa-ledger-XX)?

massa-ledger-exports/src/key.rs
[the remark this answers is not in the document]
sydhds: As an example, you can define:

`

#[derive(Debug, Clone, Serialize, Deserialize, PartialEq)]
pub enum OperationType {
    ...
}

#[derive(IntoPrimitive, Debug, Eq, PartialEq, TryFromPrimitive)]
#[repr(u32)]
enum OperationTypeId {
    ...
}
`

so the original enum does not change at all.

massa-ledger-worker/src/ledger_db.rs
```diff
@@ -266,38 +275,42 @@ impl LedgerDB {
...
             }
-            StreamingStep::Finished(_) => return Ok((ledger_part, cursor)),
+            StreamingStep::<Key>::Finished(_) => return Ok((ledger_part, cursor)),
         };
 
         // Iterates over the whole database
         for (key, entry) in db_iterator.flatten() {
             if (ledger_part.len() as u64) < (self.ledger_part_size_message_bytes) {
-                key_serializer.serialize(&key.to_vec(), &mut ledger_part)?;
+                let (_, key) = key_deserializer.deserialize(&key)?;
```
AurelienFT, line 310: What's the point of deserializing to re-serialize it ?

massa-ledger-worker/src/ledger_db.rs
```diff
@@ -266,38 +275,42 @@ impl LedgerDB {
...
         &self,
-        cursor: StreamingStep<Vec<u8>>,
-    ) -> Result<(Vec<u8>, StreamingStep<Vec<u8>>), ModelsError> {
+        cursor: StreamingStep<Key>,
+    ) -> Result<(Vec<u8>, StreamingStep<Key>), ModelsError> {
         let handle = self.db.cf_handle(LEDGER_CF).expect(CF_ERROR);
         let opt = ReadOptions::default();
         let ser = VecU8Serializer::new();
-        let key_serializer = KeySerializer::new();
+        let key_deserializer = KeyDeserializer::new(self.max_datastore_key_length);
```
AurelienFT, line 283: The deserializer could be in self to avoid multiple creations each time we call this.

massa-ledger-worker/src/ledger_db.rs
```diff
@@ -333,22 +343,24 @@ impl LedgerDB {
...
-            Ok(StreamingStep::Ongoing((*last_key).clone()))
-        } else {
-            Err(ModelsError::SerializeError(
-                "rest is not empty.".to_string(),
-            ))
+        match last_key.as_ref() {
+            Some(last_key) => {
+                if rest.is_empty() {
+                    self.write_batch(batch);
+                    Ok(StreamingStep::Finished(Some(last_key.clone())))
```
AurelienFT, line 356: This doesn't match the previous behavior it should be : 
```suggestion
                    Ok(StreamingStep::Ongoing(Some(last_key.clone())))
```

massa-ledger-worker/src/ledger_db.rs
```diff
@@ -391,17 +403,21 @@ impl LedgerDB {
...
         self.len_serializer
-            .serialize(&(key.len() as u64), &mut len_bytes)
+            .serialize(&(serialized_key.len() as u64), &mut len_bytes)
             .expect(KEY_LEN_SER_ERROR);
-        let hash = Hash::compute_from(&[&len_bytes, key, value].concat());
+        let hash = Hash::compute_from(&[&len_bytes, &serialized_key, value].concat());
         batch.ledger_hash ^= hash;
-        batch.aeh_list.insert(key.to_vec(), hash);
-        batch.write_batch.put_cf(handle, key, value);
+        batch.aeh_list.insert(serialized_key.to_vec(), hash);
```
AurelienFT, line 419: To vec here shouldn't be necessary as it's already a vec. if it's to clone it because it's moved the line above you can use the `clone()` combinator : 
```suggestion
        batch.aeh_list.insert(serialized_key.clone(), hash);
```

massa-ledger-worker/src/ledger_db.rs
```diff
@@ -435,22 +466,32 @@ impl LedgerDB {
...
+        {
+            batch.ledger_hash ^=
+                Hash::compute_from(&[&len_bytes, &serialized_key, &prev_bytes[..]].concat());
         }
-        let hash = Hash::compute_from(&[&len_bytes, key, value].concat());
+        let hash = Hash::compute_from(&[&len_bytes, &serialized_key, value].concat());
         batch.ledger_hash ^= hash;
-        batch.aeh_list.insert(key.to_vec(), hash);
-        batch.write_batch.put_cf(handle, key, value);
+        batch.aeh_list.insert(serialized_key.to_vec(), hash);
```
AurelienFT, line 493:
```suggestion
        batch.aeh_list.insert(serialized_key.clone(), hash);
```

massa-ledger-worker/src/ledger_db.rs
```diff
@@ -514,24 +568,29 @@ impl LedgerDB {
...
 
         // bytecode
-        self.delete_key(handle, batch, &bytecode_key!(addr));
+        self.delete_key(handle, batch, &Key::new(addr, KeyType::BYTECODE));
 
         // datastore
         let mut opt = ReadOptions::default();
-        opt.set_iterate_upper_bound(end_prefix(data_prefix!(addr)).unwrap());
+        let serialized_key = datastore_prefix_from_address(addr);
+        let key_deserializer = KeyDeserializer::new(self.max_datastore_key_length);
```
AurelienFT, line 579: Same as above, you could use one initialised in the self.

massa-ledger-worker/src/ledger_db.rs
```diff
@@ -514,24 +568,29 @@ impl LedgerDB {
...
+        self.delete_key(handle, batch, &Key::new(addr, KeyType::BALANCE));
 
         // bytecode
-        self.delete_key(handle, batch, &bytecode_key!(addr));
+        self.delete_key(handle, batch, &Key::new(addr, KeyType::BYTECODE));
 
         // datastore
         let mut opt = ReadOptions::default();
-        opt.set_iterate_upper_bound(end_prefix(data_prefix!(addr)).unwrap());
+        let serialized_key = datastore_prefix_from_address(addr);
```
AurelienFT, line 578: Maybe precise in the variable name that it's only a prefix

massa-ledger-worker/src/ledger_db.rs
```diff
@@ -238,7 +245,9 @@ impl LedgerDB {
         let handle = self.db.cf_handle(LEDGER_CF).expect(CF_ERROR);
 
         let mut opt = ReadOptions::default();
-        opt.set_iterate_range(data_prefix!(addr).clone()..end_prefix(data_prefix!(addr)).unwrap());
+        let serialized_key = datastore_prefix_from_address(addr);
```
AurelienFT, line 248: Could be nice to mention that it's a prefix.

massa-ledger-worker/src/ledger_db.rs
AurelienFT: As @Eitu33 is absent, I will take over on this. As the key was previously considered as byte array we had the length of the full key at start of the serialized form and we deserialized everything in one batch.
Now that the key is a more specified type we don't need the length to be encoded as we know how much byte when to deserialize and how we interpret them. We only need a length for datastore key as it's the only one that has a variable size. 
This function return the datastore key stripped and so now that we only need it for datastore variant the byte is just before the datastore key.

Something that we need to change is that this byte is useful when serialising to send it through the network in messages/ledger part but not in the database as there is no data after the key. So we need to have 2 different serializers/deserializers

massa-ledger-exports/src/key.rs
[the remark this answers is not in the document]
thomas-senechal: Fixed with https://github.com/massalabs/massa/pull/3455/commits/1740e8c506318e805d7f0ee69045665ff52861e8

massa-ledger-exports/src/key.rs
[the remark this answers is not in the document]
thomas-senechal: Fixed with https://github.com/massalabs/massa/pull/3455/commits/7b9b651ebb7beb8c425e8f8b0543302744dfae84

thomas-senechal:
@sydhds @AurelienFT 
Everything should be fixed ! 
Do not hesitate to ping me if there are still some issues !

massa-ledger-worker/src/ledger_db.rs
```diff
@@ -585,21 +648,22 @@ impl LedgerDB {
...
                 handle,
                 opt,
-                IteratorMode::From(data_prefix!(addr), Direction::Forward),
+                IteratorMode::From(&key_prefix, Direction::Forward),
             )
             .flatten()
             .map(|(key, data)| {
                 (
-                    key.split_at(ADDRESS_SIZE_BYTES + 1).1.to_vec(),
+                    key.split_at(ADDRESS_SIZE_BYTES + 1).1.to_vec(), // +1 for the type byte, +1 for the length byte
```
AurelienFT, line 666: Change this comment

AurelienFT (approved):

massa-ledger-worker/src/ledger_db.rs
thomas-senechal: Fixed with https://github.com/massalabs/massa/pull/3455/commits/573ca8243eb7b44bae408426c925c454fc4dfaba

sydhds (approved):

AurelienFT:
Tests failing aren't related to this PR and are fixed in : https://github.com/massalabs/massa/pull/3602 merging this as it's prioand blocking for #3603

PR #3460, opened 2023-01-23
Add some missing types to OpenRPC

aoudiamoncef:
  * [x] make tests compile
  * [x] make tests pass 


aoudiamoncef:
I conserved auto format from ApiCurio which is used to validate all schema.

AurelienFT (approved):

PR #3461, opened 2023-01-23
3428 implement pagination in jsonrpc endpoints

Leo-Besancon:
  * [x] make tests compile
  * [x] make tests pass 


massa-api/src/lib.rs
```diff
@@ -39,20 +39,47 @@ use massa_pos_exports::SelectorController;
...
 
 mod api;
 mod api_trait;
 mod private;
 mod public;
 
+/// Represents a Vec that can be split across Pages
+/// Cf. https://docs.rs/paginate/latest/paginate/
+#[derive(Serialize)]
+pub struct PagedVec<T> {
```
aoudiamoncef, line 60: This struct could be moved to `massa-api-exports`as it's shared between modules

massa-api/src/public.rs
```diff
@@ -404,7 +404,11 @@ impl MassaRpcServer for API<Public> {
         Ok(consensus_controller.get_cliques())
     }
 
-    async fn get_stakers(&self) -> RpcResult<Vec<(Address, u64)>> {
+    async fn get_stakers(
+        &self,
+        limit: Option<usize>,
```
aoudiamoncef, line 409: It'll be better to create a struct PageRequest

could be inspired by https://docs.spring.io/spring-data/data-commons/docs/current/api/org/springframework/data/domain/PageRequest.html

aoudiamoncef (changes requested):
Nice work, minimalistic impl of pagination without breaking changes. We have juste to improve the method params and let's goooo

Leo-Besancon:
Hmm, maybe the PageRequest param should be optional.
Right now the PageRequest is required, but its fields are optional.

aoudiamoncef:
I prefer the first impl because it's lightweight and doesn't break the api:

1- we should make the pageable struct optional, actually I need to provide empty object or it breaks my call.
2- PageVec should be serialised as classique Vec, we will add a wrapper struct with multiple informations about pagination in ApiV2. 
3- we'll merge ^^

massa-node/base_config/openrpc.json
```diff
@@ -2516,6 +2535,22 @@
                     }
                 }
             },
+            "PagedStakers": {
```
aoudiamoncef, line 2538: this should be removed

massa-node/base_config/openrpc.json
```diff
@@ -3103,6 +3138,14 @@
                     "$ref": "#/components/schemas/OperationType"
                 }
             },
+            "PagedStakers": {
```
aoudiamoncef, line 3141: this should be removed

aoudiamoncef (changes requested):
Great job, a small clean in `openrpc.json` file and it's done

massa-node/base_config/openrpc.json
```diff
@@ -394,12 +394,31 @@
                     "description": "Massa public api"
                 }
             ],
-            "params": [],
+            "params": [
+                {
+                    "name": "page_request",
```
aoudiamoncef, line 399: As it's designed to be reused, you could extract it into a component schema section and add it to contentDescriptors because it's an INPUT which is directly manipulated by the user.

aoudiamoncef (changes requested):
One last thing and it's good for me

aoudiamoncef (approved):
Great 💯

PR #3462, opened 2023-01-23
Test/block serializer args 1

sydhds:
  * [x] make tests compile
  * [x] make tests pass 


Eitu33 (commented):
This is nice 👍

Eitu33:
But is it really necessary for functions that only use a few arguments? I'm thinking about `BlockDeserializerArgs`, this would be nice but only for anything above the clippy limit imho.

sydhds:
> But is it really necessary for functions that only use a few arguments? I'm thinking about `BlockDeserializerArgs`, this would be nice but only for anything above the clippy limit imho.

Indeed for struct with fee arguments, it's not very necessary but if we adopt this PR it would be nice to stay consistent in the code no?

Eitu33:
I think what we should do is simply adopt the clippy rule, since the vast majority of function have few arguments it would add a lot of code, take time, and not bring much to the table. That said for those who are above the limit it's great.

AurelienFT:
> * [ ]  document all added functions
> * [ ]  try in sandbox /simulation/labnet
> * [ ]  unit tests on the added/changed features
>   
>   * [ ]  make tests compile
>   * [ ]  make tests pass
> * [ ]  add logs allowing easy debugging in case the changes caused problems
> * [ ]  if the API has changed, update the API specification

Can you make all the checks here ? so that we are sure it doesn't break anything as we are close to the release date of the testnet we should be even more careful to try not include any new bug

Eitu33 (dismissed):

Eitu33:
Approving the code, did not test

AurelienFT:
Postpone it to testnet_20 as it has been tested.

sydhds:
Tested in sandbox mode and all good! @AurelienFT can you approve it?

massa-bootstrap/src/client_binder.rs
```diff
@@ -19,33 +20,13 @@ use tokio::io::AsyncWriteExt;
 
 /// Bootstrap client binder
 pub struct BootstrapClientBinder {
-    max_bootstrap_message_size: u32,
+    // max_bootstrap_message_size: u32,
```
AurelienFT, line 23: Can you remove this dead code ?

AurelienFT (approved):

Issue #2607, opened 2022-05-09
Base_config/config.toml - need ipv6 listen on port 33035


rykcod:
For ipv6 nodes, in the base_config/config.toml, the monitoring service only listen on ipv4 address for the 33035 port.

AurelienFT:
@rykcod what is the monitoring service ?

rykcod:
Hi Aurélien,

In the massa-node/base_config/config.toml
    # port on which the node API listens for public requests. Can be exposed to the Internet.
    bind_public = "0.0.0.0:33035"

Its result, this service is dont listen on ipv6

PR #3463, opened 2023-01-24
Bind to IPv6 addr.

Leo-Besancon:
  * [X] make tests compile
  * [X] make tests pass 

Only binding on IPv6 addresses (bind_private = "[::]:33034", bind_public = "[::]:33035" and bind_api = "[::]:33036") still works with IPv4 only interfaces.

Cf issue #2607

### DIFF
--- a/massa-client/base_config/config.toml
+++ b/massa-client/base_config/config.toml
@@ -3,7 +3,8 @@ history_file_path = "config/.massa_history"
 timeout = 1000
 
 [default_node]
-ip = "::1"
+# The IP of your node. Works both with IPv4 (like 127.0.0.1) and IPv6 (like ::1) addresses, if the node is bound to the correct protocol.
+ip = "127.0.0.1"
 private_port = 33034
 public_port = 33035
 
--- a/massa-client/base_config/config.toml
+++ b/massa-client/base_config/config.toml
@@ -3,7 +3,7 @@ history_file_path = "config/.massa_history"
 timeout = 1000
 
 [default_node]
-ip = "127.0.0.1"
+ip = "::1"
 private_port = 33034
 public_port = 33035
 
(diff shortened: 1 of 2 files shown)


Leo-Besancon:
Postman queries went through, however the current solution breaks the client.
I'll investigate more.

sydhds (commented):
Is this a good idea to make ipv6 the default? I'm wondering if it can confuse user. Maybe documentation about supported ip address (ipv4 + ipv6) can help?

Leo-Besancon:
> Is this a good idea to make ipv6 the default? I'm wondering if it can confuse user. Maybe documentation about supported ip address (ipv4 + ipv6) can help?

Oh, you mean a user that knows what 127.0.0.1 means but only sees `ip = "::1"` in the client config file, so he is lost?

Adding a comment above like the following would be good?

```
[default_node]
# Target IP of the API. Defaults to localhost, and works both with IPv4 and IPv6 interfaces.
ip = "::1"
private_port = 33034
public_port = 33035
```

sydhds:
> > Is this a good idea to make ipv6 the default? I'm wondering if it can confuse user. Maybe documentation about supported ip address (ipv4 + ipv6) can help?
> 
> Oh, you mean a user that knows what 127.0.0.1 means but only sees `ip = "::1"` in the client config file, so he is lost?
> 
> Adding a comment above like the following would be good?
> 
> ```
> [default_node]
> # Target IP of the API. Defaults to localhost, and works both with IPv4 and IPv6 interfaces.
> ip = "::1"
> private_port = 33034
> public_port = 33035
> ```

I would even add something like:

... IPv4 (like 127.0.0.1) and IPv6 (like ::1) address.
I

sydhds (approved):

PR #3471, opened 2023-01-26
Multisig tests

Leo-Besancon:
  * [ ] make tests compile
  * [ ] make tests pass 


massa-api/src/public.rs
```diff
@@ -138,7 +138,7 @@ impl MassaRpcServer for API<Public> {
         {
             let address = address.unwrap_or_else(|| {
                 // if no addr provided, use a random one
-                Address::from_public_key(&KeyPair::generate().get_public_key())
+                Address::from_public_key(&KeyPair::generate(0).unwrap().get_public_key())
```
Ben-PH, line 141: Leave comment/use `expect` to provide context for the next person coming around to remove the unwraps

massa-bootstrap/src/tests/tools.rs
```diff
@@ -267,18 +269,18 @@ pub fn get_dummy_block_id(s: &str) -> BlockId {
     BlockId(Hash::compute_from(s.as_bytes()))
 }
 
-pub fn get_random_public_key() -> PublicKey {
-    let priv_key = KeyPair::generate();
+pub fn get_random_public_key(version: u64) -> PublicKey {
+    let priv_key = KeyPair::generate(version).unwrap();
```
Ben-PH, line 273: Could use some sort of range primitive to constrain the selected value? 

If you're interestedd in a type-theory rabit-hole, look into dependant typing/GATs

massa-consensus-worker/src/state/graph.rs
```diff
@@ -168,7 +168,7 @@ impl ConsensusState {
             self.discarded_index.insert(*block_id);
             Ok(())
         } else {
-            return Err(ConsensusError::ContainerInconsistency(format!("inconsistency inside block statuses removing stale blocks adding {} - block {} is missing", add_block_id, block_id)));
+            Err(ConsensusError::ContainerInconsistency(format!("inconsistency inside block statuses removing stale blocks adding {} - block {} is missing", add_block_id, block_id)))
```
Ben-PH, line 171: This is an ideal use-case for the let-else syntax:

```rust
let Some(BlockStatus::Active {
            a_block: active_block,
            storage: _storage,
}) = self.block_statuses.remove(block_id) else {
    // Write the "early return" line (which is what this is) where it aught to be
    return Err(ConsensusError::ContainerInconsistency(format!("inconsistency inside block statuses removing stale blocks adding {} - block {} is missing", add_block_id, block_id)));
}

// continue with the Some(...) path without an extra indentation
self.active_index.remove(block_id);
    if active_block.is_final {
        return Err(ConsensusError::ContainerInconsistency(format!("inconsistency inside block statuses removing stale blocks adding {} - block {} was already final", add_block_id, block_id)));
}
...
```

massa-execution-worker/src/context.rs
```diff
@@ -411,7 +411,14 @@ impl ExecutionContext {
...
         // hash the seed to get a unique address
-        let address = Address::SC(SCAddress(massa_hash::Hash::compute_from(&data)));
+        let address = Address::SC(
+            SCAddress::from_bytes_without_version(
+                0,
+                massa_hash::Hash::compute_from(&data).to_bytes(),
+            )
+            .expect("Could not create the sc address"),
+        );
+        //let address = Address::SC(SCAddress::from_bytes_without_version(1, &data)));
```
Ben-PH, line 421: no longer needed?

Ben-PH, line 421: Dead code?

```suggestion
```

massa-ledger-exports/src/key.rs
```diff
@@ -88,12 +90,27 @@ impl Serializer<Vec<u8>> for KeySerializer {
...
+                "The key does not contain a valid address",
+            )));
+        }
+        let (rest, addr): (&[u8], Address) = result_addr.unwrap();
+
+        buffer.extend(addr.to_prefixed_bytes());
+
+        if rest[0] == DATASTORE_IDENT {
+            if !rest.is_empty() {
+                self.vec_u8_serializer.serialize(&rest.to_vec(), buffer)?;
```
Ben-PH, line 113: Double check that the byte-vec is an encoding of a u8, as opposed to a serialized varint that happens to only need 1 byte. Might be better to do a deserialization on `rest` using something like the `nom::verify` thing?

massa-ledger-exports/src/key.rs
```diff
@@ -160,7 +177,7 @@ impl Deserializer<Vec<u8>> for KeyDeserializer {
         &self,
         buffer: &'a [u8],
     ) -> nom::IResult<&'a [u8], Vec<u8>, E> {
-        let (rest, address) = self.address_deserializer.deserialize(buffer)?;
+        let (rest, address): (&[u8], Address) = self.address_deserializer.deserialize(buffer)?;
```
Ben-PH, line 180: check to see if this explicit type is still needed.

massa-ledger-worker/src/ledger_db.rs
```diff
@@ -235,16 +235,33 @@ impl LedgerDB {
...
+            .map(|(key, _)| {
+                let res: Result<(&[u8], Address), _> =
+                    address_deserializer.deserialize::<massa_serialization::DeserializeError>(&key);
+                match res {
+                    Ok((rest, _addr)) => rest.to_vec(),
+                    Err(_) => {
+                        vec![]
+                    }
+                }
+            })
```
Ben-PH, line 264: It seems like this will map a parsing error to an empty vec, which will be included in the btree set, thus the result being `Some(...)` instead of `None`, or a set of (serialized representation of) keys that includes an empty vec(which obfuscates the error).

one (very hacky) way to handle this, is to add a `filter(|v| !v.is_empty())`, after the `map` and before the `peekable`, but there's probably a "nicer" way to do it.

massa-ledger-worker/src/ledger_db.rs
```diff
@@ -585,23 +601,22 @@ impl LedgerDB {
...
         let mut opt = ReadOptions::default();
         opt.set_iterate_upper_bound(end_prefix(data_prefix!(addr)).unwrap());
 
         self.db
-            .iterator_cf_opt(
-                handle,
-                opt,
-                IteratorMode::From(data_prefix!(addr), Direction::Forward),
-            )
+            .iterator_cf_opt(handle, opt, IteratorMode::Start)
```
Ben-PH, line 611: Is it possible to re-interoduce a lower-bound, instead of starting from the very beginning, on the db-iterator?

massa-models/src/address.rs
```diff
@@ -179,202 +276,606 @@ impl FromStr for Address {
...
+        match version {
+            <UserAddress!["0"]>::VERSION => Ok(UserAddressVariant!["0"](
+                <UserAddress!["0"]>::from_bytes_without_version(rest)?,
+            )),
+            <UserAddress!["1"]>::VERSION => Ok(UserAddressVariant!["1"](
+                <UserAddress!["1"]>::from_bytes_without_version(rest)?,
+            )),
+            _ => Err(ModelsError::AddressParseError),
+        }
+    }
```
Ben-PH, line 356: good candidate to use the `nom::alt` parser. We can pair together to make it work

massa-models/src/address.rs
```diff
@@ -179,202 +276,606 @@ impl FromStr for Address {
...
+
+        match version {
+            <SCAddress!["0"]>::VERSION => Ok(SCAddressVariant!["0"](
+                <SCAddress!["0"]>::from_bytes_without_version(rest)?,
+            )),
+            <SCAddress!["1"]>::VERSION => Ok(SCAddressVariant!["1"](
+                <SCAddress!["1"]>::from_bytes_without_version(rest)?,
+            )),
+            _ => Err(ModelsError::AddressParseError),
+        }
```
Ben-PH, line 443: Same comment as before about using `alt` and `verify`

massa-models/src/address.rs
```diff
@@ -179,202 +276,606 @@ impl FromStr for Address {
...
+        if self.read_only {
+            buff.push(0u8);
+        } else {
+            buff.push(1u8);
+        }
+        buff
+    }
+
+    fn from_bytes_without_version(data: &[u8]) -> Result<SCAddress, ModelsError> {
+        if data.len() != SLOT_KEY_SIZE + 9 {
```
Ben-PH, line 530: add comment about the magic number

massa-models/src/address.rs
```diff
@@ -179,202 +276,606 @@ impl FromStr for Address {
...
+            slot,
+            created_addr_index,
+            read_only,
+        })
     }
 }
 
+/* /!\ SCAddressV1 not prehashed! */
+impl PreHashed for Address {}
+
```
Ben-PH, line 569: This is probably dangerous. If the SCV1 being marked as prehashed, but actually isn't, then it could cause really painful bugs in the future.

massa-network-exports/src/peers.rs
```diff
@@ -145,10 +145,23 @@ impl Deserializer<BootstrapPeers> for BootstrapPeersDeserializer {
...
+    Eq,
+    PartialOrd,
+    Ord,
+    Hash,
+    Display,
+    Enum,
 )]
 pub enum PeerType {
     /// Just a peer
+    #[default]
```
Ben-PH, line 164: <3

Ben-PH:
Great effort on resolving the conflicts. I'm guessing a lot of the comments I made are premature (given it's still a draft).  Look forward to the next review request :)

massa-execution-worker/src/context.rs
Leo-Besancon: This will get resolved once we have a "switch" from one version to another. It's mainly to show what is the expected input of the SCAddressV0 and SCAddressV1 constructor.

massa-api/src/public.rs
Leo-Besancon: I'll try and see what I can do better here, thanks!

Leo-Besancon:
I think we can close this PR.
- For the MultiSig aspect, the tests were able to give us some insights about the current state of Rust crates on the subject. We may want to get back to it later, but we're not following up on it right now.
- For the versioning aspect, this PR gives an idea of how to use different versions (for Address, KeyPairs, Signatures, etc.à, and it may be useful to use it in the versioning follow-ups: https://github.com/massalabs/massa/pull/3664

PR #3472, opened 2023-01-26
Add current/next slot timestamp to `get_status`

### LINKED ISSUE #3446: Add current/next slot timestamp to `get_status`
Actually(testnet18), frontend devs have to calculate the next cycle timestamp based on current cycle:

```json
        "version": "TEST.18.0",
        "current_time": 1674125023084,
        "current_cycle": 651,
```

Expected:
```json
        "version": "TEST.18.0",
        "current_time": 1674125023084,
        "current_cycle": 651,
        "current_cycle_time": 1674125013084,
        "next_cycle_time": 1674125033084,
```

aoudiamoncef:
  * [x] make tests compile
  * [x] make tests pass 

closes #3446

### DIFF
--- a/massa-api/src/public.rs
+++ b/massa-api/src/public.rs
@@ -393,17 +393,13 @@ impl MassaRpcServer for API<Public> {
         let current_cycle_time_result = if current_cycle == 0 {
             Ok(api_settings.genesis_timestamp)
         } else {
-            let previous_cycle = current_cycle.saturating_sub(1);
-            if previous_cycle == 0 {
-                Ok(cycle_duration)
-            } else {
-                cycle_duration.checked_mul(previous_cycle)
-            }
-            .and_then(|elapsed_time_before_current_cycle| {
-                api_settings
-                    .genesis_timestamp
-                    .checked_add(elapsed_time_before_current_cycle)
-            })
+            cycle_duration.checked_mul(current_cycle).and_then(
+                |elapsed_time_before_current_cycle| {
+                    api_settings
+                        .genesis_timestamp
+                        .checked_add(elapsed_time_before_current_cycle)
+                },
+            )
         };
 
         let current_cycle_time = match current_cycle_time_result {
--- a/massa-api/src/public.rs
+++ b/massa-api/src/public.rs
@@ -381,11 +381,46 @@ impl MassaRpcServer for API<Public> {
             })
             .collect::<BTreeMap<_, _>>();
 
+        let current_cycle = last_slot
+            .unwrap_or_else(|| Slot::new(0, 0))
+            .get_cycle(api_settings.periods_per_cycle);
+
+        let cycle_duration = match api_settings.t0.checked_mul(api_settings.periods_per_cycle) {
+            Ok(cycle_duration) => cycle_duration,
+            Err(e) => return Err(ApiError::TimeError(e).into()),
+        };
+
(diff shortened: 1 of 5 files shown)


massa-api/src/public.rs
```diff
@@ -381,11 +381,46 @@ impl MassaRpcServer for API<Public> {
...
+            .get_cycle(api_settings.periods_per_cycle);
+
+        let cycle_duration = match api_settings.t0.checked_mul(api_settings.periods_per_cycle) {
+            Ok(cycle_duration) => cycle_duration,
+            Err(e) => return Err(ApiError::TimeError(e).into()),
+        };
+
+        let previous_cycle = current_cycle.saturating_sub(1);
+
+        let elapsed_time_before_current_cycle = match if previous_cycle == 0 {
```
AurelienFT, line 395: I think you should destruct this pattern. I find the `match if` pretty hard to read

massa-api/src/public.rs
```diff
@@ -381,11 +381,46 @@ impl MassaRpcServer for API<Public> {
...
+            Ok(MassaTime::from(0))
+        } else {
+            cycle_duration.checked_mul(previous_cycle)
```
AurelienFT, line 398: Maybe I'm misunderstanding but I find the calculation to be false because we have a cycle 0. Let me try to explain with an example : 
```
genesis = 1000
t0 = 100
period_per_cycle = 100
slot = (280, 3)
current_cycle = 2
previous_cycle = 1
cycle_duration = (100 * 100) = 10000
elapsed_time_before_current_cycle = 10000
```

`current_cycle_time` is `10000 + 1000 = 11000` but should be in fact `21000` because in cycle 2 already 2 cycles has elapsed.

AurelienFT (changes requested):

massa-api/src/public.rs
aoudiamoncef: image

I have `genesis_timestamp` but how can I find the duration between It and the start of cycle 0 ?

AurelienFT: Start of cycle 0 == genesis_timestamp

massa-api/src/public.rs
```diff
@@ -381,11 +381,48 @@ impl MassaRpcServer for API<Public> {
...
+
+        let cycle_duration = match api_settings.t0.checked_mul(api_settings.periods_per_cycle) {
+            Ok(cycle_duration) => cycle_duration,
+            Err(e) => return Err(ApiError::TimeError(e).into()),
+        };
+
+        let current_cycle_time_result = if current_cycle == 0 {
+            Ok(api_settings.genesis_timestamp)
+        } else {
+            let previous_cycle = current_cycle.saturating_sub(1);
```
AurelienFT, line 396: I still don't understand why the sub here. Regarding the example I shown : https://github.com/massalabs/massa/pull/3472#discussion_r1090358207 If we are in cycle number 2 then it's the third because we start at 0. But you multiply `cycle_duration` only one time L.400. So you will point to the timestamp at start of cycle 1. You want the one at start of cycle 2.

Can you check in sandbox what's the behaviour please

aoudiamoncef: Sorry, it was a mistake  from my first impl

aoudiamoncef:
bors merge

AurelienFT:
Don't use bors we have a job failing for no reason merge it with the button. Will try to sort this out before the release.

PR #3473, opened 2023-01-26
Add lazy error handling in `get_block_slot_timestamp`

aoudiamoncef:
  * [x] make tests compile
  * [x] make tests pass 


aoudiamoncef:
I checked the entire codebase to find similar code, so this the only function affected.

AurelienFT (approved):
The gas costs is broken today you can still merge it. Will be fixed ASAP

PR #3475, opened 2023-01-27
Encapsulate user/sc address split in an enum

Ben-PH:
  * [x] make tests compile
  * [x] make tests pass 
* [ ] Update labnet to populate initial rolls and ledger with new address leader
* [ ] grep other parts of the codebase for wher "AU1" is used, and update.


Ben-PH:
CI test fails because it relies on an update of the test repo. (changing the "A1..." string to "AU1...")

massa-pos-exports/src/pos_changes.rs
```diff
@@ -99,7 +99,7 @@ impl Serializer<PoSChanges> for PoSChangesSerializer {
         self.u64_serializer
             .serialize(&(value.roll_changes.len() as u64), buffer)?;
         for (addr, roll) in value.roll_changes.iter() {
-            buffer.extend(addr.to_bytes());
+            buffer.extend(addr.prefixed_bytes());
```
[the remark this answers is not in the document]
Ben-PH: Thanks for the catch.

This should use the serializer. If I was to make some assumptions, the use of extend was chosen due to type being a raw-wrapper around an array. This PR doesn't change that all that much, however that might not be the case for too long. With that change, the use of a serializer to handle this step will be important

Ben-PH:
Unit tests pass locally. This PR depends on an update to `massa-unit-test-src`. The fix for local is to buuild that repo and put the result into the appropriate directory. As CI only builds from the repo, unedited, updating it is required to pass CI

massa-bootstrap/src/tests/tools.rs
```diff
@@ -238,7 +238,13 @@ pub fn get_random_final_state_bootstrap(
         sorted_ledger.insert(get_random_address(), get_random_ledger_entry());
     }
     // insert the last possible address to prevent the last cursor to move when testing the changes
-    sorted_ledger.insert(Address::from_bytes(&[255; 32]), get_random_ledger_entry());
+
+    let mut bytes = [255; 33];
+    bytes[0] = 1;
```
AurelienFT, line 243: It would be better to have a comment on why we are setting the first byte to a magic value

massa-execution-worker/src/context.rs
```diff
@@ -411,7 +411,7 @@ impl ExecutionContext {
             data.push(1u8);
         }
         // hash the seed to get a unique address
-        let address = Address(massa_hash::Hash::compute_from(&data));
+        let address = Address::User(UserAddress(massa_hash::Hash::compute_from(&data)));
```
AurelienFT, line 414: This should be an SC address here. I saw that in Address you used the same type for `UserAddress` and `SCAddress`. It's ok to define the structure SCAddress in the second time but we should open an issue to not forget before merging this but in this line we could use the variant SCAddress to avoid having a wrong line of code merged.

massa-pos-exports/src/pos_changes.rs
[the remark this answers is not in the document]
AurelienFT: It was just because we didn't took the time to make an `AddressSerializer` now that you made one it's nice to use it :)

Leo-Besancon:
Sure, I'll look at the implementation! 👍

massa-execution-worker/src/context.rs
Ben-PH: I misunderstood your request. Now has `Address::SC(SCAddress(...))`

Ben-PH:
I'll merge once @AurelienFT is happy with the remaining unresolved conversations.

AurelienFT (approved):
I resolved everything. Great work !!

PR #3479, opened 2023-01-30
Add thread number to Address debug output

Ben-PH:
  * [x] make tests compile
  * [x] make tests pass 

### DIFF
--- a/massa-models/src/address.rs
+++ b/massa-models/src/address.rs
@@ -1,5 +1,6 @@
 // Copyright (c) 2022 MASSA LABS <info@massa.net>
 
+use crate::config::THREAD_COUNT;
 use crate::error::ModelsError;
 use crate::prehash::PreHashed;
 use massa_hash::{Hash, HashDeserializer};
@@ -43,7 +44,13 @@ impl std::fmt::Display for Address {
 
 impl std::fmt::Debug for Address {
     fn fmt(&self, f: &mut std::fmt::Formatter) -> std::fmt::Result {
-        write!(f, "{}", self)
+        write!(
+            f,
+            "{} Thread {} of {}",
+            self,
+            self.get_thread(THREAD_COUNT),
+            THREAD_COUNT
+        )
     }
 }
 


Ben-PH:
Only places I could see things generated in the labnet logs were generated from this code here (it would seem)
https://github.com/massalabs/massa/blob/main/massa-pos-worker/src/draw.rs#L95

Eitu33 (approved):

massa-models/src/address.rs
```diff
@@ -43,7 +44,13 @@ impl std::fmt::Display for Address {
 
 impl std::fmt::Debug for Address {
     fn fmt(&self, f: &mut std::fmt::Formatter) -> std::fmt::Result {
-        write!(f, "{}", self)
+        write!(
+            f,
+            "{} Thread {} of {}",
+            self,
+            self.get_thread(THREAD_COUNT),
```
Eitu33, line 51: Actually this might be an issue when using test configs, since the same `config::THREAD_COUNT` will always be taken, but there really isn't any way to take this value as input other than saving in `Address`(which is not an option).

Eitu33: That could potentially be error inducing for a dev in the future, and we have no real solution for that.

Eitu33: @AurelienFT considering this do we want to have that?

Eitu33: Maybe re-implementing it in the test exports could be a solution even if it's not a great one.

Eitu33:
Please do not merge until above discussion is resolved

Eitu33:
This has been discussed and we've come to the conclusion that it would be better to not merge this for the following reasons:

- Trivial implementation is error inducing as it was raised in the discussion above
- Avoiding that would require a more complex testing-only implementation that can save a config parameter or take one on the fly
- `get_thread` already exists if a dev needs to the thread of an address, this is just a bonus

@Ben-PH I will let you close the PR unless you are against this

PR #3482, opened 2023-01-30
Provide context for Address dbg impl

Ben-PH:
  * [ ] make tests compile
  * [ ] make tests pass 


Ben-PH:
see: https://github.com/massalabs/massa/pull/3479#issuecomment-1408694720

AurelienFT (approved):

PR #3486, opened 2023-02-01
Reinstate pool tests

Ben-PH:
  * [ ] make tests compile
  * [ ] make tests pass 

### DIFF
--- a/massa-pool-worker/src/tests/scenario.rs
+++ b/massa-pool-worker/src/tests/scenario.rs
@@ -65,7 +65,7 @@ fn test_simple_get_operations() {
                     Ok(ControllerMsg::UnexecutedOpsAmong { response_tx, .. }) => {
                         response_tx.send(unexecuted_ops.clone()).unwrap();
                     }
-                    Ok(op) => panic!("expected `ControllerMsg::UnexecutedOpsAmong`, got {:?}", op),
+                    Ok(op) => panic!("Expected `ControllerMsg::UnexecutedOpsAmong`, got {:?}", op),
                     Err(_) => panic!("execution never called"),
                 }
                 match execution_receiver.recv_timeout(Duration::from_millis(100)) {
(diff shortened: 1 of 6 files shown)


Ben-PH:
67d2921760f420f0e22c5d82032aeb15ba5a5dcc adds some `dbg!`s, and this is the typical output:

```
     Running unittests src/lib.rs (target/debug/deps/massa_pool_worker-c92c862378c0b4ac)

running 1 test
[massa-pool-worker/src/tests/operation_pool_tests.rs:129] thread = 0
[massa-pool-worker/src/operation_pool.rs:224] &slot = Slot {
    period: 31,
    thread: 0,
}
[massa-execution-exports/src/test_exports/mock.rs:217] "never seen this not err..." = "never seen this not err..."
[massa-execution-exports/src/test_exports/mock.rs:218] response_rx.recv_timeout(Duration::from_millis(100)) = Err(
    Timeout,
)
thread 'tests::operation_pool_tests::test_pool' panicked at 'called `Result::unwrap()` on an `Err` value: Timeout', massa-execution-exports/src/test_exports/mock.rs:218:68
```

Still need to work out the source of the problem...

Ben-PH:
At this point, two of the methods are unignored, but the third one is not clear for me yet. @AurelienFT what do you think of merging these two so far, and do the rest in a follow up PR. It at least makes these two available in the meantime.

Ben-PH:
Basically, needed to comment out the balance check to get things through.

Ben-PH:
test is flakey: run this inside `massa-pool-worker` to (eventually) get a fail;

`while cargo test test_pool; do :; done`

AurelienFT:
I didn't follow well everything that has been going on but to give you maybe some context that might help you. We always run the test at root with feature testing (to have exports only when test cfg (no other way with cargo yet)).

Ben-PH:
updated script:

```sh
COUNTER = 0;
while cargo test test_pool -F testing --release ; do
    COUNTER=$[$COUNTER +1]
    :
done

echo $COUNTER;
```

@Eitu33 had a look at it with me, and it seems that this comes down to concurrent nature of the way the tests are setup?

The means to fix it is similar to the other two tests: Use the mock-receiver to intercept requests for data, and reply with hardcoded/intercepted data.

With that in mind, I'll do the final tidy up then mark for review.

massa-pool-worker/src/tests/scenario.rs
```diff
@@ -114,12 +123,22 @@ fn launch_basic_get_block_operation_execution_mock(
         use ControllerMsg::GetFinalAndCandidateBalance as GetFinal;
         use ControllerMsg::UnexecutedOpsAmong as Unexecuted;
```
Eitu33, line 124: Not related to these changes but this is a confusing and useless rename

Eitu33 (approved):

massa-pool-worker/src/tests/scenario.rs
Ben-PH: I've added a point to the follow up issue #3506 . That sufficient?

PR #3492, opened 2023-02-02
Update sandbox version

sydhds:
  * [ ] make tests compile
  * [ ] make tests pass 

### DIFF
--- a/massa-models/src/config/constants.rs
+++ b/massa-models/src/config/constants.rs
@@ -62,7 +62,7 @@ lazy_static::lazy_static! {
     /// node version
     pub static ref VERSION: Version = {
         if cfg!(feature = "sandbox") {
-            "SAND.0.20"
+            "SAND.20.0"
         } else {
             "TEST.20.0"
         }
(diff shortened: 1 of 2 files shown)


massa-models/src/config/constants.rs
```diff
@@ -62,7 +62,7 @@ lazy_static::lazy_static! {
     /// node version
     pub static ref VERSION: Version = {
         if cfg!(feature = "sandbox") {
-            "SAND.0.0"
+            "SAND.0.20"
```
Eitu33, line 65:
```suggestion
            "SAND.20.0"
```

It would be better to put the corresponding testnet version and update the compatibility check since this is going to be merge into the release branch

sydhds: Done

massa-models/src/version.rs
```diff
@@ -189,7 +189,7 @@ impl Deserializer<Version> for VersionDeserializer {
 impl Version {
     /// true if instance and major are the same
     pub fn is_compatible(&self, other: &Version) -> bool {
-        self.instance == other.instance && self.major == other.major && other.minor >= 1
+        self.instance == other.instance && self.major == other.major && other.major >= 1
```
Eitu33, line 192:
```suggestion
        self.instance == other.instance && self.major == other.major
```

PR #3495, opened 2023-02-03
Fix small typos in CLI

aoudiamoncef:
  * [x] make tests compile
  * [x] make tests pass 


Leo-Besancon (approved):
Boostrap just seems a better world, not sure if I want to approve this ;)

PR #3502, opened 2023-02-06
Add doc-test with testing feature to CI steps

Ben-PH:
  * [ ] make tests compile
  * [ ] make tests pass 

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -117,6 +117,10 @@ jobs:
         with:
           command: nextest
           args: run --features testing --retries 10
+      - uses: actions-rs/cargo@v1
+        with:
+          command: test
+          args: --features testing --doc
 
   build: # quick hack because bors wrongly detect matrix jobs status
     needs: full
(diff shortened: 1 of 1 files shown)


AurelienFT:
@Ben-PH Why did you moved it out ? Because it will not benefit from the cache from other tasks but it also not be run on the matrix with multiple OS.

Ben-PH:
My thinking is that doctests are at a much higher level, and are platform agnostic, so I thought just the one OS is sufficient.

Losing the cache, yeah, probably not the best.

Writing good CI scripts is very much in the "something I need to develop as a skill" basket for me, so I'm playing around a bit.

Some of the things in my mind:

- Run it parallel to the unit tests might be good. Yes they are a unit test of sorts, but a different class of tests.
- No need to check across platforms: Doctests are not meant to be thorough, that's for the unit tests. Doc tests serve a dual purpose of sniff-testing for correctness at a basic level, and providing use-case-examples in the documentation. Leave the x-platform stuff to the full-test job
- didn't think that a separate doc-test would take so long to DL and compile.

On balance, the second commit doesn't seem to be a good idea, no? Lesson to be learned.

sydhds:
Not sure to see the benefits of separating doctests from other unit tests...

Ben-PH:
Was thinking of having them in parallel with the unit tests, but the gains are absolute minimal, the costs massive. Not worth it, I think.

sydhds (approved):

PR #3507, opened 2023-02-07
Tidy up and restructure of pool unit-tests

Ben-PH:
  * [x] make tests compile
  * [x] make tests pass 

### DIFF
--- a/massa-pool-worker/src/tests/operation_pool_tests.rs
+++ b/massa-pool-worker/src/tests/operation_pool_tests.rs
@@ -17,24 +17,19 @@
 //! latest period given his own thread. All operation which doesn't fit these
 //! requirements are "irrelevant"
 //!
+use crate::tests::tools::OpGenerator;
+
 use super::tools::{create_some_operations, operation_pool_test, pool_test};
 use massa_execution_exports::test_exports::MockExecutionControllerMessage;
-use massa_models::{
-    address::Address,
-    amount::Amount,
-    operation::{Operation, OperationSerializer, OperationType, SecureShareOperation},
-    prehash::PreHashMap,
-    secure_share::SecureShareContent,
-    slot::Slot,
-};
+use massa_models::{amount::Amount, prehash::PreHashMap, slot::Slot};
 use massa_pool_exports::PoolConfig;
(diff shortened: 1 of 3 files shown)


massa-pool-worker/src/tests/operation_pool_tests.rs
```diff
@@ -52,174 +42,164 @@ fn test_add_irrelevant_operation() {
...
-                ops, response_tx, ..
-            }) => {
-                response_tx.send(ops).unwrap();
+        |mut pool_manager, mut pool, execution_receiver, storage_base| {
+            // generate (id, transactions, range of validity) by threads
+            let mut thread_tx_lists = vec![Vec::new(); pool_config.thread_count as usize];
+
+            for i in 0..18 {
+                let expire_period: u64 = 40 + i;
+                let op = OpGenerator::default()
```
Ben-PH, line 65: Ideally this code could be handled by a helper method, as similar work is done in the other test functions. This is not so straight forward, as there are subtle differences between the tests that would make this doable, but in my view "clever for the sake of clever", and not of significant benefit in this context.

AurelienFT:
This PR Is still in draft what's the current state ?

Ben-PH:
Needed to ask you about: 

a) the flakey tests. They fail every now and then, but generally, the CI accounts for that with its retries. I believe it will reliably fail if there is a bug that it's checking for.
b) It was mentioned that it would be nice to have a helper method that handles the setup based on the way it's called, with an iteration count. The nuances of how things are setup differently makes this... tricky. Sure it can be done, but I'm not sure it's a great use of time, and not sure if it would make the test code more readable-maintainable anyway.

Not sure what the CI failure is about, though.

AurelienFT:
Thanks for precisions @Ben-PH let's review it and merge it as it is for now and iterate in the futur to avoid letting this PR die. Can you update on testnet_20 to fix the gas_calibration check and put it as ready ?

Ben-PH:
re-ran the failed CI test. Passed. marking as ready.

AurelienFT (approved):
That's really nice, we cover a lot more cases now. Nice work !

PR #3511, opened 2023-02-08
Add client error message on bootstrap rejected because of blacklist/whitelist

### LINKED ISSUE #3292: Better error message on bootstrap rejected because of blacklist/whitelist
```
2022-12-01T17:38:55.403601Z  INFO massa_node: Node version : TEST.17.0
2022-12-01T17:38:55.435090Z  INFO massa_bootstrap::client: Start bootstrapping from 195.134.167.226:31245
2022-12-01T17:38:55.435808Z  WARN massa_bootstrap::client: Error while bootstrapping: io error: early eof
```

"early eof" is not very informative of the problem which is lack of whitelisting.

Maybe do it like we do elsewhere: send an error message from the server if there are slots available, otherwise just close

aoudiamoncef:
  * [x] make tests compile
  * [x] make tests pass 

closes #3292


aoudiamoncef:
@Eitu33 

This was quickly tested locally and with two physical nodes

Results with a simple telnet calls:

blacklist:
```shell
telnet 192.168.1.190 31245
Trying 192.168.1.190...
Connected to 192.168.1.190.
Escape character is '^]'.
?q??-?h?,Gs?S%?`???9????'?Q[?g????T?T"y+>F?^ox?R????{??" IP 192.168.1.190 is blacklisted.Connection closed by foreign host.
```

not in whitelist:
```shell
telnet 192.168.1.190 31245
Trying 192.168.1.190...
Connected to 192.168.1.190.
Escape character is '^]'.
??S?.
_???;R??M?o?Z?????c?	?t?f??)???cy?a?k?a5?N?f??A?A whitelist exists and the IP 192.168.1.190 is not whitelisted.Connection closed by foreign host.
```

massa-bootstrap/src/server.rs
```diff
@@ -300,6 +318,49 @@ impl BootstrapServer {
...
+        duplex: Duplex,
+        remote_address: SocketAddr,
+        whitelist: &Option<HashSet<IpAddr>>,
+        blacklist: &Option<HashSet<IpAddr>>,
+    ) -> io::Result<(Duplex, SocketAddr)> {
+        let ip = normalize_ip(remote_address.ip());
+        // whether the peer IP address is blacklisted, send back an error message
+        if let Some(ip_list) = &blacklist && ip_list.contains(&ip) {
+            let config = self.bootstrap_config.clone();
+            let mut server = BootstrapServerBinder::new(duplex, self.keypair.clone(), config.max_bytes_read_write, config.max_bootstrap_message_size, config.thread_count, config.max_datastore_key_length, config.randomness_size_bytes, config.consensus_bootstrap_part_size);
```
Eitu33, line 334: Creating a new `BootstrapServerBinder` for one message is not optimal, please take it as a function argument. This will also avoid a lot of duplicated code.

aoudiamoncef: Before calling this function, there is no instance of it. It could be not used at all, this why it's created only if needed. 
It's implemented that way in the rest of the `run`function.

aoudiamoncef: May be we should create a single instance and use it in `run` function ?

Eitu33: I see, the code is still seriously duplicated though. The only difference between the two cases are the error messages, the binder creation code can be written only once and the conditional statements should only take care of the messages content.

aoudiamoncef: i'll refactor this ASAP. We could share a single instance of `BootstrapServerBinder`:

```rust
let (mut server, remote_addr) = if res_connection.is_ok() {
                        let (dplx, remote_addr) = res_connection.unwrap();
                        let server = BootstrapServerBinder::new(dplx, self.keypair.clone(), self.bootstrap_config.max_bytes_read_write, self.bootstrap_config.max_bootstrap_message_size, self.bootstrap_config.thread_count, self.bootstrap_config.max_datastore_key_length, self.bootstrap_config.randomness_size_bytes, self.bootstrap_config.consensus_bootstrap_part_size);
                        // we didn't test whether Ip is allowed
```

aoudiamoncef: done

massa-bootstrap/src/server.rs
```diff
@@ -187,14 +188,38 @@ impl BootstrapServer {
                 }
 
                 // listener
-                res_connection = listener.accept(&whitelist, &blacklist) => {
-                    let (dplx, remote_addr) = if res_connection.is_ok() {
-                        res_connection.unwrap()
+                res_connection = listener.accept() => {
+                        let (mut server, remote_addr) = if res_connection.is_ok() {
```
Eitu33, line 192: Indentation seems wrong here, maybe you didn't run fmt but its probably the formatter bugging out because of the select or the test scopes. If that's the case you have to do it manually.

Eitu33: Otherwise lgtm, will merge when this is fixed 👍

Eitu33: (And CI is passing)

massa-bootstrap/src/server.rs
```diff
@@ -187,14 +188,38 @@ impl BootstrapServer {
...
+
+                        #[cfg(not(test))]
+                        {
+                            // check whether incoming peer IP is allowed or return an error which is ignored
+                            let res = self.is_ip_allowed(remote_addr, server, &whitelist, &blacklist).await;
+                            if res.is_ok() {
+                                res.unwrap()
+                            } else {
+                                continue
+                            }
```
Ben-PH, line 217: This might be a nice place to use let-else:

```suggestion
                            let Ok(res) = self.is_ip_allowed(remote_addr, server, &whitelist, &blacklist).await else {
                                continue;
                            };
```

Ben-PH: (wrote this out manually in github. Not sure if the `;` on line 213 is needed)

aoudiamoncef:
```
                        let (mut server, remote_addr) = if let Ok(res) = res_connection {
                            let server = BootstrapServerBinder::new(
                                res.0,
                                self.keypair.clone(),
                                self.bootstrap_config.max_bytes_read_write,
                                self.bootstrap_config.max_bootstrap_message_size,
                                self.bootstrap_config.thread_count,
                                self.bootstrap_config.max_datastore_key_length,
                                self.bootstrap_config.randomness_size_bytes,
                                self.bootstrap_config.consensus_bootstrap_part_size);
                        // we didn't test whether Ip is allowed
                        #[cfg(test)]
                        {
                            (server, res.1)
                        }

                        #[cfg(not(test))]
                        {
                            // check whether incoming peer IP is allowed or return an error which is ignored
                            let Ok(res) = self.is_ip_allowed(res.1, server, &whitelist, &blacklist).await else {
                                continue;
                            };
                            res
                        }
                    } else {
                        continue
                    };
```

@Eitu33 @Ben-PH is it more readable like this ? (not sure)

aoudiamoncef: done

massa-bootstrap/src/server.rs
aoudiamoncef: done

massa-bootstrap/src/server.rs
Ben-PH: I think so, yeah. The dropping of the `unwrap` is nice in the old line 193 is a nice extra.

aoudiamoncef: IMHO, it's already done ✅

Eitu33:
We can merge without gas costs checks passing for now but the ubuntu CI fails also on unused, please fix before we merge

aoudiamoncef:
@Eitu33 done 💯

Eitu33 (approved):

PR #3549, opened 2023-02-09
Initial vesting

modship:
  * [x] make tests compile
  * [x] make tests pass 


massa-node/base_config/initial_vesting.json
```diff
@@ -0,0 +1,22 @@
+{
+    "Address": [
+        {
+            "slot": {
+                "start_period": 0,
+                "end_period": 99,
+                "thread": 0
```
AurelienFT, line 7: A slot is a `period` and a `thread` you can't have this 3 fields in your slot. I think it's better to have a `start_slot` and `end_slot` and both have `period` and `thread`.

massa-execution-worker/src/execution.rs
```diff
@@ -1341,4 +1356,54 @@ impl ExecutionState {
     pub fn get_address_future_deferred_credits(&self, address: &Address) -> BTreeMap<Slot, Amount> {
         context_guard!(self).get_address_future_deferred_credits(address, self.config.thread_count)
     }
+
+    /// Initialize the hashmap of addresses from the vesting file
+    fn init_vesting_registry(
+        config: &ExecutionConfig,
+    ) -> Result<PreHashMap<Address, Vec<VestingRange>>, ExecutionError> {
+        let mut hashmap: PreHashMap<Address, Vec<VestingRange>> = serde_json::from_str(
```
AurelienFT, line 1364: Instead of having mut here and mut in the `for` why not using a map on your initialization to have no mutability

massa-execution-worker/src/execution.rs
```diff
@@ -1341,4 +1356,54 @@ impl ExecutionState {
...
+        .map_err(|err| {
+            ExecutionError::InitVestingError(format!(
+                "error on deserialize initial vesting file  {}",
+                err
+            ))
+        })?;
+
+        for v in hashmap.values_mut() {
+            let mut current_index = 1;
+            let temp_list = v.clone();
```
AurelienFT, line 1381: Instead of making a clone and index fetching for the next could you use `.next()` method of an element ?

massa-execution-worker/src/execution.rs
modship: if we want to modify end_slot on a VestingRange we need mut on map() too

massa-execution-worker/src/execution.rs
modship: i don't think so, next() need a mutable reference on the same iter.

Fox example this do not compile because iter_vesting.next() take the ownership on iter_vesting. 

```rust
.map(|mut res: PreHashMap<Address, Vec<VestingRange>>| {
            let _ = res.values_mut().map(|value| {
                let mut iter_vesting = value.iter_mut();

                for vesting_slot in iter_vesting {
                    if let Some(next) = iter_vesting.next() {
                        if let Ok(prev_slot) = next.start_slot
                            .get_prev_slot(config.thread_count) {
                            vesting_slot.end_slot = prev_slot;
                        } else {
                            error!("error on get prev slot for init vesting : {}", next.start_slot)
                        }
                    } else {
                        vesting_slot.end_slot = Slot {
                            thread: config.thread_count,
                            period: u64::MAX,
                        };
                    }
                }
            });
            res
        })

```

AurelienFT: I proposed a refactor here : https://github.com/massalabs/massa/pull/3549/commits/aecfe3ae70c795afcee948d0b74e1c4442df834e

AurelienFT:
I think it's better to fix the CI before saying it's ready for review to be "more sure" that we review something working

massa-execution-worker/src/context.rs
```diff
@@ -575,24 +596,46 @@ impl ExecutionContext {
...
+            if check_rights && !self.has_write_rights_on(from_addr) {
+                return Err(ExecutionError::RuntimeError(format!(
+                    "spending from address {} is not allowed in this context",
+                    from_addr
+                )));
+            }
+
+            if let Some(current_slot) = current_slot {
+                // control vesting min_balance for sender address
+                if let Some(vesting_range) = VestingRange::find_vesting_range(
```
AurelienFT, line 619: For me `find_vesting_range()` shouldn't be in `VestingRange` as it don't use any of the member of the structure and doesn't do anything related to the structure. It should be on the type `PreHashMap<Addresss, VestingRange>` but if you don't do a special type for it you can put it in the context.

massa-execution-worker/src/context.rs
```diff
@@ -575,24 +596,46 @@ impl ExecutionContext {
...
+                    from_addr,
+                    current_slot,
+                    &self.vesting_registry,
+                ) {
+                    let new_balance = self
```
AurelienFT, line 624: The minimum balance for an address shouldn't be only the balance but the formula specified in the discussion : 
`min_balance = (rolls * roll_price) + balance + deferred_credits`

massa-models/src/slot.rs
```diff
@@ -344,3 +348,46 @@ impl std::fmt::Display for IndexedSlot {
         writeln!(f, "Slot: {}, Index: {}", self.slot, self.index)
     }
 }
+
+/// Represent a vesting range
+#[derive(Clone, Copy, Deserialize, Serialize, Debug)]
+pub struct VestingRange {
```
AurelienFT, line 354: I think it should be on a separate file.

massa-execution-worker/src/execution.rs
```diff
@@ -385,13 +402,25 @@ impl ExecutionState {
...
         };
 
+        // control vesting max_rolls for buyer address
+        if let Some(vesting_range) =
+            VestingRange::find_vesting_range(&buyer_addr, current_slot, &self.vesting_registry)
+        {
+            let rolls = self.get_final_and_candidate_rolls(&buyer_addr);
+            // (candidate_rolls + amount to buy)
+            if (rolls.1 + roll_count) >= vesting_range.max_rolls {
+                return Err(ExecutionError::VestingError("max rolls".to_string()));
```
AurelienFT, line 420: I think specifying the values in the error could help the user

massa-execution-worker/src/context.rs
```diff
@@ -575,24 +596,46 @@ impl ExecutionContext {
...
+                        .get_balance(from_addr)
+                        .ok_or_else(|| ExecutionError::RuntimeError(format!("spending address {} not found", from_addr)))?
+                        .checked_sub(amount)
+                        .ok_or_else(|| {
+                            ExecutionError::RuntimeError(format!("failed check transfer {} from spending address {} due to insufficient balance {}", amount, from_addr, self
+                                .get_balance(from_addr).unwrap_or_default()))
+                        })?;
+                    if new_balance < vesting_range.min_balance {
+                        return Err(ExecutionError::VestingError("min_balance".to_string()));
```
AurelienFT, line 633: I think specifying the values in the error could help the user.

AurelienFT (commented):
Nice job for your first PR ! Well done. Few adjustements to do. Also before re-asking the second review after fix can you tests complex cases with rolls when trying transfers etc on sandbox and/or UT

massa-execution-worker/src/context.rs
AurelienFT: Ok then you can just do a function inside of anything private in this file.

AurelienFT:
@modship Can you fix the tests and make some real world test with a sandbox node ? (We can call to setup it together)

modship:
> @modship Can you fix the tests and make some real world test with a sandbox node ? (We can call to setup it together)

@AurelienFT  call me when you want

massa-execution-worker/src/context.rs
```diff
@@ -575,24 +596,69 @@ impl ExecutionContext {
...
+                    self.config.periods_per_cycle,
+                    *current_slot,
+                );
+                let Some(exec_info) = vec.first() else {
+                    return Err(ExecutionError::VestingError(format!("can not get address info cycle for {}", from_addr)));
+                };
+
+                let rolls_value = exec_info
+                    .active_rolls
+                    .map(|rolls| Amount::from_raw(rolls * self.config.roll_price.to_raw()))
```
AurelienFT, line 639:
```suggestion
                    .map(|rolls| self.config.roll_price.saturating_mul_u64(rolls))
```

massa-node/base_config/initial_vesting.json
```diff
@@ -0,0 +1,46 @@
...
+        },
+        {
+            "slot": {
+                "period": 30,
+                "thread": 6
+            },
+            "min_balance": "0",
+            "max_rolls": 18446744073709551615
+        }
+    ]
```
AurelienFT, line 45: Can you increase the periods number of each vesting range so that we have few days on each (a period is 16 secondes) ? And also add a address with coins vested until 1 milion period without any roll authorized so that we have a usecase that is similar to decentralization pool ?

massa-execution-worker/src/execution.rs
```diff
@@ -385,13 +398,27 @@ impl ExecutionState {
...
             OperationType::RollBuy { roll_count } => roll_count,
             _ => panic!("unexpected operation type"),
         };
 
+        // control vesting max_rolls for buyer address
+        if let Some(vesting_range) = self.find_vesting_range(&buyer_addr, &current_slot) {
+            let rolls = self.get_final_and_candidate_rolls(&buyer_addr);
+            // (candidate_rolls + amount to buy)
+            let max_rolls = rolls.1 + roll_count;
+            if max_rolls >= vesting_range.max_rolls {
```
AurelienFT, line 414: Should be >

massa-node/base_config/initial_vesting.json
modship: I set in vesting file : 
First address range is 0 - 5 - 10 Days
The second is 0 - 5 days
The third is 0 - 185 days (1_000_000 period)

AurelienFT (approved):
Well done ! Very nice job. You can merge it even if gas costs fail

modship:
> Well done ! Very nice job. You can merge it even if gas costs fail

Ok thx :)

PR #3554, opened 2023-02-10
Benph/bootstrap thread migration

Ben-PH:
  * [ ] make tests compile
  * [ ] make tests pass 

### DIFF
--- a/massa-bootstrap/src/server.rs
+++ b/massa-bootstrap/src/server.rs
@@ -64,32 +64,31 @@ pub async fn start_bootstrap_server(
     version: Version,
 ) -> Result<Option<BootstrapManager>, BootstrapError> {
     massa_trace!("bootstrap.lib.start_bootstrap_server", {});
-    if let Some(bind) = bootstrap_config.bind {
-        let (manager_tx, manager_rx) = mpsc::channel::<()>(1);
+    let Some(bind) = bootstrap_config.bind else {
+        return Ok(None);
+    };
+    let (manager_tx, manager_rx) = mpsc::channel::<()>(1);
 
-        let join_handle = tokio::spawn(async move {
-            BootstrapServer {
-                consensus_controller,
-                network_command_sender,
-                final_state,
-                establisher,
-                manager_rx,
-                bind,
-                keypair,
-                version,
-                ip_hist_map: HashMap::with_capacity(bootstrap_config.ip_list_max_size),
-                bootstrap_config,
-            }
-            .run()
-            .await
-        });
-        Ok(Some(BootstrapManager {
-            join_handle,
-            manager_tx,
-        }))
-    } else {
-        Ok(None)
-    }
+    let join_handle = tokio::spawn(async move {
+        BootstrapServer {
+            consensus_controller,
+            network_command_sender,
+            final_state,
+            establisher,
+            manager_rx,
+            bind,
+            keypair,
+            version,
+            ip_hist_map: HashMap::with_capacity(bootstrap_config.ip_list_max_size),
+            bootstrap_config,
+        }
+        .run()
+        .await
+    });
+    Ok(Some(BootstrapManager {
+        join_handle,
+        manager_tx,
+    }))
 }
 
 struct BootstrapServer {
@@ -188,12 +187,12 @@ impl BootstrapServer {
 
                 // listener
                 res_connection = listener.accept(&whitelist, &blacklist) => {
-                    let (dplx, remote_addr) = if res_connection.is_ok() {
-                        res_connection.unwrap()
-                    } else {
+                    let Ok((dplx, remote_addr)) =  res_connection else {
                         continue;
                     };
-                    if bootstrap_sessions.len() < self.bootstrap_config.max_simultaneous_bootstraps.try_into().map_err(|_| BootstrapError::GeneralError("Fail to convert u32 to usize".to_string()))? {
+
+                    let max_bootstraps = self.bootstrap_config.max_simultaneous_bootstraps.try_into().map_err(|_| BootstrapError::GeneralError("Fail to convert u32 to usize".to_string()))?;
+                    if bootstrap_sessions.len() < max_bootstraps {
 
                         massa_trace!("bootstrap.lib.run.select.accept", {"remote_addr": remote_addr});
                         let now = Instant::now();
@@ -212,11 +211,29 @@ impl BootstrapServer {
                         match self.ip_hist_map.entry(remote_addr.ip()) {
                             hash_map::Entry::Occupied(mut occ) => {
                                 if now.duration_since(*occ.get()) <= per_ip_min_interval {
-                                    let mut server = BootstrapServerBinder::new(dplx, self.keypair.clone(), self.bootstrap_config.max_bytes_read_write, self.bootstrap_config.max_bootstrap_message_size, self.bootstrap_config.thread_count, self.bootstrap_config.max_datastore_key_length, self.bootstrap_config.randomness_size_bytes, self.bootstrap_config.consensus_bootstrap_part_size);
-                                    let _ = match tokio::time::timeout(self.bootstrap_config.write_error_timeout.into(), server.send(BootstrapServerMessage::BootstrapError {
-                                        error:
-                                        format!("Your last bootstrap on this server was {} ago and you have to wait {} before retrying.", format_duration(occ.get().elapsed()), format_duration(per_ip_min_interval.saturating_sub(occ.get().elapsed())))
-                                    })).await {
+                                    let mut server = BootstrapServerBinder::new(
+                                        dplx,
+                                        self.keypair.clone(),
+                                        self.bootstrap_config.max_bytes_read_write,
+                                        self.bootstrap_config.max_bootstrap_message_size,
+                                        self.bootstrap_config.thread_count,
+                                        self.bootstrap_config.max_datastore_key_length,
+                                        self.bootstrap_config.randomness_size_bytes,
+                                        self.bootstrap_config.consensus_bootstrap_part_size
+                                    );
+                                    let send_timeout = tokio::time::timeout(
+                                        self.bootstrap_config.write_error_timeout.into(),
+                                        server.send(
+                                            BootstrapServerMessage::BootstrapError {
+                                                error: format!(
+                                                    "Your last bootstrap on this server was {} ago and you have to wait {} before retrying.",
+                                                    format_duration(occ.get().elapsed()),
+                                                    format_duration(per_ip_min_interval.saturating_sub(occ.get().elapsed()))
+                                                )
+                                            }
+                                        )).await;
+
+                                    let _ = match send_timeout {
                                         Err(_) => Err(std::io::Error::new(std::io::ErrorKind::TimedOut, "bootstrap error too early retry bootstrap send timed out").into()),
                                         Ok(Err(e)) => Err(e),
                                         Ok(Ok(_)) => Ok(()),
@@ -257,9 +274,29 @@ impl BootstrapServer {
                         let config = self.bootstrap_config.clone();
 
                         bootstrap_sessions.push(async move {
-                            let mut server = BootstrapServerBinder::new(dplx, keypair, config.max_bytes_read_write, config.max_bootstrap_message_size, config.thread_count, config.max_datastore_key_length, config.randomness_size_bytes, config.consensus_bootstrap_part_size);
+                            let mut server = BootstrapServerBinder::new(
+                                dplx,
+                                keypair,
+                                config.max_bytes_read_write,
+                                config.max_bootstrap_message_size,
+                                config.thread_count,
+                                config.max_datastore_key_length,
+                                config.randomness_size_bytes,
+                                config.consensus_bootstrap_part_size
+                            );
+
                             debug!("awaiting on bootstrap of peer {}", remote_addr);
-                            match tokio::time::timeout(config.bootstrap_timeout.into(), manage_bootstrap(&config, &mut server, data_execution, version, consensus_command_sender, network_command_sender)).await {
+                            match tokio::time::timeout(
+                                config.bootstrap_timeout.into(),
+                                manage_bootstrap(
+                                    &config,
+                                    &mut server,
+                                    data_execution,
+                                    version,
+                                    consensus_command_sender,
+                                    network_command_sender
+                                )
+                            ).await {
                                 Ok(mgmt) => match mgmt {
                                     Ok(_) => {
                                         info!("bootstrapped peer {}", remote_addr)
@@ -282,7 +319,16 @@ impl BootstrapServer {
                         massa_trace!("bootstrap.session.started", {"active_count": bootstrap_sessions.len()});
                     } else {
                         let config = self.bootstrap_config.clone();
-                        let mut server = BootstrapServerBinder::new(dplx, self.keypair.clone(), config.max_bytes_read_write, config.max_bootstrap_message_size, config.thread_count, config.max_datastore_key_length, config.randomness_size_bytes, config.consensus_bootstrap_part_size);
+                        let mut server = BootstrapServerBinder::new(
+                            dplx,
+                            self.keypair.clone(),
+                            config.max_bytes_read_write,
+                            config.max_bootstrap_message_size,
+                            config.thread_count,
+                            config.max_datastore_key_length,
+                            config.randomness_size_bytes,
+                            config.consensus_bootstrap_part_size
+                        );
                         let _ = match tokio::time::timeout(config.clone().write_error_timeout.into(), server.send(BootstrapServerMessage::BootstrapError {
                             error: "Bootstrap failed because the bootstrap server currently has no slots available.".to_string()
                         })).await {
@@ -295,6 +341,7 @@ impl BootstrapServer {
                 }
             }
         }
+
         // wait for bootstrap sessions to finish
         while bootstrap_sessions.next().await.is_some() {}
 
--- a/massa-bootstrap/src/server.rs
+++ b/massa-bootstrap/src/server.rs
@@ -25,6 +25,7 @@ use tracing::{debug, info, warn};
 
 use crate::{
     error::BootstrapError,
+    establisher::types::Duplex,
     messages::{BootstrapClientMessage, BootstrapServerMessage},
     server_binder::BootstrapServerBinder,
     tools::normalize_ip,
@@ -273,49 +274,19 @@ impl BootstrapServer {
                         let keypair = self.keypair.clone();
                         let config = self.bootstrap_config.clone();
 
-                        bootstrap_sessions.push(async move {
-                            let mut server = BootstrapServerBinder::new(
+                        bootstrap_sessions.push(
+                            push_bootstrap_session(
                                 dplx,
                                 keypair,
-                                config.max_bytes_read_write,
-                                config.max_bootstrap_message_size,
-                                config.thread_count,
-                                config.max_datastore_key_length,
-                                config.randomness_size_bytes,
-                                config.consensus_bootstrap_part_size
-                            );
+                                config,
+                                remote_addr,
+                                data_execution,
+                                version,
+                                consensus_command_sender,
+                                network_command_sender
+                            )
+                        );
 
-                            debug!("awaiting on bootstrap of peer {}", remote_addr);
-                            match tokio::time::timeout(
-                                config.bootstrap_timeout.into(),
-                                manage_bootstrap(
-                                    &config,
-                                    &mut server,
-                                    data_execution,
-                                    version,
-                                    consensus_command_sender,
-                                    network_command_sender
-                                )
-                            ).await {
-                                Ok(mgmt) => match mgmt {
-                                    Ok(_) => {
-                                        info!("bootstrapped peer {}", remote_addr)
-                                    },
-                                    Err(BootstrapError::ReceivedError(error)) => debug!("bootstrap serving error received from peer {}: {}", remote_addr, error),
-                                    Err(err) => {
-                                        debug!("bootstrap serving error for peer {}: {}", remote_addr, err);
-                                        // We allow unused result because we don't care if an error is thrown when sending the error message to the server we will close the socket anyway.
-                                        let _ = tokio::time::timeout(config.write_error_timeout.into(), server.send(BootstrapServerMessage::BootstrapError { error: err.to_string() })).await;
-                                    },
-                                }
-                                Err(_timeout) => {
-                                    debug!("bootstrap timeout for peer {}", remote_addr);
-                                    // We allow unused result because we don't care if an error is thrown when sending the error message to the server we will close the socket anyway.
-                                    let _ = tokio::time::timeout(config.write_error_timeout.into(), server.send(BootstrapServerMessage::BootstrapError { error: format!("Bootstrap process timedout ({})", format_duration(config.bootstrap_timeout.to_duration())) })).await;
-                                }
-                            }
-
-                        });
                         massa_trace!("bootstrap.session.started", {"active_count": bootstrap_sessions.len()});
                     } else {
                         let config = self.bootstrap_config.clone();
@@ -346,6 +317,81 @@ impl BootstrapServer {
         while bootstrap_sessions.next().await.is_some() {}
 
         Ok(())
+    }
+}
+
+/// TODO: Doc-comment me
+async fn push_bootstrap_session(
+    dplx: Duplex,
+    keypair: KeyPair,
+    config: BootstrapConfig,
+    remote_addr: SocketAddr,
+    data_execution: Arc<RwLock<FinalState>>,
+    version: Version,
+    consensus_command_sender: Box<dyn ConsensusController>,
+    network_command_sender: NetworkCommandSender,
+) {
+    let mut server = BootstrapServerBinder::new(
+        dplx,
+        keypair,
+        config.max_bytes_read_write,
+        config.max_bootstrap_message_size,
+        config.thread_count,
+        config.max_datastore_key_length,
+        config.randomness_size_bytes,
+        config.consensus_bootstrap_part_size,
+    );
+
+    debug!("awaiting on bootstrap of peer {}", remote_addr);
+    match tokio::time::timeout(
+        config.bootstrap_timeout.into(),
+        manage_bootstrap(
+            &config,
+            &mut server,
+            data_execution,
+            version,
+            consensus_command_sender,
+            network_command_sender,
+        ),
+    )
+    .await
+    {
+        Ok(mgmt) => match mgmt {
+            Ok(_) => {
+                info!("bootstrapped peer {}", remote_addr)
+            }
+            Err(BootstrapError::ReceivedError(error)) => debug!(
+                "bootstrap serving error received from peer {}: {}",
+                remote_addr, error
+            ),
+            Err(err) => {
+                debug!("bootstrap serving error for peer {}: {}", remote_addr, err);
+                // We allow unused result because we don't care if an error is thrown when
+                // sending the error message to the server we will close the socket anyway.
+                let _ = tokio::time::timeout(
+                    config.write_error_timeout.into(),
+                    server.send(BootstrapServerMessage::BootstrapError {
+                        error: err.to_string(),
+                    }),
+                )
+                .await;
+            }
+        },
+        Err(_timeout) => {
+            debug!("bootstrap timeout for peer {}", remote_addr);
+            // We allow unused result because we don't care if an error is thrown when
+            // sending the error message to the server we will close the socket anyway.
+            let _ = tokio::time::timeout(
+                config.write_error_timeout.into(),
+                server.send(BootstrapServerMessage::BootstrapError {
+                    error: format!(
+                        "Bootstrap process timedout ({})",
+                        format_duration(config.bootstrap_timeout.to_duration())
+                    ),
+                }),
+            )
+            .await;
+        }
     }
 }
 
--- a/massa-node/src/main.rs
+++ b/massa-node/src/main.rs
@@ -791,6 +791,7 @@ async fn run(args: Args) -> anyhow::Result<()> {
     // load or create wallet, asking for password if necessary
     let node_wallet = load_wallet(args.password, &SETTINGS.factory.staking_wallet_path)?;
 
+    // HYPOTHESIS: This loop only runs once; why would you need `launch` in the loop?
     loop {
         let (
             consensus_event_receiver,


Ben-PH:
This implementation keeps async around, but uses `tokio::spawn` to run the the handling of a bootstrap connection, instead of pushing it onto a `FuturesUnordered` and keeping it on the same thread. Remov ing the async completely is non-trivial:

- Bandwidth throttling uses contructs that are coupled to an async context.
- Interconnected parts with the client code.
- use of timeouts, e.g. limiting the connection time so clients don't DoS the server.

These can, of course, be overcome in the `(•_•)     ( •_•)>⌐■-■     (⌐■_■)` future.

Instead of checking the length of the (now removed) `FuturesUnordered` list, we track the number of active threads through a counter. It's incremented on spawn, and decremented when it handles the `Complete` message.

It still uses the select macro with these branches:

- controller message: break if the message is a `Stop` decrement active bootstraps if the message is a `Complete`
- cache updater: As before, runs the black/white list updater on the primary thread. Will look into moving this into dedicated thread...
- listener: the main change here is that the once the server decides to provide the _actual_ bootstrapping services to a client, that is done on a separate thread using `tokio::spawn`

Ben-PH:
Issues with implementation at this point:

- OBO errors need to be tested for the thread-count. Either find a more ergonomic/idiomatic way to keep the count correct automagically, or unit test.
- I'm not 100% sure that the listen-branch is cancel proof/safe. If there's an `.await` at a pathalogical moment, that could introduce some weird behavior.
- Things are still async: related to previous point, but addressed in earlier comment
- ADD TO ME

massa-bootstrap/src/server.rs
```diff
@@ -168,48 +175,57 @@ impl BootstrapServer {
...
+                    let Ok((mut server, remote_addr)) = self.is_ip_allowed(remote_addr, server, &whitelist, &blacklist).await else {
+                        continue;
                     };
 
-                    if bootstrap_sessions.len() < self.bootstrap_config.max_simultaneous_bootstraps.try_into().map_err(|_| BootstrapError::GeneralError("Fail to convert u32 to usize".to_string()))? {
+                    let max_bootstraps: usize = self.bootstrap_config.max_simultaneous_bootstraps.try_into().map_err(|_| BootstrapError::GeneralError("Fail to convert u32 to usize".to_string()))?;
+                    // TODO: double-check OBO errors here, or find a better way to track count
+                    if Arc::strong_count(&bootstrap_sessions_counter) < max_bootstraps {
+
+                        let per_ip_min_interval = self.bootstrap_config.per_ip_min_interval.to_duration();
```
Ben-PH, line 228: revert me

massa-bootstrap/src/server.rs
```diff
@@ -269,31 +266,24 @@ impl BootstrapServer {
...
+                            self.manager_tx.clone(),
+                            bootstrap_count_token,
+                            config,
+                            remote_addr,
+                            data_execution,
+                            version,
+                            consensus_command_sender,
+                            network_command_sender
+                        ));
+                        // increment the number of active bootstraps. Will be decremented when handling the
```
Ben-PH, line 280: delete me

massa-bootstrap/src/server.rs
```diff
@@ -269,31 +266,24 @@ impl BootstrapServer {
...
+                        // Complete message that is sent when complete.
```
Ben-PH, line 281: delete me

massa-bootstrap/src/server.rs
```diff
@@ -269,31 +266,24 @@ impl BootstrapServer {
...
+
+                        massa_trace!("bootstrap.session.started", {"active_count": Arc::strong_count(&bootstrap_sessions_counter) - 1});
                     } else {
-                        let _ = match tokio::time::timeout(self.bootstrap_config.write_error_timeout.into(), server.send(BootstrapServerMessage::BootstrapError {
+                        let config = self.bootstrap_config.clone();
```
Ben-PH, line 285: do we still need this clone?

massa-bootstrap/src/server.rs
```diff
@@ -373,6 +411,75 @@ impl BootstrapServer {
...
+            &config,
+            &mut server,
+            data_execution,
+            version,
+            consensus_command_sender,
+            network_command_sender,
+        ),
+    )
+    .await;
+    // This drop allows the server to accept new connections without having to complete the error notifications
```
Ben-PH, line 440: without -> before

massa-bootstrap/src/server.rs
```diff
@@ -373,6 +411,75 @@ impl BootstrapServer {
...
+    drop(arc_counter);
+    let _ = controller_tx.send(BSInternalMessage::Complete).await;
```
Ben-PH, line 442: consider reverting the introduction of the complete message, and provide the logging in the arc-drop

AurelienFT:
> Instead of checking the length of the (now removed) FuturesUnordered list, we track the number of active threads through an Arc<()>, and looking at its strong-count. Running a bootstrap requires passing in a clone, and the runner will drop the clone on completion.

It's not better to have a vector with the handle on the tasks ?

Ben-PH:
That would require some sort of mechanism to pop the join handle from the collection automagically. 

The reason I went with the Arc is because there is a tight coupling between the strong count of the arc, and the number of active threads. I originally went with a plain old `u32`, which worked nicely, but keeping the value in line with the number of active threads (active threads == active bootstrap sessions) would make the code less readable/maintainable. Although this approach still requires manual book-keeping, it is constrained to "only needing to pass in the a clone of a specific variable, and don't forget to account for the root copy". The built-in mechanisms handle the rest.

I'm sure there's a better way, such as a mechanism that will automagically push/pop join handles from a collection in O(1) based on their ownership. Thread-pooling comes to mind. We might end up going there on the next iteration of this task, but this seemed appropriate to begin with.

Ben-PH:
The Arc has a (minor) added benefit of freeing up a bootstrap slot as soon as it completes/fails:

```rust
    let res = tokio::time::timeout(
        config.bootstrap_timeout.into(),
        manage_bootstrap(
            &config,
            &mut server,
            data_execution,
            version,
            consensus_command_sender,
            network_command_sender,
        ),
    )
    .await;
    // This drop allows the server to accept new connections before having to complete the error notifications
    drop(arc_counter);
    let _ = controller_tx.send(BSInternalMessage::Complete).await;
    match res {
```

the `match res` includes some branches inside a `tokio::timeout` context that sends errors to the client. In the cases of an error with the bootstrap, or a timeout of the bootstrap, we do not have to wait until the message sending is completed for that slot to be available for the next client.

It's a small and perhaps rarely needed optimisation, but it's something that works for now.

The next design iterations should consider applying such optimisations automagically. My initial thinking, is to couple the count with the `manage_bootstrap` scope inside here: https://github.com/massalabs/massa/pull/3554/files#diff-2f2a0d279d4c015c632b27bdc7bd2fb62cf753fa8811dead19aadc3133666156R431

...instead of inside the `run_bootstrap` scope from which the `manage_bootstrap` scope is created.

This would all be made redundant, though, if we were to create a dedicated thread pool for the bootstrap server, spawn the threads within that pool, and count the spawned threads. That should work in a bubble, but I don't think we want to go down that road yet...

massa-bootstrap/src/server.rs
```diff
@@ -35,14 +33,22 @@ use crate::{
...
-    manager_tx: mpsc::Sender<()>,
+    manager_tx: mpsc::Sender<BSInternalMessage>,
+}
+
+#[derive(Debug)]
+enum BSInternalMessage {
+    // Signals the Manager to stop
+    Stop,
+    // Client has completed a bootstrap
+    Complete,
```
Ben-PH, line 44: I don't think we need this enum for the time-being. The only use for this seems to be for running a massa_trace. This could be integrated into a dropping of a `Arc<()>` wrapper. Would be ideal to not use the Arc-pattern though.

massa-bootstrap/src/server.rs
```diff
@@ -305,12 +298,60 @@ impl BootstrapServer {
                 }
             }
         }
-        // wait for bootstrap sessions to finish
-        while bootstrap_sessions.next().await.is_some() {}
```
Ben-PH, line 309: As sessions are now spawned into a thread that runs in the background, I _don't think_ this is required, as the execution context (tokio runtime) of the spawned threads remain active.

This removal would need to be revisited if we opt for a design that has a dedicated execution context (might be useful for tracking the number of active sessions, for example).

massa-bootstrap/src/server.rs
AurelienFT: Yes I don't think it's useful if we have the arc counter

AurelienFT (commented):
Before reviewing in depth. I want to be sure that I'm well understanding the PR. Here you are just splitting the code and introducing some channels for the thread architecture for the future or I am missing something ?

Ben-PH:
I'm not 100% sure. Given that there's a more generalised goal of removing `async`. If that is, indeed, a desired outcome, we we might need to retain this channel pattern.

With that said, Probably better off reverting the channel for this PR...

massa-bootstrap/src/server.rs
Ben-PH: I'll revert. Probably only still in there due to sentimental attachment for my more ambitious attempt that a backtracked from :(

massa-bootstrap/src/server.rs
Ben-PH: done, though without the the arc-drop logging. That feels like a premature commitment, given the goal on further iterations

massa-bootstrap/src/server.rs
Ben-PH: reversion done.

AurelienFT:
I saw that you created a second PR. What's the workflow you wanna have ? Merge this one first and continue on the other one or close this one and start over in the other one ?

AurelienFT:
I also agree with that did you mange to test it in labnet ? Can you resolve conflict also and we will review

Ben-PH:
labnet seems to be fine. Haven't stress-tested it though.

massa-bootstrap/src/server.rs
```diff
@@ -305,12 +279,66 @@ impl BootstrapServer {
...
         Ok(())
     }
 
+    /// Helper method to check IP's bootstrap attempt history, consuming the server in the process.
+    ///
+    /// On success, will return the server for later use.
+    /// On failure, includes a side effect of sending an error message to the client.
+    ///
+    /// The intended use of the `Result` is to signal a `continue` to the calling context on failure.
+    /// ...I know, not the best. This function is largely a cut-paste of code that existed in the main-loop `BootstrapServer::run()` method, and where that code would invoke a `continue`, here we return `Err`
```
AurelienFT, line 292: Try to avoid using "I" in public comments

massa-bootstrap/src/server.rs
```diff
@@ -269,29 +254,18 @@ impl BootstrapServer {
...
-                                Err(_timeout) => {
-                                    debug!("bootstrap timeout for peer {}", remote_addr);
-                                    // We allow unused result because we don't care if an error is thrown when sending the error message to the server we will close the socket anyway.
-                                    let _ = tokio::time::timeout(config.write_error_timeout.into(), server.send(BootstrapServerMessage::BootstrapError { error: format!("Bootstrap process timedout ({})", format_duration(config.bootstrap_timeout.to_duration())) })).await;
-                                }
-                            }
-
-                        });
-                        massa_trace!("bootstrap.session.started", {"active_count": bootstrap_sessions.len()});
+                        tokio::spawn(run_bootstrap_session(
```
AurelienFT, line 257: If we stop the node this task will block the join of the main thread ?

AurelienFT:
While working on network I discovered wait group that could help to replace the arc counter : https://docs.rs/crossbeam/latest/crossbeam/sync/struct.WaitGroup.html

Ben-PH:
> While working on network I discovered wait group that could help to replace the arc counter : https://docs.rs/crossbeam/latest/crossbeam/sync/struct.WaitGroup.html

I can't see how. The idea behind the arc counter is to prevent servicing too many bootstraps at once. Is addressing this question what you had in mind? https://github.com/massalabs/massa/pull/3554#discussion_r1109750465

AurelienFT:
Yes sorry, I forgotten about the benefit from arc to be able to keep a counter. Yes waitGroup could help waiting all bootstrap thread before quitting.

AurelienFT:
Can you address the last changes so that we can move on on this PR ?

massa-bootstrap/src/server.rs
Ben-PH: I've brought in a lot of follow-up work done in benph/bootstrap-de-async branch.

The approach taken is that this is now a `std::thread::spawn`, with `run_bootstrap_session` now a non-`async` function.

`run_bootstrap_session`, however, has a lot of async stuff that still needs to be removed. In the meantime, the compromise is to follow this pattern:

```rust
fn run_bootstrap_session(...) -> ... {
    tokio::Runtime::new().block_on(/* do the actual session here */);
}
```

It's not pretty, and if there's a better way to approach the idea in my mind (explained below), I'm all ears.

The goal is as follows:

- the `launch()` function is the top-level runtime context
- the `start_bootstrap_server` function runs in that context, but that is (close to) being non-`async`, and can be thought of as a such
- this function, in turn, spawns a thread running the servers main-loop
- this main-loop spawns a thread for each bootstrap session to be run
 - _inside_ this session, a tokio runtime is created so the remaining async code base can work
- As far as the the execution stack/contexts outside of this thread are concerned, there exists a thread for each bootstrap session, and it has the usuall block/unblock behaviour. They are oblivious to the fact that this is often done by that threads `tokio` `Runtime` context.

massa-bootstrap/src/server.rs
```diff
@@ -14,363 +42,451 @@ use massa_signature::KeyPair;
...
+    update_handle: tokio::task::JoinHandle<Result<(), String>>,
+    listen_handle: tokio::task::JoinHandle<Result<(), Box<BootstrapError>>>,
+    main_handle: std::thread::JoinHandle<Result<(), Box<BootstrapError>>>,
+    stopper_tx: crossbeam::channel::Sender<()>,
 }
 
 impl BootstrapManager {
     /// stop the bootstrap server
-    pub async fn stop(self) -> Result<(), BootstrapError> {
+    pub fn stop(self) -> Result<(), Box<BootstrapError>> {
```
Ben-PH, line 72: This box is to address a clippy lint.

massa-bootstrap/src/server.rs
```diff
@@ -14,363 +42,451 @@ use massa_signature::KeyPair;
...
-            join_handle,
-            manager_tx,
-        }))
-    } else {
-        Ok(None)
-    }
+    // TODO(low prio): See if a zero capacity channel model can work
+    let (stopper_tx, stopper_rx) = crossbeam::channel::bounded::<()>(1);
+
+    let runtime = Runtime::new().expect("Failed to create a bootstrap runtime");
```
Ben-PH, line 101: once the listener and updater are synchronous, this runtime will be redundant

massa-bootstrap/src/server.rs
```diff
@@ -14,363 +42,451 @@ use massa_signature::KeyPair;
...
+        return Err(BootstrapError::GeneralError("Fail to convert u32 to usize".to_string()).into());
+    };
+    // This is the primary interface between the async-listener, and the (soon to be) sync worker
+    let (listener_tx, listener_rx) =
+        crossbeam::channel::bounded::<(Duplex, SocketAddr)>(max_bootstraps);
+
+    let allow_block_list = SharedAllowBlockList::new(
+        config.bootstrap_whitelist_path.clone(),
+        config.bootstrap_blacklist_path.clone(),
+    )
```
Ben-PH, line 117: Change this to a CoW?

massa-bootstrap/src/server.rs
```diff
@@ -14,363 +42,451 @@ use massa_signature::KeyPair;
...
+    .map_err(BootstrapError::GeneralError)?;
+
+    let update_handle = runtime.handle().clone().spawn(BootstrapServer::run_updater(
+        allow_block_list.clone(),
+        config.cache_duration.into(),
+    ));
```
Ben-PH, line 123: This is where I'm a bit hazey about how to best use a created handle. Will exist on a thread one it's non-`async`, but will need to add-in some sort of stop-control. Perhaps an `Arc<AtomicBool>` that can break the loop and that would allow for a join on the thread.

massa-bootstrap/src/server.rs
```diff
@@ -14,363 +42,451 @@ use massa_signature::KeyPair;
...
+            allow_block_list,
+            keypair,
+            version,
+            ip_hist_map: HashMap::with_capacity(config.ip_list_max_size),
+            bootstrap_config: config,
+            runtime,
+        }
+        .run_loop(max_bootstraps)
+    });
+    // does `runtime` get dropped here?
```
Ben-PH, line 145: Not so sure about this one. Perhaps it's lifetime is bound to the join-handles that were spawned from this runtime?

massa-bootstrap/src/server.rs
```diff
@@ -14,363 +42,451 @@ use massa_signature::KeyPair;
...
+        }
+    }
+    async fn run_listener(
+        mut listener: Listener,
+        listener_tx: Sender<(Duplex, SocketAddr)>,
+    ) -> Result<(), Box<BootstrapError>> {
+        loop {
+            let msg = listener.accept().await.map_err(BootstrapError::IoError)?;
+            let Ok(_) = listener_tx.send(msg) else {
+                todo!("handle send error. even better, make this channel a tokio channel");
```
Ben-PH, line 186: missed this todo!

massa-bootstrap/src/server.rs
```diff
@@ -14,363 +42,451 @@ use massa_signature::KeyPair;
...
+                if unlikely(stop.is_ok()) {
                     massa_trace!("bootstrap.lib.run.select.manager", {});
-                    break
+                    break;
+                } else if unlikely(stop == Err(crossbeam::channel::TryRecvError::Disconnected)) {
+                    return Err(BootstrapError::GeneralError(
+                        "Unexpected stop-channel disconnection".to_string(),
+                    )
+                    .into());
+                };
```
Ben-PH, line 244: move this into helper method

massa-bootstrap/src/server.rs
```diff
@@ -14,363 +42,451 @@ use massa_signature::KeyPair;
...
+                Err(try_rcv_err) => match try_rcv_err {
+                    crossbeam::channel::TryRecvError::Empty => continue,
+                    crossbeam::channel::TryRecvError::Disconnected => {
+                        return Err(BootstrapError::GeneralError(
+                            "listener recv channel disconnected unexpectedly".to_string(),
+                        )
+                        .into())
+                    }
                 },
+            };
```
Ben-PH, line 258: move into helper method

massa-bootstrap/src/server.rs
```diff
@@ -14,363 +42,451 @@ use massa_signature::KeyPair;
...
+            let mut server = BootstrapServerBinder::new(
+                dplx,
+                self.keypair.clone(),
+                (&self.bootstrap_config).into(),
+            );
+
+            // check whether incoming peer IP is allowed.
+            // TODO: confirm error handled according to the previous `is_ip_allowed` fn
+            if let Err(error_msg) = self.allow_block_list.is_ip_allowed(&remote_addr) {
+                let _ = match self.runtime.block_on(server.send_error(error_msg.clone())) {
```
Ben-PH, line 271: is this the correct use of block_on?

massa-bootstrap/src/server.rs
```diff
@@ -14,363 +42,451 @@ use massa_signature::KeyPair;
...
+                    )
+                    .into()),
+                    Ok(Err(e)) => Err(e),
+                    Ok(Ok(_)) => Ok(()),
+                };
+                // not exactly sure what to do here
+                // return Err(std::io::Error::new(
+                //     std::io::ErrorKind::PermissionDenied,
+                //     error_msg,
+                // ));
```
Ben-PH, line 284: revert comment

massa-bootstrap/src/server.rs
```diff
@@ -14,363 +42,451 @@ use massa_signature::KeyPair;
...
-                }
+                debug!("did not bootstrap {}: no available slots", remote_addr);
             }
         }
-        // wait for bootstrap sessions to finish
-        while bootstrap_sessions.next().await.is_some() {}
-
+        // abort listener and updater here?
+        // TODO: clean up the listener and updater here.
+        // TODO: do we drop(self.runtime) here?
```
Ben-PH, line 393: address these

massa-bootstrap/src/server.rs
```diff
@@ -14,363 +42,451 @@ use massa_signature::KeyPair;
...
-                )
-                .into()),
-                Ok(Err(e)) => Err(e),
-                Ok(Ok(_)) => Ok(()),
-            };
-            return Err(std::io::Error::new(
-                std::io::ErrorKind::PermissionDenied,
-                error_msg,
-            ));
+/// To be called from a `tokio::spawn` invocation
```
Ben-PH, line 423:
```suggestion
/// To be called from a `thread::spawn` invocation
```

massa-bootstrap/src/server.rs
```diff
@@ -14,363 +42,451 @@ use massa_signature::KeyPair;
...
+///
+/// Runs the bootstrap management in a tokio::timeout context. A failed bootstrap session
+/// will fail silently locally, but will send a message to the client with an error message.
+///
+/// The arc_counter variable is used as a proxy to keep track the number of active bootstrap
+/// sessions.
```
Ben-PH, line 429: update comment

massa-bootstrap/src/server.rs
```diff
@@ -14,363 +42,451 @@ use massa_signature::KeyPair;
...
+    mut server: BootstrapServerBinder,
+    arc_counter: Arc<()>,
+    config: BootstrapConfig,
+    remote_addr: SocketAddr,
+    data_execution: Arc<RwLock<FinalState>>,
+    version: Version,
+    consensus_command_sender: Box<dyn ConsensusController>,
+    network_command_sender: NetworkCommandSender,
+) {
+    debug!("awaiting on bootstrap of peer {}", remote_addr);
```
Ben-PH, line 441: update debug

massa-bootstrap/src/server.rs
```diff
@@ -14,363 +42,451 @@ use massa_signature::KeyPair;
...
+                &mut server,
+                data_execution,
+                version,
+                consensus_command_sender,
+                network_command_sender,
+            ),
+        )
+        .await;
+        // This drop allows the server to accept new connections before having to complete the error notifications
+        massa_trace!("bootstrap.session.finished", {
```
Ben-PH, line 457: confirm the number we want, and that it's correct

massa-bootstrap/src/server/allow_block_list.rs
```diff
@@ -0,0 +1,113 @@
...
+
+use crate::tools::normalize_ip;
+
+/// A wrapper around the allow/block lists that allows efficient sharing between threads
+// TODO: don't clone the path-bufs...
+#[derive(Clone)]
+pub(crate) struct SharedAllowBlockList {
+    inner: Arc<RwLock<AllowBlockListInner>>,
+    allow_path: PathBuf,
+    block_path: PathBuf,
```
Ben-PH, line 20: change to CoW

massa-bootstrap/src/server/allow_block_list.rs
```diff
@@ -0,0 +1,113 @@
...
+        Ok(Self {
+            inner: Arc::new(RwLock::new(AllowBlockListInner {
+                allow_list,
+                block_list,
+            })),
+            allow_path,
+            block_path,
+        })
+    }
+    pub(crate) fn update(&mut self) -> Result<(), String> {
```
Ben-PH, line 36: comment me

massa-bootstrap/src/server/allow_block_list.rs
```diff
@@ -0,0 +1,113 @@
...
+            }
+            if block_delta {
+                mut_inner.block_list = new_block;
+            }
+        }
+        Ok(())
+    }
+    #[cfg(test)]
+    // TODO we didn't test whether the peer IP address is banned
+    pub(crate) fn is_ip_allowed(&self, _remote_addr: &SocketAddr) -> Result<(), String> {
```
Ben-PH, line 60: mmove the cfg annotation into the method itself as an early-return statement

Ben-PH (commented):
Self-review

massa-bootstrap/src/server.rs
Ben-PH: Done

AurelienFT:
@modship can you also make a review to familiarise with the bootstrap and ask all the questions you have

massa-bootstrap/src/server.rs
AurelienFT: You moved it to the thread of the bootstrap server no ?

massa-bootstrap/src/server.rs
```diff
@@ -14,363 +42,470 @@ use massa_signature::KeyPair;
...
+    /// Ok(Ok(())) listener closed without issue
+    /// Ok(Err((dplx, address))) listener accepted a connection then tried sending on a disconnected channel
+    /// Err(..) Error accepting a connection
+    async fn run_listener(
+        mut listener: Listener,
+        listener_tx: Sender<BsConn>,
+    ) -> Result<Result<(), BsConn>, Box<BootstrapError>> {
+        loop {
+            let msg = listener.accept().await.map_err(BootstrapError::IoError)?;
+            match listener_tx.send(msg) {
```
AurelienFT, line 197: Why does the listener send back the connection to the main loop instead of using a shared data to check if he can accept it and spawn the thread for the connection himself? It will avoid this communication

massa-bootstrap/src/server.rs
```diff
@@ -14,363 +42,470 @@ use massa_signature::KeyPair;
...
-            tokio::select! {
-                // managed commands
-                _ = self.manager_rx.recv() => {
-                    massa_trace!("bootstrap.lib.run.select.manager", {});
-                    break
-                },
+            // block until we have a connection to work with, or break out of main-loop
+            let Some((dplx, remote_addr)) = self.receive_connection(&mut selector).map_err(BootstrapError::GeneralError)? else {break};
+            // claim a slot in the max_bootstrap_sessions
+            let bootstrap_count_token = bootstrap_sessions_counter.clone();
```
AurelienFT, line 222: I think it's better to take a slot only if the bootstrap started

massa-bootstrap/src/server.rs
AurelienFT: What do you means by correct ? You are waiting for the message to be sent. It's ok. I just find it strange to do all this check for allowance, cooldown in the main loop as we will take time to do it and take even more time to send the error and doing this time we are not doing anything.

If we follow something more like the architecture here : https://github.com/massalabs/massa/discussions/3149
The listener create threads for people that connect and it's on each of these thread that we verify the cooldown, allowance and send back the errors. So we don't block the main listening if a peer is not allowed take time to receive his error

massa-bootstrap/src/server.rs
AurelienFT, line 197: Shouldn't be better to use `try_send` so that if we are full in the queue for the main loop to manage it it avoid people waiting to connect for ages ?

massa-bootstrap/src/server/allow_block_list.rs
```diff
@@ -0,0 +1,114 @@
...
+impl AllowBlockListInner {
+    #[allow(clippy::result_large_err)]
+    #[allow(clippy::type_complexity)]
+    fn load_allow_block_lists(
+        allowlist_path: &Path,
+        blocklist_path: &Path,
+    ) -> Result<(Option<HashSet<IpAddr>>, Option<HashSet<IpAddr>>), String> {
+        let allow_list = Self::load_list(allowlist_path)?;
+        let block_list = Self::load_list(blocklist_path)?;
+        Ok((allow_list, block_list))
```
modship, line 92: i personally think you should name this black_list and white_list like your error message instead of allow/block.

massa-bootstrap/src/server/allow_block_list.rs
```diff
@@ -0,0 +1,114 @@
...
+    pub(crate) fn is_ip_allowed(&self, remote_addr: &SocketAddr) -> Result<(), String> {
+        #[cfg(test)]
+        return Ok(());
+
+        let ip = normalize_ip(remote_addr.ip());
+        // whether the peer IP address is blacklisted
+        let read = self.inner.read();
+        if let Some(ip_list) = &read.block_list && ip_list.contains(&ip) {
+            massa_trace!("bootstrap.lib.run.select.accept.refuse_blacklisted", {"remote_addr": remote_addr});
+            Err(format!("IP {} is blacklisted", &ip))
```
modship, line 72: Should create specific Error instead of String

massa-bootstrap/src/server/allow_block_list.rs
```diff
@@ -0,0 +1,114 @@
...
+        allowlist_path: &Path,
+        blocklist_path: &Path,
+    ) -> Result<(Option<HashSet<IpAddr>>, Option<HashSet<IpAddr>>), String> {
+        let allow_list = Self::load_list(allowlist_path)?;
+        let block_list = Self::load_list(blocklist_path)?;
+        Ok((allow_list, block_list))
+    }
+
+    fn load_list(list_path: &Path) -> Result<Option<HashSet<IpAddr>>, String> {
+        let Ok(list) = std::fs::read_to_string(list_path) else {
```
modship, line 96: Print an error when std::fs::read_to_string fail

massa-bootstrap/src/server.rs
```diff
@@ -14,363 +42,470 @@ use massa_signature::KeyPair;
...
-            massa_trace!("bootstrap.lib.run.select.accept.refuse_not_whitelisted", {"remote_addr": remote_addr});
-            Some(format!("A whitelist exists and the IP {} is not whitelisted", &ip))
+    /// this is a bit complicated, but the simple is bad:
+    /// ```rust-ignore
+    /// if self.stopper_rx.try_recv().is_ok() {
+    ///     break;
+    /// }
+    /// // if the server is stopped while blocked here, this risks never unblocking
+    /// // OR: would serve as an OBO error; Server doesn't actually stop working until after
+    /// // it's done processing one more connection _after_ it's been stopped.
```
AurelienFT, line 367: I don't understand this sentence

massa-bootstrap/src/server.rs
```diff
@@ -14,363 +42,470 @@ use massa_signature::KeyPair;
...
-        let ip = normalize_ip(remote_addr.ip());
-        // whether the peer IP address is blacklisted
-        let not_allowed_msg = if let Some(ip_list) = &blacklist && ip_list.contains(&ip) {
-            massa_trace!("bootstrap.lib.run.select.accept.refuse_blacklisted", {"remote_addr": remote_addr});
-            Some(format!("IP {} is blacklisted", &ip))
-        // whether the peer IP address is not present in the whitelist
-        } else if let Some(ip_list) = &whitelist && !ip_list.contains(&ip){
-            massa_trace!("bootstrap.lib.run.select.accept.refuse_not_whitelisted", {"remote_addr": remote_addr});
-            Some(format!("A whitelist exists and the IP {} is not whitelisted", &ip))
+    /// this is a bit complicated, but the simple is bad:
```
AurelienFT, line 360: Why you don't use the select macro of crossbeam ?

AurelienFT (commented):
Nice first version and preparation work for fully removal.

massa-bootstrap/src/server/allow_block_list.rs
Ben-PH: The reason being is that it would be a `BootstrapError::GeneralError(String)` every time.

How about we `struct BootStrapGenErrorMsg(String);` and pass that up?

massa-bootstrap/src/server.rs
Ben-PH: :thinking:

massa-bootstrap/src/server.rs
Ben-PH: The select macro is non-deterministic in which branch it runs.

this approach allows us to wake up on the listen channel without actually consuming anything, and then checking for a stop signal.

massa-bootstrap/src/server.rs
AurelienFT: I don't understand your emoji 😆 . It's related to https://github.com/massalabs/massa/pull/3554#discussion_r1114320260

massa-bootstrap/src/server/allow_block_list.rs
Ben-PH: done

massa-bootstrap/src/server/allow_block_list.rs
Ben-PH: done, though instead of logging/printing, it propogates the error up (I'll log the errors where the errors are eventually captured)

massa-bootstrap/src/server.rs
Ben-PH: @AurelienFT  do you feel this is resolved?

massa-bootstrap/src/server.rs
Ben-PH: does this clarify?
```rust
    /// // if the server is stopped while blocked on the next line, this risks never unblocking
    /// //
    /// // An approach that somehow forces a recv in the event that the stop signal is sent,
    /// // it would serve as an OBO error; Server doesn't actually stop working until after
    /// // it's done processing one more connection _after_ it's been stopped.
    /// //
    /// // ...any bug fix at that point would be a downstream hack to an upstream
    /// // bug.
```

massa-bootstrap/src/server.rs
Ben-PH: In principal, I like the idea. Would have to think about the approach:

- do we immediately drop?
- do we give a short delay (maybe a second) to give the server a chance to free up a slot in the channel?
- do we make the channel unbounded so that the message is guaranteed to be processed (eventually. Would be vulnerable to clients spamming requests in such a way that flood the channel, but not the slots)
- do we give a long-ish delay, then signal the client to immediately try the next server?

...perhaps these questions will be redundant if there's a solution that involves making that channel redundant.

Ben-PH: If the `.await` wasn't needed, then we could put the `.accept()` at the top of the main-loop. That could work well under the assumption that the time taken to sanity-check the client before the bootstrap thread is spawned is sufficiently short.

...and if we are not happy to make that assumption, we could move a whole bunch of it into the handler itself (though that might mean having to set-up something to handle the join() results on that. We should be doing that anyway, but might be a bit premature for this PR)

If we are happy to block on this accept inside the main-loop, we could do something like this?

```rust
let acceptance = Runtime::new()?.block_on(/*loop-body here */);
```

massa-bootstrap/src/server.rs
Ben-PH: I'll spend some time now trying to move all that into the spawned thread.

massa-bootstrap/src/server/allow_block_list.rs
modship: Not possible to replace GeneralError by BlacklistError/WhitelistError ?

Ben-PH: Even if this falls under the scope of not being a `GeneralError`, we would need to add another variant, which I feel is out of scope for this specific PR

massa-bootstrap/src/server.rs
Ben-PH: > I don't understand your emoji laughing . It's related to [#3554 (comment)](https://github.com/massalabs/massa/pull/3554#discussion_r1114320260)

it's the thinking one. Intended to communicate that "...huh" moment that gives food for thought.

I've since made a `close_and_send_error` helper that

- consumes the server binder
- spawns a thread, and makes an async runtime in it
- sends the message
- runs a passed-in closure (which is typically to run a logging macro)

Can we resolve this thread, and move discussion to https://github.com/massalabs/massa/pull/3554#discussion_r1114320260 ?, (make a reference to this thread over there)

massa-bootstrap/src/server.rs
Ben-PH: If we were to move everything into the handler thread, it would need mutable borrow of the server, which I really don't want to do at this stage as

a) I suspect that will cause a lot of issues on read/write contention
b) if we manage to put the error-sending off-thread (which I've got something almost done for that already), then the thread-spawning pre-amble should be acceptably quick.

At some stage, we will refine what gets moved into the thread, which might open the door to reducing the accept-to-spawn time

massa-bootstrap/src/server.rs
```diff
@@ -14,363 +42,470 @@ use massa_signature::KeyPair;
...
-        let mut bootstrap_sessions = FuturesUnordered::new();
-        let cache_timeout = self.bootstrap_config.cache_duration.to_duration();
-        let (mut whitelist, mut blacklist) = reload_whitelist_blacklist(
-            &self.bootstrap_config.bootstrap_whitelist_path,
-            &self.bootstrap_config.bootstrap_blacklist_path,
-        )?;
-        let mut cache_interval = tokio::time::interval(cache_timeout);
+    fn run_loop(&mut self, max_bootstraps: usize) -> Result<(), Box<BootstrapError>> {
+        // Use the strong-count of this variable to track the session count
+        let bootstrap_sessions_counter: Arc<()> = Arc::new(());
```
modship, line 212: Is that really an advantage? rather than storing a counter

massa-bootstrap/src/server.rs
Ben-PH: I managed to move the error-sending stuff into a thread using the servers dedicated runtime.

massa-bootstrap/src/server.rs
Ben-PH: Yeah.

One pain-point was that it needs to drop in a non-blocking context

a) using the runtime to send errors back to the client from a spawned thread frees up some nasty blocks
b) the server drop coincides with a good time to drop the runtime.

massa-bootstrap/src/server/allow_block_list.rs
Ben-PH: edit: propogating the error up is not what we want.

I'll add a debug! log

massa-bootstrap/src/server.rs
AurelienFT: Yes

massa-bootstrap/src/server.rs
AurelienFT: It's very clear like that ! Thank you

massa-bootstrap/src/server.rs
AurelienFT: I think for sending the error back to the send it should just best-effort but it's debatable

massa-bootstrap/src/server.rs
AurelienFT: It's still confusing for me maybe I'm not the brightest light but "forces a recv in the event that the stop signal is sent" => it's not "in the channel that the stop signal is sent"?

massa-bootstrap/src/server.rs
modship: you convinced me ! 
Yes a dedicated type can be a good idea for a better use.

massa-bootstrap/src/server.rs
AurelienFT: Ok but I think in the future it would be nice that the threads are spawn directly by the listener so that the main loop don't have to make all the verif things etc.. and we accept very fast. We can close this for now as it would be more appropriate to discuss in the thread PR.

AurelienFT (approved):
I didn't look deeply on the architecture as it's an intermediate PR. I'm just worrying that the bootstrap still works well. Are you make to make a testing session on Friday morning @modship @Ben-PH where we try to bootstrap on labnet server with only two slots etc.. To be sure everything works ?

Also for the second part it would be very nice that you make some kind of diagram as we talk @Ben-PH and as it has been done for the network refactoring for the architecture. So that it will be more easier to discuss and take decisions on architecture with this document than directly in the code.
It's more clear when we manage things one after the other.

massa-bootstrap/src/server/allow_block_list.rs
modship: personally if i provide a file path, i want the software load it correctly, if he fail i want to know it. 
So i think error level for the log is better here. Maybe not ?

massa-bootstrap/src/server/white_black_list.rs
```diff
@@ -0,0 +1,114 @@
...
+    }
+
+    /// Checks if the white/black list is up to date with a read-lock
+    /// Creates a new list, and replaces the old one in a write-lock
+    pub(crate) fn update(&mut self) -> Result<(), String> {
+        let read_lock = self.inner.read();
+        let (new_white, new_black) =
+            WhiteBlackListInner::load_white_black_lists(&self.white_path, &self.black_path)?;
+        let white_delta = new_white != read_lock.white_list;
+        let black_delta = new_black != read_lock.black_list;
```
modship, line 44:
```rust 
let should_update = new_black.ne(&read_lock.black_list);
``` 
doesn"t work correctly ?

massa-bootstrap/src/server/allow_block_list.rs
Ben-PH: I would actually like to do something with the Err enums more broadly. It makes sense in many places to return the inner-value, and not an enum instance, but that is problematic with a lot of inner values being a `String`, and thus not conveying much meaning at the type-level.

I had a quick go at changing `String` to `GenErrString` or similar, and it quickly became apparant that it's a bit out of scope (involves the client side a lot, needs a little bit of design thought put into it, etc).

Would you be happy to put this into a follow-up issue?

massa-bootstrap/src/server/allow_block_list.rs
Ben-PH: You're right. debug! is definately not the right one.

What do you think about `warn!`? In this context, the intent to have a white/black list of IPs is not available, and aside from that, everything is a-okay.

massa-bootstrap/src/server/white_black_list.rs
Ben-PH: I prefer to use `!=` as I find it less cognitive load for reading. using the method means that, at first glance, it could be _any_ method, then your brain has to process that it's a comparison operation, and it can take a non-zero cognitive load to process which is "left" and which is "right". the `!=` does away with a lot of that effort.

massa-bootstrap/src/server.rs
Ben-PH: > I'm not the brightest light

Even if that were true, I don't think it's relevant: This explanation is almost as incoherent as a Charles Manson rant...

In any case, it's probably better to get rid of it. It's commenting as to how the code _isn't_, which reminds me of a lesson I learned in my educator training: Don't teach by demonstrating how things _shouldn't_ be done. That's just bad.

Do you feel that, with just the comments on how it _does_ work, that things are (or could be made to be) clear enough?

massa-bootstrap/src/server.rs
Ben-PH: 100% agree.

making the listener non-async is a bit non-trivial, as it ties in with the client code base and the mocks. 

I think making de-async task is coupled with the client. Will probably be much more efficient to do them together in another PR.

massa-bootstrap/src/server/allow_block_list.rs
modship: i think bootstrap error module should return globally `BootstrapError` which is an enum and the load function of file should return for exemple `BootstrapError::InitListError(String)`with the error detailed string in it.

massa-bootstrap/src/server/allow_block_list.rs
modship: i don't want everything is ok if i provide a file and he failed to load my whitelist/blacklist. I think the end user must not miss this information.

maybe @AurelienFT would have an opinion too

Personally i can `panic!` here :boom:

massa-bootstrap/src/server/allow_block_list.rs
modship: @Ben-PH 

I have make an example of what i mean here : https://github.com/massalabs/massa/commit/2cb1de8d7a3a9ec45a1d296840754175dbbe81b2

modship: and this for is_ip_allowed : https://github.com/massalabs/massa/commit/8167af85e95e0d2bb8e27c0591d35141f3ba3adf

massa-bootstrap/src/server/allow_block_list.rs
AurelienFT: The precedent behaviour was : if there is no whitelist than everyone is allowed. If there is an existing whitelist but empty everyone is forbidden

So in the case of the whilelist no being here isn't an issue. This is our way to have "everyone allow" and "nobody allow" with this file.

massa-bootstrap/src/server.rs
AurelienFT: Yes we can close this discussion

massa-bootstrap/src/server/allow_block_list.rs
Ben-PH: Definitely agree that these are improvements, and the review comments provided. Thanks heaps for going ahead and doing it as well.

I feel there could be improvements to the overall design of the error type system here, but worrying about that is probably an nonconstructive distraction.

Go ahead and bring them in.

massa-bootstrap/src/server.rs
```diff
@@ -14,365 +42,516 @@ use massa_signature::KeyPair;
...
-                        })).await {
-                            Err(_) => Err(std::io::Error::new(std::io::ErrorKind::TimedOut, "bootstrap error no available slots send timed out").into()),
-                            Ok(Err(e)) => Err(e),
-                            Ok(Ok(_)) => Ok(()),
-                        };
-                        debug!("did not bootstrap {}: no available slots", remote_addr);
-                    }
-                }
+                // load cache if absent
+                // if bootstrap_data.is_none() {
```
Ben-PH, line 345: @AurelienFT  this appears to be commented out in by you here: https://github.com/massalabs/massa/commit/b6b40ebb605734b9fa66e3d9e056dbfab8bf5f6c#diff-2f2a0d279d4c015c632b27bdc7bd2fb62cf753fa8811dead19aadc3133666156R113

can we get rid of this commented out code?

massa-bootstrap/src/server/allow_block_list.rs
Ben-PH: thanks for adding those changes. I've Boxed the returns to satisfy clippy, I've also made #3608 as a follow-up

massa-bootstrap/src/server/allow_block_list.rs
Ben-PH: Not exactly sure what the ideal resolution for this is, but I'm happy to `panic!` here as the panic would be triggered on startup.

Would be problematic if in the future we would like to change the file while things are running, or if you change the name of the file in the file-system.

AurelienFT: Don't change the behavior in this PR (for this month). You can just add a warning.

massa-bootstrap/src/server.rs
AurelienFT: Yes

massa-bootstrap/src/server/allow_block_list.rs
Ben-PH: From what I can work out, the testnet 20 behavior was that the bootstrap server would die silently and independantly of the rest of the system by `?` returning errors, ending up in the task-handler. The error is captured in a `let _ = self.join_handle.await?;`, and the returning Err is handled with `.expect("bootstrap server shutdown failed")`.

The behavior here should be the same, with some extra `error!` and `info!` tracing macros thrown in for good measure.

massa-bootstrap/src/server.rs
Ben-PH: Done

massa-bootstrap/src/server.rs
Ben-PH: I think I'm getting a better handle on things now.

key-word: _**better**_

massa-bootstrap/src/server.rs
Ben-PH: This should only block on the thread hosting the main-loop, and not the binary-wide-system.

Ungraceful shutdowns of active sessions is fine in the short-term.

massa-bootstrap/src/server.rs
```diff
@@ -1,5 +1,33 @@
...
+//!
+//! # Listener
+//!
+//! Runs in the server-dedication tokio async runtime
+//! Accepts bootstrap connections in an async-loop
+//! Upon connection, pushes the accepted connection onto a channel for the worker loop to consume
+//!
+//! # Updater
+//!
+//! Runs in the same runtime as the listener.
```
Ben-PH, line 12:
```suggestion
//! Runs on a dedicated thread. Signal sent my manager stop method terminates the thread.
```

massa-bootstrap/src/server.rs
```diff
@@ -14,365 +42,508 @@ use massa_signature::KeyPair;
...
 };
 
+/// Abstraction layer over data produced by the listener, and transported
+/// over to the worker via a channel
+type BsConn = (Duplex, SocketAddr);
 /// handle on the bootstrap server
 pub struct BootstrapManager {
-    join_handle: JoinHandle<Result<(), BootstrapError>>,
-    manager_tx: mpsc::Sender<()>,
+    update_handle: std::thread::JoinHandle<Result<(), Box<BootstrapError>>>,
```
Ben-PH, line 67:
```suggestion
    update_handle: std::thread::JoinHandle<Result<(), Box<BootstrapError>>>,
    // need to preserve the listener handle up to here to prevent it being destroyed
```

massa-bootstrap/src/server.rs
```diff
@@ -14,365 +42,508 @@ use massa_signature::KeyPair;
...
+    update_stopper_tx: crossbeam::channel::Sender<()>,
 }
 
 impl BootstrapManager {
     /// stop the bootstrap server
-    pub async fn stop(self) -> Result<(), BootstrapError> {
+    pub async fn stop(self) -> Result<(), Box<BootstrapError>> {
         massa_trace!("bootstrap.lib.stop", {});
-        if self.manager_tx.send(()).await.is_err() {
+        dbg!("stopping");
```
Ben-PH, line 78: remove all the `dbg!`s

massa-bootstrap/src/server.rs
```diff
@@ -14,365 +42,508 @@ use massa_signature::KeyPair;
...
+        if dbg!(self.listen_stopper_tx.send(()).is_err()) {
             warn!("bootstrap server already dropped");
         }
-        let _ = self.join_handle.await?;
-        Ok(())
+        if dbg!(self.update_stopper_tx.send(()).is_err()) {
+            warn!("bootstrap ip-list-update already dropped");
```
Ben-PH, line 83:
```suggestion
            warn!("bootstrap ip-list-updater already dropped");
```

massa-bootstrap/src/server.rs
```diff
@@ -14,365 +42,508 @@ use massa_signature::KeyPair;
...
+        }
+        // TODO?: handle join errors.
+        // TODO: examine dead-lock potential
```
Ben-PH, line 86: clear this TODO

massa-bootstrap/src/server.rs
```diff
@@ -14,365 +42,508 @@ use massa_signature::KeyPair;
...
+
+    // TODO(low prio): See if a zero capacity channel model can work
+    let (listen_stopper_tx, listen_stopper_rx) = crossbeam::channel::bounded::<()>(1);
+    let (update_stopper_tx, update_stopper_rx) = crossbeam::channel::bounded::<()>(1);
 
-        let join_handle = tokio::spawn(async move {
+    let Ok(max_bootstraps) = config.max_simultaneous_bootstraps.try_into() else {
+        return Err(BootstrapError::GeneralError("Fail to convert u32 to usize".to_string()).into());
+    };
+    let Ok(bs_server_runtime) = runtime::Builder::new_multi_thread()
```
Ben-PH, line 120:
```suggestion
    };
    
    // We use a runtime dedicated to the bootstrap part of the system
    // This should help keep it isolated from the rest of the overall binary
    let Ok(bs_server_runtime) = runtime::Builder::new_multi_thread()
```

massa-bootstrap/src/server.rs
```diff
@@ -14,365 +42,508 @@ use massa_signature::KeyPair;
...
+        .build() else {
+            return Err(Box::new(BootstrapError::GeneralError("Failed to creato bootstrap async runtime".to_string())));
+        };
+
+    let listener = establisher
+        .get_listener(listen_addr)
+        .await
+        .map_err(BootstrapError::IoError)?;
+
+    // This is the primary interface between the async-listener, and the (soon to be) sync worker
```
Ben-PH, line 134:
```suggestion
    // This is the primary interface between the async-listener, and the "sync" worker
```

massa-bootstrap/src/server.rs
```diff
@@ -14,365 +42,508 @@ use massa_signature::KeyPair;
...
+                    crossbeam::channel::TryRecvError::Empty => {}
+                    crossbeam::channel::TryRecvError::Disconnected => {
+                        return Err(BootstrapError::GeneralError(
+                            "update stopper unexpected disconnect".to_string(),
+                        )
+                        .into());
+                    }
                 },
+            }
+            // TODO: loop interval here is tick + update time. Implement a state-based,
```
Ben-PH, line 236:
```suggestion
            // TODO: loop interval here is sleep + update time. Implement a state-based,
```

massa-bootstrap/src/server.rs
```diff
@@ -14,365 +42,508 @@ use massa_signature::KeyPair;
...
 
-                // Whitelist cache timeout
-                _ = cache_interval.tick() => {
-                    (whitelist, blacklist) = reload_whitelist_blacklist(&self.bootstrap_config.bootstrap_whitelist_path, &self.bootstrap_config.bootstrap_blacklist_path)?;
+    /// Listens on a channel for incoming connections, and loads them onto a crossbeam channel
+    /// for the main-loop to process.
+    ///
+    /// Ok(Ok(())) listener closed without issue
+    /// Ok(Err((dplx, address))) listener accepted a connection then tried sending on a disconnected channel
+    /// Err(..) Error accepting a connection
```
Ben-PH, line 247:
```suggestion
    /// Err(..) Error accepting a connection
    /// TODO: Integrate the listener into the bootstrap-main-loop
```

massa-bootstrap/src/server.rs
```diff
@@ -14,365 +42,508 @@ use massa_signature::KeyPair;
...
-                    };
+        // Use the strong-count of this variable to track the session count
+        let bootstrap_sessions_counter: Arc<()> = Arc::new(());
+        let per_ip_min_interval = self.bootstrap_config.per_ip_min_interval.to_duration();
+        let mut selector = Select::new();
+        selector.recv(&self.listen_stopper_rx);
+        selector.recv(&self.listener_rx);
+        // TODO: Work out how to integration-test this
+        loop {
+            // block until we have a connection to work with, or break out of main-loop
```
Ben-PH, line 287:
```suggestion
            // block until we have a connection to work with, or break out of main-loop
            // if a stop-signal is received
```

massa-bootstrap/src/server.rs
```diff
@@ -14,365 +42,508 @@ use massa_signature::KeyPair;
...
     }
 }
 
+/// To be called from a `thread::spawn` invocation
+///
+/// Runs the bootstrap management in a dedicated tokio runtime context. A failed bootstrap session
+/// will fail silently locally, but will send a message to the client with an error message.
+///
+/// The arc_counter variable is used as a proxy to keep track the number of active bootstrap
+/// sessions.
```
Ben-PH, line 482:
```suggestion
/// Runs the bootstrap management in a dedicated thread, handling the async by using
/// a multi-thread-aware tokio runtime (the bs-main-loop runtime, to be exact). When this
/// function blocks in the `block_on`, it should thread-block, and switch to another session
///
/// The arc_counter variable is used as a proxy to keep track the number of active bootstrap
/// sessions.
```

Ben-PH (commented):
Self review

modship (approved):

PR #3557, opened 2023-02-10
Fix check args in CLI

aoudiamoncef:
  * [x] make tests compile
  * [x] make tests pass 

### DIFF
--- a/massa-client/src/cmds.rs
+++ b/massa-client/src/cmds.rs
@@ -1113,7 +1113,7 @@ impl Command {
                 }
             }
             Command::read_only_execute_smart_contract => {
-                if parameters.len() < 2 || parameters.len() > 5 {
+                if parameters.len() < 2 || parameters.len() > 4 {
                     bail!("wrong number of parameters");
                 }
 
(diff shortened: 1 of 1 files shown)


massa-client/src/cmds.rs
```diff
@@ -1113,7 +1113,7 @@ impl Command {
                 }
             }
             Command::read_only_execute_smart_contract => {
-                if parameters.len() != 2 && parameters.len() <= 4 {
+                if parameters.len() < 2 || parameters.len() > 5 {
```
Leo-Besancon, line 1116: Unless I'm tired, it should be
```rust
if parameters.len() < 2 || parameters.len() > 4
```
right?
If you have 5 params it should bail ( read_only_execute_smart_contract PathToBytecode MaxGas Address IsFinal )

aoudiamoncef: Copy and paste is evil when you don't have automated test. I'll recheck for the 100 time

aoudiamoncef: done

Leo-Besancon: Nice! Improving the client, one PR at a time 👍

Leo-Besancon (approved):

PR #3562, opened 2023-02-14
Show operation execution status

bilboquet:
  * [x] make tests compile
  * [x] make tests pass 

### DIFF
--- a/massa-api/src/public.rs
+++ b/massa-api/src/public.rs
@@ -495,41 +495,41 @@ impl MassaRpcServer for API<Public> {
                 .collect()
         };
 
-        // keep only the ops found in storage
+        // keep only the ops id (found in storage)
         let ops: Vec<OperationId> = storage_info.iter().map(|(op, _)| op.id).collect();
 
-        // ask pool whether it carries the operations
-        let in_pool = self.0.pool_command_sender.contains_operations(&ops);
-
         let api_cfg = self.0.api_settings.clone();
-        let consensus_controller = self.0.consensus_controller.clone();
         if ops.len() as u64 > api_cfg.max_arguments {
             return Err(ApiError::BadRequest("too many arguments".into()).into());
         }
 
-        // check finality by cross-referencing Consensus and looking for final blocks that contain the op
-        let is_final: Vec<bool> = {
-            let involved_blocks: Vec<BlockId> = storage_info
-                .iter()
-                .flat_map(|(_op, bs)| bs.iter())
-                .unique()
-                .cloned()
-                .collect();
+        // ask pool whether it carries the operations
+        let in_pool = self.0.pool_command_sender.contains_operations(&ops);
 
-            let involved_block_statuses = consensus_controller.get_block_statuses(&involved_blocks);
+        let (speculative_op_exec_statuses, final_op_exec_statuses) =
+            self.0.execution_controller.get_op_exec_status();
 
-            let block_statuses: PreHashMap<BlockId, BlockGraphStatus> = involved_blocks
-                .into_iter()
-                .zip(involved_block_statuses.into_iter())
-                .collect();
-            storage_info
-                .iter()
-                .map(|(_op, bs)| {
-                    bs.iter()
-                        .any(|b| block_statuses.get(b) == Some(&BlockGraphStatus::Final))
-                })
-                .collect()
-        };
+        // compute operation finality and operation execution status from *_op_exec_statuses
+        let (is_operation_final, statuses): (Vec<Option<bool>>, Vec<Option<bool>>) = ops
+            .iter()
+            .map(|op| {
+                match (
+                    final_op_exec_statuses.get(op),
+                    speculative_op_exec_statuses.get(op),
+                ) {
+                    // op status found in the final hashmap, so the op is "final"(first value of the tuple: Some(true))
+                    // and we keep its status (copied as the second value of the tuple)
+                    (Some(val), _) => (Some(true), Some(*val)),
+                    // op status NOT found in the final hashmap but in the speculative one, so the op is "not final"(first value of the tuple: Some(false))
+                    // and we keep its status (copied as the second value of the tuple)
+                    (None, Some(val)) => (Some(false), Some(*val)),
+                    // op status not defined in any hashmap, finality and status are unknow hence (None, None)
+                    (None, None) => (None, None),
+                }
+            })
+            .collect::<Vec<(Option<bool>, Option<bool>)>>()
+            .into_iter()
+            .unzip();
 
         // gather all values into a vector of OperationInfo instances
         let mut res: Vec<OperationInfo> = Vec::with_capacity(ops.len());
@@ -537,18 +537,22 @@ impl MassaRpcServer for API<Public> {
             ops.into_iter(),
             storage_info.into_iter(),
             in_pool.into_iter(),
-            is_final.into_iter()
+            is_operation_final.into_iter(),
+            statuses.into_iter(),
         );
-        for (id, (operation, in_blocks), in_pool, is_final) in zipped_iterator {
+        for (id, (operation, in_blocks), in_pool, is_operation_final, op_exec_status) in
+            zipped_iterator
+        {
             res.push(OperationInfo {
                 id,
                 in_pool,
-                is_final,
+                is_operation_final,
                 thread: operation
                     .content_creator_address
                     .get_thread(api_cfg.thread_count),
                 operation,
                 in_blocks: in_blocks.into_iter().collect(),
+                op_exec_status,
             });
         }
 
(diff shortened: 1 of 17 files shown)


AurelienFT:
Nice job ! Can you fix the CI and place the PR "ready for review" when you want to get a review ? :)

massa-executed-ops/src/ops_changes.rs
```diff
@@ -98,13 +108,20 @@ impl Deserializer<ExecutedOpsChanges> for ExecutedOpsChangesDeserializer {
                     context("operation id", |input| {
                         self.operation_id_deserializer.deserialize(input)
                     }),
+                    context("operation execution status", |input| {
```
AurelienFT, line 111: It's possible here to make directly a sub tuple in the tuple parser above. (Optional)

massa-execution-worker/src/execution.rs
```diff
@@ -193,6 +193,68 @@ impl ExecutionState {
...
+                err
+            )));
+        }
+
+        // from here, the op is considered as executed (fees transferred)
+
+        // add operation to executed ops list
+        context.insert_executed_op(
+            operation_id,
+            None, // execution status not know yet
```
AurelienFT, line 239: Ok now I see why you have placed a None value. But I think you can remove the insert of the executed op here and add it only after the execution so that you already have the status. (Add it after the reset of the snapshot in case of error to avoid re-writing). It will allow you to remove the Option in the executed structures and will simplify a lot the rest of the code.
I don't see anywhere in the code where it could cause problems.

massa-models/src/operation.rs
```diff
@@ -1273,6 +1273,75 @@ impl Deserializer<Vec<SecureShareOperation>> for OperationsDeserializer {
...
+impl OperationExecutionStatus {
+    /// ctor
+    pub fn new(status: Option<bool>) -> Self {
+        Self { status }
+    }
+}
+
+/// Serializer for bool
+#[derive(Debug, Default)]
+pub struct BoolSerializer {}
```
AurelienFT, line 1295: This is a more general structure I think it could be saved in an other folder

massa-models/src/operation.rs
```diff
@@ -1273,6 +1273,75 @@ impl Deserializer<Vec<SecureShareOperation>> for OperationsDeserializer {
...
+        Self {}
+    }
+}
+
+impl Serializer<bool> for BoolSerializer {
+    fn serialize(&self, value: &bool, buffer: &mut Vec<u8>) -> Result<(), SerializeError> {
+        match value {
+            true => buffer.push(b'1'),
+            false => buffer.push(b'0'),
+        }
```
AurelienFT, line 1309: I propose : 
```suggestion
    buffer.push(*value as u8);
```

massa-models/src/operation.rs
```diff
@@ -1273,6 +1273,75 @@ impl Deserializer<Vec<SecureShareOperation>> for OperationsDeserializer {
...
+    }
+}
+
+impl Deserializer<bool> for BoolDeserializer {
+    fn deserialize<'a, E: ParseError<&'a [u8]> + ContextError<&'a [u8]>>(
+        &self,
+        buffer: &'a [u8],
+    ) -> IResult<&'a [u8], bool, E> {
+        context("Failed bool deserialization", |input: &'a [u8]| {
+            Ok((&buffer[1..], {
```
AurelienFT, line 1331: I think you could use the `tag` combinator of nom. (Optional)

AurelienFT (commented):
Nice work ! 💯  I think few simplifications could be done.

massa-models/src/operation.rs
bilboquet: done, thx

massa-models/src/operation.rs
bilboquet: done

massa-api/src/public.rs
```diff
@@ -495,60 +495,62 @@ impl MassaRpcServer for API<Public> {
...
-                    bs.iter()
-                        .any(|b| block_statuses.get(b) == Some(&BlockGraphStatus::Final))
-                })
-                .collect()
-        };
+        // compute operation finality and operation execution status from *_op_exec_statuses
+        let (is_operation_final, statuses): (Vec<Option<bool>>, Vec<Option<bool>>) = ops
+            .iter()
+            .map(|op| {
+                let is_final = final_op_exec_statuses.get(op).and_then(|status| *status);
```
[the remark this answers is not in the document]
bilboquet: it was poorly named :/
I hope it's clearer in the upcoming commit.
In fact we have 2 sets (hashmaps) of execution statuses. 
One set is for speculative executions, the other for final execution.
so the "finality" is computed against the source set of the status (aka if the status is found in the final execution one, operation is "final," if it is found in the speculative one the operation is "not final")
during the same pass I unify the statuses into one set for later use (when we iterate to build the OperationInfo)

massa-api/src/public.rs
```diff
@@ -495,60 +495,62 @@ impl MassaRpcServer for API<Public> {
...
+                let is_speculative = speculative_op_exec_statuses
+                    .get(op)
+                    .and_then(|status| *status);
+                match (is_final, is_speculative) {
+                    (Some(true), _) => (Some(true), Some(true)),
+                    (Some(false), _) => (Some(true), Some(false)),
```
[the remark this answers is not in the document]
bilboquet: I hope the answer to the previous question also solve this one, I'll add some comments in the code

massa-execution-exports/src/controller_traits.rs
```diff
@@ -51,6 +51,17 @@ pub trait ExecutionController: Send + Sync {
         addresses: &[Address],
     ) -> Vec<(Option<Amount>, Option<Amount>)>;
 
+    /// Get the execution status of operation that have been executed both speculatively or finaly
+    ///
+    /// # Return value
+    /// * `(speculative_statuses, final_statuses)`
+    fn get_op_exec_status(
```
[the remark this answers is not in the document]
bilboquet: done

massa-execution-worker/src/execution.rs
```diff
@@ -193,6 +193,59 @@ impl ExecutionState {
...
+        if let Err(err) =
+            context.transfer_coins(Some(sender_addr), None, operation.content.fee, false)
+        {
+            return Err(ExecutionError::IncludeOperationError(format!(
+                "could not spend fees: {}",
+                err
+            )));
+        }
+
+        // from here, the op is considered as executed (fees transferred)
```
AurelienFT, line 232: This comment should be changed

massa-models/src/operation.rs
```diff
@@ -1273,6 +1273,23 @@ impl Deserializer<Vec<SecureShareOperation>> for OperationsDeserializer {
     }
 }
 
+/// Execution status of an operation
+/// Wrap an optional bool.
+#[derive(Debug, Default)]
+pub struct OperationExecutionStatus {
```
AurelienFT, line 1279: This structure doesn't seems to be used anywhere.

massa-api/src/public.rs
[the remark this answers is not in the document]
AurelienFT: Thanks for clarify, it's better with the comments

massa-models/src/operation.rs
bilboquet: removed

massa-execution-worker/src/execution.rs
bilboquet: yeah, I didn't want to drop it, but, with respect to the changes, it's not that correct anymore…
I hope it's more clear now

massa-executed-ops/src/executed_ops.rs
```diff
@@ -39,6 +39,8 @@ pub struct ExecutedOps {
     pub ops: PreHashSet<OperationId>,
```
damip, line 39: how about just changing this into a `PreHashMap<OperationId, Result<(), String>>` instead of having another container that we must not forget to update/clear ?

massa-execution-worker/src/active_history.rs
```diff
@@ -259,4 +259,19 @@ impl ActiveHistory {
...
+    /// A hashmap with
+    /// * the operation id as the key
+    /// * and a bool as the value: true: execution succeeded, false: execution failed
+    pub fn get_op_exec_status(&self) -> HashMap<OperationId, bool> {
+        self.0
+            .iter()
+            .flat_map(|exec_output| exec_output.state_changes.executed_ops_changes.clone())
+            .map(|(op_id, (status, _))| (op_id, status))
+            .collect()
+    }
```
damip, line 276:
```rust
    /// Check the outcome of the execution of an operation
    pub fn get_op_exec_status(&self, operation_id: &OperationId) -> HistorySearchResult<Result<(), String>> {
        for item in self.0.iter().rev() {
            if let Some(outcome) = item.state_changes.executed_ops_changes.get(operation_id) {
                return HistorySearchResult::Present(outcome);
            }
        }
        HistorySearchResult::NoInfo
    }
```

PR #3571, opened 2023-02-15
Improve logging

AurelienFT:
The goal of this PR is to clean up a bit the logs in debug to be able to run a node in debug with a buffer of 3 days in logs without having the disk being full

### DIFF
--- a/massa-final-state/src/final_state.rs
+++ b/massa-final-state/src/final_state.rs
@@ -97,10 +97,6 @@ impl FinalState {
             as usize;
         for cycle_info in self.pos_state.cycle_history.iter().skip(n) {
             hash_concat.extend(cycle_info.cycle_global_hash.to_bytes());
-            debug!(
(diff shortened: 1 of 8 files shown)


massa-final-state/src/final_state.rs
```diff
@@ -110,10 +104,6 @@ impl FinalState {
         }
         // 5. executed operations hash
         hash_concat.extend(self.executed_ops.hash.to_bytes());
-        debug!(
```
Eitu33, line 113: You left the cycle history hash debug logs right above this, was it intended?

Eitu33 (approved):

massa-final-state/src/final_state.rs
AurelienFT: Good catch. Thank you

Eitu33 (approved):

AurelienFT:
bors merge

PR #3573, opened 2023-02-15
Fix bootstrap server deserialize error

### LINKED ISSUE #3572: Serialization error server side bootstrap
```
bootstrap serving error for peer 38.242.236.160:37782: general bootstrap error: Parsing Error: Failed BootstrapClientMessage deserialization / Failed last_consensus_step deserialization / StreamingStep / ongoing data / Failed PreHashSet<_> deserialization / length / Failed u64 deserialization / Fail / Input: [98, 111
```

AurelienFT:
Fix #3572. 
The bootstrap part is retrieved using the `list_required_blocks` and so it retrieves the last two periods of the network which is 64 blocks and now the max is 50 which may cause #3572.

I'm not fully sure of this if you can confirm @Eitu33

### DIFF
--- a/massa-models/src/config/constants.rs
+++ b/massa-models/src/config/constants.rs
@@ -116,7 +116,7 @@ pub const DEFERRED_CREDITS_BOOTSTRAP_PART_SIZE: u64 = 100;
 /// Maximum executed ops per slot in a bootstrap batch
 pub const EXECUTED_OPS_BOOTSTRAP_PART_SIZE: u64 = 10;
 /// Maximum number of consensus blocks in a bootstrap batch
-pub const CONSENSUS_BOOTSTRAP_PART_SIZE: u64 = 50;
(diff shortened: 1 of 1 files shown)


Eitu33:
@AurelienFT 

Multiple things:
* `list_required_blocks` will return significantly more than `period * 2` blocks because it retrieves the parents and grand parents of the latest blocks per thread
* `CONSENSUS_BOOTSTRAP_PART_SIZE` only describes the max size of a consensus blocks payload, if length of what is returned by `list_required_blocks` is over this size it will simply split it in multiple payloads. So the only impact that constant should have is the amount of bootstrap messages it takes to get all the consensus blocks
* I do not think that the error you linked (https://github.com/massalabs/massa/issues/3572) would be fixed by the change contained in this PR

AurelienFT:
Ok so I close this do you have any thoughts on the error I linked ?

Eitu33:
Not off the top of my head but we could look into it, might need a bit more context though.

Issue #3412, opened 2023-01-12
Crash in a second bootstrap attempt


AurelienFT:
```
2023-01-12T18:36:35.744037Z  INFO massa_node: Node version : TEST.18.0
2023-01-12T18:36:35.858795Z  INFO massa_bootstrap::client: Start bootstrapping from 198.27.74.5:31245
2023-01-12T18:36:36.027402Z  WARN massa_bootstrap::client: Error received from bootstrap server: Bootstrap failed because the bootstrap server currently has no slots available.
2023-01-12T18:36:36.027510Z  INFO massa_bootstrap::client: Bootstrap from server 198.27.74.5:31245 failed. Your node will try to bootstrap from another server in 1s.
2023-01-12T18:36:37.028517Z  INFO massa_bootstrap::client: Start bootstrapping from 158.69.120.215:31245
2023-01-12T18:36:37.191766Z  WARN massa_bootstrap::client: Error received from bootstrap server: Bootstrap failed because the bootstrap server currently has no slots available.
2023-01-12T18:36:37.191790Z  INFO massa_bootstrap::client: Bootstrap from server 158.69.120.215:31245 failed. Your node will try to bootstrap from another server in 1s.
2023-01-12T18:36:38.193057Z  INFO massa_bootstrap::client: Start bootstrapping from 92.169.36.208:31245
2023-01-12T18:36:38.205857Z  WARN massa_bootstrap::client: Error while connecting to bootstrap server: io error: Connection refused (os error 111)
2023-01-12T18:36:38.205920Z  INFO massa_bootstrap::client: Bootstrap from server 92.169.36.208:31245 failed. Your node will try to bootstrap from another server in 1s.
2023-01-12T18:36:39.206375Z  INFO massa_bootstrap::client: Start bootstrapping from 54.36.174.177:31245
2023-01-12T18:37:54.072917Z  INFO massa_bootstrap::client: Slot is too old retry bootstrap from scratch
2023-01-12T18:37:54.072929Z  WARN massa_bootstrap::client: Error while bootstrapping: general bootstrap error: Slot too old
2023-01-12T18:37:54.072958Z  INFO massa_bootstrap::client: Bootstrap from server 54.36.174.177:31245 failed. Your node will try to bootstrap from another server in 1s.
2023-01-12T18:37:55.074365Z  INFO massa_bootstrap::client: Start bootstrapping from 51.75.60.228:31245
[...]

PR #3576, opened 2023-02-15
Fix bootstrap retry

AurelienFT:
  * [x] make tests compile
  * [x] make tests pass 

### DIFF
--- a/massa-async-pool/src/pool.rs
+++ b/massa-async-pool/src/pool.rs
@@ -51,6 +51,14 @@ impl AsyncPool {
         }
     }
 
+    /// Resets the pool to its initial state
+    ///
+    /// USED ONLY FOR BOOTSTRAP
+    pub fn reset(&mut self) {
+        self.messages.clear();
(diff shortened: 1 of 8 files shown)


AurelienFT:
Fix #3412

AurelienFT:
Tried to make a bootstrap test but it's too complicated and takes too much time. Will test it in labnet with modified node.

AurelienFT:
I have tested on labnet with tests line of https://github.com/massalabs/massa/pull/3576/commits/31e2e60f2381f0cd5d891d7e3c7e88319cb5e341 this allow to make the first bootstrap fail and wait so much that the second fail for too old slot and after reset everything went well and the final hash at slot was good

Eitu33 (approved):
Do we have an open issue about the possibility of retrying bootstrap from the downloaded state of a previously failed bootstrap or is that not something we want to handle?

AurelienFT:
> Do we have an open issue about the possibility of retrying bootstrap from the downloaded state of a previously failed bootstrap or is that not something we want to handle?

It should already be possible if it's not a too old slot. For a too old slot I don't think it's possible

PR #3579, opened 2023-02-16
Fix bootstrap server error on retry

Eitu33:


AurelienFT:
Fix #3572

AurelienFT (commented):
I think 300 is too magical we could put the number of block we keep in consensus. I think we only keep 10 periods (variable `force_keep_final_periods` in config.toml) so 10*32 = 320 could be great with a comment explaining this

Eitu33:
I disagree because the `force_keep_final_periods` is in the dynamic config and that would make the payload limit accessible  to the user. We do not want that.

AurelienFT (approved):
You convinced me that there is no good solution. Let's keep it like that for now and change in the future if needed

PR #3580, opened 2023-02-16
Fix get call coins.

### LINKED ISSUE #3575: Context.transferedCoins() ==> returns value 0 even if the transfer during the call carried out properly
**Context :** 

- Innonet 18.
- .env and all settings done properly
- A SC using `Context.transferedCoins()` to know how much coins have been sent during the call. The SC is deployed at : "A12mFyHgM8SA5oYzSiXM2gm6SEw2rFxosXa9Pj67ZaWFGBSWvxMx"
- A caller.ts transferring 10 coins to the SC using the call function.

**Issue :** 
The return of the `Context.transferedCoins()` function is always 0 even if the transfer has been done.

**The result :**
```bash
Wallet_Balance : 845.20375
Caller_Address : A1W85B2UJhua6P1J98nrWD9SVGAGEuseRkbaLEKMLVSiGqC1VhA || tokens sent to the SC during the call : 0
Wallet_Balance : 835.20375
```

I precise that I have checked and the balance of the smart contract is correctly incremented of 10 coins, thus the transfer of the coins during the call is effective. My issue is only to know how to get the value of how many coins have been transferred during the call?

If someone has an idea plz? :)

**the SC code :** 
```ts
import { generateEvent, Context} from "@massalabs/massa-as-sdk";

export function test(param: StaticArray<u8>): void {      
    generateEvent("Caller_Address : " + Context.caller().toByteString() + " || tokens sent to the SC during the call : " + Context.transferedCoins().toString());  
}
```

**the caller.ts code :**
```ts
import { Args, ClientFactory, WalletClient, ProviderType, IProvider, EOperationStatus, ICallData } from '@massalabs/massa-web3';

const base_account = await WalletClient.getAccountFromSecretKey('secret_key_hidden');

const addr_sc = "A1sTJ6vXFkcrpun2M5PsE899iPzRKiL1mzXtWHiXRwYWmFnzkaw";

const Client = await ClientFactory.createCustomClient(
  [
    { url: "https://inno.massa.net/test18", type: ProviderType.PUBLIC } as IProvider,
    // This IP is false but we don't need private for this script so we don't want to ask one to the user
    // but massa-web3 requires one
    { url: "https://inno.massa.net/test18", type: ProviderType.PRIVATE } as IProvider,
  ],
  true,
  base_account,
);


async function DisplayEvent(opIds: string) {

  // Wait the end of deployment
  await Client.smartContracts().awaitRequiredOperationStatus(opIds, EOperationStatus.FINAL);

  const event = await Client.smartContracts().getFilteredScOutputEvents({
    emitter_address: null,
    start: null,
    end: null,
    original_caller_address: null,
    original_operation_id: opIds,
    is_final: null,
  });

  if (event.length) {
    // This prints the deployed SC address
    event.forEach((e) => {
      console.log(e.data);
    });
  }
  else {
    console.log('Call done. No events has been generated');
  }
}

async function CallScFunction(func: string, coins_to_send: number) {
  const res = await Client.smartContracts().callSmartContract(
    {
      fee: 0,
      maxGas: 4_200_000_000,
      coins: coins_to_send,
      targetAddress: addr_sc,
      functionName: func,
      parameter: (new Args()).serialize(),
    } as ICallData,
    base_account
  );
  await DisplayEvent(res.toString());
}

// get the balance of the wallet
let info = await Client.wallet().walletInfo();
let balance_wallet = info.at(0)?.final_balance;
console.log("Wallet_Balance : " + balance_wallet);

//send 10 coins during the call of the 'test' function of the sc
await CallScFunction("test", 10 * 1000000000);

//update the wallet balance
info = await Client.wallet().walletInfo();
balance_wallet = info.at(0)?.final_balance;
console.log("Wallet_Balance : " + balance_wallet);
```

AurelienFT:
  * [x] make tests compile
  * [x] make tests pass 

Fix #3575


massa-execution-worker/src/execution.rs
```diff
@@ -582,7 +582,7 @@ impl ExecutionState {
                 },
                 ExecutionStackElement {
                     address: target_addr,
-                    coins: Default::default(),
+                    coins: coins,
```
sydhds, line 585: clippy here

massa-execution-worker/src/tests/scenarios_mandatories.rs
```diff
@@ -267,6 +268,110 @@ mod tests {
...
+            Default::default(),
+            block_storage.clone(),
+        );
+        std::thread::sleep(Duration::from_millis(100));
+        // Get the events that give us the gas usage (refer to source in ts) without fetching the first slot because it emit a event with an address.
+        let events = controller.get_filtered_sc_output_event(EventFilter {
+            start: Some(Slot::new(2, 0)),
+            ..Default::default()
+        });
+        println!("events: {:#?}", events);
```
sydhds, line 370: Should test the message content (and check for the coin amount)?

massa-execution-worker/src/tests/scenarios_mandatories.rs
```diff
@@ -344,7 +449,9 @@ mod tests {
             ..Default::default()
         });
 
-        println!("events: {:?}", events);
+        assert!(events[0]
+            .data
+            .contains("tokens sent to the SC during the call : 10000000000"));
```
sydhds, line 454: Should not be hardcoded?

sydhds (changes requested):

massa-execution-worker/src/execution.rs
AurelienFT: Done

massa-execution-worker/src/tests/scenarios_mandatories.rs
AurelienFT: Done

massa-execution-worker/src/tests/scenarios_mandatories.rs
AurelienFT: Yes nice idea. Should be done now :) Thank you !

sydhds (approved):

PR #3582, opened 2023-02-16
Benph/bootstrap de async

Ben-PH:
  * [ ] make tests compile
  * [ ] make tests pass 


Ben-PH:
@gterzian I plan to take a similar approach to your network refactor in some respects. I think it would be constructive at some point to pair up on it.

No hurry, still have a decent amount of homework to do, but when you have a good moment, let me know?

Issue #3584, opened 2023-02-17
retry intermittant failures on reinstated tests


Ben-PH:
some tests were reinstated in #3507

the reinstated tests occassionally fail for reasons outside the scope of what is being tested (#3509) . This is handled in CI through our use of cargo nextest with automatic retries, but locally, you could run into some false fails.

Until I find an upstream fix (i.e. you can run cargo test -F testing, and things just work), the downstream sollution is to retry those tests manually, or use nextest run --retries 5. 

A intermittant fix should allow `cargo test` to run `nextest run --retries 5`

PR #3602, opened 2023-02-23
Fix tests pool failing

### LINKED ISSUE #3509: Intermittant failures in pool tests
**Describe the bug**
Pool tests in `massa-pool-worker` fail intermittently. It appears that the component that should be receiving the data from the mocker isn't getting the data it needs

**To Reproduce**

on testnet_20 run the following in a bash shell:

```bash
COUNTER = 0;
while cargo test test_pool -F testing --release ; do
    COUNTER=$[$COUNTER +1]
    :
done

echo $COUNTER;
```

It loops, failing with an `echo $COUNTER` output > 1 (usually

**Expected behavior**
Either a Fail where `echo $COUNTER` outputs `1`, or an infinite loop 

**Desktop (please complete the following information):**

Ubuntu, but that should only be a bash-script thing

**Additional context**
In CI this shouldn't be an issue, as our tests have up to 10 retries, and it would be incredibly rare for this to fail more than 3 or 4 times in a row.

The logic being tested seems to be correct, that is, if the code being tested is incorrect, it will always fail.

AurelienFT:
  * [x] make tests compile
  * [x] make tests pass 

The tests were failing because `add_operations` is managed in an other thread and so the tests don't wait that the ops are inserted before trying to fetch them. So sometimes it was fast enough sometimes not. With the new sleep it's always fast enough.

The `test_pool` failed after this change because it was wrong by design. Fixed it.

Fix #3509 and #3584

### DIFF
--- a/massa-pool-worker/src/tests/operation_pool_tests.rs
+++ b/massa-pool-worker/src/tests/operation_pool_tests.rs
@@ -31,6 +31,7 @@ fn test_add_operation() {
         let op_gen = OpGenerator::default().expirery(2);
         storage.store_operations(create_some_operations(10, &op_gen));
         operation_pool.add_operations(storage);
+        std::thread::sleep(Duration::from_millis(100));
         assert_eq!(operation_pool.storage.get_op_refs().len(), 10);
     });
 }
@@ -46,6 +47,7 @@ fn test_add_irrelevant_operation() {
         storage.store_operations(create_some_operations(10, &op_gen));
         operation_pool.notify_final_cs_periods(&vec![51; thread_count.into()]);
         operation_pool.add_operations(storage);
+        std::thread::sleep(Duration::from_millis(100));
         assert_eq!(operation_pool.storage.get_op_refs().len(), 0);
     });
 }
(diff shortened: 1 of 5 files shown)


Ben-PH:
LGTM. I'm gonna try something locally before I approve, but the essence of it is something like this:

```bash
for ((i = 0 ; i < 100 ; i++)); do
	cargo test --release #argument only for these tests here
done
```

AurelienFT:
I will fix the tests that are not related to this PR but as it's a test fixing PR let's make everything here.

Ben-PH:
Would it be constructive to add a comment to one of the sleeps that provides this context (not a merge-blocking suggestion)

> The tests were failing because add_operations is managed in an other thread and so the tests don't wait that the ops are inserted before trying to fetch them. So sometimes it was fast enough sometimes not. With the new sleep it's always fast enough.

AurelienFT:
It shouldn't fail anymore will check on that also

massa-pool-worker/src/tests/operation_pool_tests.rs
```diff
@@ -62,17 +64,15 @@ fn test_pool() {
             // generate (id, transactions, range of validity) by threads
             let mut thread_tx_lists = vec![Vec::new(); pool_config.thread_count as usize];
 
+            let mut storage = storage_base.clone_without_refs();
```
Ben-PH, line 67: Might be out of scope for this PR, but it might be helpful to think about "why clone_without_refs, and not just clone, in this particular context".

AurelienFT: Issue made #3605

Ben-PH (approved):
LGTM, Just apply the same comment change "Allow some time for...", and for CI to finish and merge it.

PR #3603, opened 2023-02-23
Add version bytes for ledger persistancy

Leo-Besancon:
  * [x] make tests compile
  * [x] make tests pass 


Leo-Besancon:
Oh, two tests are failing right now:
```
massa_bootstrap tests::scenarios::test_bootstrap_server
 TRY 11 FAIL [   0.145s] massa_execution_worker tests::scenarios_mandatories::tests::test_operation_execution_status
```

massa-execution-worker/src/speculative_ledger.rs
```diff
@@ -123,7 +124,7 @@ impl SpeculativeLedger {
     ///
     /// # Returns
     /// `Some(Vec<u8>)` if the address was found, otherwise None
-    pub fn get_bytecode(&self, addr: &Address) -> Option<Vec<u8>> {
+    pub fn get_bytecode(&self, addr: &Address) -> Option<Bytecode> {
```
AurelienFT, line 127: I find a behavior a bit tricky there. We fetch the bytecode from ledger db and so it already have the version encoded inside the VecU8 and if we use his serializer it will re-add the version. It think the fact that we have, or not the version number of the bytecode should be uniform and comprehensible

AurelienFT (commented):
Checked only the bytecode side. Will check the key side after.

massa-ledger-worker/src/ledger.rs
```diff
@@ -106,9 +107,16 @@ impl LedgerController for FinalLedger {
...
     /// A copy of the found bytecode, or None if the ledger entry was not found
-    fn get_bytecode(&self, addr: &Address) -> Option<Vec<u8>> {
+    fn get_bytecode(&self, addr: &Address) -> Option<Bytecode> {
+        let bytecode_deserializer = BytecodeDeserializer::new(u64::MAX);
         self.sorted_ledger
             .get_sub_entry(addr, LedgerSubEntry::Bytecode)
+            .map(|bytes| {
+                bytecode_deserializer
+                    .deserialize::<DeserializeError>(&bytes)
+                    .expect("critical: invalid balance format")
```
AurelienFT, line 117:
```suggestion
                    .expect("critical: invalid bytecode format")
```

AurelienFT (commented):
The key side seems ok

massa-ledger-exports/src/key.rs
```diff
@@ -130,6 +133,84 @@ impl Key {
     }
 }
 
+impl Serializer<KeyType> for KeyTypeSerializer {
```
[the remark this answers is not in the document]
Leo-Besancon: Fixed by commit: https://github.com/massalabs/massa/pull/3603/commits/eb64db00ba9c8bb41fe91a72e6f08e5928a5403b

massa-execution-worker/src/speculative_ledger.rs
Leo-Besancon: Right now the bytecode version is never included when we manipulate the `Vec<u8>` (needed in e.g. the interface in the execution module), so it should not be a problem until the `Bytecode`struct is trully versioned.
		
If I understand your comment correctly, I could preemptively implement the following methods to make the interface between `Vec<u8>` without version and `Bytecode` easier?
		
```rust
impl Bytecode {
	pub fn from_raw_bytes(version: u64, bytes: Vec<u8>) -> Self {}
	pub fn into_raw_bytes_without_version(&self) -> Vec<u8> {}			
}
```

AurelienFT: It's then if there is never the version in the `Vec<u8>` but is that a behavior that we want ? For example in `raw_get_bytecode_for` we will return a bytecode without version maybe in the future we will modify the shape of the bytecode by adding an extra byte and the user will not have any version byte to implement the new layout.

If you don't find an easy and uniform way now it's not a problem. The main goal of this PR is to able to keep the ledger for the next episodes and so changing the return format of an ABI isn't a problem.

massa-ledger-worker/src/ledger.rs
```diff
@@ -106,9 +107,16 @@ impl LedgerController for FinalLedger {
     ///
     /// # Returns
     /// A copy of the found bytecode, or None if the ledger entry was not found
-    fn get_bytecode(&self, addr: &Address) -> Option<Vec<u8>> {
+    fn get_bytecode(&self, addr: &Address) -> Option<Bytecode> {
+        let bytecode_deserializer = BytecodeDeserializer::new(u64::MAX);
```
AurelienFT, line 111: We are currently using the `max_datastore_value_length` for the max size of the bytecode also (maybe add 4 for the version bytes)

massa-execution-worker/src/speculative_ledger.rs
```diff
@@ -123,7 +124,7 @@ impl SpeculativeLedger {
     ///
     /// # Returns
     /// `Some(Vec<u8>)` if the address was found, otherwise None
```
AurelienFT, line 126:
```suggestion
    /// `Some(Bytecode)` if the address was found, otherwise None
```

massa-execution-worker/src/speculative_ledger.rs
Leo-Besancon: Ah I see what you mean, we might want to pass along the version byte everywhere (even the runtime, that only handles `Vec<u8>` and not `Bytecode`.
I'll look into it tomorrow, if I can't find a good way to handle that fast enough we will merge the current code then.

massa-ledger-worker/src/ledger.rs
AurelienFT: You should use the one in self.config so that we can change it in tests etc

massa-execution-worker/src/speculative_ledger.rs
AurelienFT: 👍

massa-ledger-worker/src/ledger.rs
Leo-Besancon: ? self.config (a LedgerConfig) does not have the max_datastore_value_length field right now. Only max_key_length and max_ledger_part_size.
Should I change the LedgerConfig to have this field available?

AurelienFT: Yes we avoid using the constants directly (sorry was thinking it was already in the struct)

Leo-Besancon: Resolved by https://github.com/massalabs/massa/pull/3603/commits/80e887d469b085f41d9f4f67cb363ab6fbfc74a4

AurelienFT (approved):
Thanks a lot. Well done.

massa-ledger-exports/src/ledger_changes.rs
```diff
@@ -248,9 +249,8 @@ impl LedgerEntryUpdateDeserializer {
                 Included(Amount::MIN),
                 Included(Amount::MAX),
             )),
-            bytecode_deserializer: SetOrKeepDeserializer::new(VecU8Deserializer::new(
-                Included(u64::MIN),
-                Included(max_datastore_value_length),
+            bytecode_deserializer: SetOrKeepDeserializer::new(BytecodeDeserializer::new(
+                max_datastore_value_length + 4,
```
sydhds, line 253: No hardcoding...

sydhds (changes requested):

massa-models/src/bytecode.rs
```diff
@@ -0,0 +1,72 @@
...
+use massa_serialization::{
+    Deserializer, SerializeError, Serializer, U64VarIntDeserializer, U64VarIntSerializer,
+};
+use nom::error::{ContextError, ParseError};
+use nom::IResult;
+use serde::{Deserialize, Serialize};
+use std::ops::Bound::Included;
+
+/// Current version of the bytecode
+pub const BYTECODE_VERSION: u64 = 0;
```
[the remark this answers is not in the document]
Leo-Besancon: The current versioning macros use u64 for the version: https://github.com/massalabs/transition/blob/main/transition/src/lib.rs#L4

massa-ledger-exports/src/ledger_changes.rs
Leo-Besancon: Ok, should I add `const MAX_VERSION_BYTE_SIZE: u64 = 4` in Bytecode?
Or should I just take the max_datastore_value_length as arg and hide the magic value in the BytecodeDeserializer code?

massa-models/src/bytecode.rs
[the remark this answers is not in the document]
sydhds: Do we really need that much version? u64 seems way to much no?

massa-ledger-exports/src/ledger_changes.rs
sydhds: If I'm not mistaken you use a u64 for version and the size of it is 8 no?

Leo-Besancon: Ah indeed :)

massa-models/src/bytecode.rs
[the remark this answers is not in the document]
Leo-Besancon: We would probably be fine with u16 I think. But as its serialized as a varint it's not much of an issue IMO.

massa-ledger-exports/src/ledger_changes.rs
Leo-Besancon: As the max_datastore_value_length only covers the inner `vec<u8>` length, the magic value is not needed, we deserialize the version bytes beforehand. Fixed in https://github.com/massalabs/massa/pull/3603/commits/c5dbdd89580157213f6ff4d7ef0e53311d8e88e0

sydhds (approved):

PR #3615, opened 2023-02-27
Refactor(bootstrap_list) : improve log

modship:
remove the error log when try to load a file and every update. add warn log on init load file
add info log every time the list has updated.

  * [X] make tests compile
  * [X] make tests pass 


AurelienFT (approved):
init_list and load_list could be removed to use only one of them but I'm not sure what design is the best so I approve also like that.

PR #3617, opened 2023-02-27
Benph/bs mock tcpstream as duplex

Ben-PH:
Not to be merged.

This is a pre-cursor to de-asyncing the bootstrapping system. These changes set up the mocks to use `tokio::net::TcpStream` for the duplex, routing via "localhost:{Os-chosen port}"

The purpose of this PR is for reviewers to confirm that this is (effectively) just a straight-up refactor. The thinking being is that when a TcpStream is connected on the same local host, it's effectively a `DuplexStream`, only it goes via the OS instead of managed in-mem.

For the purpose of testing, the performance implications should be irrelevant.

### DIFF
--- a/massa-bootstrap/src/client.rs
+++ b/massa-bootstrap/src/client.rs
@@ -377,10 +377,8 @@ async fn connect_to_server(
     pub_key: &PublicKey,
 ) -> Result<BootstrapClientBinder, BootstrapError> {
     // connect
-    let mut connector = establisher
-        .get_connector(bootstrap_config.connect_timeout)
-        .await?; // cancellable
-    let socket = connector.connect(*addr).await?; // cancellable
+    let mut connector = establisher.get_connector(bootstrap_config.connect_timeout)?;
+    let socket = connector.connect(*addr).await?;
     Ok(BootstrapClientBinder::new(
         socket,
         *pub_key,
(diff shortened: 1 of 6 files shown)


Ben-PH:
@Eitu33  This change seems to be re-binding on the same hard-coded local-host port. I have put in some links to provide code-context. 

In effect, these changes the `Duplex` stream from tokios `in-memory channel, process-owned and managed` `DuplexStream` struct to the async version of `port-to-port channel, OS owned and managed` `TcpStream`. 

 CI failure is triggered here here: https://github.com/massalabs/massa/pull/3617/files#diff-701bbaa494a0d030064ad0c25890ce2a12734b3f532b3254116e42968ff873c7R71-R77

The address passed in is defined here: https://github.com/massalabs/massa/pull/3617/files#diff-8cd38f915cf6177aacea92ef29b97d8ceea3c69dc1d21ba95b55301b2ca258fdR304-R307

The need to have a known port number is here: https://github.com/massalabs/massa/blob/babc3fead36dd2d1651eeebd1af69665796554cf/massa-bootstrap/src/tests/scenarios.rs#L186-L191

It passes for me locally with `cargo test -F testing`, but CI [fails](https://github.com/massalabs/massa/actions/runs/4285712181/jobs/7464357804#step:6:782), and @modship has triggered a similar failure weirdness which he can explain.

modship (approved):

Ben-PH:
closing in favor of #3685 and assosciated PRs

PR #3618, opened 2023-02-27
fix(bootstrap) : fix update loop

### LINKED ISSUE #3616: Bootstrap : SIGINT signal
On testnet_20 when i send SIGINT to my node he take some times to start interrupt himself. (config.bootstrap.cache_duration)

image

It is necessary to split the list updater and stop signal receiver.

modship:
  * [X] make tests compile
  * [X] make tests pass 

fix #3616

### DIFF
--- a/massa-bootstrap/src/server.rs
+++ b/massa-bootstrap/src/server.rs
@@ -234,7 +234,7 @@ impl BootstrapServer<'_> {
                 recv(stopper) -> res => {
                     match res {
                         Ok(()) => return Ok(()),
-                        Err(e) => return Err(Box::new(BootstrapError::GeneralError(format!("update stopper error : {}", e.to_string())))),
+                        Err(e) => return Err(Box::new(BootstrapError::GeneralError(format!("update stopper error : {}", e)))),
                     }
                 },
                 recv(ticker) -> _ => {list.update()?;},
--- a/massa-bootstrap/src/server.rs
+++ b/massa-bootstrap/src/server.rs
@@ -25,9 +25,11 @@
 //! 5. All checks have passed: spawn a thread on which to run the bootstrap session
 //!    This thread creates a new tokio runtime, and runs it with `block_on`
 mod white_black_list;
+
 use white_black_list::*;
 
-use crossbeam::channel::{Receiver, Select, SendError, Sender};
+use crossbeam::channel::{tick, Receiver, Select, SendError, Sender};
+use crossbeam::select;
 use humantime::format_duration;
 use massa_async_pool::AsyncMessageId;
 use massa_consensus_exports::{bootstrapable_graph::BootstrapableGraph, ConsensusController};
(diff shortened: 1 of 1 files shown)


massa-bootstrap/src/server.rs
```diff
@@ -63,14 +65,15 @@ use crate::{
 /// Abstraction layer over data produced by the listener, and transported
 /// over to the worker via a channel
 type BsConn = (Duplex, SocketAddr);
+
 /// handle on the bootstrap server
 pub struct BootstrapManager {
-    update_handle: std::thread::JoinHandle<Result<(), Box<BootstrapError>>>,
```
Ben-PH, line 68: While in the process of moving over to native threads and de-asyncing the code-base, I would like to be absolutely explicit about which threads are being used. Might be good to remove the std::thread import, actually.

massa-bootstrap/src/server.rs
```diff
@@ -224,23 +227,18 @@ impl BootstrapServer<'_> {
...
-            match stopper.try_recv() {
-                Ok(()) => return Ok(()),
-                Err(e) => match e {
-                    crossbeam::channel::TryRecvError::Empty => {}
-                    crossbeam::channel::TryRecvError::Disconnected => {
-                        return Err(BootstrapError::GeneralError(
-                            "update stopper unexpected disconnect".to_string(),
-                        )
-                        .into());
+            select! {
```
Ben-PH, line 233:
```suggestion
            crossbeam::select! {
```
Similar in principal to the thread comment

massa-bootstrap/src/server.rs
```diff
@@ -224,23 +227,18 @@ impl BootstrapServer<'_> {
...
+                recv(stopper) -> res => {
+                    match res {
+                        Ok(()) => return Ok(()),
+                        Err(e) => return Err(Box::new(BootstrapError::GeneralError(format!("update stopper error : {}", e)))),
                     }
                 },
+                recv(ticker) -> _ => {list.update()?;},
```
Ben-PH, line 240: add a try_recv on the stopper here: if both are ready, then it's not guaranteed that it will choose the stopper.

In this scenario, the reality is that it's almost never gonna happen. The value will be more in self-documenting the code: "if we _can_ recieve on the stopper, that takes ultimate priority"

Ben-PH (changes requested):
Nicely done, thanks!

Keeping the namespace use is a bit ugly, but is intended to help with the cognitive load for the de-async stuff: much less need to think about _which_ libraries thread/sender/selector is in play. Once the system is settled into being sync, they can be cleaned up.

Ben-PH:
It will feel sluggish and frustrating to users. From a "perceived quality" basis I think there's a strong argument to have it in testnet_20, or at least a 20.1 bump

massa-bootstrap/src/server.rs
modship: I can put them back but my ide displays warnings

massa-bootstrap/src/server.rs
```diff
@@ -63,14 +65,15 @@ use crate::{
...
+
 /// handle on the bootstrap server
 pub struct BootstrapManager {
-    update_handle: std::thread::JoinHandle<Result<(), Box<BootstrapError>>>,
+    update_handle: thread::JoinHandle<Result<(), Box<BootstrapError>>>,
     // need to preserve the listener handle up to here to prevent it being destroyed
-    _listen_handle: std::thread::JoinHandle<Result<Result<(), BsConn>, Box<BootstrapError>>>,
-    main_handle: std::thread::JoinHandle<Result<(), Box<BootstrapError>>>,
-    listen_stopper_tx: crossbeam::channel::Sender<()>,
-    update_stopper_tx: crossbeam::channel::Sender<()>,
```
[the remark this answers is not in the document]
modship: same

massa-bootstrap/src/server.rs
modship: in fact, priority is not specialy needed here. it's better but not needed. 
it's works well for now and there is no more waiting time for stop

modship:
> Is it mandatory to merge this for testnet_20 ? Or it could wait 21 ?

i think it is better if it's merge for testnet_20.
That fix SIGINT && node_stop command from massa-client too

massa-bootstrap/src/server.rs
Ben-PH: reasonably put. resolved.

massa-bootstrap/src/server.rs
Ben-PH: Given the value of the suggested change assumes it doesn't affect the ergonomics for others (aside from the fact that it's a bit ugly in the meantime) it's better to move things towards being nice in the general case, rather than making things nice for my specific use-case.

Happy with the changes you made, given your explanation. Thanks!

Ben-PH (approved):
Thanks for the follow-ups, and well done on the feature, I think it'll do a lot for user experience in the testnet.

AurelienFT:
Ok you can merge it then. It doesn't seems to modify too important things

PR #3619, opened 2023-02-28
Initial code for versioning & versioning factory models

sydhds:
  * [ ] make tests compile
  * [ ] make tests pass 


massa-versioning-worker/src/versioning.rs
```diff
@@ -0,0 +1,1057 @@
...
+    Block,
+    VM,
+}
+
+/// MIP info (name,  version,
+#[derive(Clone, Debug)]
+pub struct MipInfo {
+    /// MIP name or descriptive name
+    pub name: String,
+    /// Network (or global) version
```
Leo-Besancon, line 31: Add some context (for other devs, to be able to differentiate with the `VERSION` const that is e.g. TEST.20), for example:

```suggestion
    /// Network (or global) version, to be included in block headers
```

massa-versioning-worker/src/versioning.rs
```diff
@@ -0,0 +1,1057 @@
...
+    pub version: u32,
+    /// Component concerned by this versioning (e.g. a new Block version)
+    pub component: MipComponent,
+    /// Component version
+    pub component_version: u32,
+    /// a timestamp at which the version gains its meaning (e.g. accepted in block header)
+    pub start: MassaTime,
+    /// a timestamp at which the deployment is considered failed (timeout > start)
```
Leo-Besancon, line 39:
```suggestion
    /// a timestamp at which the deployment is considered failed or active (timeout > start)
```

massa-versioning-worker/src/versioning.rs
```diff
@@ -0,0 +1,1057 @@
...
+
+machine!(
+    /// State machine for a Versioning component that tracks the deployment state
+    #[allow(missing_docs)]
+    #[derive(Clone, Copy, Debug, PartialEq)]
+    enum MipState {
+        /// Initial state
+        Defined,
+        /// Past start
+        Started { threshold: Amount },
```
Leo-Besancon, line 88: Maybe explain the threshold param.

massa-versioning-worker/src/versioning.rs
```diff
@@ -0,0 +1,1057 @@
...
+    /// from VersioningInfo.timeout
+    pub timeout: MassaTime,
+    /// % of past blocks with this version
+    pub threshold: Amount,
+    /// Current time (timestamp)
+    pub now: MassaTime,
+}
+
+impl Default for Advance {
+    fn default() -> Self {
```
Leo-Besancon, line 143: Given the content of the impl, I think this should only be useful for tests, maybe remove this?

massa-versioning-worker/src/versioning.rs
```diff
@@ -0,0 +1,1057 @@
...
+        };
+
+        for (adv, _state) in self.history.iter().skip(1) {
+            advance_msg.now = adv.now;
+            advance_msg.threshold = adv.threshold;
+            vsh.on_advance(&advance_msg);
+        }
+
+        // XXX: is there always full eq? Can we have slight variation here?
+        vsh == *self
```
Leo-Besancon, line 322: I think this is correct, because you recreate the only things that should evolve in the Advance message (now and threshold)

massa-versioning-worker/src/versioning.rs
```diff
@@ -0,0 +1,1057 @@
...
+                Err(StateAtError::EmptyHistory)
+            }
+        }
+    }
+}
+
+/// Error returned by MipStateHistory::state_at
+#[allow(missing_docs)]
+#[derive(Error, Debug, PartialEq)]
+pub enum StateAtError {
```
Leo-Besancon, line 406: Maybe rename to StateError? This way we can expand the usage to other functions, such as update_with, see below

massa-versioning-worker/src/versioning.rs
```diff
@@ -0,0 +1,1057 @@
...
+}
+
+// Store
+
+/// Database for all MIP info
+#[derive(Debug, Clone)]
+pub struct MipStore(pub Arc<RwLock<MipStoreRaw>>);
+
+/*
+impl Default for MipStore {
```
Leo-Besancon, line 422: Remove this commented out section if we never want to create a default empty store.

massa-versioning-worker/src/versioning.rs
```diff
@@ -0,0 +1,1057 @@
...
+        Self(Arc::new(RwLock::new(Default::default())))
+    }
+}
+*/
+
+impl MipStore {
+    // TODO: merge get_version_current / get_version_to_announce so we don't iter twice?
+    //       & rename to get_network_versions(...) -> (u32, u32)
+    //       Move all others funcs from Factory and named them like:
+    //       get_component_version_XXX (best, all, ...)
```
Leo-Besancon, line 433: Merging the two functions is a good idea if we can do both find_map cleanly. I think we should keep the API simple, especially if we also add the functions per components

massa-versioning-worker/src/versioning.rs
```diff
@@ -0,0 +1,1057 @@
...
+}
+
+/// Store of all versioning info
+#[derive(Debug, Clone, Default)]
+pub struct MipStoreRaw(pub(crate) BTreeMap<MipInfo, MipStateHistory>);
+
+impl MipStoreRaw {
+    /// Update our store with another (usually after a bootstrap where we received another store)
+    /// Return true if update is successful, false if something is wrong on given store raw
+    fn update_with(&mut self, store_raw: &MipStoreRaw) -> bool {
```
Leo-Besancon, line 484: Replace with a `Result<(), StateError>`?

massa-versioning-worker/src/versioning.rs
```diff
@@ -0,0 +1,1057 @@
...
+        // iter over items in given store:
+        // -> 2 cases: VersioningInfo is already in self store -> add to 'to_update' list
+        //             VersioningInfo is not in self.store -> add to 'to_add' list
+        //
+
+        // TODO: Check component_version always increase...
```
Leo-Besancon, line 490: Couldn't we want to have a new MIP that reverts a component version? Why is that a problem?

massa-versioning-worker/src/versioning.rs
```diff
@@ -0,0 +1,1057 @@
...
+        Self {
+            state: Default::default(),
+            history,
+        }
+    }
+
+    /// Advance state
+    pub fn on_advance(&mut self, input: &Advance) {
+        let now = input.now;
+        // Check that input.now is not after last item in history
```
Leo-Besancon, line 267:
```suggestion
        // Check that input.now is after last item in history
```

massa-versioning-worker/src/versioning.rs
```diff
@@ -0,0 +1,1057 @@
...
+            .history
+            .last_key_value()
+            .map(|(adv, _)| adv.now < now)
+            .unwrap_or(false);
+
+        if is_forward {
+            // machines crate (for state machine) does not support passing ref :-/
+            let state = self.state.on_advance(input.clone());
+            // Update history as well
+            if state != self.state {
```
Leo-Besancon, line 279: Here, is there a reason to update the history if we go from Started {threshold: 2} to Started {threshold: 3}?
Or should we consider it as duplicate as long as the threshold is not above the const?
This could add a lot of states to the history which is not optimal.
Another option, only keep the first "Started" and the latest one, removing all the states in between. The way we still have some info about "where we currently are"

massa-versioning-worker/src/versioning.rs
```diff
@@ -0,0 +1,1057 @@
...
+        // Optim: this avoids iterating over history (cheap to retrieve first item)
+        let last = self.history.last_key_value().unwrap(); // safe to unwrap
+        if ts > last.0.now {
+            lower_bound = Some(last);
+            is_after_last = true;
+        }
+
+        if !is_after_last {
+            // We are in between two states in history, find bounds
+            for (adv, state_id) in self.history.iter() {
```
Leo-Besancon, line 357: If you think this is well-enough optimized, can you add your reasoning for using the .iter() instead of .range() (with e.g. a link to your benchmarks ;) )

massa-versioning-worker/src/versioning_factory.rs
```diff
@@ -0,0 +1,402 @@
...
+    UnimplementedVersion(u32), // programmer error, forgot to implement new version in factory traits
+    #[error("Version: {0} is not active yet")]
+    OnStateNotReady(u32),
+    #[error("Could not create object of type {0}: {1}")]
+    OnCreate(String, String),
+}
+
+#[derive(Clone)]
+/// Strategy to use when creating a new object from a factory
+pub enum FactoryStrategy {
```
Leo-Besancon, line 24: Maybe rename FactoryStrategy::Best to LatestActive and FactoryStrategy::At() to ActiveAt() to be more explicit?

massa-versioning-worker/src/versioning_factory.rs
```diff
@@ -0,0 +1,402 @@
...
+                    _ => None,
+                }
+            })
+            .unwrap_or(0);
+
+        Ok(version)
+    }
+
+    /// Get all versions in 'Active state' for the associated MipComponent
+    fn get_versions(&self) -> Vec<u32> {
```
Leo-Besancon, line 101: As you have `get_all_versions` below, I think it's clearer if you have `get_all_active_versions` here

Leo-Besancon (changes requested):
Very nice PR, with quite a few tricky edge-cases you took care of!
My comments mainly intend to improve the code readability.

massa-versioning-worker/src/versioning.rs
sydhds: This is used in MipStateHistory::new. As it allow use to store an advance in history for state Defined

massa-versioning-worker/src/versioning.rs
sydhds: I renamed it to Optim as I do not see an obvious way to do this and it's easier to read like that. So let's keep it that way for now?

massa-versioning-worker/src/versioning.rs
sydhds: I think StateError is a misleading name here for something related to MipStoreRaw. Why did you suggest to return a Result here? Any special case or error in mind?

massa-versioning-worker/src/versioning.rs
sydhds: From what I see, this is prone to error (for user & dev) so I'd vote to disallow it. In Versioning followup, there is a proposal to add another state like Disabled in order to disable a version.
Does it sound ok to you?

massa-versioning-worker/src/versioning.rs
sydhds: done

massa-versioning-worker/src/versioning.rs
sydhds: done

massa-versioning-worker/src/versioning.rs
sydhds: done

massa-versioning-worker/src/versioning.rs
sydhds: Removed the XXX comment

massa-versioning-worker/src/versioning.rs
sydhds: Please see update_with comment answer :)

massa-versioning-worker/src/versioning.rs
sydhds: done

massa-versioning-worker/src/versioning.rs
sydhds: done

massa-versioning-worker/src/versioning_factory.rs
sydhds: This adds some 'State info' for the dev that will implement a factory and I'm not a big fan of this... Maybe we can add some doc in Factory trait for create method to make it clear the dev?

massa-versioning-worker/src/versioning_factory.rs
sydhds: Again this add some 'State info' to the dev, let's talk about it on Monday :)

massa-versioning-worker/src/versioning.rs
sydhds: The benchmark was a way of exercising with Criterion; not sure if this is really relevant here as we will have only at max 4 entries in history (Defined, Stated, LockedIn, Active).
If the code is not clear or too long to read, I can switch to the range version. What do you think?

Leo-Besancon: > (Defined, Stated, LockedIn, Active).

This is what I'm unsure about. Doesn't the current code say the history can contain a lot of different Started states, with different thresholds?
Because if history contains 4 states, I agree with you the iterator is fine. If you can have history an history that contains (Defined, Started{a}, Started{b}, Started{c}, ..., LockedIn, Active), then it is a problem.

massa-versioning-worker/src/versioning.rs
Leo-Besancon: Sure, it is fine as it is now :)

massa-versioning-worker/src/versioning.rs
Leo-Besancon: Yeah ok, we may need to think a bit more about that workflow in the follow up I think, but your reasoning does make sense.

massa-versioning-worker/src/versioning.rs
Leo-Besancon: After discussion, this code will be changed to only keep the first Started state.

massa-versioning-worker/src/versioning.rs
Leo-Besancon: This conversation will be resolved by only keeping one Started state in the history.

massa-versioning-worker/src/versioning_factory.rs
Leo-Besancon: Your concern is valid, let's try to improve the function naming without "manually" exposing the state we are in. We should also desambiguate the "network version" and the "component version" as much as possible :)

massa-versioning-worker/src/versioning.rs
Leo-Besancon: We'll keep it to a follow up if needed.

massa-versioning-worker/src/versioning_factory.rs
sydhds: done

massa-versioning-worker/src/versioning_factory.rs
sydhds: done

massa-versioning-worker/src/versioning.rs
sydhds: fixed

massa-versioning-worker/src/versioning.rs
sydhds: fixed

massa-versioning-worker/src/versioning.rs
sydhds: As discussed, this might be improved in the future if we need it. For now, a bool is good enough.

massa-versioning-worker/src/versioning.rs
sydhds: I've removed the Default impl as it might confuse other devs

Leo-Besancon (approved):
Well done :)

massa-versioning-worker/src/versioning.rs
```diff
@@ -0,0 +1,1074 @@
...
+        self.start.hash(state);
+        self.timeout.hash(state);
+    }
+}
+
+machine!(
+    /// State machine for a Versioning component that tracks the deployment state
+    #[allow(missing_docs)]
+    #[derive(Clone, Copy, Debug, PartialEq)]
+    enum MipState {
```
modship, line 84: can you explain the meaning of "MIP" in comment :innocent:

massa-versioning-worker/src/versioning.rs
```diff
@@ -0,0 +1,1074 @@
...
+impl Failed {
+    /// Update state (will always stay in state Failed)
+    pub fn on_advance(self, _input: Advance) -> Failed {
+        Failed {}
+    }
+}
+
+/// Wrapper of MipState (in order to keep state history)
+#[derive(Debug, Clone, PartialEq)]
+pub struct MipStateHistory {
```
AurelienFT, line 230: I don't find this structure very clear. And what is his purpose. Why we are not using a MipState directly ?

EDIT: Reading the rest of the code this allow to have the history of each state of the version. Is that useful ? Would it be send through the bootstrap ?

massa-versioning-worker/src/versioning_factory.rs
```diff
@@ -0,0 +1,405 @@
...
+    type Arguments;
+
+    /// Return the MipComponent associated with return type (Output)
+    /// e.g. if type Output = Address, should return MipComponent::Address
+    fn get_component() -> MipComponent;
+    /// Access to the MipStore
+    fn get_versioning_store(&self) -> MipStore;
+
+    /// Get best component version (aka last active for the factory component)
+    fn get_best_component_version(&self) -> u32 {
```
AurelienFT, line 58: Shouldn't be `latest` ?

massa-versioning-worker/src/versioning_factory.rs
```diff
@@ -0,0 +1,405 @@
...
+    }
+
+    #[derive(Debug)]
+    enum Address {
+        V0(AddressV0),
+        V1(AddressV1),
+    }
+
+    //
+    struct AddressArgs {
```
AurelienFT, line 184: Why this is not an enum with `V0Args()` and `V1Args()` variant to avoid optional and regroup everything like that ?

massa-versioning-worker/src/versioning_factory.rs
```diff
@@ -0,0 +1,405 @@
...
+            .rev()
+            .find_map(|(vi, vsh)| {
+                (vi.component == component && vsh.state == state_active)
+                    .then_some(vi.component_version)
+            })
+            .unwrap_or(0)
+    }
+
+    /// Get best version at given timestamp (e.g. slot)
+    fn get_best_component_version_at(&self, ts: MassaTime) -> Result<u32, FactoryError> {
```
AurelienFT, line 76: Shouldn't be `latest` ?

massa-versioning-worker/src/versioning_factory.rs
```diff
@@ -0,0 +1,405 @@
...
+    OnStateNotReady(u32),
+    #[error("Could not create object of type {0}: {1}")]
+    OnCreate(String, String),
+}
+
+#[derive(Clone)]
+/// Strategy to use when creating a new object from a factory
+pub enum FactoryStrategy {
+    /// use get_best_version (see Factory trait)
+    Best,
```
AurelienFT, line 26: I'm agree with @Leo-Besancon I think it should be more easy to understand with `Latest` saying something like `latest possible to create`

massa-versioning-worker/src/versioning_factory.rs
```diff
@@ -0,0 +1,405 @@
...
+    /// Require to create an object with this specific version
+    Exact(u32),
+    /// Create an object given a timestamp (e.g slot)
+    At(MassaTime),
```
AurelienFT, line 30: What's the usecase of this ?

massa-versioning-worker/src/versioning.rs
```diff
@@ -0,0 +1,1074 @@
...
+    pub(crate) state: MipState,
+    history: BTreeMap<Advance, MipStateTypeId>,
+}
+
+impl MipStateHistory {
+    /// Create
+    pub fn new(defined: MassaTime) -> Self {
```
AurelienFT, line 237: When creating the history of all the versions, I find it more understandable to have either an empty history or a version state to initialize. I don't fully understand the parameter `defined`

massa-versioning-worker/src/versioning.rs
```diff
@@ -0,0 +1,1074 @@
...
+
+        let history = BTreeMap::from([(advance, state_id)]);
+        Self {
+            state: Default::default(),
+            history,
+        }
+    }
+
+    /// Advance state
+    pub fn on_advance(&mut self, input: &Advance) {
```
AurelienFT, line 256: If I understand well this check if the Mip can advance. This should be called often no ? If that's the case maybe we should document it more

AurelienFT (commented):
I didn't fully understand especially the history part so this is my first review. Will make an other when I will have more context after addressing my first questions.

massa-versioning-worker/src/versioning.rs
Leo-Besancon: The history is useful to know the state of a proposal for a given timestamp. For example, when we receive a block header in the protocol-worker with a timestamp `block_timestamp < now()`, we check the history to see if we need to keep or reject the block header.
We need to know if the "active version" of the block is a version that was in active state at `block_timestamp`, and not at `now()`.

AurelienFT: Ok why we don't save only the list of the current active version and the timestamp when they became active ?

massa-models/src/config/constants.rs
```diff
@@ -218,6 +218,12 @@ pub const NETWORK_NODE_COMMAND_CHANNEL_SIZE: usize = 10_000;
 /// network node event channel size
 pub const NETWORK_NODE_EVENT_CHANNEL_SIZE: usize = 10_000;
 
+//
+// Constants used in versioning
+//
+/// Threshold to accept a new versioning
+pub const VERSIONING_THRESHOLD_TRANSITION_ACCEPTED: Amount = Amount::from_mantissa_scale(75, 0);
```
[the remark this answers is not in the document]
sydhds: I've added this as a follow up: https://github.com/massalabs/massa/issues/3631

massa-versioning-worker/src/versioning.rs
```diff
@@ -0,0 +1,1074 @@
...
+
+/// MIP info (name,  version,
+#[derive(Clone, Debug)]
+pub struct MipInfo {
+    /// MIP name or descriptive name
+    pub name: String,
+    /// Network (or global) version (to be included in block header)
+    pub version: u32,
+    /// Component concerned by this versioning (e.g. a new Block version)
+    pub component: MipComponent,
```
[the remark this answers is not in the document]
sydhds: For now, yes we can only have one component update by version. An improvement can be found here: https://github.com/massalabs/massa/issues/3631

massa-versioning-worker/src/versioning.rs
```diff
@@ -0,0 +1,1074 @@
...
+
+/// A message to update the `MipState`
+#[derive(Clone, Debug)]
+pub struct Advance {
+    /// from VersioningInfo.start
+    pub start_timestamp: MassaTime,
+    /// from VersioningInfo.timeout
+    pub timeout: MassaTime,
+    /// % of past blocks with this version
+    pub threshold: Amount,
```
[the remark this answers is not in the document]
sydhds: Because we want to use float and Amount is more meaningful than u64 and easy to ser/der?

massa-versioning-worker/src/versioning_factory.rs
sydhds: The goal here was to keep the unit test code easy to understand and clear but at the end the dev that will implement AddressArgs is free to do whatever he wants to :)

massa-versioning-worker/src/versioning.rs
sydhds: It does not check, it really advances the state (e.g. for instance from defined to started)... I've updated the doc

massa-versioning-worker/src/versioning_factory.rs
sydhds: One use case is when you can to create a new block but in the future. If the state for component version 2 is locked in, in the future it can be active and you will want to use this version (2)

massa-versioning-worker/src/versioning.rs
sydhds: @modship can you check the doc in lib.rs and report? :)

sydhds:
> I didn't fully understand especially the history part so this is my first review. Will make an other when I will have more context after addressing my first questions.

I've added some doc in lib.rs. Can you check and report if it's not clear enough?

massa-versioning-worker/src/versioning.rs
```diff
@@ -0,0 +1,1074 @@
...
+#[allow(missing_docs)]
+#[derive(Clone, Debug, PartialEq, Eq, Hash, TryFromPrimitive, IntoPrimitive)]
+#[repr(u32)]
+pub enum MipComponent {
+    Address,
+    Block,
+    VM,
+}
+
+/// MIP info (name,  version,
```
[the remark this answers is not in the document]
sydhds: fixed

massa-versioning-worker/src/versioning.rs
```diff
@@ -0,0 +1,1074 @@
...
+        self.version.hash(state);
+        self.component.hash(state);
+        self.start.hash(state);
+        self.timeout.hash(state);
+    }
+}
+
+machine!(
+    /// State machine for a Versioning component that tracks the deployment state
+    #[allow(missing_docs)]
```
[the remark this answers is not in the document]
sydhds: removed

massa-versioning-worker/src/versioning.rs
sydhds: I've added some doc in lib.rs about when we used the history

massa-versioning-worker/src/versioning_factory.rs
sydhds: done

massa-versioning-worker/src/versioning_factory.rs
sydhds: done

massa-versioning-worker/src/versioning.rs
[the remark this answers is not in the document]
Leo-Besancon: Thinking about this, I think it may be better to change:
```rust
pub component: MipComponent,
pub component_version: u64,
```
to
```rust
pub components: HashMap<MipComponent, u32>
```
Because we may need to update multiple components at once.

What BIP-9 proposes (and what would be in the follow-up) would be to decouple different component upgrades. E.g. you can vote to change both the Bytecode version and the Address version even though they are unrelated.

massa-versioning-worker/src/versioning.rs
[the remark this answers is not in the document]
AurelienFT: OK

massa-versioning-worker/src/versioning_factory.rs
AurelienFT: Okok

massa-versioning-worker/src/test_helpers/versioning_helpers.rs
```diff
@@ -0,0 +1,54 @@
...
+    let start = versioning_info.start;
+    let timeout = versioning_info.timeout;
+
+    if matches!(at_state, MipState::Error) {
+        todo!()
+    }
+
+    let mut state = MipStateHistory::new(start.saturating_sub(MassaTime::from(1)));
+
+    if matches!(at_state, MipState::Defined(_)) {
```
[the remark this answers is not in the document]
AurelienFT: Ok didn't read it well

massa-versioning-worker/src/versioning.rs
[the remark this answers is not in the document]
sydhds: Do you think we need to do this in this PR or in the follow up? @Leo-Besancon

massa-versioning-worker/src/test_helpers/versioning_helpers.rs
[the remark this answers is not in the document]
sydhds: So we can resolve it ? ;-)

massa-versioning-worker/src/versioning.rs
```diff
@@ -0,0 +1,1080 @@
...
+                        // Trying to downgrade state' (e.g. trying to go from 'active' -> 'defined')
+                        should_merge = false;
+                        break;
+                    }
+                }
+            } else {
+                // Versioning info (from right) is not in self.0 (left)
+                // Need to check if we add this to 'to_add' list
+
+                let last_v_info_ = to_add
```
AurelienFT, line 521: Don't really understand the logic on this code. Why do you check only the last element of the store ? It could be possible that we want to add a version that start before another existing ?

AurelienFT (commented):
Thanks for addressing comments ! I think it's way more clear. Hype to see the usage in the rest of the code

massa-versioning-worker/src/versioning.rs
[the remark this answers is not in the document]
sydhds: So let's do it in the next PR :)

massa-versioning-worker/src/versioning.rs
sydhds: The only case I see is running a older version of the code and bootstrapping to a server running a higher version of the code. In that cases, we can add some new MipInfo but it will always be after existing one. I do not see any case where you can add a version that start before another existing...

AurelienFT (approved):

PR #3629, opened 2023-03-02
Update rocksdb to v0.20

litchipi:
Related issue: #3567 

I checked most of the TODO list under becuase it doesn't really apply to a simple version bump.

  * [x] make tests compile
  * [x] make tests pass 


modship:
> Woooow so gooood ! 😆

that rocks~~db~~

PR #3637, opened 2023-03-03
Remove `test`/`not(test)` `#[cfg(...)]` annotations from bootstrap establisher system

Ben-PH:
This PR is a spin-off of the work being done to de-async the bootstrap server.

All functions are unchanged, so it should be (pretty close) to a straight up refactor.

The removal has several advantages:

- rust-analyzer won't be confused anymore about whether to go to the `Mock` or the `Default` definition of the types
- Allows one to be explicit about having a setup be mocked or not, regardless of the build-context.
- The mock communication duplex-type is the same as deployed type.

Problems:
- the `'static` constraint in this code. This will not always be needed (listener will soon not need to be moved into a thread-spawn), but should be looked into none-the-less
```rust
pub trait BSEstablisher {
    // TODO: this is needed for thread spawning. Once the listener is on-thread, the static
    // lifetime can be thrown away.
    // TODO: use super-advanced lifetime/GAT/other shenanigans to
    // make the listener compatable with being moved into a thread
    type Listener: BSListener + Send + 'static;
    type Connector: BSConnector;
    fn get_listener(&mut self, addr: SocketAddr) -> io::Result<Self::Listener>;
    fn get_connector(&mut self, timeout_duration: MassaTime) -> io::Result<Self::Connector>;
}
```
- non-test code would have to account for the mocking stuff. The optimizer should be able to handle that, though.
- [x] non-test code doesn't exclude test-code (can be fixed)

  * [x] make tests compile
  * [x] make tests pass 

### DIFF
--- a/massa-bootstrap/src/client.rs
+++ b/massa-bootstrap/src/client.rs
@@ -17,9 +17,10 @@ use tracing::{debug, info, warn};
 use crate::{
     client_binder::BootstrapClientBinder,
     error::BootstrapError,
+    establisher::{BSConnector, BSEstablisher},
     messages::{BootstrapClientMessage, BootstrapServerMessage},
     settings::IpType,
-    BootstrapConfig, Establisher, GlobalBootstrapState,
+    BootstrapConfig, GlobalBootstrapState,
 };
 
 /// This function will send the starting point to receive a stream of the ledger and will receive and process each part until receive a `BootstrapServerMessage::FinalStateFinished` message from the server.
@@ -371,7 +372,7 @@ async fn send_client_message(
 }
 
 async fn connect_to_server(
-    establisher: &mut Establisher,
+    establisher: &mut impl BSEstablisher,
     bootstrap_config: &BootstrapConfig,
     addr: &SocketAddr,
     pub_key: &PublicKey,
@@ -417,7 +418,7 @@ fn filter_bootstrap_list(
 pub async fn get_state(
     bootstrap_config: &BootstrapConfig,
     final_state: Arc<RwLock<FinalState>>,
-    mut establisher: Establisher,
+    mut establisher: impl BSEstablisher,
     version: Version,
     genesis_timestamp: MassaTime,
     end_timestamp: Option<MassaTime>,
--- a/massa-bootstrap/src/client.rs
+++ b/massa-bootstrap/src/client.rs
@@ -377,10 +377,8 @@ async fn connect_to_server(
     pub_key: &PublicKey,
 ) -> Result<BootstrapClientBinder, BootstrapError> {
     // connect
-    let mut connector = establisher
-        .get_connector(bootstrap_config.connect_timeout)
-        .await?; // cancellable
-    let socket = connector.connect(*addr).await?; // cancellable
+    let mut connector = establisher.get_connector(bootstrap_config.connect_timeout)?;
+    let socket = connector.connect(*addr).await?;
     Ok(BootstrapClientBinder::new(
         socket,
         *pub_key,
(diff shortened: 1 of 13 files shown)


massa-bootstrap/src/client.rs
```diff
@@ -371,16 +372,14 @@ async fn send_client_message(
 }
 
 async fn connect_to_server(
-    establisher: &mut Establisher,
+    establisher: &mut impl BSEstablisher,
```
Ben-PH, line 375: Monomorphisation will cause this function to be compiled once for each impl. In non-test builds, It's only the default establisher.

If we want to make many different establishers, we can use this pattern:
```rust
fn foo(many_impl: impl ToString) {
    fn concrete_foo(the_string) {
        // do the string thing
    }
    concrete_foo(many_impl.to_string())
```

This is an ugly pattern, and is one of rusts flaws that this optimisation needs a manual opt-in, but is available if needed (can explain whats happening if needed)

massa-bootstrap/src/establisher.rs
```diff
@@ -1,110 +1,112 @@
...
+use std::{io, net::SocketAddr};
 
-#[cfg(test)]
-pub mod types {
-    pub type Duplex = crate::tests::mock_establisher::Duplex;
+/// duplex connection
+pub type Duplex = tokio::net::TcpStream;
 
-    pub type Listener = crate::tests::mock_establisher::MockListener;
+/// duplex connection
```
Ben-PH, line 9:
```suggestion
```

massa-bootstrap/src/tests/mock_establisher.rs
```diff
@@ -1,19 +1,18 @@
 // Copyright (c) 2022 MASSA LABS <info@massa.net>
-
-use massa_models::config::{CHANNEL_SIZE, MAX_DUPLEX_BUFFER_SIZE};
+use async_trait::async_trait;
```
Ben-PH, line 2: Once everything is non-async and using threads, this will go away again

massa-bootstrap/src/tests/tools.rs
```diff
@@ -64,7 +64,7 @@ use tokio::io::AsyncReadExt;
 use tokio::io::AsyncWriteExt;
 use tokio::{sync::mpsc::Receiver, time::sleep};
 
-pub const BASE_BOOTSTRAP_IP: IpAddr = IpAddr::V4(Ipv4Addr::new(169, 202, 0, 10));
+pub const BASE_BOOTSTRAP_IP: IpAddr = IpAddr::V4(Ipv4Addr::new(127, 0, 0, 1));
```
Ben-PH, line 67: works well in conjunction with port 0.

massa-bootstrap/src/tests/tools.rs
```diff
@@ -301,7 +291,10 @@ pub fn get_bootstrap_config(bootstrap_public_key: NodeId) -> BootstrapConfig {
         write_timeout: 1000.into(),
         read_error_timeout: 200.into(),
         write_error_timeout: 200.into(),
-        bootstrap_list: vec![(SocketAddr::new(BASE_BOOTSTRAP_IP, 16), bootstrap_public_key)],
+        bootstrap_list: vec![(
+            SocketAddr::new(BASE_BOOTSTRAP_IP, 8069),
```
Ben-PH, line 295: ...nice

(port 16 wouldn't work)

massa-node/src/main.rs
```diff
@@ -236,7 +236,7 @@ async fn launch(
         res = get_state(
             &bootstrap_config,
             final_state.clone(),
-            massa_bootstrap::types::Establisher::default(),
+            massa_bootstrap::DefaultEstablisher::default(),
```
Ben-PH, line 239: Default establisher implements BSEstablisher trait now...

Ben-PH (commented):
Self review + comments for reviewers

Eitu33 (approved):

Ben-PH:
@Eitu33 thanks for the review. Ill merge once the CI passes, which should be fixed by #3676

modship (approved):

Issue #3644, opened 2023-03-06
Add validate_address()


litchipi:
**Technical specification**

-   ABI on massa-sc-runtime side : `assembly_script_validate_address(address: &str) -> i32`
-   Behavior of the function: Validate an address, returns 1 if correct, 0 if not

**Related pull request links**

-   massa-sc-runtime: https://github.com/massalabs/massa-sc-runtime/pull/238

@massalabs/run

PR #3640, opened 2023-03-03
Implement new ABI "get_matching_keys_for"

litchipi:
Closes: #3625 
PR in the `massa-sc-runtime` repo: https://github.com/massalabs/massa-sc-runtime/pull/237

New ABI to get all the keys in a datastore matching a given prefix

* [ ] document all added functions
* [ ] try in sandbox /simulation/labnet
* [ ] unit tests on the added/changed features
  * [ ] make tests compile
  * [ ] make tests pass 
* [ ] add logs allowing easy debugging in case the changes caused problems
* [ ] if the API has changed, update the API specification

PR #3645, opened 2023-03-06
Implement ABI validate_address()

litchipi:
Closes: https://github.com/massalabs/massa/issues/3644
Closes: https://github.com/massalabs/massa-sc-runtime/issues/142
Massa-sc-runtime PR: https://github.com/massalabs/massa-sc-runtime/pull/238

  * [ ] make tests compile
  * [ ] make tests pass 

### DIFF
--- a/massa-execution-worker/src/interface_impl.rs
+++ b/massa-execution-worker/src/interface_impl.rs
@@ -551,6 +551,10 @@ impl Interface for InterfaceImpl {
         Ok(addr.to_string())
     }
 
+    fn validate_address(&self, address: &str) -> Result<bool> {
+        Ok(massa_models::address::Address::from_str(address).is_ok())
+    }
+
     /// Verifies a signature
     ///
     /// # Arguments
--- a/massa-execution-worker/src/tests/scenarios_mandatories.rs
+++ b/massa-execution-worker/src/tests/scenarios_mandatories.rs
@@ -1916,4 +1916,90 @@ mod tests {
         // stop the execution controller
         manager.stop();
     }
+
+    #[test]
+    #[serial]
+    fn validate_address() {
+        let vesting = get_initials_vesting(false);
+        // setup the period duration and the maximum gas for asynchronous messages execution
+        let exec_cfg = ExecutionConfig {
+            t0: 100.into(),
+            max_async_gas: 100_000,
+            cursor_delay: 0.into(),
+            initial_vesting_path: vesting.path().to_path_buf(),
+            ..ExecutionConfig::default()
+        };
+        // get a sample final state
+        let (sample_state, _keep_file, _keep_dir) = get_sample_state().unwrap();
+
+        // init the storage
+        let mut storage = Storage::create_root();
+        // start the execution worker
+        let (mut manager, controller) = start_execution_worker(
+            exec_cfg.clone(),
+            sample_state.clone(),
+            sample_state.read().pos_state.selector.clone(),
+        );
+        // initialize the execution system with genesis blocks
+        init_execution_worker(&exec_cfg, &storage, controller.clone());
+
+        // keypair associated to thread 0
+        let keypair =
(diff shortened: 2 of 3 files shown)


massa-execution-worker/src/interface_impl.rs
```diff
@@ -534,6 +534,10 @@ impl Interface for InterfaceImpl {
         Ok(addr.to_string())
     }
 
+    fn validate_address(&self, address: &str) -> Result<bool> {
```
sydhds, line 537: Any reason to return here a Result as a bool looks enough?

sydhds (changes requested):
Looks good to me but there is no unit test. Is this intentional?

litchipi:
Not intentional, I'll try to create some

massa-execution-worker/src/interface_impl.rs
litchipi: Indeed, was just mimicking the other functions, but can take it out

litchipi: No I can't, because in the trait definition in `src/types.rs`, it fails as apparently `unimplemented!` will type check as `Result` for clippy

litchipi:
(Waiting for https://github.com/massalabs/massa-as-sdk/pull/164 to be merged before creating a unit test)

litchipi:
@sydhds Added a unit test, but I copied it from `datastore_manipulations` a bit stupidly so please could you review the code and tell me what are the un-necessary parts of it ?

Also I'll wait for your review on https://github.com/massalabs/massa-unit-tests-src/pull/33 before merging it (and this one also)

massa-execution-worker/src/interface_impl.rs
```diff
@@ -534,6 +534,10 @@ impl Interface for InterfaceImpl {
...
+        Ok(massa_models::address::Address::from_str(address).is_err())
```
sydhds, line 538: Wondering if it should not be is_ok() instead of is_err(). validate should return true if all went well no?

litchipi: Oops, I changed it but forgot to push it

massa-execution-worker/src/tests/scenarios_mandatories.rs
```diff
@@ -1890,4 +1890,84 @@ mod tests {
...
+        std::thread::sleep(
+            exec_cfg
+                .t0
+                .saturating_add(MassaTime::from_millis(50))
+                .into(),
+        );
+
+        let events = controller.get_filtered_sc_output_event(EventFilter::default());
+        // match the events
+        assert!(!events.is_empty(), "2 events were expected");
```
sydhds, line 1952: Why not matching directly the wanted length?

massa-execution-worker/src/tests/scenarios_mandatories.rs
```diff
@@ -1890,4 +1890,84 @@ mod tests {
...
+        println!("0: {:?}", events[0].data);
```
sydhds, line 1953: Still needed?

litchipi: Well as it's only printed if the test fails, I feel it's nice to see why it failed, but no it can be deleted if you think it's better

massa-execution-worker/src/tests/scenarios_mandatories.rs
litchipi: Fixed

massa-execution-worker/src/tests/scenarios_mandatories.rs
sydhds: If this is useful when the test fail maybe it should be include in the assert msg?

litchipi:
As I merged the unit tests for #3640 in the same unit tests release, it'll have to wait till it's merged before this can pass

AurelienFT (approved):

AurelienFT:
@sydhds can you resolve your comment ? as this PR blocks for testnet_21 to work well

sydhds (approved):

PR #3646, opened 2023-03-06
Mid term double stake protection

AurelienFT:
  * [x] make tests compile
  * [x] make tests pass 

Didn't tested it yet. We don't have unit tests for consensus. I will test in labnet if solution is validated.

### DIFF
--- a/massa-consensus-worker/src/state/process.rs
+++ b/massa-consensus-worker/src/state/process.rs
@@ -112,8 +112,11 @@ impl ConsensusState {
                         block_id
                     )));
                 };
-                // Verify that we didn't already received more than 2 blocks for this slot
-                let entry = self.blocks_per_slot.entry(header.content.slot).or_default();
+                // Verify that we haven't already received 2 blocks for this slot
+                let entry = self
+                    .nonfinal_active_blocks_per_slot
+                    .entry(header.content.slot)
+                    .or_default();
                 if !entry.contains(&header.id) {
                     if entry.len() > 1 && !self.wishlist.contains_key(&block_id) {
                         warn!(
@@ -153,9 +156,9 @@ impl ConsensusState {
                     .get(&block_id)
                     .cloned()
                     .expect("incoming block not found in storage");
-                // Verify that we didn't already received more than 2 blocks for this slot
+                // Verify that we haven't already received 2 blocks for this slot
                 let entry = self
-                    .blocks_per_slot
+                    .nonfinal_active_blocks_per_slot
                     .entry(stored_block.content.header.content.slot)
                     .or_default();
                 if !entry.contains(&stored_block.id) {
--- a/massa-consensus-worker/src/state/process.rs
+++ b/massa-consensus-worker/src/state/process.rs
@@ -20,7 +20,7 @@ use massa_models::{
 use massa_signature::PublicKey;
 use massa_storage::Storage;
 use massa_time::MassaTime;
-use tracing::log::{debug, info};
+use tracing::log::{debug, info, warn};
 
 use super::ConsensusState;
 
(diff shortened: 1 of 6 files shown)


massa-consensus-worker/src/state/process.rs
```diff
@@ -112,6 +112,19 @@ impl ConsensusState {
                         block_id
                     )));
                 };
+                // Verify that we didn't already received more than 2 blocks for this slot
```
damip, line 115:
```suggestion
                // Verify that we haven't already received 2 blocks for this slot
```

massa-consensus-worker/src/worker/init.rs
```diff
@@ -225,6 +232,13 @@ impl ConsensusWorker {
                     genesis_block_ids.iter().map(|h| (*h, 0)).collect();
                 write_shared_state.best_parents =
                     genesis_block_ids.iter().map(|v| (*v, 0)).collect();
+                for (thread, b) in genesis_block_ids.iter().enumerate() {
```
damip, line 235: no need to bother with this, we discard incoming blocks that are at or before the genesis period

massa-consensus-worker/src/worker/init.rs
```diff
@@ -200,6 +200,13 @@ impl ConsensusWorker {
                     final_blocks.iter().map(|(b, _)| b.block_id).collect();
                 write_shared_state.best_parents = latest_final_blocks_periods.clone();
                 write_shared_state.latest_final_blocks_periods = latest_final_blocks_periods;
+                for (b, _) in &final_blocks {
```
damip, line 203: no need to bother with this, we discard incoming blocks that are at or before the genesis period

massa-consensus-worker/src/state/prune.rs
```diff
@@ -135,30 +137,41 @@ impl ConsensusState {
...
                 None
             })
             .collect();
         discard_hashes.sort_unstable();
         discard_hashes.truncate(self.discarded_index.len() - self.config.max_discarded_blocks);
-        for (_, block_id) in discard_hashes.iter() {
+        for (_, slot, block_id) in discard_hashes.iter() {
             self.block_statuses.remove(block_id);
             self.discarded_index.remove(block_id);
+            match self.blocks_per_slot.entry(*slot) {
```
damip, line 166: Instead of clearing there, we should remove all slots in blocks_per_slot at or below the latest final slot of every thread.

massa-consensus-worker/src/worker/init.rs
AurelienFT: Done.

massa-consensus-worker/src/worker/init.rs
AurelienFT: Done.

massa-consensus-worker/src/state/prune.rs
AurelienFT: Done.

massa-consensus-worker/src/state/prune.rs
```diff
@@ -332,6 +332,21 @@ impl ConsensusState {
         Ok(())
     }
 
+    fn prune_nonfinal_blocks_per_slot(&mut self) {
```
damip, line 335: This is implemented in a very heavy way. Here is a higher performance option:

```
self.nonfinal_active_blocks_per_slot.retain(|s, _| s.period > self.latest_final_blocks_periods[s.thread as usize]);
```

AurelienFT: Done.

damip (approved):
LGTM :) gg

Issue #3650, opened 2023-03-06
Massa node crash if IPV6 is disabled on server.


AN09111976:
Hi,

2023-03-06T14:36:43.442323Z  INFO massa_bootstrap::client: Bootstrap from server 158.69.120.215:31245 failed. Your node will try to bootstrap from another server in 1m.
2023-03-06T14:37:43.442960Z  INFO massa_bootstrap::client: Start bootstrapping from [2001:41d0:1004:67::]:31245
2023-03-06T14:37:43.445509Z  WARN massa_bootstrap::client: Error while connecting to bootstrap server: io error: Address family not supported by protocol (os error 97)
2023-03-06T14:37:43.445649Z  INFO massa_bootstrap::client: Bootstrap from server [2001:41d0:1004:67::]:31245 failed. Your node will try to bootstrap from another server in 1m.
2023-03-06T14:38:43.446929Z  INFO massa_bootstrap::client: Start bootstrapping from 149.202.89.125:31245
2023-03-06T14:38:50.912205Z  INFO massa_bootstrap::client: State bootstrap complete
2023-03-06T14:38:50.945931Z  INFO massa_bootstrap::client: Successful bootstrap
2023-03-06T14:38:50.947159Z  INFO massa_network_worker: The node_id of this node is: [REMOVED]
thread 'main' panicked at 'could not start network controller: IOError(Os { code: 97, kind: Uncategorized, message: "Address family not supported by protocol" })', massa-node/src/main.rs:301:10
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace

On the server IPV6 is disabled globally, but in the massa-node config set IPV4 address explicitly:

[network]
bind = "0.0.0.0:31244"
routable_ip="A.B.C.D"

[bootstrap]
bind = "0.0.0.0:31245"

AurelienFT:
@Leo-Besancon this is related to the changes you made here : https://github.com/massalabs/massa/blob/main/massa-network-exports/src/establisher.rs#L87 ? How can use only ipv4 now ?

AN09111976:
Thanks for the answer. So now is it mandatory to have IPV6 enabled ?  For some reasons, I don't need IPV6 running on the server, what should I do ?

Leo-Besancon:
Hello @AN09111976, can you please tell me your system and what happens if  you keep the default config values for the network and bootstrap bind? ( `bind = "[::]:31244"` and `bind = "[::]:31245"`)

[...]

PR #3653, opened 2023-03-06
Fix bind ipv4 only network

Leo-Besancon:
  * [x] make tests compile
  * [x] make tests pass 

### DIFF
--- a/massa-bootstrap/src/establisher.rs
+++ b/massa-bootstrap/src/establisher.rs
@@ -78,6 +78,8 @@ pub mod types {
         /// * `addr`: `SocketAddr` we want to bind to.
         pub async fn get_listener(&mut self, addr: SocketAddr) -> io::Result<DefaultListener> {
             // Create a socket2 TCP listener to manually set the IPV6_V6ONLY flag
+            // This is needed to get the same behavior on all OS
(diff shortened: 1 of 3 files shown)


Leo-Besancon:
Context: issue https://github.com/massalabs/massa/issues/3650
The fix for ipv4 network on windows previously made here (https://github.com/massalabs/massa/pull/3590) was working on windows, but not on ubuntu with IPv6 disabled at a system level.
Here, someone with IPv6 disabled (and not "just" no IPv6 interface exposed) can bind to 0.0.0.0 for the network and bootstrap controllers.

Should I add some more comments on the config.toml bind settings?

aoudiamoncef (approved):
It seems good for me. Just add a small comment to explain the context. 

If you have a specific behaviour linked to a specific OS  in the future, take a look at  https://doc.rust-lang.org/reference/conditional-compilation.html#target_os, It may help.

Leo-Besancon:
> It seems good for me. Just add a small comment to explain the context.
> 
> If you have a specific behaviour linked to a specific OS in the future, take a look at https://doc.rust-lang.org/reference/conditional-compilation.html#target_os, It may help.

Great, will do, thanks for the link!

PR #3654, opened 2023-03-06
Fix vesting file (old timestamp)

### LINKED ISSUE #3651: Node panic on boot (SANDBOX)
Node crash on startup (SANDBOX) : 

image

Caused by unwrap and get_latest_block_slot_at_timestamp() used with a passed timestamp

modship:
  * [X] make tests compile
  * [X] make tests pass 

Fix #3651


massa-execution-worker/src/execution.rs
```diff
@@ -1470,16 +1470,25 @@ impl ExecutionState {
...
-                        prev.end_slot = next_range_slot.get_prev_slot(config.thread_count)?;
+                        if let Some(next_range_slot) =
+                            get_slot_at_timestamp(config, next.timestamp)?
+                        {
+                            prev.end_slot = next_range_slot.get_prev_slot(config.thread_count)?;
+                        }
 
                         Ok(prev)
                     })
+                    // we don't need range with StartSlot(0,0) && EndSlot(0,0)
```
AurelienFT, line 1486: We need the range with the start timestamp at genesis otherwise from genesis to X there is no vesting.

AurelienFT: No problem.

AurelienFT (approved):
Nice fix

PR #3657, opened 2023-03-08
Fix bind ipv4 only network (#3653)

Leo-Besancon:
* initial fix

* Do not set ipv6 flag if ipv4 only

* fmt

* Added documentation and comments

  * [ ] make tests compile
  * [ ] make tests pass 


AurelienFT (approved):
Tested in labnet.

AurelienFT:
bors merge

PR #3666, opened 2023-03-09
Feature/denunciation reloaded

sydhds:
  * [x] make tests compile
  * [x] make tests pass 

### DIFF
--- a/massa-models/src/block_header.rs
+++ b/massa-models/src/block_header.rs
@@ -89,14 +89,14 @@ impl SecureShareContent for BlockHeader {
         content: &Self,
         content_serialized: &[u8],
         content_creator_pub_key: &PublicKey,
-    ) -> Result<Hash, SerializeError> {
+    ) -> Hash {
         let de_data = BlockHeaderDenunciationData::new(content.slot);
 
         let mut hash_data = Vec::new();
         hash_data.extend(content_creator_pub_key.to_bytes());
         hash_data.extend(de_data.to_bytes());
         hash_data.extend(Hash::compute_from(content_serialized).to_bytes());
-        Ok(Hash::compute_from(&hash_data))
+        Hash::compute_from(&hash_data)
     }
 }
 
@@ -482,8 +482,7 @@ impl BlockHeaderDenunciationData {
     /// Get byte array
     pub fn to_bytes(&self) -> Vec<u8> {
         let mut buf = Vec::new();
-        buf.extend(self.slot.period.to_le_bytes());
-        buf.push(self.slot.thread);
+        buf.extend(self.slot.to_bytes_key());
         buf
     }
 }
--- a/massa-models/src/block_header.rs
+++ b/massa-models/src/block_header.rs
@@ -1,15 +1,19 @@
 use crate::block_id::BlockId;
 use crate::config::THREAD_COUNT;
+use crate::denunciation::{DenunciationData, DenunciationDataSerializer};
 use crate::endorsement::{
     Endorsement, EndorsementDeserializerLW, EndorsementId, EndorsementSerializer,
     EndorsementSerializerLW, SecureShareEndorsement,
 };
-use crate::secure_share::{SecureShare, SecureShareDeserializer, SecureShareSerializer};
+use crate::secure_share::{
+    SecureShare, SecureShareContent, SecureShareDeserializer, SecureShareSerializer,
+};
 use crate::slot::{Slot, SlotDeserializer, SlotSerializer};
 use massa_hash::{Hash, HashDeserializer};
 use massa_serialization::{
     Deserializer, SerializeError, Serializer, U32VarIntDeserializer, U32VarIntSerializer,
 };
+use massa_signature::PublicKey;
 use nom::branch::alt;
 use nom::bytes::complete::tag;
 use nom::error::{context, ContextError, ParseError};
@@ -79,6 +83,25 @@ impl BlockHeader {
 
 /// BlockHeader wrapped up alongside verification data
 pub type SecuredHeader = SecureShare<BlockHeader, BlockId>;
+
+impl SecureShareContent for BlockHeader {
+    fn compute_hash(
+        content: &Self,
+        content_serialized: &[u8],
+        content_creator_pub_key: &PublicKey,
+    ) -> Result<Hash, SerializeError> {
+        let de_data = DenunciationData::BlockHeader(content.slot);
+        let de_data_serializer = DenunciationDataSerializer::new();
+        let mut de_data_ser = Vec::new();
+        de_data_serializer.serialize(&de_data, &mut de_data_ser)?;
+
+        let mut hash_data = Vec::new();
+        hash_data.extend(content_creator_pub_key.to_bytes());
+        hash_data.extend(de_data_ser);
+        hash_data.extend(Hash::compute_from(content_serialized).to_bytes());
+        Ok(Hash::compute_from(&hash_data))
+    }
+}
 
 impl SecuredHeader {
     /// gets the header fitness
--- a/massa-models/src/block_header.rs
+++ b/massa-models/src/block_header.rs
@@ -84,6 +84,7 @@ impl BlockHeader {
 pub type SecuredHeader = SecureShare<BlockHeader, BlockId>;
 
 impl SecureShareContent for BlockHeader {
+    /// Compute hash for Block header in SecuredHeader - taking care of Denunciation verification
     fn compute_hash(
         content: &Self,
         content_serialized: &[u8],
--- a/massa-models/src/block_header.rs
+++ b/massa-models/src/block_header.rs
@@ -4,9 +4,7 @@ use crate::endorsement::{
     Endorsement, EndorsementDeserializerLW, EndorsementId, EndorsementSerializer,
     EndorsementSerializerLW, SecureShareEndorsement,
 };
-use crate::secure_share::{
-    SecureShare, SecureShareContent, SecureShareDeserializer, SecureShareSerializer,
-};
+use crate::secure_share::{SecureShare, SecureShareDeserializer, SecureShareSerializer};
 use crate::slot::{Slot, SlotDeserializer, SlotSerializer};
 use massa_hash::{Hash, HashDeserializer};
 use massa_serialization::{
@@ -96,8 +94,6 @@ impl SecuredHeader {
             .map_err(|er| format!("{}", er).into())
     }
 }
-
-impl SecureShareContent for BlockHeader {}
 
 /// Serializer for `BlockHeader`
 pub struct BlockHeaderSerializer {
--- a/massa-models/src/block_header.rs
+++ b/massa-models/src/block_header.rs
@@ -1,6 +1,5 @@
 use crate::block_id::BlockId;
 use crate::config::THREAD_COUNT;
-use crate::denunciation::{DenunciationData, DenunciationDataSerializer};
 use crate::endorsement::{
     Endorsement, EndorsementDeserializerLW, EndorsementId, EndorsementSerializer,
     EndorsementSerializerLW, SecureShareEndorsement,
@@ -90,14 +89,11 @@ impl SecureShareContent for BlockHeader {
         content_serialized: &[u8],
         content_creator_pub_key: &PublicKey,
     ) -> Result<Hash, SerializeError> {
-        let de_data = DenunciationData::BlockHeader(content.slot);
-        let de_data_serializer = DenunciationDataSerializer::new();
-        let mut de_data_ser = Vec::new();
-        de_data_serializer.serialize(&de_data, &mut de_data_ser)?;
+        let de_data = BlockHeaderDenunciationData::new(content.slot);
 
         let mut hash_data = Vec::new();
         hash_data.extend(content_creator_pub_key.to_bytes());
-        hash_data.extend(de_data_ser);
+        hash_data.extend(de_data.to_bytes());
         hash_data.extend(Hash::compute_from(content_serialized).to_bytes());
         Ok(Hash::compute_from(&hash_data))
     }
@@ -468,5 +464,25 @@ impl std::fmt::Display for BlockHeader {
             writeln!(f, "\tNo endorsements found")?;
         }
         Ok(())
+    }
+}
+
+/// A denunciation data for block header
+pub struct BlockHeaderDenunciationData {
+    slot: Slot,
+}
+
+impl BlockHeaderDenunciationData {
+    /// Create a new DenunciationData for block hedader
+    pub fn new(slot: Slot) -> Self {
+        Self { slot }
+    }
+
+    /// Get byte array
+    pub fn to_bytes(&self) -> Vec<u8> {
+        let mut buf = Vec::new();
+        buf.extend(self.slot.period.to_le_bytes());
+        buf.push(self.slot.thread);
+        buf
     }
 }
--- a/massa-models/src/denunciation.rs
+++ b/massa-models/src/denunciation.rs
@@ -4,7 +4,6 @@ use crate::block::Block;
 use crate::block_header::{BlockHeader, BlockHeaderSerializer, SecuredHeader};
 use crate::endorsement::{Endorsement, EndorsementSerializer, SecureShareEndorsement};
 use crate::operation::Operation;
-use crate::secure_share::Id;
 use crate::slot::{Slot, SlotSerializer};
 
 use massa_hash::Hash;
@@ -44,17 +43,6 @@ pub struct DenunciationDataSerializer {
     slot_serializer: SlotSerializer,
     index_serializer: U32VarIntSerializer,
 }
-
-/*
-impl Default for DenunciationDataSerializer {
-    fn default() -> Self {
-        Self {
-            slot_serializer: Default::default(),
-            index_serializer: Default::default(),
-        }
-    }
-}
-*/
 
 impl DenunciationDataSerializer {
     /// Create a new `DenunciationDataSerializer`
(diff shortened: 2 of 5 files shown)


massa-models/src/denunciation.rs
```diff
@@ -0,0 +1,659 @@
...
+        }))
+    }
+}
+
+/// Create a new Denunciation from 2 SecureHeader
+impl TryFrom<(&SecuredHeader, &SecuredHeader)> for Denunciation {
+    type Error = DenunciationError;
+
+    fn try_from((s_bh1, s_bh2): (&SecuredHeader, &SecuredHeader)) -> Result<Self, Self::Error> {
+        // Cannot use the same endorsement twice
```
bilboquet, line 341: isn't a test missing here?

```suggestion
// Cannot use the same endorsement twice
if s_bh1 == s_bh2 {
    return Err(DenunciationError::InvalidInput);
}
```

massa-models/src/denunciation.rs
```diff
@@ -0,0 +1,659 @@
...
+        // In order to create a Denunciation, there should be the same
+        // slot, index & public key
+        if s_bh1.content.slot != s_bh2.content.slot
+            || s_bh1.content_creator_pub_key != s_bh2.content_creator_pub_key
+            || s_bh1.serialized_data == s_bh2.serialized_data
+        {
+            return Err(DenunciationError::InvalidInput);
+        }
+
+        // Check sig of s_e2 but with s_e1.public_key, s_e1.slot, s_e1.index
```
bilboquet, line 351: Typo in the comment

```suggestion
// Check sig of s_bh2 but with s_bh1.public_key, s_bh1.slot, s_bh1.index
```

massa-models/src/denunciation.rs
bilboquet: Ok, you may forget this because
`s_bh1.serialized_data == s_bh2.serialized_data` => `s_bh1 == s_bh2`
Maybe I was confused buy the comment

massa-models/src/denunciation.rs
```diff
@@ -0,0 +1,659 @@
...
+            s_block_header_2,
+            s_block_header_3,
+        );
+    }
+
+    #[test]
+    fn test_block_header_denunciation() {
+        // Create an block header denunciation and check if it is valid
+        let (_slot, _keypair, s_block_header_1, s_block_header_2, s_block_header_3) =
+            gen_block_headers_for_denunciation();
```
bilboquet, line 593: You may want to add (and extend?) something like this:

```suggestion
    #[test]
    fn test_block_header_denunciation_invalid() {
        // Create a block header denunciation
        let (_slot, _keypair, s_block_header_1, _s_block_header_2, _s_block_header_3) =
            gen_block_headers_for_denunciation();

        let denunciation = Denunciation::try_from((&s_block_header_1, &s_block_header_1));
        assert!(matches!(denunciation, Err(DenunciationError::InvalidInput)));
    }

```

massa-models/src/denunciation.rs
```diff
@@ -0,0 +1,731 @@
...
+
+    #[test]
+    fn test_endorsement_denunciation_is_for() {
+        let (slot, keypair, s_endorsement_1, s_endorsement_2, s_endorsement_3) =
+            gen_endorsements_for_denunciation();
+
+        let denunciation: Denunciation = (&s_endorsement_1, &s_endorsement_2).try_into().unwrap();
+
+        assert_eq!(denunciation.is_for_endorsement(), true);
+        assert_eq!(denunciation.is_valid().unwrap_or(false), true);
```
bilboquet, line 554: In the context of tests, unwrap() should succeed, if it fails it's because we have change something. Here unwarp_or(false) may hide unwanted changes.
You may want to replace by:

```suggestion
        assert_eq!(denunciation.is_valid().unwrap(), true);
```
(also the case in other tests below)

massa-models/src/denunciation.rs
```diff
@@ -0,0 +1,731 @@
...
+            }
+        };
+        Ok(())
+    }
+}
+
+// Denounceable trait impl
+
+impl Denounceable for Endorsement {
+    fn is_denounceable() -> bool {
```
damip, line 76: isn't it enough to implem the Denounceable trait to make the object denouncable?

massa-models/src/denunciation.rs
sydhds: Indeed there is no PartialEq impl for SecuredHeader so I decided to check the serialized_data (which is enough in our case)

massa-models/src/denunciation.rs
```diff
@@ -0,0 +1,659 @@
...
+}
+
+/// A Serializer for `DenunciationData`
+#[derive(Default)]
+pub struct DenunciationDataSerializer {
+    slot_serializer: SlotSerializer,
+    index_serializer: U32VarIntSerializer,
+}
+
+/*
```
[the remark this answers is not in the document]
sydhds: Removed

massa-models/src/denunciation.rs
sydhds: done

massa-models/src/denunciation.rs
```diff
@@ -0,0 +1,731 @@
...
+
+impl EndorsementDenunciation {
+    /// Compute hash for Endorsement content
+    fn compute_content_hash(content: &Endorsement) -> Result<Hash, SerializeError> {
+        let mut buf = Vec::new();
+        let endorsement_serializer = EndorsementSerializer::new();
+        endorsement_serializer.serialize(content, &mut buf)?;
+        Ok(Hash::compute_from(&buf))
+    }
+
```
damip, line 117: Currently we avoid computing the hash twice and being sensitive to signature malleability by computing `hash(contents without signature)` to get the ID of the object, then this hash is the one directly signed/verified.

We should try to keep this behavior and simply redefine the ID computation as: `hash(denunciation_field1 + denunciation_field2 + hash(contents without signature))`.

massa-models/src/denunciation.rs
sydhds: Not really because we need to impl this for Operation and Block (because they can be included in SecureShare struct). For these 2 objects, we use the default implentation of the trait

massa-models/src/denunciation.rs
sydhds: fixed

massa-models/src/denunciation.rs

massa-models/src/block_header.rs
```diff
@@ -97,8 +95,6 @@ impl SecuredHeader {
     }
 }
 
-impl SecureShareContent for BlockHeader {}
```
damip, line 100: I would have left the implementation here instead of moving it to denunciation.rs, because that's actually the way the ID of BlockHeader is computed

massa-models/src/secure_share.rs
```diff
@@ -81,6 +78,19 @@ where
         })
     }
 
+    /// Compute hash
+    #[allow(unused_variables)]
+    fn compute_hash(
+        content: &Self, // use only for Denounce able object
+        content_serialized: &[u8],
+        content_creator_pub_key: &PublicKey,
+    ) -> Result<Hash, SerializeError> {
```
damip, line 87: for simplicity, this does not need to return an error

massa-models/src/secure_share.rs
```diff
@@ -137,6 +147,16 @@ where
         let creator_address = Address::from_public_key(&creator_public_key);
         let mut serialized_full_data = creator_public_key.to_bytes().to_vec();
         serialized_full_data.extend(&content_serialized);
```
damip, line 149: this is not needed

massa-models/src/secure_share.rs
sydhds: I need to return this kind of error because I use a DenunciationDataSerializer (for Endorsement & Block header)

massa-models/src/block_header.rs
sydhds: moved

massa-models/src/secure_share.rs
sydhds: removed

massa-models/src/block_header.rs
```diff
@@ -82,6 +84,25 @@ impl BlockHeader {
 /// BlockHeader wrapped up alongside verification data
 pub type SecuredHeader = SecureShare<BlockHeader, BlockId>;
 
+impl SecureShareContent for BlockHeader {
+    fn compute_hash(
```
damip, line 88: documentation

massa-models/src/denunciation.rs
```diff
@@ -0,0 +1,685 @@
...
+enum Denunciation {
+    Endorsement(EndorsementDenunciation),
+    BlockHeader(BlockHeaderDenunciation),
+}
+
+#[allow(dead_code)]
+impl Denunciation {
+    /// Check if it is a Denunciation of several endorsements
+    fn is_for_endorsement(&self) -> bool {
```
damip, line 157: to simplify code I wouldn't implement this helper because it is already an explicit enum

massa-models/src/denunciation.rs
```diff
@@ -0,0 +1,685 @@
...
+        matches!(self, Denunciation::Endorsement(_))
+    }
+
+    /// Check if it is a Denunciation of several block headers
+    fn is_for_block_header(&self) -> bool {
```
damip, line 162: same here

massa-models/src/endorsement.rs
```diff
@@ -166,7 +168,24 @@ impl SecureShareEndorsement {
 /// Wrapped endorsement
 pub type SecureShareEndorsement = SecureShare<Endorsement, EndorsementId>;
 
-impl SecureShareContent for Endorsement {}
+impl SecureShareContent for Endorsement {
+    fn compute_hash(
```
damip, line 172: TODO: doc

massa-models/src/endorsement.rs
```diff
@@ -166,7 +168,24 @@ impl SecureShareEndorsement {
...
+        content: &Self,
+        content_serialized: &[u8],
+        content_creator_pub_key: &PublicKey,
+    ) -> Result<Hash, SerializeError> {
+        let de_data = DenunciationData::Endorsement((content.slot, content.index));
```
damip, line 177: I find it a bit weird to use a Serializer for DenunciationData because:
* it doesn't really serialize (for example the variant number of the enum is not encoded)
* it can return an Error which complicates everything upstream, but compute_hash does not need to return a Result

damip:
I find it a bit weird to use a Serializer for DenunciationData because:
* it doesn't really serialize (for example the variant number of the enum is not encoded)
* it can return an Error which complicates everything upstream, but compute_hash does not need to return a Result

This can be simplified by having two strucutres (we don't need an enum there):
* EndorsementDenunciationData with a to_bytes() method
* BlockHeaderDenunciationData with a to_bytes() method

I would also define the EndorsementDenunciationData within the endorsement.rs file to avoid circular dependencies between denunciation.rs and endorsement.rs. Would do the same for the block header.

What do you think ?

sydhds:
> I find it a bit weird to use a Serializer for DenunciationData because:
> 
> * it doesn't really serialize (for example the variant number of the enum is not encoded)
> * it can return an Error which complicates everything upstream, but compute_hash does not need to return a Result
> 
> This can be simplified by having two strucutres (we don't need an enum there, does not serve any genericity purpose anyway):
> 
> * EndorsementDenunciationData with public fields and a to_bytes() method
> * BlockHeaderDenunciationData with public fields and a to_bytes() method
> 
> I would also define the EndorsementDenunciationData within the endorsement.rs file to avoid circular dependencies between denunciation.rs and endorsement.rs. Would do the same for the block header.
> 
> What do you think ?


Done

massa-models/src/endorsement.rs
```diff
@@ -166,7 +167,25 @@ impl SecureShareEndorsement {
...
 pub type SecureShareEndorsement = SecureShare<Endorsement, EndorsementId>;
 
-impl SecureShareContent for Endorsement {}
+impl SecureShareContent for Endorsement {
+    fn compute_hash(
+        content: &Self,
+        content_serialized: &[u8],
+        content_creator_pub_key: &PublicKey,
+    ) -> Result<Hash, SerializeError> {
+        // let de_data = DenunciationData::Endorsement((content.slot, content.index));
```
bilboquet, line 176: Old commented code

massa-models/src/denunciation.rs
sydhds: It think it really simplifies the code especially in unit test

massa-models/src/block_header.rs
sydhds: done

massa-models/src/endorsement.rs
sydhds: done

massa-models/src/endorsement.rs
sydhds: done

massa-models/src/block_header.rs
```diff
@@ -451,3 +467,23 @@ impl std::fmt::Display for BlockHeader {
...
+    pub fn new(slot: Slot) -> Self {
+        Self { slot }
+    }
+
+    /// Get byte array
+    pub fn to_bytes(&self) -> Vec<u8> {
+        let mut buf = Vec::new();
+        buf.extend(self.slot.period.to_le_bytes());
+        buf.push(self.slot.thread);
+        buf
```
damip, line 487: you can use directly `self.slot.to_bytes_key()`

massa-models/src/endorsement.rs
```diff
@@ -385,9 +400,31 @@ impl Deserializer<Endorsement> for EndorsementDeserializerLW {
...
+    /// Create a new denunciation data for endorsement
+    pub fn new(slot: Slot, index: u32) -> Self {
+        Self { slot, index }
+    }
+
+    /// Get byte array
+    pub fn to_bytes(&self) -> Vec<u8> {
+        let mut buf = Vec::new();
+        buf.extend(self.slot.period.to_le_bytes());
+        buf.push(self.slot.thread);
```
damip, line 419: here as well, self.slot.to_bytes_key()

massa-models/src/secure_share.rs
damip: now the Result can be removed

massa-models/src/secure_share.rs
```diff
@@ -135,8 +145,15 @@ where
             serialized_data[..serialized_data.len() - rest.len()].to_vec()
         };
         let creator_address = Address::from_public_key(&creator_public_key);
-        let mut serialized_full_data = creator_public_key.to_bytes().to_vec();
-        serialized_full_data.extend(&content_serialized);
+        let hash = Self::compute_hash(&content, &content_serialized, &creator_public_key).map_err(
```
damip, line 148: compute_hash should never fail anymore

massa-models/src/denunciation.rs
```diff
@@ -0,0 +1,642 @@
...
+                            .verify_signature(&hash, &s_block_header.signature)
+                            .is_ok())
+                } else {
+                    Ok(false)
+                }
+            }
+        }
+    }
+
+    fn is_valid(&self) -> Result<bool, DenunciationError> {
```
damip, line 193: is this also only for tests ?

massa-models/src/denunciation.rs
```diff
@@ -0,0 +1,642 @@
...
+
+/// Create a new Denunciation from 2 SecureShareEndorsement
+impl TryFrom<(&SecureShareEndorsement, &SecureShareEndorsement)> for Denunciation {
+    type Error = DenunciationError;
+
+    fn try_from(
+        (s_e1, s_e2): (&SecureShareEndorsement, &SecureShareEndorsement),
+    ) -> Result<Self, Self::Error> {
+        // Cannot use the same endorsement twice
+        if s_e1 == s_e2 {
```
damip, line 254: we can compare IDs

damip: I don't remember, maybe this is only useful for tests ?

massa-models/src/denunciation.rs
```diff
@@ -0,0 +1,642 @@
...
+        // slot, index & public key
+        if s_e1.content.slot != s_e2.content.slot
+            || s_e1.content.index != s_e2.content.index
+            || s_e1.content_creator_pub_key != s_e2.content_creator_pub_key
+        {
+            return Err(DenunciationError::InvalidInput);
+        }
+
+        // Check sig of s_e1 with s_e1.public_key, s_e1.slot, s_e1.index
+        let s_e1_hash_content = EndorsementDenunciation::compute_content_hash(&s_e1.content)?;
```
damip, line 268: we shouldn't have to recompute this, because we have the ID of the object. But maybe it's only for tests (?)

massa-models/src/denunciation.rs
```diff
@@ -0,0 +1,642 @@
...
+/// Create a new Denunciation from 2 SecureHeader
+impl TryFrom<(&SecuredHeader, &SecuredHeader)> for Denunciation {
+    type Error = DenunciationError;
+
+    fn try_from((s_bh1, s_bh2): (&SecuredHeader, &SecuredHeader)) -> Result<Self, Self::Error> {
+        // Cannot use the same block header twice
+        // In order to create a Denunciation, there should be the same slot, index & public key
+        if s_bh1.content.slot != s_bh2.content.slot
+            || s_bh1.content_creator_pub_key != s_bh2.content_creator_pub_key
+            || s_bh1.serialized_data == s_bh2.serialized_data
```
damip, line 310: I guess it's only for tests because this comparison is heavy

massa-models/src/secure_share.rs
sydhds: Removed

massa-models/src/block_header.rs
sydhds: Done

massa-models/src/secure_share.rs
sydhds: Fixed

massa-models/src/denunciation.rs
sydhds: Add a TODO so this will be done in a follow up

massa-models/src/denunciation.rs
sydhds: Add a TODO so this will be done in a follow up

massa-models/src/denunciation.rs
sydhds: Add a TODO so this will be done in a follow up

massa-models/src/denunciation.rs
sydhds: Add a TODO so this will be done in a follow up

massa-models/src/denunciation.rs
modship: or symply 
```rust
assert!(denunciation.is_valid().unwrap());
```

PR #3675, opened 2023-03-10
MipStoreRaw::update_with now return enum

sydhds:
  * [ ] make tests compile
  * [ ] make tests pass 

### DIFF
--- a/massa-versioning-worker/src/versioning.rs
+++ b/massa-versioning-worker/src/versioning.rs
@@ -461,20 +461,14 @@ impl<const N: usize> TryFrom<[(MipInfo, MipState); N]> for MipStore {
     }
 }
 
-#[derive(Debug, PartialEq)]
-pub enum UpdateWithResult {
-    Success(Vec<MipInfo>, Vec<MipInfo>),
-    Fail,
-}
-
 /// Store of all versioning info
 #[derive(Debug, Clone, Default)]
 pub struct MipStoreRaw(pub(crate) BTreeMap<MipInfo, MipState>);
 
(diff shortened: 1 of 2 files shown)


massa-versioning-worker/src/versioning.rs
```diff
@@ -461,18 +461,27 @@ impl<const N: usize> TryFrom<[(MipInfo, MipState); N]> for MipStore {
...
+    Fail,
+}
+
 /// Store of all versioning info
 #[derive(Debug, Clone, Default)]
 pub struct MipStoreRaw(pub(crate) BTreeMap<MipInfo, MipState>);
 
 impl MipStoreRaw {
     /// Update our store with another (usually after a bootstrap where we received another store)
     /// Return true if update is successful, false if something is wrong on given store raw
```
Leo-Besancon, line 476: Change doc comments for return value

massa-versioning-worker/src/versioning.rs
```diff
@@ -461,18 +461,27 @@ impl<const N: usize> TryFrom<[(MipInfo, MipState); N]> for MipStore {
     }
 }
 
+#[derive(Debug, PartialEq)]
+pub enum UpdateWithResult {
+    Success(Vec<MipInfo>, Vec<MipInfo>),
```
Leo-Besancon, line 466: Add a comment to understand the "update"/"added" order here (otherwise we don't understand what the two Vec<MipInfo> are).

Leo-Besancon (approved):
You really don't like the `Result<>` Rust enum :p.
Good addition!

PR #3685, opened 2023-03-14
Bootstrap/sync accept

Ben-PH:
  * [x] make tests compile
  * [x] make tests pass 


Ben-PH:
@modship @Eitu33 This PR pre-empts some changes that lead into #3730 . It might be nice to include them as well, but it wouldn't be an issue at all to do them separately

massa-bootstrap/src/tests/scenarios.rs
```diff
@@ -177,24 +183,22 @@ async fn test_bootstrap_server() {
...
+        assert_eq!(
+            conn_addr, expect_conn_addr,
+            "client connected to wrong bootstrap ip"
+        );
+        waker.store(true, Ordering::Relaxed);
+        remote_rw
+    });
 
     // connect to bootstrap
     let remote_addr = std::net::SocketAddr::from_str("82.245.72.98:10000").unwrap(); // not checked
```
modship, line 200: what is this address ?

Ben-PH: it's tcp stream of the mocked client. When calling `get_state` the requests _to_ the bootstrap are sent _from_ this stream (At least, that's my best understanding. This system is pretty tough to reason about. #3730 removes it alltogether, pretty much.

massa-bootstrap/src/establisher.rs
```diff
@@ -1,48 +1,61 @@
...
+        // However, if IPv6 is disabled system-wide, you may need to bind to an IPv4 address instead.
+        let domain = match addr.is_ipv4() {
+            true => socket2::Domain::IPV4,
+            _ => socket2::Domain::IPV6,
+        };
+
+        let socket = socket2::Socket::new(domain, socket2::Type::STREAM, None)?;
+
+        if addr.is_ipv6() {
+            socket.set_only_v6(false)?;
```
modship, line 44: sure ?

massa-bootstrap/src/server.rs
```diff
@@ -98,19 +102,16 @@ impl BootstrapManager {
...
     final_state: Arc<RwLock<FinalState>>,
     config: BootstrapConfig,
-    mut establisher: impl BSEstablisher,
+    listener: impl BSListener + Send + 'static,
     keypair: KeyPair,
     version: Version,
-) -> Result<Option<BootstrapManager>, Box<BootstrapError>> {
+) -> Result<Option<BootstrapManager<TcpStream>>, Box<BootstrapError>> {
     massa_trace!("bootstrap.lib.start_bootstrap_server", {});
-    let Some(listen_addr) = config.listen_addr else {
```
modship, line 111: we should keep that no ?

massa-bootstrap/src/server.rs
```diff
@@ -251,11 +249,13 @@ impl BootstrapServer<'_> {
...
         mut listener: impl BSListener,
-        listener_tx: crossbeam::channel::Sender<BsConn>,
-    ) -> Result<Result<(), BsConn>, Box<BootstrapError>> {
+        listener_tx: crossbeam::channel::Sender<BsConn<TcpStream>>,
+    ) -> Result<Result<(), BsConn<TcpStream>>, Box<BootstrapError>> {
         loop {
-            let msg = listener.accept().await.map_err(BootstrapError::IoError)?;
-            match listener_tx.send(msg) {
+            let (msg, addr) = listener.accept().map_err(BootstrapError::IoError)?;
+            msg.set_nonblocking(true).unwrap();
```
modship, line 256: you should map_err to BootstrapError

massa-bootstrap/src/establisher.rs
Ben-PH: ...hmmm. probably not, now that you mention it. At some point I removed the Establisher trait/struct entirely, and some wires must've gotten crossed in my head.

The fix will be to have this either in the establisher OR here, but not both.

massa-bootstrap/src/server.rs
Ben-PH: it's changed from an establisher (with a `get_listener` call further down) to just passing in a listener.

listen address is no longer needed, because it was only used to get the listener. This now happens before you call the bootstrap start

massa-network-exports/src/establisher.rs
```diff
@@ -102,7 +102,9 @@ mod types {
             // Number of connections to queue, set to the hardcoded value used by tokio
             socket.listen(1024)?;
 
-            Ok(DefaultListener(TcpListener::from_std(socket.into())?))
+            let socket: std::net::TcpListener = socket.into();
+            socket.set_nonblocking(true).unwrap();
```
modship, line 106: should propagate error

modship (commented):
there is still a lot of unwrap() :)

massa-bootstrap/src/establisher.rs
modship: ok, same at line 103

massa-network-exports/src/establisher.rs
Ben-PH: done https://github.com/massalabs/massa/pull/3685/commits/210e52e368844ba0859b46d7f0224a035fed021f

Ben-PH:
If im reading the CI failure right, it's failing on a file that I haven't changed (looking at the PR) and is different to what I have on my file locally... that's _weird_

massa-bootstrap/src/server.rs
Ben-PH: This change here can provide additional context https://github.com/massalabs/massa/pull/3685/commits/840cfa5d0346c2ca083ea5556e436e81c375ec9f

massa-bootstrap/src/establisher.rs
Ben-PH: done: [0721d10](https://github.com/massalabs/massa/pull/3685/commits/0721d10d0ec91df983bce12e059cfcc8d62842c9)

Ben-PH: note: the setting to non-blocking is deferred until the `run_listener` method

AurelienFT:
@Ben-PH I have fixed it on testnet_21 just now. You can update.

modship (approved):

massa-bootstrap/src/establisher.rs
```diff
@@ -1,91 +1,35 @@
...
+    net::{SocketAddr, TcpListener, TcpStream},
+};
 
 /// duplex connection
-pub type Duplex = tokio::net::TcpStream;
+pub trait Duplex:
+// static because need to send between threads :(
+    'static  + Send + tokio::io::AsyncReadExt + tokio::io::AsyncWriteExt + std::marker::Unpin
+{
+}
```
Eitu33, line 13: Requiring a static lifetime here is fine on the server side as it will stay up for the whole duration of the program. But this trait is also used on the client side which is only going to exist for one bootstrap duration, and `'static` objects will outlive it even though we have no need for it. One solution could be to use reference counters.

massa-bootstrap/src/client.rs
```diff
@@ -371,17 +371,20 @@ async fn send_client_message(
...
+) -> Result<BootstrapClientBinder<tokio::net::TcpStream>, Box<BootstrapError>> {
+    let socket = connector.connect(*addr).map_err(|e| Box::new(e.into()))?;
+    socket
+        .set_nonblocking(true)
+        .map_err(|e| Box::new(BootstrapError::IoError(e)))?;
     Ok(BootstrapClientBinder::new(
-        socket,
+        // this from_std will panic if this method doesn't exist within an async runtime...
+        tokio::net::TcpStream::from_std(socket)
+            .map_err(|e| Box::new(BootstrapError::IoError(e)))?,
```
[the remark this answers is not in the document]
Ben-PH: how about something like this:

"from_std will panic, unless called within a tokio-runtime with IO not enabled. This is currently not a problem, and won't be unless these conditions are not met"?

massa-bootstrap/src/establisher.rs
Ben-PH: tried playing around with that, and couldn't get rid of the static.even with use of Arc, but the 'static constraint can be moved to constraints on implementations that only apply in a server-context (not client)

can be found here: [f040213](https://github.com/massalabs/massa/pull/3685/commits/f040213525793787e28c85a81d743f234207c545)

massa-bootstrap/src/establisher.rs
```diff
@@ -1,91 +1,35 @@
...
-use std::{io, net::SocketAddr};
+use std::{
+    io,
+    net::{SocketAddr, TcpListener, TcpStream},
+};
 
 /// duplex connection
-pub type Duplex = tokio::net::TcpStream;
+pub trait Duplex:
+// static because need to send between threads :(
```
modship, line 10: update comment (static)

Ben-PH: fixed in 281bd5a

massa-bootstrap/src/client.rs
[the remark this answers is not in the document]
Ben-PH: How does this sound? [0708ac4](https://github.com/massalabs/massa/pull/3685/commits/0708ac4184016494a6049bb9d3fd21c8534dd6b2)

Eitu33 (approved):

PR #3489, opened 2023-02-01
Testnet 20

AurelienFT:
Merged in this testnet:

- #3475 
- #3549
- #3562 
- #3462 
- #3492 
- #3502 
- #3495 
- #3556 
- #3511 
- #3498 
- #3566 
- #3557 
- #3576 
- #3579 
- #3507 
- #3585 
- #3587 
- #3580 
- #3590 
- #3549 
- #3455 
- #3601 
- #3602 
- #3606 
- #3588 
- #3603 
- #3554

PR #3695, opened 2023-03-20
Use saturating sub as appropriate for thread-count

Ben-PH:
  * [x] make tests compile
  * [x] make tests pass 

* [x] confirm `.get() - 1` only: no subtractions that can potentially underflow

https://github.com/massalabs/massa/pull/3489#discussion_r1142152635

### DIFF
--- a/massa-models/src/slot.rs
+++ b/massa-models/src/slot.rs
@@ -178,7 +178,7 @@ impl Slot {
             .ok_or(ModelsError::PeriodOverflowError)?;
         Ok(Slot {
             period,
-            thread: thread_count - 1,
+            thread: thread_count.saturating_sub(1),
         })
     }
 
@@ -202,7 +202,7 @@ impl Slot {
     pub const fn max(thread_count: u8) -> Slot {
         Slot {
             period: u64::MAX,
-            thread: thread_count,
+            thread: thread_count.saturating_sub(1),
         }
     }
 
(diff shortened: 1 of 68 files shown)


Ben-PH:
I'll see what I can do with `NonZeroU8`:

Ben-PH:
I've changed thread-count to use `NonZeroU8` wherever I could find `thread_count: u8`.

This introduces one line of unsafe code:

https://github.com/massalabs/massa/pull/3695/files#diff-e151d5d22edfdbbd0bc0fb2bb32a88393e99a976e67ee8be309af16d59816332R89-R91


I'll self review for arithmatic issues

Ben-PH (commented):
todo: go through "as usize/u64" and use `.into()` were possible

damip:
Note: doing thread_count.get() -1 is defeating the purpose of the type that explicitly does not implement the "+" and "-" operators for very good reasons. Use saturating_sub or checked_sub.

damip:
This is a NON-PRIORITARY refactoring

massa-models/src/slot.rs
```diff
@@ -199,10 +200,10 @@ impl Slot {
     }
 
     /// returns the maximal slot
-    pub const fn max(thread_count: u8) -> Slot {
+    pub const fn max(thread_count: NonZeroU8) -> Slot {
         Slot {
             period: u64::MAX,
-            thread: thread_count,
+            thread: thread_count.get() - 1,
```
damip, line 206: this is the only PRIORITARY change

PR #3697, opened 2023-03-20
Testnet 20_2 fix vesting

modship:
  * [x] make tests compile
  * [x] make tests pass 


massa-execution-worker/src/execution.rs
```diff
@@ -1470,16 +1470,25 @@ impl ExecutionState {
...
+                        {
+                            prev.end_slot = next_range_slot.get_prev_slot(config.thread_count)?;
+                        }
 
                         Ok(prev)
                     })
+                    // we don't need range with StartSlot(0,0) && EndSlot(0,0)
+                    // this can happen when the timestamp is passed
+                    .filter(|a| {
+                        !(a.as_ref().unwrap().end_slot == Slot::min()
```
damip, line 1489: is this unwrap safe ? if so, it needs to be explained in a comment

massa-execution-worker/src/execution.rs
```diff
@@ -1470,16 +1470,25 @@ impl ExecutionState {
...
+                            && a.as_ref().unwrap().start_slot == Slot::min())
```
damip, line 1490: is this unwrap safe ? if so, it needs to be explained in a comment

damip (approved):

PR #3698, opened 2023-03-20
Drop then recreate objects on bootstrap (fixes some comments form #3489)

bilboquet:
  * [ ] make tests compile
  * [ ] make tests pass 

### DIFF
--- a/massa-bootstrap/src/client.rs
+++ b/massa-bootstrap/src/client.rs
@@ -170,7 +170,7 @@ async fn stream_final_state_and_consensus(
 
                     // TODO:
                     // Building a new FinalState outside of this function would probably be better.
-                    // There are to much deps to manage for the moment.
+                    // There are too much deps to manage for the moment.
                     let mut write_final_state = global_bootstrap_state.final_state.write();
                     write_final_state.reset();
 
--- a/massa-bootstrap/src/client.rs
+++ b/massa-bootstrap/src/client.rs
@@ -170,7 +170,7 @@ async fn stream_final_state_and_consensus(
 
                     // TODO:
                     // Building a new FinalState outside of this function would probably be better.
-                    // There are too much deps to manage for the moment.
+                    // There are too many deps to manage for the moment.
                     let mut write_final_state = global_bootstrap_state.final_state.write();
                     write_final_state.reset();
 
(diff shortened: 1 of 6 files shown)


massa-bootstrap/src/client.rs
```diff
@@ -168,8 +168,13 @@ async fn stream_final_state_and_consensus(
                         last_ops_step: StreamingStep::Started,
                         last_consensus_step: StreamingStep::Started,
                     };
+
+                    // TODO:
+                    // Building a new FinalState outside of this function would probably be better.
+                    // There are to much deps to manage for the moment.
```
sydhds, line 174: to much -> too much?

massa-pos-exports/src/pos_final_state.rs
```diff
@@ -64,12 +64,22 @@ impl PoSFinalState {
...
-    /// Reset the state of the PoS final state
-    ///
-    /// USED ONLY FOR BOOTSTRAP
-    pub fn reset(&mut self) {
-        self.cycle_history.clear();
-        self.deferred_credits = DeferredCredits::default();
+    /// Inspired by mem::take()
+    /// This function reinitialize in place the given object.
+    /// It reuses what it can from the given reference.
+    /// The implementation ensure that all fields are tracked.
```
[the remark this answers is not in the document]
bilboquet: done

massa-final-state/src/final_state.rs
```diff
@@ -86,9 +86,9 @@ impl FinalState {
     pub fn reset(&mut self) {
```
bilboquet, line 86: Tradeoff here see the followup in  #3728

damip: Why not just recreate the final_state object everytime we attempt a bootstrap ?

bilboquet: IIRC I did not find a way with minimal modifications mainly because the final_state is behind a & mut.

AurelienFT: After analyzing the code. The final state is created before the bootstrap and sent as a `RwLock` to the bootstrap but I don't think it's necessary as it's not used anywhere before getting filled by the bootstrap. So we could create the first one before the bootstrap or inside, and return the filled one to be used in the rest of the code.

As the code that launch the bootstrap is under a `tokio::select` I think it should be better to continue on that PR after bootstrap de-async

PR #3710, opened 2023-03-21
Check operation count returned in block factory

sydhds:
  * [ ] make tests compile
  * [ ] make tests pass 

### DIFF
--- a/massa-factory-worker/src/block_factory.rs
+++ b/massa-factory-worker/src/block_factory.rs
@@ -223,11 +223,13 @@ impl BlockFactoryWorker {
         .expect("error while producing block header");
 
         // create block
+        let block_ = Block {
+            header,
+            operations: op_ids.into_iter().collect()
+        };
+
         let block = Block::new_verifiable(
-            Block {
-                header,
-                operations: op_ids.into_iter().collect(),
-            },
+            block_,
             BlockSerializer::new(), // TODO reuse self.block_serializer
             block_producer_keypair,
         )
--- a/massa-factory-worker/src/block_factory.rs
+++ b/massa-factory-worker/src/block_factory.rs
@@ -38,7 +38,7 @@ impl BlockFactoryWorker {
         wallet: Arc<RwLock<Wallet>>,
         channels: FactoryChannels,
         factory_receiver: mpsc::Receiver<()>,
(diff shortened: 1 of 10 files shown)


massa-protocol-worker/src/protocol_network.rs
```diff
@@ -353,6 +353,13 @@ impl ProtocolWorker {
 
         // Check operation_list against expected operations hash from header.
         if header.content.operation_merkle_root == Hash::compute_from(&total_hash) {
+
+            if operation_ids.len() > self.config.max_operations_per_block as usize {
+                warn!("Node id {} sent us a operation list for block id {} but the operations we already have in our records exceed max operations per block constant.", from_node_id, block_id);
```
damip, line 358: it's not the operations we have in our records, but the list of operation_ids that were sent

damip:
TODO: also add the operation count check on receiving the full operations for a block (not only the list)

sydhds:
> TODO: also correct the deserialization limit problem that was found in the issue

done

massa-factory-worker/src/block_factory.rs
```diff
@@ -195,6 +195,16 @@ impl BlockFactoryWorker {
 
         // gather operations and compute global operations hash
         let (op_ids, op_storage) = self.channels.pool.get_block_operations(&slot);
+
+        let Ok(max_ops_usize) = usize::try_from(self.cfg.max_operations_per_block) else {
```
damip, line 199: I think we can just use "self.cfg.max_operations_per_block as usize" so that it panics if the config contains something really weird

damip (approved):

damip:
LGTM with nits

massa-factory-worker/src/block_factory.rs
Ben-PH: proposition statement: "having `as $INT_TYPE` throughout the code is a smell. use `.into()` or `.try_into()` instead"

Should we avoid the use of `as` instead of using the conversion methods, and setup some sort of standard to handle fallable conversion errors? (outside of scope here, of course)

Ben-PH: conversion use reverted

damip (approved):

PR #3712, opened 2023-03-21
End of free gas

### LINKED ISSUE #3700: Wasm instructions gas cost
We should never be zero because someone could just make an infinite loop of just those instructions.

_Originally posted by @damip in https://github.com/massalabs/massa/pull/3489#discussion_r1142034058_

See also : 
https://github.com/massalabs/massa/pull/3489#discussion_r1142034182
https://github.com/massalabs/massa/pull/3489#discussion_r1142034269
https://github.com/massalabs/massa/pull/3489#discussion_r1142034734

Related to https://github.com/massalabs/gas-calibration/issues/4

aoudiamoncef:
closes #3700

  * [x] make tests compile
  * [x] make tests pass 

### DIFF
--- a/massa-node/base_config/gas_costs/abi_gas_costs.json
+++ b/massa-node/base_config/gas_costs/abi_gas_costs.json
@@ -1,11 +1,11 @@
 {
-  "assembly_script_abort": 0,
+  "assembly_script_abort": 1,
   "assembly_script_address_from_public_key": 294,
   "assembly_script_append_data": 198,
   "assembly_script_append_data_for": 231,
   "assembly_script_call": 10000,
   "assembly_script_caller_has_write_access": 146,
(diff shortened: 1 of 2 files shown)


massa-node/base_config/gas_costs/abi_gas_costs.json
```diff
@@ -1,11 +1,11 @@
 {
-  "assembly_script_abort": 0,
+  "assembly_script_abort": 1,
```
Leo-Besancon, line 2: Does it make sence to have the abort be at 1? It's probably fine but I'm not sure how it all fits ;). E.g if abort is auto called when out of gas or something weird like that.

Leo-Besancon (approved):

damip (approved):

PR #3716, opened 2023-03-22
Fix problem joining the threads in the tests

litchipi:
See https://github.com/massalabs/PeerNet/issues/14

I wonder why the issue is in Peernet tho.

I used a `Arc<RwLock<bool>>` as a thread-safe flag to notify all the launched threads that we are stopping.
It can be really easily embedded into any `select!` using a `default(Duration(...)) =>` that breaks the loop of the thread.
Using this, the cargo test now works


AurelienFT (commented):
I think using a `AtomicBool` instead of a `RwLock<bool>` could be better (https://stackoverflow.com/a/52466851). Also instead of making the channel not blocking it couldn't be possible (for handlers for example) to drop all senders and so the `recv()` will return with an error and this error is used to kill the thread. Because your solution, if the thread as a lot of things to recv can make the stop taking a lot of time. (Reference in [Rust Atomics and Locks](https://marabos.nl/atomics/atomics.html#example-stop-flag) :

> This simple solution works great as long as the flag is regularly checked by the background thread. If it gets stuck in some_work() for a long time, that can result in an unacceptable delay between the stop command and the program quitting.

)

litchipi:
Ok I see.
However all the `Sender` are passed to `PeerNetManager` through its config, so I cannot really `drop` them here easily.
I just cannot drop the manager entirely as it provokes panics everywhere.
Did you have a vague idea of a fix in mind ? Because this seams blocking on an code architecture level

litchipi:
Related: https://github.com/massalabs/PeerNet/pull/26

PR #3717, opened 2023-03-22
Feature/versioning ser der 3

sydhds:
  * [ ] make tests compile
  * [ ] make tests pass 

### DIFF
--- a/massa-versioning-worker/src/versioning_ser_der.rs
+++ b/massa-versioning-worker/src/versioning_ser_der.rs
@@ -1,15 +1,16 @@
+use std::mem;
 use std::ops::Bound::{Excluded, Included};
 
-use nom::error::context;
-use nom::sequence::tuple;
-use nom::Parser;
 use nom::{
+    error::context,
     error::{ContextError, ParseError},
-    IResult,
+    sequence::tuple,
+    IResult, Parser,
 };
 
-use crate::versioning::{MipComponent, MipInfo};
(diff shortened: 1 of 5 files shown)


massa-versioning-worker/src/versioning_ser_der.rs
```diff
@@ -170,10 +179,463 @@ impl Deserializer<MipInfo> for MipInfoDeserializer {
...
+    amount_deserializer: AmountDeserializer,
+}
+
+impl ComponentStateDeserializer {
+    /// Creates a new ``
+    pub fn new() -> Self {
+        Self {
+            state_deserializer: U32VarIntDeserializer::new(
+                Included(0),
+                Excluded(COMPONENT_STATE_VARIANT_COUNT + 1),
```
Leo-Besancon, line 245: If the +1 is to account for the Error, why don't we use `COMPONENT_STATE_ID_VARIANT_COUNT` here?
More generally, I think having two different counts may be prone to errors (especially has they differ by just one, and the Error state is added automatically!).
At the very least, needs more documentations.

Leo-Besancon (approved):

massa-versioning-worker/src/versioning_ser_der.rs
sydhds: Good catch! I've removed this +1 :)

Leo-Besancon: Ok great!

PR #3718, opened 2023-03-22
Minor fix to optim denunciation creation from 2 SecuredHeader

sydhds:
  * [ ] make tests compile
  * [ ] make tests pass 


bilboquet (approved):
ok for me.

PR #3720, opened 2023-03-22
Add EndorsementDenunciation ser / der + unit test

sydhds:
  * [ ] make tests compile
  * [ ] make tests pass 


massa-models/src/denunciation.rs
```diff
@@ -358,16 +390,341 @@ pub enum DenunciationError {
     Serialization(#[from] SerializeError),
 }
 
+// Serialization / Deserialization
+
+/// Serializer for `EndorsementDenunciation`
+struct EndorsementDenunciationSerializer {
```
bilboquet, line 396:
```suggestion
#[derive(Default)]
struct EndorsementDenunciationSerializer {
```

massa-models/src/denunciation.rs
```diff
@@ -358,16 +390,341 @@ pub enum DenunciationError {
...
+        }
+    }
+}
+
+impl Default for EndorsementDenunciationSerializer {
+    fn default() -> Self {
+        Self::new()
+    }
+}
+
```
bilboquet, line 418: use autogenerated default()

massa-models/src/denunciation.rs
```diff
@@ -358,16 +390,341 @@ pub enum DenunciationError {
...
+                    signature_2,
+                }
+            },
+        )
+        .parse(buffer)
+    }
+}
+
+/// Serializer for `Denunciation`
+pub struct DenunciationSerializer {
```
bilboquet, line 618: Same comment as above about using #[devive(Default)];

PR #3721, opened 2023-03-23
End of free gas

aoudiamoncef:
  * [ ] make tests compile
  * [ ] make tests pass 


Leo-Besancon (approved):

massa-node/base_config/gas_costs/wasm_gas_costs.json
```diff
@@ -2,12 +2,12 @@
   "Wasm:Drop": 8,
   "Wasm:GlobalGet": 28,
   "Wasm:GlobalSet": 26,
-  "Wasm:I32Add": 0,
```
AurelienFT, line 5: This is not a problem if they are 0 as the amount of gas for a wasm operator is the mean of the values in this file

AurelienFT:
Already made in main and merged in testnet_21

PR #3725, opened 2023-03-24
Module cache v2

Eitu33:
* [x] try in sandbox / simulation / labnet
* [x] unit tests on the added features
* [x] make previous tests compile & pass


Eitu33:
For anyone reviewing this: there as still some minor work to be discussed & done, it is marked as `IMPORTANT TODO` in the code. Feel free to comment on theses lines if you have something to add on the subject.

massa-module-cache/src/hd_cache.rs
```diff
@@ -0,0 +1,296 @@
...
+        self.db.write(batch).expect(CRUD_ERROR);
+
+        self.entry_count = self.entry_count.saturating_add(1);
+    }
+
+    /// Sets the initialization cost of a given module separately
+    ///
+    /// # Arguments
+    /// * `hash`: hash associated to the module for which we want to set the cost MUST exist else
+    ///   exit with error: i.e. insert has been called before with the same hash and it has not
```
bilboquet, line 114: comment is outdated

massa-module-cache/Cargo.toml
```diff
@@ -0,0 +1,29 @@
...
+massa_serialization = { path = "../massa-serialization" }
+massa-sc-runtime = { git = "https://github.com/massalabs/massa-sc-runtime", branch = "caching-followup", features = [
+    "testing",
+] }
+
+[dev-dependencies]
+tempfile = "3.3"
+
+[features]
+testing = []
```
AurelienFT, line 29: Feature testing for massa-sc-runtime should be included only here and not in the dependencies in normal case

massa-execution-worker/src/execution.rs
```diff
@@ -677,7 +686,7 @@ impl ExecutionState {
             Ok(Response { init_cost, .. }) => {
                 self.module_cache
                     .write()
-                    .save_module(&bytecode, module, init_cost);
+                    .set_init_cost(&bytecode, init_cost);
```
AurelienFT, line 689: Something that I personally don't understand but that may be correct.

You set the `init_cost` only in case of success. In case of the HD cache, the module has been saved above so it's possible that the module stay save but without the init cost as they are saved in two different keys ?

massa-module-cache/src/hd_cache.rs
```diff
@@ -0,0 +1,296 @@
...
+    /// Try to remove as much as self.amount_to_snip entries from the db
+    fn snip(&mut self) {
+        let mut iter = self.db.raw_iterator();
+        let mut batch = WriteBatch::default();
+        let mut snipped_entries_count: usize = 0;
+
+        while snipped_entries_count < self.snip_amount {
+            // generate a random key
+            let mut rbytes = Vec::new();
+            rand::thread_rng().fill_bytes(&mut rbytes);
```
AurelienFT, line 176: I don't really understand this behavior. `rbytes` have no size so you are filling a empty vector it will result to not filling anything ?

massa-module-cache/src/hd_cache.rs
```diff
@@ -0,0 +1,296 @@
...
+            let key = *Hash::compute_from(&rbytes).to_bytes();
+
+            // take the upper existing key
+            iter.seek_for_prev(key);
```
AurelienFT, line 180: If we are sniping for a module key we want to remove the module and the metadata key no ?

massa-ledger-exports/src/ledger_changes.rs
```diff
@@ -526,6 +526,30 @@ impl LedgerChanges {
         self.0.get(addr)
     }
 
+    /// Retrieves all the bytcode updates contained in the current changes
+    pub fn get_bytecode_updates(&self) -> Vec<Bytecode> {
+        let mut v = Vec::new();
+        for (_addr, change) in self.0.iter() {
+            match change {
+                SetUpdateOrDelete::Set(LedgerEntry { bytecode, .. }) => {
+                    // IMPORTANT TODO: investigate on why the len check is needed
```
AurelienFT, line 535: If we create a new address then a `Set` will be created with all fields filled and bytecode will be empty.

massa-module-cache/src/controller.rs
```diff
@@ -0,0 +1,159 @@
...
+                ModuleInfo::Invalid
+            }
+        }
+    }
+
+    /// Save a new or an already existing module in the cache
+    pub fn save_module(&mut self, bytecode: &[u8]) {
+        let hash = Hash::compute_from(bytecode);
+        if let Some(hd_module_info) =
+            self.hd_cache
```
AurelienFT, line 68: What is this check for ? I assume it's for optimizing and don't try to re-add a module that is already saved in one of the caches but if the module already exist in both then in this case it will fall in this `if` and re-save it in `lru_cache`

massa-module-cache/Cargo.toml
Eitu33: Done

massa-module-cache/src/hd_cache.rs
Eitu33: Updated 👍

massa-execution-worker/src/execution.rs
Eitu33: Yes that is the current behaviour, but your question reminded me that because we always compile with `max_block_gas` as input we should be able to know the init cost even after a VM error. This requires multiple changes, I will resolve the other comments.

massa-module-cache/src/hd_cache.rs
Eitu33: Yes that was a mistake I fixed it

massa-module-cache/src/hd_cache.rs
Eitu33: Good catch, it has been corrected

massa-ledger-exports/src/ledger_changes.rs
Eitu33: The comment has been updated

massa-module-cache/src/controller.rs
Eitu33: In a LRU, if you insert an already existing entry it moves it up in the cache

massa-module-cache/src/hd_cache.rs
```diff
@@ -0,0 +1,300 @@
...
+            // check iterator validity
+            if !iter.valid() {
+                continue;
+            }
+
+            // unwrap justified by above conditional statement
+            // read data key first because of seek_for_prev
+            let data_key = iter.key().unwrap();
+            batch.delete(data_key);
+            iter.prev();
```
AurelienFT, line 189: As you already take the first key that match your prefix shouldn't you go to the next to get the metadata key instead of prev one?

Eitu33: I literally put a comment about this bro, but now that I see it maybe using `data` is misleading. We find the key from the randomly generated one using `seek_for_prev` so the first key we'll get will always be the metadata (aka `data_key`) after which we get the `module_key`.

massa-execution-worker/src/execution.rs
Eitu33: This has been implemented now

massa-module-cache/src/hd_cache.rs
AurelienFT: Resolved on the whiteboard for me

massa-module-cache/src/hd_cache.rs
```diff
@@ -0,0 +1,301 @@
...
+const CRUD_ERROR: &str = "critical: rocksdb crud operation failed";
+const DATA_SER_ERROR: &str = "critical: metadata serialization failed";
+const DATA_DESER_ERROR: &str = "critical: metadata deserialization failed";
+const MOD_SER_ERROR: &str = "critical: module serialization failed";
+const MOD_DESER_ERROR: &str = "critical: module deserialization failed";
+const MODULE_IDENT: u8 = 0u8;
+const DATA_IDENT: u8 = 1u8;
+
+/// Module key formatting macro
+#[macro_export]
```
Leo-Besancon, line 21: I think the key formating in the ledger was refactored. Maybe we can unify the keys formatting at some point.

massa-module-cache/src/hd_cache.rs
```diff
@@ -0,0 +1,301 @@
...
+        if self.entry_count >= self.max_entry_count {
+            self.snip();
+        }
+
+        let mut ser_metadata = Vec::new();
+        let ser_module = match module_info {
+            ModuleInfo::Invalid => {
+                self.meta_ser
+                    .serialize(&ModuleMetadata::Invalid, &mut ser_metadata)
+                    .expect(DATA_SER_ERROR);
```
Leo-Besancon, line 85: Should we panic in case of an insert error in the cache? I would have thought it is not an unrecoverable error.

massa-node/base_config/config.toml
```diff
@@ -54,10 +54,16 @@
...
-    # max number of compiled modules in the cache
-    max_module_cache_size = 1000
     # path to the initial vesting file
     initial_vesting_path = "base_config/initial_vesting.json"
+    # path to the hard drive cache storage
+    hd_cache_path = "storage/cache/rocks_db"
+    # maximum number of entries we want to keep in the LRU cache
+    lru_cache_size = 1000
+    # maximum number of entries we want to keep in the HD cache
+    hd_cache_size = 10_000
```
Leo-Besancon, line 64: How where these values set?

Leo-Besancon (commented):
Everything seems good to me, just have a few questions :)

massa-node/base_config/config.toml
Eitu33: These are "doigt mouillé"-computed values, we need to estimate them alongside the costs of ABIs that perform bytecode modifications.

massa-module-cache/src/hd_cache.rs
Eitu33: The only causes of a module metadata serialization failure are:
* Internal programmer mistake
* Corrupted database

These errors were initially recovered but it was changed since they are critical and should never result in anything else than a panic.

massa-module-cache/src/hd_cache.rs
Eitu33: I actually disagree on how the key formatting was done in the ledger, I was not able to review the PR because I was sick at that time. Furthermore we found behavioural issues when going over the testnet 19 to 20 diff changes if I remember correctly. In any case I agree with you that we need it to be unified, but I think we need to define what we'd like to do and why (which was not done for the ledger implementation). I consider this to be out of this PR scope.

massa-module-cache/src/hd_cache.rs
Leo-Besancon: Ok, that makes sense!

massa-module-cache/src/hd_cache.rs
Leo-Besancon: Ok, I was asking because the network restart follow-up will almost certainly need to implement something similar, so I'll be sure to include you in the design discussion ;)

massa-module-cache/src/config.rs
```diff
@@ -0,0 +1,19 @@
...
+    pub hd_cache_path: PathBuf,
+    /// Gas costs used to:
+    /// * setup `massa-sc-runtime` metering on compilation
+    /// * debit compilation costs
+    pub gas_costs: GasCosts,
+    /// Default gas for compilation
+    pub compilation_gas: u64,
+    /// Maximum number of entries we want to keep in the LRU cache
+    pub lru_cache_size: u32,
+    /// Maximum number of entries we want to keep in the HD cache
```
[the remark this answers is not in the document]
Leo-Besancon: > Having an entry limitation rather than a Gb limitation makes the hd cache code significantly more convoluted (especially the pruning function) and I don't find its benefits to make up for that.

Ok, that's what I thought happened.

> Also the Gb limitation can be deduced from the entry count because we define a bytecode maximum size. We could have the config parameter unit to be Gb (or Mb) and compute the number of hd cache entries in a worst case scenario.

Yeah we could do something like that. Personnaly I don't think this workaround is wanted (if most modules are small, it seems weird to say "max cache of 10Gb" while using <1Gb for example, even though it is "to specs").

> Compilation costs for cached modules will be associated to the ABIs that produce a bytecode update. Since the gas cost is payed up front, we do not know the CPU cost of compilation yet, so we would have to calibrate a gas cost for those ABIs that take into consideration all of these issues. This has not been done yet, for the time being we need to agree on an approximation, later it will become necessary to define these costs properly. It was considered to make the user pay post compilation but the solutions make storing a bytecode on the blockchain too complicated for users.

Thank you for the explanation!

massa-execution-worker/src/execution.rs
```diff
@@ -563,28 +568,24 @@ impl ExecutionState {
...
             &*self.execution_interface,
             module,
-            *max_gas,
+            remaining_gas,
             self.config.gas_costs.clone(),
-        ) {
-            Ok(_response) => {}
-            Err(err) => {
-                return Err(ExecutionError::RuntimeError(format!(
-                    "module execution error in execute_executesc_op: {}",
```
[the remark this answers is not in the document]
Eitu33: Added context to the VM error, see https://github.com/massalabs/massa/pull/3725/commits/099880715c76acc546dfcbbc72ca060055395c35

AurelienFT (approved):
Nice work let's see the improvements in number of CallSC on this testnet !

PR #3734, opened 2023-03-28
Initial code for DenunciationFactory & DenunciationPool

sydhds:
  * [ ] make tests compile
  * [ ] make tests pass 


massa-factory-worker/src/denunciation_factory.rs
```diff
@@ -0,0 +1,260 @@
...
+    endorsement_pool_receiver: Receiver<SecureShareEndorsement>,
+
+    // TODO: optim with PreHashSet?
+    // internal for endorsements (or secure share endorsements)
+    /// Internal storage for endorsement -
+    /// store at max 1 endorsement per entry, as soon as we have 2 we produce a Denunciation
+    endorsements_by_slot_index: HashMap<(Slot, u32), Vec<SecureShareEndorsement>>,
+
+    // internal for block header (or secured header)
+    /// Internal storage for endorsement -
```
modship, line 36: same comment as endorsements_by_slot_index

massa-factory-worker/src/denunciation_factory.rs
```diff
@@ -0,0 +1,260 @@
...
+                    let header_1 = secured_headers.get(0).unwrap();
+                    let header_2 = secured_headers.get(1).unwrap();
+                    let de_ = Denunciation::try_from((header_1, header_2));
+                    if let Err(e) = de_ {
+                        debug!("Denunciation factory cannot create denunciation from block headers: {}", e);
+                        return;
+                    }
+                    Some(de_.unwrap()) // Already checked for error
+                } else {
+                    // Already 2 entries - so a Denunciation has already been created
```
modship, line 99: possible reach here if secured_headers.len() == 0 ? (don't know if possible to have len = 0)

massa-factory-worker/src/denunciation_factory.rs
```diff
@@ -0,0 +1,260 @@
...
+                    let header_1 = secure_share_endorsements.get(0).unwrap();
+                    let header_2 = secure_share_endorsements.get(1).unwrap();
+                    let de_ = Denunciation::try_from((header_1, header_2));
+                    if let Err(e) = de_ {
+                        debug!("Denunciation factory cannot create denunciation from block headers: {}", e);
+                        return;
+                    }
+                    Some(de_.unwrap()) // Already checked for error
+                } else {
+                    // Already 2 entries - so a Denunciation has already been created
```
modship, line 166: same note that line 99

massa-factory-worker/src/denunciation_factory.rs
```diff
@@ -0,0 +1,260 @@
...
+                    // TODO / OPTIM?: use [] ?
+                    // safe to unwrap as we just checked the vec len
+                    let header_1 = secure_share_endorsements.get(0).unwrap();
+                    let header_2 = secure_share_endorsements.get(1).unwrap();
+                    let de_ = Denunciation::try_from((header_1, header_2));
+                    if let Err(e) = de_ {
+                        debug!("Denunciation factory cannot create denunciation from block headers: {}", e);
+                        return;
+                    }
+                    Some(de_.unwrap()) // Already checked for error
```
modship, line 164: use match instead of if Err + unwrap() ?

massa-factory-worker/src/denunciation_factory.rs
```diff
@@ -0,0 +1,260 @@
...
+                            warn!("Denunciation factory cannot receive from endorsement pool receiver: {}", e);
+                            break;
+                        }
+                    }
+                }
+                recv(self.factory_receiver) -> msg_ => {
+                    if let Err(e) = msg_ {
+                        warn!("Denunciation factory receive an error from factory receiver: {}", e);
+                    }
+                    // factory_receiver send () and is supposed to be a STOP message
```
modship, line 254: small log when the factory stop ?

massa-models/src/denunciation.rs
```diff
@@ -714,6 +755,93 @@ impl Deserializer<Denunciation> for DenunciationDeserializer {
...
+    }
+}
+
+impl DenunciationId {
+    /// denunciation id to bytes
+    pub fn to_bytes(&self) -> &[u8; DENUNCIATION_ID_SIZE_BYTES] {
+        self.0.to_bytes()
+    }
+
+    /// endorsement id into bytes
```
modship, line 812:
```suggestion
    /// denunciation id into bytes
```

massa-node/src/main.rs
```diff
@@ -505,7 +518,13 @@ async fn launch(
         protocol: ProtocolCommandSender(protocol_command_sender.clone()),
         storage: shared_storage.clone(),
     };
-    let factory_manager = start_factory(factory_config, node_wallet.clone(), factory_channels);
+    let factory_manager = start_factory(
+        factory_config,
+        node_wallet.clone(),
+        factory_channels,
+        denunciation_factory_receiver,
```
modship, line 525: maybe they can be put in FactoryChannels ?

massa-factory-worker/src/denunciation_factory.rs
```diff
@@ -0,0 +1,260 @@
...
+
+    // TODO: optim with PreHashSet?
+    // internal for endorsements (or secure share endorsements)
+    /// Internal storage for endorsement -
+    /// store at max 1 endorsement per entry, as soon as we have 2 we produce a Denunciation
+    endorsements_by_slot_index: HashMap<(Slot, u32), Vec<SecureShareEndorsement>>,
+
+    // internal for block header (or secured header)
+    /// Internal storage for endorsement -
+    /// store at max 1 endorsement per entry, as soon as we have 2 we produce a Denunciation
```
bilboquet, line 37: max => most

massa-factory-worker/src/denunciation_factory.rs
```diff
@@ -0,0 +1,260 @@
...
+    // FIXME: Store per 'Slot' then per 'PubKey'
+    block_header_by_slot: HashMap<Slot, Vec<SecuredHeader>>,
```
bilboquet, line 39: why a Vec<> ? ^^

massa-factory-worker/src/denunciation_factory.rs
```diff
@@ -0,0 +1,260 @@
...
+// TODO: rework these values
+const DENUNCIATION_FACTORY_ENDORSEMENT_CACHE_MAX_LEN: usize = 4096;
+const DENUNCIATION_FACTORY_BLOCK_HEADER_CACHE_MAX_LEN: usize = 4096;
+
+/// Structure gathering all elements needed by the factory thread
+#[allow(dead_code)]
+pub(crate) struct DenunciationFactoryWorker {
+    cfg: FactoryConfig,
+    channels: FactoryChannels,
+    factory_receiver: Receiver<()>,
```
bilboquet, line 25: The type is not very meaningful to me.
Particularly for this one, I'd like a comment.

massa-factory-worker/src/denunciation_factory.rs
```diff
@@ -0,0 +1,260 @@
...
+                self.block_header_by_slot.len()
+            );
+            return;
+        }
+
+        let denunciation_: Option<Denunciation> = match self.block_header_by_slot.entry(key) {
+            Entry::Occupied(mut eo) => {
+                let secured_headers = eo.get_mut();
+                // Store at max 2 SecuredHeader's
+                if secured_headers.len() == 1 {
```
bilboquet, line 85:
```suggestion
                if secured_headers.len() >= 1 {
```?

File: massa-models/src/denunciation.rs [hunk 12]
```diff
@@ -258,6 +265,40 @@ impl Denunciation {
...
+        // we will reject Denunciation older than 3 cycle compared to the current slot
+        let cycle = slot.get_cycle(periods_per_cycle);
+        let next_cycle = slot_at_now.get_cycle(periods_per_cycle);
+
+        if (next_cycle - cycle) > denunciation_expire_cycle_delta {
+            return true;
+        }
+
+        false
+    }
```
bilboquet, line 301: shorter one?
```suggestion
    (next_cycle - cycle) > denunciation_expire_cycle_delta
```

massa-storage/src/lib.rs
```diff
@@ -473,6 +505,63 @@ impl Storage {
         }
         Storage::internal_claim_refs(&ids, &mut owners, &mut self.local_used_endorsements);
     }
+
+    /// Store denunciation
+    pub fn store_denunciation(&mut self, denunciation: Denunciation) {
```
damip, line 510: Is it necessary to have denunciations in the storage ? I think they should just be added to the block header, they are not meant to propagate individually

massa-consensus-worker/src/controller.rs
```diff
@@ -281,6 +281,15 @@ impl ConsensusController for ConsensusControllerImpl {
                 .block_header_sender
                 .send(header.clone().content);
         }
+
+        if let Err(e) = self
+            .channels
+            .denunciation_factory_sender
+            .send(header.clone())
+        {
+            warn!("Cannot send header to denunciation factory: {}", e);
```
damip, line 290: Note that the consensus already has logic to detect double-staking: https://github.com/massalabs/massa/pull/3646/files#diff-08965c1c08d1437142b10be633e041151c0a7d8385e4d05363387c1a169e1204R115-R130

massa-factory-worker/src/denunciation_factory.rs
damip, line 39: same principle as https://github.com/massalabs/massa/pull/3734/files#r1152992033

massa-factory-worker/src/denunciation_factory.rs
```diff
@@ -0,0 +1,260 @@
...
+                };
+                factory.run();
+            })
+            .expect("failed to spawn thread : denunciation-factory")
+    }
+
+    /// Process new secured header (~ block header)
+    fn process_new_secured_header(&mut self, secured_header: SecuredHeader) {
+        let key = secured_header.content.slot;
+
```
damip, line 72: If the public key of the header creator is not checked, someone can spam with headers coming from non-PoS-drawn pubkeys and cause a problem

massa-factory-worker/src/denunciation_factory.rs
```diff
@@ -0,0 +1,260 @@
...
+                    // TODO / OPTIM?: use [] ?
+                    // safe to unwrap as we just checked the vec len
+                    let header_1 = secured_headers.get(0).unwrap();
+                    let header_2 = secured_headers.get(1).unwrap();
+                    let de_ = Denunciation::try_from((header_1, header_2));
+                    if let Err(e) = de_ {
+                        debug!("Denunciation factory cannot create denunciation from block headers: {}", e);
+                        return;
+                    }
+                    Some(de_.unwrap()) // Already checked for error
```
damip, line 97: the unwrap can be avoided using a match on de_

massa-factory-worker/src/denunciation_factory.rs
```diff
@@ -0,0 +1,260 @@
...
+            let mut de_storage = self.channels.storage.clone_without_refs();
+            de_storage.store_denunciation(denunciation);
+            self.channels.pool.add_denunciations(de_storage);
+        }
+
+        self.cleanup_cache();
+    }
+
+    /// Process new secure share endorsement (~ endorsement)
+    fn process_new_secure_share_endorsement(
```
damip, line 124: same remark as https://github.com/massalabs/massa/pull/3734/files#r1153362179

massa-models/src/denunciation.rs
```diff
@@ -714,6 +755,93 @@ impl Deserializer<Denunciation> for DenunciationDeserializer {
...
+/// endorsement id
+#[derive(Clone, Copy, Eq, PartialEq, Ord, PartialOrd, Hash)]
+pub struct DenunciationId(Hash);
+
+const DENUNCIATION_ID_PREFIX: char = 'D';
+const DENUNCIATION_ID_VERSION: u64 = 0;
+
+impl PreHashed for DenunciationId {}
+
+impl Id for DenunciationId {
```
damip, line 772: I think that the whole concept of DenunciationID is not necessary. All denunciations targeting a given slot (or pair slot/index) are identical

massa-factory-worker/src/denunciation_factory.rs
sydhds: done

massa-factory-worker/src/denunciation_factory.rs
sydhds: done

massa-storage/src/lib.rs
sydhds: Remove use of Storage in pool

massa-models/src/denunciation.rs
```diff
@@ -714,6 +755,93 @@ impl Deserializer<Denunciation> for DenunciationDeserializer {
...
+
+impl From<&Denunciation> for DenunciationId {
+    fn from(value: &Denunciation) -> Self {
+        match value {
+            Denunciation::Endorsement(endo_de) => {
+                let mut to_hash = Vec::new();
+                to_hash.extend(endo_de.public_key.to_bytes());
+                to_hash.extend(endo_de.slot.to_bytes_key());
+                to_hash.extend(endo_de.index.to_le_bytes());
+                DenunciationId(Hash::compute_from(&to_hash))
```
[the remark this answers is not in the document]

massa-models/src/denunciation.rs

massa-pool-worker/src/denunciation_pool.rs
```diff
@@ -0,0 +1,141 @@
...
+
+                let now = MassaTime::now().expect("could not get current time");
+                // get closest slot according to the current absolute time
+                let slot_now = get_closest_slot_to_timestamp(
+                    self.config.thread_count,
+                    self.config.t0,
+                    self.config.genesis_timestamp,
+                    now,
+                );
+                if Denunciation::is_expired(
```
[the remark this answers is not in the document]
sydhds: is_expired() is already implemented for Denunciation ;)

massa-models/src/denunciation.rs
```diff
@@ -714,6 +755,93 @@ impl Deserializer<Denunciation> for DenunciationDeserializer {
...
+    pub fn to_bytes(&self) -> &[u8; DENUNCIATION_ID_SIZE_BYTES] {
+        self.0.to_bytes()
+    }
+
+    /// endorsement id into bytes
+    pub fn into_bytes(self) -> [u8; DENUNCIATION_ID_SIZE_BYTES] {
+        self.0.into_bytes()
+    }
+
+    /// endorsement id from bytes
```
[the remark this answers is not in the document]
sydhds: fixed

massa-models/src/denunciation.rs
sydhds: fixed

massa-factory-worker/src/denunciation_factory.rs
sydhds: fixed

massa-factory-worker/src/denunciation_factory.rs
sydhds: It's not even possible because there is a match against Entry::Vacant

massa-pool-worker/src/denunciation_pool.rs
```diff
@@ -118,24 +84,15 @@ impl DenunciationPool {
             now,
         );
 
-        for (de_id, de_slot) in self.denunciations_cache.iter() {
-            if Denunciation::is_expired(
-                de_slot,
+        // remove all denunciations that are expired
```
bilboquet, line 87: Nice change!

massa-node/src/main.rs
sydhds: Good idea. I've added a note to do that in a follow up :)

massa-factory-worker/src/denunciation_factory.rs
sydhds: done

massa-factory-worker/src/denunciation_factory.rs
sydhds: done

massa-factory-worker/src/denunciation_factory.rs
```diff
@@ -0,0 +1,260 @@
...
+    cfg: FactoryConfig,
+    channels: FactoryChannels,
+    factory_receiver: Receiver<()>,
+    consensus_receiver: Receiver<SecuredHeader>,
+    endorsement_pool_receiver: Receiver<SecureShareEndorsement>,
+
+    // TODO: optim with PreHashSet?
+    // internal for endorsements (or secure share endorsements)
+    /// Internal storage for endorsement -
+    /// store at max 1 endorsement per entry, as soon as we have 2 we produce a Denunciation
```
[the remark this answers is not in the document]
sydhds: reworked

massa-factory-worker/src/denunciation_factory.rs
sydhds: reworked

massa-factory-worker/src/denunciation_factory.rs
sydhds: added

massa-factory-worker/src/denunciation_factory.rs
```diff
@@ -0,0 +1,303 @@
...
+
+        if let Some(denunciation) = denunciation_ {
+            info!(
+                "Created a new block header denunciation : {:?}",
+                denunciation
+            );
+
+            self.channels.pool.add_denunciation(denunciation);
+        }
+
```
bilboquet, line 128: if think it can be simplified to this 

```suggestion
match self.block_header_by_slot.entry(key) {
    Entry::Occupied(mut eo) => {
        match eo.get_mut() {
            BlockHeaderDenunciationStatus::Accumulating(header_1_) => {
                let header_1: &SecuredHeader = header_1_;
                match Denunciation::try_from((header_1, &secured_header)) {
                    Ok(de) => {
                        eo.insert(BlockHeaderDenunciationStatus::DenunciationEmitted);
                        info!("Created a new block header denunciation : {:?}", de);
                        self.channels.pool.add_denunciation(de);
                    }
                    Err(e) => {
                        debug!("Denunciation factory cannot create denunciation from endorsements: {}", e);
                    }
                }
            }
            // Denunciation has already been created
            BlockHeaderDenunciationStatus::DenunciationEmitted => {
            }
        }
    }
    Entry::Vacant(ev) => {
        ev.insert(BlockHeaderDenunciationStatus::Accumulating(secured_header));
    }
};

```

massa-factory-worker/src/denunciation_factory.rs
```diff
@@ -0,0 +1,303 @@
...
+        };
+
+        if let Some(denunciation) = denunciation_ {
+            info!(
+                "Created a new endorsement denunciation : {:?}",
+                denunciation
+            );
+            self.channels.pool.add_denunciation(denunciation);
+        }
+
```
bilboquet, line 217: Same simplification should also apply here

```suggestion
match self.endorsements_by_slot_index.entry(key) {
    Entry::Occupied(mut eo) => {
        match eo.get_mut() {
            EndorsementDenunciationStatus::Accumulating(secure_endo_1_) => {
                let secure_endo_1: &SecureShareEndorsement = secure_endo_1_;
                match Denunciation::try_from((secure_endo_1, &secure_share_endorsement)) {
                    Ok(de) => {
                        eo.insert(EndorsementDenunciationStatus::DenunciationEmitted);
                        info!("Created a new endorsement denunciation : {:?}", de);
                        self.channels.pool.add_denunciation(de);
                    }
                    Err(e) => {
                        debug!("Denunciation factory cannot create denunciation from endorsements: {}", e);
                    }
                }
            }
            EndorsementDenunciationStatus::DenunciationEmitted => {}
        }
    }
    Entry::Vacant(ev) => {
        ev.insert(EndorsementDenunciationStatus::Accumulating(
            secure_share_endorsement,
        ));
    }
}

```

massa-factory-worker/src/denunciation_factory.rs
sydhds: Not a big fan to have the process of denunciation creation and the storage all in one. I'd rather keep my way, clearly seeing the steps...

massa-consensus-worker/src/controller.rs
damip: discussed orally

massa-models/src/denunciation.rs
sydhds: It is used in DenunciationPool in order to have a common hashmap key as it covers both use case (e.g. for block header denunciation & endorsement denunciation).

massa-factory-worker/src/denunciation_factory.rs
```diff
@@ -0,0 +1,260 @@
...
+        self.cleanup_cache();
+    }
+
+    fn cleanup_cache(&mut self) {
+        let now = MassaTime::now().expect("could not get current time");
+
+        // get closest slot according to the current absolute time
+        let slot_now = get_closest_slot_to_timestamp(
```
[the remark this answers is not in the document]
sydhds: Fixed

massa-factory-worker/src/denunciation_factory.rs
```diff
@@ -0,0 +1,317 @@
...
+        block_header_denunciation_interest: DenunciationInterest,
+    ) {
+        let de_i_orig = block_header_denunciation_interest.clone();
+        let de_i = match block_header_denunciation_interest {
+            DenunciationInterest::Endorsement(_) => {
+                return;
+            }
+            DenunciationInterest::BlockHeader(de_i) => de_i,
+        };
+
```
bilboquet, line 69: feel free to reject directly

```suggestion
let DenunciationInterest::BlockHeader(de_i) = block_header_denunciation_interest else {
    return;
};

```

massa-storage/src/denunciation_indexes.rs
```diff
@@ -0,0 +1,38 @@
+use massa_models::denunciation::Denunciation;
+use massa_models::{denunciation::DenunciationId, prehash::PreHashMap};
+
```
damip, line 3: this whole file can be deleted

massa-models/src/denunciation.rs
```diff
@@ -714,6 +731,203 @@ impl Deserializer<Denunciation> for DenunciationDeserializer {
...
+    pub slot: Slot,
+    hash: Hash,
+    signature: Signature,
+}
+
+/// Lightweight data for Denunciation creation
+/// (avoid storing heavyweight secured header or secure share endorsement)
+#[allow(missing_docs)]
+#[derive(Debug, Clone)]
+pub enum DenunciationInterest {
```
damip, line 850: "interest" is a weird name. Maybe DenunciationData

massa-models/src/denunciation.rs
```diff
@@ -714,6 +731,203 @@ impl Deserializer<Denunciation> for DenunciationDeserializer {
...
+    Endorsement(EndorsementDenunciationInterest),
+    BlockHeader(BlockHeaderDenunciationInterest),
+}
+
+impl TryFrom<&SecureShareEndorsement> for DenunciationInterest {
+    type Error = DenunciationError;
+
+    fn try_from(value: &SecureShareEndorsement) -> Result<Self, Self::Error> {
+        let hash = EndorsementDenunciation::compute_content_hash(&value.content)?;
```
damip, line 859: It's not necessary to reserialize everything compute the hash, manage potential errors etc...
SecureShare objects store their hash in their ID, just recover it from there like we do for signature verification.

massa-models/src/denunciation.rs
```diff
@@ -714,6 +731,203 @@ impl Deserializer<Denunciation> for DenunciationDeserializer {
...
+            },
+        ))
+    }
+}
+
+impl TryFrom<&SecuredHeader> for DenunciationInterest {
+    type Error = DenunciationError;
+
+    fn try_from(value: &SecuredHeader) -> Result<Self, Self::Error> {
+        let hash = BlockHeaderDenunciation::compute_content_hash(&value.content)?;
```
damip, line 877: same comment as above

massa-factory-worker/src/denunciation_factory.rs
damip: see my comment above to completely avoid this

massa-factory-worker/src/denunciation_factory.rs
```diff
@@ -0,0 +1,317 @@
...
+            self.cfg.genesis_timestamp,
+            now,
+        );
+
+        let cycle_of_header = de_i.slot.get_cycle(self.cfg.periods_per_cycle);
+        let cycle_now = slot_now.get_cycle(self.cfg.periods_per_cycle);
+
+        // Do not fulfill the cache with block header too much in the future
+        // Note that cache is also purged each time a slot becomes final
+        if cycle_now - cycle_of_header > self.cfg.denunciation_items_max_cycle_delta {
```
damip, line 85: Problems:
* this does not do what the comment says
  * the comment says it rejects headers that are too much in the future
  * the code rejects headers that are too far in the past compare to now, which is not what we want
* artithmetic underflow if the header cycle is in the future => very important general rule: always think twice before using unprotected arithmetic operators
* lack of precision (granularity = cycles)

 Solutions:
* let's express all durations in periods instead of cycles
* first check if `de.slot.period < last_final_periods[de.slot.thread].saturating_sub(max_denunciation_periods) ` 
  * if true => ignore because the element is too old and cannot be denounced anymore anyway
* then check if `de.slot.period.saturating_sub(now.slot.period) > periods_per_cycle` 
  * if true => ignore because the element is too much in the future

massa-factory-worker/src/denunciation_factory.rs
```diff
@@ -0,0 +1,317 @@
...
```
damip, line 178: same remarks as for the header

massa-models/src/denunciation.rs
sydhds: This is not the full hash with public key / denunciation data / hash(content) but only this hash(content) so we need to recompute it here...
Otherwise the test that check Denunciation made from 2 block headers versus 2 DenunciationInterest will fail...

massa-models/src/denunciation.rs
sydhds: There is already a DenunciationData struct (e.g. BlockHeaderDenunciationData) so I think it will only confuse devs. DenunciationInterest as it name says only carry the interesting data in order to create a Denunciation so I would say this is a valid name ?

massa-models/src/denunciation.rs
sydhds: See reply above

massa-models/src/denunciation.rs
```diff
@@ -258,6 +246,35 @@ impl Denunciation {
...
+        denunciation_expire_cycle_delta: u64,
+    ) -> bool {
+        // If the Slot is final, a Denunciation can still be made for 1 cycle
+        if slot.period <= last_cs_final_periods[slot.thread as usize] {
+            let cycle_of_slot = slot.get_cycle(periods_per_cycle);
+
+            let slot_final = Slot::new(last_cs_final_periods[slot.thread as usize], slot.thread);
+            let cycle_of_final = slot_final.get_cycle(periods_per_cycle);
+
+            (cycle_of_final - cycle_of_slot) > denunciation_expire_cycle_delta
```
damip, line 273: let's count in periods instead of cycles so this whole function becomes:

```
is_expired(...) -> bool {
   last_cs_final_periods[slot.thread].checked_sub(slot.period) > Some(denunciation_expire_periods)
}

```

massa-factory-worker/src/denunciation_factory.rs
```diff
@@ -0,0 +1,317 @@
...
+                };
+                factory.run();
+            })
+            .expect("failed to spawn thread : denunciation-factory")
+    }
+
+    /// Process new secured header (~ block header)
+    fn process_new_secured_header(
+        &mut self,
+        block_header_denunciation_interest: DenunciationInterest,
```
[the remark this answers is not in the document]
sydhds: I'd rather keep the enum as it is because:
* It is used directly in internal caching
* There is a Denunciation try_from impl for DenunciationInterest and not for each particular variant

massa-models/src/denunciation.rs
```diff
@@ -714,6 +731,203 @@ impl Deserializer<Denunciation> for DenunciationDeserializer {
...
+        }
+    }
+}
+
+// End Denunciation Id
+
+// Denunciation interest
+
+/// DenunciationInterest variant for endorsement
+#[allow(missing_docs)]
```
damip, line 826: we should never use allow missing_docs nor fake empty docs to avoid warnings => it guarantees we will forget to document this :)

massa-models/src/denunciation.rs
damip: Can't we just use the following locally in the pool ?

```rust
#[derive(Debug, Copy, Hash, PartialOrd, Ord, PartialEq, Eq)]
pub struct DenunciationKey {
    Header {
       slot: Slot
    },
    Endorsement {
        slot: Slot,
        index: Index
    }
}
```

that way we don't need to hash, nor to implement serialization, deserialization etc...

massa-factory-worker/src/denunciation_factory.rs
```diff
@@ -0,0 +1,317 @@
+use std::collections::hash_map::Entry;
+use std::collections::HashMap;
+use std::thread;
+
```
damip, line 4: interesting question: could we completely remove de denunciation_factory and only have the denunciation pool in which the state for each DenunciationKey is either `Accumulating(DenunciationData)` or `Denounced(Denunciation)`. The pruning is the same, and when the module is asked for denunciations to include in a block, it only filters out the ones that are ready (Denounced)

massa-models/src/denunciation.rs
sydhds: fixed

massa-models/src/denunciation.rs
damip: naming is hard :) I would vote for DenunciationPrecursor because you need to mix 2 of them to make a Denunciation https://en.wikipedia.org/wiki/Precursor_(chemistry)

massa-models/src/denunciation.rs
damip: kinda hurts to have to reserialize the whole data strucutre here...  maybe we should think of a way to store this hash somewhere ?

massa-factory-worker/src/denunciation_factory.rs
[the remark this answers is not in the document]
damip: ok then let's keep it, it's not very important

massa-models/src/denunciation.rs
sydhds: Nice naming :)

massa-models/src/denunciation.rs
sydhds: I've added a TODO to change this

massa-models/src/denunciation.rs
sydhds: I've added a TODO for this

massa-models/src/denunciation.rs
sydhds: Added a TODO

massa-factory-worker/src/denunciation_factory.rs
sydhds: I'm in favor of merging it as it is and remove the DenunciationFactory when we add the Denunciation in Block headers (next PR) so we have the whole picture in mind

massa-models/src/denunciation.rs
sydhds: Renamed

massa-factory-worker/src/denunciation_factory.rs
sydhds: Fixed

massa-models/src/denunciation.rs
damip: I proposed a correction here: https://github.com/massalabs/massa/pull/3755

massa-models/src/denunciation.rs
damip: Note: this key will also be used in ExecutedDenunciations

massa-models/src/denunciation.rs
damip: ok, is there a followup issue with all the TODOs ? because if we just leave them in the code we might forget them

sydhds: I've crated an issue here: https://github.com/massalabs/massa/issues/3758 :)

damip (approved):

PR #3735, opened 2023-03-29
add max_coins to executesc

damip:
  * [x] make tests compile
  * [x] make tests pass 
* [ ] evaluate impact on tooling and prepare measures to tackle it


massa-proto/proto/massa/api/v1/operation.proto
```diff
@@ -63,10 +63,12 @@ message RollSell {
 message ExecuteSC {
   // Smart contract bytecode.
   bytes data = 1;
+  // The maximum of coins that could be spent by the operation sender
```
aoudiamoncef, line 66: In the mainet, we couldn't do this, we always have to add the end of type to avoid breaking changes.

AurelienFT (approved):
Tested on sandbox works well. Will launch a labnet when merged

massa-proto/proto/massa/api/v1/operation.proto
AurelienFT: Ok but for this month it's still ok to do this ?

aoudiamoncef (approved):
It's good for me. 

We are missing the doc generation, but I'll rebase and do it in the my PR which brings auto update docs

PR #3744, opened 2023-03-30
Add network version stats in MipStoreRaw

sydhds:
  * [ ] make tests compile
  * [ ] make tests pass 


massa-versioning-worker/src/versioning.rs
```diff
@@ -581,24 +618,105 @@ impl MipStoreRaw {
...
+
+    fn update_network_version_stats(
+        &mut self,
+        slot_timestamp: MassaTime,
+        network_versions: Option<(u32, u32)>,
+    ) {
+        if let Some((_current_network_version, announced_network_version)) = network_versions {
+            let removed_version_ = match self.stats.latest_announcements.len() {
+                n if n > self.stats.config.block_count_considered => {
+                    self.stats.latest_announcements.pop_front()
```
Leo-Besancon, line 637: Why have you decided not to use CircularBuffers?

massa-versioning-worker/src/versioning.rs
```diff
@@ -448,31 +446,70 @@ impl MipStore {
...
-#[derive(Debug, Clone, Default, PartialEq)]
-pub struct MipStoreRaw(pub(crate) BTreeMap<MipInfo, MipState>);
+#[derive(Debug, Clone, PartialEq)]
+pub struct MipStoreRaw {
+    pub(crate) store: BTreeMap<MipInfo, MipState>,
+    pub(crate) stats: MipStoreStats,
+}
 
 impl MipStoreRaw {
     /// Update our store with another (usually after a bootstrap where we received another store)
```
Leo-Besancon, line 515: Maybe add a TODO flag to also update the MipStoreStats?
And if it's too big, split it in two methods, to update the store and the stats separately.

Leo-Besancon (commented):
This is just a first quick pass, I'll do another one tomorrow ;).

massa-versioning-worker/src/versioning.rs
sydhds: I've tried but then I cannot use #derive(Clone, PartialEq) and that was problematic in many places :-/

massa-versioning-worker/src/versioning.rs
sydhds: From what I see, the update_with will replace the stats. Not sure if there are checks that can be done?

massa-execution-worker/Cargo.toml
```diff
@@ -34,6 +34,7 @@ massa_ledger_exports = { path = "../massa-ledger-exports" }
 massa_pos_worker = { path = "../massa-pos-worker", optional = true }
 massa_pos_exports = { path = "../massa-pos-exports" }
 massa_final_state = { path = "../massa-final-state" }
+massa_versioning_worker = { path = "../massa-versioning-worker" }
```
Leo-Besancon, line 37: I think it's time to rename this module to massa-versioning, because it's not in it's own thread anymore.

massa-execution-worker/src/execution.rs
```diff
@@ -1079,6 +1088,31 @@ impl ExecutionState {
...
+                } else {
+                    warn!("Unable to get slot timestamp for slot: {} in order to update mip_store stats", slot);
+                }
+            } else {
+                warn!(
+                    "Could not find block id: {} in storage in order to update mip_store stats",
+                    block_id
+                );
+            }
+        }
```
Leo-Besancon, line 1115: As it's not the main goal of `execute_final_slot`, I think it makes sense to put this in it's own method `update_versioning_stats(exec_target)`.

massa-versioning-worker/src/versioning_ser_der.rs
```diff
@@ -589,13 +591,196 @@ impl Deserializer<MipState> for MipStateDeserializer {
...
+
+impl Serializer<MipStoreStats> for MipStoreStatsSerializer {
+    fn serialize(&self, value: &MipStoreStats, buffer: &mut Vec<u8>) -> Result<(), SerializeError> {
+        // config
+        let cfg_1 = u32::try_from(value.config.block_count_considered).map_err(|e| {
+            SerializeError::GeneralError(format!("Could not convert to u32: {}", e))
+        })?;
+        let cfg_2 = u32::try_from(value.config.counters_max).map_err(|e| {
+            SerializeError::GeneralError(format!("Could not convert to u32: {}", e))
+        })?;
```
Leo-Besancon, line 626: Do you think it makes sense to serialize the config values along the struct? Usually in the codebase it's serialized as is and the config is passed in the deserializer args to check the validity of the structure. But maybe here it makes sense.

Leo-Besancon (changes requested):

massa-versioning-worker/src/versioning.rs
Leo-Besancon: Aw to bad!

massa-versioning-worker/src/versioning.rs
Leo-Besancon: Mhh probably yes. For checks I think it depends if the config is included in the stats or not (see my comment below), if it's not we can check that the buffer length and max counter values are consistant with what we receive.

massa-execution-worker/Cargo.toml
sydhds: I'm in favor of doing it in another PR (next one?) what do you think?

Leo-Besancon: Sure, it's certainly not breaking for this one!

massa-execution-worker/src/execution.rs
sydhds: done

massa-versioning-worker/src/versioning_ser_der.rs
sydhds: Removed cfg ser / der :)

Leo-Besancon (approved):
Great!

PR #3748, opened 2023-03-31
Documentation Improvements Massa-Hash

dr-chain:
- Explaining the need for bitwise XOR
 - Explaining the need for Hash comparison
 - Explaining the motivations for choosing base58 encoding
 - Explaining the motivations for choosing blake3 as the hashing algorithm

### DIFF
--- a/massa-hash/src/hash.rs
+++ b/massa-hash/src/hash.rs
@@ -15,6 +15,33 @@ use std::{
 };
 
 /// Hash wrapper, the underlying hash type is `Blake3`
+///
+/// The motivations for selecting Blake3 were-
+/// Speed: Blake3 is significantly faster than other popular hashing algorithms, such as SHA-256 and SHA-3.
+/// This is largely due to its ability to leverage modern CPU architectures and instruction sets, as well as its optimized implementation.
+///
+/// Security: Blake3 is designed to be highly secure and resistant to a wide range of attacks, including collision attacks, 
+/// length-extension attacks, and timing attacks. It also offers better resistance to side-channel attacks than many other hashing algorithms.
+///
+/// Flexibility: Blake3 is highly flexible and can be used in a variety of applications, including as a general-purpose hash function, 
+/// as a key derivation function, and as a message authentication code. It also supports a wide range of input sizes and can produce output
+/// of any desired length.
+
+/// Scalability: Blake3 can efficiently take advantage of multiple cores and SIMD (single instruction, multiple data) instructions, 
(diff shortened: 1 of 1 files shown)


Eitu33 (dismissed):

Eitu33:
Great 👍

massa-hash/src/hash.rs
```diff
@@ -84,6 +133,16 @@ impl Hash {
     /// let hash = Hash::compute_from(&"hello world".as_bytes());
     /// let serialized: String = hash.to_bs58_check();
     /// ```
+    /// All massa addresses, whenever presented as user-readable texts,
```
AurelienFT, line 136: This is not true. Addresses are not only base58 there is prefix also in front of them. I also think that speaking of address in `massa-hash` could lead to confusion as Addresses doesn't exist in this crate.

We could speak only of hashes

massa-hash/src/hash.rs
```diff
@@ -24,6 +51,10 @@ impl PartialOrd for Hash {
     }
 }
 
+/// This function is useful for the BTreeMap where the order of the addresses is to be maintained.
```
Eitu33, line 54: Just a tiny specification that it is not only used for addresses

dr-chain: Fixed

massa-hash/src/hash.rs
dr-chain: Fixed

AurelienFT (dismissed):

massa-hash/src/hash.rs
```diff
@@ -24,6 +51,11 @@ impl PartialOrd for Hash {
     }
 }
 
+/// This function is generally useful for data structures that performs ordering and where hashes are used
+/// as keys. In massa, it is used for the BTreeMap where the order of the addresses is to be maintained.
+/// This function helps to have a single coherent BTreeMap which is then used to perform the draw
+/// See Pos-Worker for more details.
```
Eitu33, line 57: What I was trying to say is that we have other use cases in Massa itself than just addresses, like block ids, operation ids, basically anything that has an underlying hash representation.

dr-chain: Ok, I am gonna edit it accordingly.

Eitu33:
@dr-chain Also as a general rule we wait for the comment emitter to resolve the conversation, to make sure that the person is satisfied with the change / response. This avoids creating new conversations or reopening old ones 👍

dr-chain:
bors merge

PR #3757, opened 2023-04-03
improve error and readability 

modship:
  * [ ] make tests compile
  * [ ] make tests pass 

### DIFF
--- a/massa-bootstrap/src/client.rs
+++ b/massa-bootstrap/src/client.rs
@@ -32,31 +32,13 @@ fn stream_final_state_and_consensus<D: Duplex>(
     global_bootstrap_state: &mut GlobalBootstrapState,
 ) -> Result<(), BootstrapError> {
     if let BootstrapClientMessage::AskBootstrapPart { .. } = &next_bootstrap_message {
-        match client.send_timeout(
+        client.send_timeout(
             next_bootstrap_message,
             Some(cfg.write_timeout.to_duration()),
-        ) {
-            Err(BootstrapError::TimedOut(_)) => Err(std::io::Error::new(
-                std::io::ErrorKind::TimedOut,
-                "bootstrap ask ledger part send timed out",
-            )
-            .into()),
-            Err(e) => Err(e),
-            Ok(_) => Ok(()),
-        }?;
+        )?;
+
         loop {
-            let msg = match client.next_timeout(Some(cfg.read_timeout.to_duration())) {
-                Err(BootstrapError::TimedOut(_)) => {
-                    return Err(std::io::Error::new(
-                        std::io::ErrorKind::TimedOut,
-                        "final state bootstrap read timed out",
-                    )
-                    .into());
-                }
-                Err(e) => return Err(e),
-                Ok(msg) => msg,
-            };
-            match msg {
+            match client.next_timeout(Some(cfg.read_timeout.to_duration()))? {
                 BootstrapServerMessage::BootstrapPart {
                     slot,
                     ledger_part,
@@ -170,14 +152,12 @@ fn stream_final_state_and_consensus<D: Duplex>(
                     return Err(BootstrapError::GeneralError(String::from("Slot too old")));
                 }
                 BootstrapServerMessage::BootstrapError { error } => {
-                    return Err(std::io::Error::new(std::io::ErrorKind::InvalidData, error).into())
+                    return Err(BootstrapError::GeneralError(error))
                 }
                 _ => {
-                    return Err(std::io::Error::new(
-                        std::io::ErrorKind::InvalidData,
-                        "unexpected message",
-                    )
-                    .into())
+                    return Err(BootstrapError::GeneralError(
+                        "unexpected message".to_string(),
+                    ))
                 }
             }
         }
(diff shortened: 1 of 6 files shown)


massa-bootstrap/src/server.rs
```diff
@@ -691,15 +679,7 @@ pub async fn stream_bootstrap_information<D: Duplex>(
...
-        ) {
-            Err(BootstrapError::TimedOut(_)) => Err(std::io::Error::new(
-                std::io::ErrorKind::TimedOut,
-                "bootstrap ask ledger part send timed out",
-            )
-            .into()),
-            Err(e) => Err(e),
-            Ok(_) => Ok(()),
-        }?;
+        )?;
```
Ben-PH, line 682: Really nice, but need we need to preserve the assoscieted error message.

massa-bootstrap/src/client_binder.rs
```diff
@@ -100,7 +100,7 @@ impl<D: Duplex> BootstrapClientBinder<D> {
                 self.remote_pubkey.verify_signature(&msg_hash, &sig)?;
                 let (_, msg) = message_deserializer
                     .deserialize::<DeserializeError>(&sig_msg_bytes[HASH_SIZE_BYTES..])
-                    .map_err(|err| BootstrapError::GeneralError(format!("{}", err)))?;
+                    .map_err(|err| BootstrapError::DeserializeError(format!("{}", err)))?;
```
Ben-PH, line 103: good thinking

Ben-PH, line 103: Nice thinking!

massa-bootstrap/src/server.rs
```diff
@@ -375,7 +373,10 @@ impl<D: Duplex, C: NetworkCommandSenderTrait + Clone> BootstrapServer<'_, D, C>
     /// - 3.a. double check the stop-signal is absent
     /// - 3.b. If present, fall-back to the stop behaviour
     /// - 3.c. If absent, all's clear to rock-n-roll.
-    fn receive_connection(&self, selector: &mut Select) -> Result<Option<BsConn<D>>, String> {
+    fn receive_connection(
+        &self,
+        selector: &mut Select,
+    ) -> Result<Option<BsConn<D>>, BootstrapError> {
```
Ben-PH, line 379: Good catch

massa-bootstrap/src/server.rs
```diff
@@ -603,16 +608,7 @@ pub async fn stream_bootstrap_information<D: Duplex>(
...
-                Err(BootstrapError::TimedOut(_)) => Err(std::io::Error::new(
-                    std::io::ErrorKind::TimedOut,
-                    "SlotTooOld message send timed out",
-                )
-                .into()),
-                Err(e) => Err(e),
-                Ok(_) => Ok(()),
-            }?;
-            return Ok(());
+            return server.send_msg(write_timeout, BootstrapServerMessage::SlotTooOld);
```
Ben-PH, line 611: This would be perfect if we could have something like 

`return server.send_msg(...).update_timeout_err("SlotTooOld message send timed out")`;

(same for other parts where the similar message setting is happening)

massa-bootstrap/src/server_binder.rs
```diff
@@ -148,7 +148,7 @@ impl<D: Duplex> BootstrapServerBinder<D> {
         )
         .map_err(|e| match e {
             BootstrapError::IoError(e) if e.kind() == ErrorKind::WouldBlock => {
-                BootstrapError::IoError(ErrorKind::TimedOut.into())
+                BootstrapError::TimedOut(ErrorKind::TimedOut.into())
```
Ben-PH, line 151: I think we can use the `impl From<std::io::Error> for BootstrapError` here, no?

Ben-PH (commented):
nicely done. We shouldn't discard the error messages. 

it might make sense to for the IoError variant to look like this:
```rust
enum BootstrapError {
    IoError {
        kind: io::ErrorKind,
        msg: &'static str,
    },
    TimedOut(&'static str),
    ...
}
````

That would make things cleaner in some ways. The io::Error interface, when it comes to messages, is probably high performance, but it's somewhat complicated.

Edit: happy with the fix from #3763 ... better the sort of thing i had in mind.

File: massa-bootstrap/src/server_binder.rs (hunk 5 above)
Comment (modship) [reply massa-bootstrap/src/server_binder.rs:151]: hm don't think , `e` is now `BootstrapError` and not `IoError`

File: massa-bootstrap/src/server.rs (hunk 4 above)
Comment (modship) [reply massa-bootstrap/src/server.rs:611]: close with https://github.com/massalabs/massa/pull/3763

File: massa-bootstrap/src/client.rs [hunk 6]
```diff
@@ -170,14 +152,12 @@ fn stream_final_state_and_consensus<D: Duplex>(
                     return Err(BootstrapError::GeneralError(String::from("Slot too old")));
                 }
                 BootstrapServerMessage::BootstrapError { error } => {
-                    return Err(std::io::Error::new(std::io::ErrorKind::InvalidData, error).into())
+                    return Err(BootstrapError::GeneralError(error))
```
Ben-PH, line 155: Could you please double check this change. At face value, it doesn't appear correct.

Ben-PH (commented):
Once that one comment is clarified, should be good to go.

massa-bootstrap/src/client.rs
Ben-PH: In discord discussion, I can now see this is the case:

- it successfully gets the next message in the pipeline from the server
- the message is an error message
- it takes the form of a string
- we need to return that string as an error message

...makes sense to me to make it a ::GeneralError now.

...maybe add a comment to this matching to that effect?

massa-bootstrap/src/client.rs
```diff
@@ -170,14 +152,12 @@ fn stream_final_state_and_consensus<D: Duplex>(
                     return Err(BootstrapError::GeneralError(String::from("Slot too old")));
                 }
                 BootstrapServerMessage::BootstrapError { error } => {
```
Ben-PH, line 154:
```suggestion
                // At this point, we have succesfully received the next message from the server, and it's an error-message String
                BootstrapServerMessage::BootstrapError { error } => {
```

PR #3764, opened 2023-04-04
Disable tonic build by default

aoudiamoncef:
  * [x] make tests compile
  * [x] make tests pass 


AurelienFT (approved):

aoudiamoncef:
@AurelienFT I've added a small fix which was detected by @sydhds on Linux.

PR #3765, opened 2023-04-04
Add mip store bootstrap

sydhds:


massa-proto/build.rs
```diff
@@ -7,6 +7,9 @@ fn main() -> Result<(), Box<dyn std::error::Error>> {
     //TODO add download external protos files instead of doing it manually
     let protos = find_protos("proto/massa/");
 
+    let mut prost_config = prost_build::Config::default();
+    prost_config.protoc_arg("--experimental_allow_proto3_optional");
```
Leo-Besancon, line 11: Are these changes still necessary with https://github.com/massalabs/massa/pull/3764/files merged?

massa-node/src/main.rs
```diff
@@ -339,7 +341,13 @@ async fn launch(
         block_count_considered: MIP_STORE_STATS_BLOCK_CONSIDERED,
         counters_max: MIP_STORE_STATS_COUNTERS_MAX,
     };
-    let mip_store = MipStore::try_from(([], mip_stats_config)).unwrap();
+    let mut mip_store =
+        MipStore::try_from(([], mip_stats_config)).expect("Cannot create an empty MIP store");
```
Leo-Besancon, line 345: I'm not sure I remember the design choice here. When we want to add a MIP, the devs would just include it in the store init here?
And after that the update_with below will check compatibilities between our MIP store and the MIP store of our bootstrap server?

Leo-Besancon (commented):
It seems clean to me👍. I think it would be worth it to have a second opinion (maybe @Ben-PH?) on the bootstrap module additions, in case I missed something there!

Ben-PH (approved):
This shouldn't cause any issues with the bootstrap system from what I can tell.

massa-node/src/main.rs
Leo-Besancon: Ok!

Leo-Besancon (approved):
Nice, I think Versioning is coming together nicely!

Issue #3709, opened 2023-03-21
Block operation checks


damip:
The following points need attention:
* [x] let's set back max_operations_per_block to 5000 instead of 10000
* [x] ensure that the block size limit is respected at block creation
* [x] ensure that the block size limit is checked at block retrieval
* [x] ensure that the block operation count limit is respected at block creation
* [x] ensure that the block operation count limit is checked at block retrieval
* [x] ensure that blocks fit within the constaints of message sizes (for example MAX_MESSAGE_SIZE)
* [x] ensure that the retrieval of block operations fits within the constaints of message sizes (for example MAX_MESSAGE_SIZE or  MAX_OPERATIONS_PER_MESSAGE)
* [x] re-read block (de-)serialization and make sure it is consistent
* [x] re-read operation (de-)serialization and make sure it is consistent
* [x] re-read block retrieval messages (de-)serialization and make sure it is consistent

sydhds:
Related PR : https://github.com/massalabs/massa/pull/3710

sydhds:
block operation count check: done

AurelienFT:
@sydhds Can you tick the things that has been done so that we know if we can closed or not ?

sydhds:
> @sydhds Can you tick the things that has been done so that we know if we can closed or not ?

Done. Do you want me to check the others tasks and make a PR if needed? @AurelienFT

sydhds:
"ensure that the block size limit is respected at block creation"

From what I see, in the block factory, the max_block_size is used when retrieving operation but we've already added some endorsements without checking the size of them. Need a fix here

sydhds:
Need to clarify the diff and rename between:
* MAX_OPERATIONS_PER_BLOCK (set to 5000) -> used in message deser for ?
* MAX_OPERATIONS_PER_MESSAGE (set to 1024) -> used in message deser for operation_replies

AurelienFT:
`MAX_OPERATIONS_PER_MESSAGE` should be used for naked operation propagation system

sydhds:
> `MAX_OPERATIONS_PER_MESSAGE` should be used for naked operation propagation system

Yep but I think the naming is misleading, I'll try to propose a better and more explicit name for it :)

sydhds:
block_header serialize does not limit the number of endorsements with a constant but with u32::max, maybe we should pass a limit to the serializer here?

sydhds:
[...]

PR #3769, opened 2023-04-05
Rename OperationsSerializer to OperationRepliesDeserializer

sydhds:
  * [ ] make tests compile
  * [ ] make tests pass 

### DIFF
--- a/massa-models/src/config/constants.rs
+++ b/massa-models/src/config/constants.rs
@@ -27,6 +27,8 @@ pub const MAX_MESSAGE_SIZE: u32 = 1048576000;
 /// Max number of hash in the message `AskForBlocks`
 pub const MAX_ASK_BLOCKS_PER_MESSAGE: u32 = 128;
 /// Max number of operations per message
+/// This is the max number for Operation reply (e.g. for Operation propagation in network)
+/// For max number of operations that can be included in block -> MAX_OPERATIONS_PER_BLOCK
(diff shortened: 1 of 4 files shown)


AurelienFT:
Maybe what I explained [here](https://github.com/massalabs/massa/issues/3709#issuecomment-1495621310) was badly said. I wanted to say that the operation deserializer is always a deserializer and his name should remain generic as it could be used in multiple contexts. It's also used in the operations from block, in both cases it's just the limit of the length that change so the limit of the length should be name `max_length` for example and don't mention `message` in it.

sydhds:
> Maybe what I explained [here](https://github.com/massalabs/massa/issues/3709#issuecomment-1495621310) was badly said. I wanted to say that the operation deserializer is always a deserializer and his name should remain generic as it could be used in multiple contexts. It's also used in the operations from block, in both cases it's just the limit of the length that change so the limit of the length should be name `max_length` for example and don't mention `message` in it.

OperationsSerializer (which has been renamed to OperationRepliesSerializer) is only used in Message so I think it make sense to rename it to a more explicit name. Note this must not be confused with OperationSerializer which is indeed generic.

AurelienFT:
>  is only used in Message 

Yes but in two differents messages one for received naked ops and one in block. And in the future this serializer could be used in other module as it just serialize a bunch of ops I don't see why it couldn't as generic as `OperationSerializer` ?

PR #3774, opened 2023-04-05
Bootstrap server binding read-message in a single `read_exact`

Ben-PH:
This allows the bootstrap server to perform a complete read in one go, without incremental consumption of the stream (...internally, read_exact is an incremental read, but sorting that out is another issue. This statement assumes read_exact functions non-incrementally)

This allows state-changes in the server to be complete: The server consumes a message completely, or not at all.

  * [x] make tests compile
  * [x] make tests pass 


AurelienFT (commented):
I don't really understand why you changed one read_exact with a peek and not the others to read bigger after. Even with the comment this change is unclear to me

EDIT: Ok I see, it's to say atomic on the read

AurelienFT (approved):

PR #3775, opened 2023-04-05
Try Fix  Network restart at cycle 2

Leo-Besancon:
  * [ ] make tests compile
  * [ ] make tests pass 


Leo-Besancon:
Use the following PR instead of this one: https://github.com/massalabs/massa/pull/3777

PR #3776, opened 2023-04-05
Migrate to latest version of `jsonrpsee`

### LINKED ISSUE #3759: Testnet 21 not exiting on SIGINT - API engine suspected
**Describe the bug**
On testnet 21, SIGINT (ctrl-c) handling causes a deadlock/blockage while stopping the various processes

**To Reproduce**

deploy labnet on testnet 21 (haven't checked with sandbox) 
run a node that bootstraps off a labnet server
once the client is up and running, provide a SIGINT (ctrl-c works well)
observe that "Stopping consensus worker..." is logged, with no follow-up

**Expected behavior**
The consensus worker exits correctly, followed by clean exit out of the program shortly after

**Additional context**
with these implementations on `ConsensusControllerImpl` in `massa-consensus-worker/src/controller.rs`

```rust
impl Clone for ConsensusControllerImpl {
    fn clone(&self) -> Self {
        warn!("cloning {}", self.name);
        Self {
            command_sender: self.command_sender.clone(),
            channels: self.channels.clone(),
            shared_state: self.shared_state.clone(),
            bootstrap_part_size: self.bootstrap_part_size.clone(),
            broadcast_enabled: self.broadcast_enabled.clone(),
            name: self.name.clone(),
        }
    }
}
impl Drop for ConsensusControllerImpl {
    fn drop(&mut self) {
        warn!("dropping ctrler imple: {}", self.name);
    }
}
```

as well as changes such as these in `massa-node/src/main.rs`
```diff
+    consensus_controller.set_name("base for protocol_controller");
+    let mut cloned_1 = consensus_controller.clone();
+    cloned_1.set_name("protocol_controller");
+    consensus_controller.set_name("base-again 1");
 
     let protocol_manager = start_protocol_controller(
         protocol_config,
         protocol_receivers,
         protocol_senders.clone(),
-        consensus_controller.clone(),
+        cloned_1,
````

a cleaned up version of the logs will look like this:

```
WARN massa_consensus_worker::controller: cloning base for protocol_controller    
WARN massa_consensus_worker::controller: cloning base for factory_channel    
WARN massa_consensus_worker::controller: cloning factory_channels    
WARN massa_consensus_worker::controller: cloning base for bootstrap_server    
WARN massa_consensus_worker::controller: cloning base for massa-api    

WARN massa_node: sending signal to stop
WARN massa_node: signal sent
INFO massa_node: interrupt signal received

INFO massa_api: API finished cleanly
INFO massa_api: API finished cleanly
INFO massa_api: API finished cleanly

INFO massa_factory_worker::manager: stopping factory...
WARN massa_consensus_worker::controller: dropping ctrler imple: factory_channels    
WARN massa_consensus_worker::controller: dropping ctrler imple: factory_channels    
INFO massa_factory_worker::manager: factory stopped

INFO massa_protocol_exports::protocol_controller: stopping protocol controller...
WARN massa_consensus_worker::controller: dropping ctrler imple: protocol_controller    
INFO massa_protocol_worker::protocol_worker: protocol worker finished cleanly
INFO massa_protocol_exports::protocol_controller: protocol controller stopped

INFO massa_pool_worker::controller_impl: stopping pool workers...
INFO massa_pool_worker::controller_impl: pool workers stopped

INFO massa_bootstrap::server: ip white/blacklist updater exited cleanly

WARN massa_consensus_worker::controller: dropping ctrler imple: bootstrap_server  
```

Note: there are 5 clones of the consensus-controller, so we would expect 6 drops (one for each clone, plus the original), 

BUT

I count only 4 drops. It looks like the instances that are provided to the public api and massa api are not dropped.

This is important, because the consensus manager stops after `join`ing on the thread that's spawned to host the `ConsensusWorker::run()` execution. The content of which exits its main-loop by way of detecting that the channel it receives on is disconnected.

The disconnection will only occur once all the consensus controllers that are cloned from that same base-controller are dropped, as each clone will also clone the sender side of that channel.

### PR DESCRIPTION
**Pull request by aoudiamoncef**
  * [x] make tests compile
  * [x] make tests pass 

fixes #3759

### CODE REVIEW

**Comment by aoudiamoncef** [comment]
@AurelienFT wi this upgrade, we could revert consensus stop hack

File: massa-node/src/main.rs [hunk 1]
```diff
@@ -902,20 +902,25 @@ async fn stop(
...
 
-    // stop public API
-    api_public_handle.stop();
-
-    // stop private API
-    api_private_handle.stop();
-
-    // stop Massa API
-    api_handle.stop();
+    info!("Start stopping API's: gRPC, EXPERIMENTAL, PUBLIC, PRIVATE");
```
modship, line 905: You can also do this for stop all API in same time : 
```rust
    info!("Start stopping API's: gRPC, EXPERIMENTAL, PUBLIC, PRIVATE");

    // stop Massa gRPC API
    if let Some(handle) = grpc_handle {
        handle.stop();
    }

    join!(
        api_handle.stop(),
        api_public_handle.stop(),
        api_private_handle.stop(),
    )
    .await;

    info!("Stopped API's : gRPC, EXPERIMENTAL, PUBLIC, PRIVATE");

```

aoudiamoncef: Why not but we are trying to spot where we have an issue, which make sequential execution better

modship (approved):

massa-node/src/main.rs
modship: in fact we should not have bug here :grimacing: . We can separate if bug appears. 

Or keep it like that the optimisation is not significant

AurelienFT (approved):

massa-node/src/main.rs
aoudiamoncef: Even in the info message, there is an order

PR #3777, opened 2023-04-06
Test fix net restart

Leo-Besancon:
  * [ ] make tests compile
  * [ ] make tests pass 


AurelienFT (approved):
I approve let's wait the labnet test before merge

PR #3778, opened 2023-04-06
Remove un-necessary serialization of message ID in messages

litchipi:
Apparently the serialization of the message ID was left over, making the tests fail.

Test are failling as usual now (never succeeds to join)


AurelienFT (approved):
Nice catch

PR #3786, opened 2023-04-06
Refactor serializing of message

AurelienFT:
- Fix the peer management message serialization to match the others on Massa

- Fix the ID serialization to have one id for all messages of all handlers but then be able to split them in all the handlers after.

Peernet modifications : https://github.com/massalabs/PeerNet/pull/25

### DIFF
--- a/massa-protocol-worker-2/src/messages.rs
+++ b/massa-protocol-worker-2/src/messages.rs
@@ -2,7 +2,6 @@ use crossbeam::channel::Sender;
 use massa_serialization::{
     DeserializeError, Deserializer, Serializer, U64VarIntDeserializer, U64VarIntSerializer,
 };
-use num_enum::{IntoPrimitive, TryFromPrimitive};
 use peernet::{
     error::{PeerNetError, PeerNetResult},
     messages::{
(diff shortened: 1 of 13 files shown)


massa-protocol-worker-2/src/messages.rs
```diff
@@ -1,11 +1,231 @@
...
+use crate::handlers::{peer_handler::{PeerManagementMessage, PeerManagementMessageSerializer}, block_handler::{BlockMessage, BlockMessageSerializer}, endorsement_handler::{EndorsementMessage, EndorsementMessageSerializer}, operation_handler::{OperationMessage, OperationMessageSerializer}};
+
+pub enum Message {
+    BlockMessage(BlockMessage),
+    EndorsementMessage(EndorsementMessage),
+    OperationMessage(OperationMessage),
+    PeerManagementMessage(PeerManagementMessage),
+}
+
+impl From<BlockMessage> for Message {
```
litchipi, line 19: Maybe it would be nice to create some kind of macro to generate all these boring code ?

massa-protocol-worker-2/src/messages.rs
```diff
@@ -1,11 +1,231 @@
...
+                        PeerNetError::HandlerError.error(
+                            "MessagesSerializer",
+                            Some(format!("Failed to serialize message id: {}", err)),
+                        )
+                    })
+            },
+            Message::OperationMessage(message) => {
+                self.id_serializer
+                    .serialize(
+                        &(MessageCategoryId::OperationHandler as u64 + message.get_id() as u64),
```
litchipi, line 137: I don't really like that implementation, it's very error prone.
If we add some more messages to `EndorsementMessage`, then the IDs could overlap..
A better way here would be to convert `MessageCategoryId` to `u32`, and each message type to `u32`, and create a `let id: u64 = ((msg_cat_id << 32) + msg_id) as u64;` or something of the sort.
We will never realistically get more than `2**32` messages types / categories, and it will prevent any overlap (that may be easy to provoke, and hard to debug)

massa-protocol-worker-2/src/messages.rs
AurelienFT: Added a comment

PR #3796, opened 2023-04-07
Remove the dependency on tokio in the bootstrap module

Ben-PH:
  * [x] make tests compile
  * [x] make tests pass 


massa-network-exports/src/network_controller.rs
```diff
@@ -450,3 +450,13 @@ impl NetworkManager {
         Ok(())
     }
 }
+
+/// Used by the bootstrap server to run async tasks, allowing the bootstrap module to
+/// remove the tokio dependency.
+pub fn make_runtime() -> tokio::runtime::Runtime {
```
modship, line 456: ** maybe we can remove the tokio rt by using async-std :: block_on

massa-bootstrap/src/server.rs
```diff
@@ -746,7 +729,8 @@ async fn manage_bootstrap<C: NetworkCommandSenderTrait>(
                     server.send_msg(
                         write_timeout,
                         BootstrapServerMessage::BootstrapPeers {
-                            peers: network_command_sender.get_bootstrap_peers().await?,
+                            peers: rt_hack
+                                .block_on(network_command_sender.get_bootstrap_peers())?,
```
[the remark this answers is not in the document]
Ben-PH: the idea is more about removing any sort of dependency on a runtime. This is still async because of the requirements of the network command system. here, we are depending on the network command system, not an async runtime (well, we still do, but only via the network command system we already depend on)

changing one runtime for another would defeat the purpose.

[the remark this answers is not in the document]
modship: yeah i know, i suggest that for removing tokio dependency

massa-network-exports/src/network_controller.rs
Ben-PH: that would require adding an async-std dependency.

Once `get_bootstrap_peers` is not async, that would allow this to be removed as well.

modship: ok :)

modship (approved):

PR #3801, opened 2023-04-11
Bootstrap/miod io3

Ben-PH:
  * [ ] make tests compile
  * [ ] make tests pass 


massa-bootstrap/src/establisher.rs
```diff
@@ -1,14 +1,18 @@
 // Copyright (c) 2022 MASSA LABS <info@massa.net>
 use massa_time::MassaTime;
+use mio::{
+    net::{TcpListener, TcpStream as MioTcpStream},
```
modship, line 4: in my branch we deal with std::net::TcpStream instead of mio::net::TcpStream

massa-bootstrap/src/server_binder.rs
```diff
@@ -270,6 +290,102 @@ impl BootstrapServerBinder {
...
+                    _ => BootstrapError::IoError(e),
+                })?;
+            if self.events.is_empty() {
+                return Err(BootstrapError::TimedOut(std::io::Error::new(
+                    ErrorKind::TimedOut,
+                    "timed out",
+                )));
+            }
+            for event in self.events.iter() {
+                if event.token() == BINDING_EVENT {
```
modship, line 364: here, when you listen for your stream, you can receive `BINDING_EVENT` for other TcpStream ?

massa-bootstrap/src/establisher.rs
```diff
@@ -44,19 +54,48 @@ impl DefaultListener {
 
         // Number of connections to queue, set to the hardcoded value used by tokio
         socket.listen(1024)?;
-        socket.set_nonblocking(false)?;
-        Ok(DefaultListener(socket.into()))
+        socket.set_nonblocking(true)?;
+        let std_sock = socket.into();
+        let mut mio_listener = mio::net::TcpListener::from_std(std_sock);
```
modship, line 59: `mio::net::TcpListener::from_std` is needed to have `std::net::TcpStream`, if not used we should use `mio::net::TcpListener::bind(...)`

massa-bootstrap/src/server_binder.rs
Ben-PH: As it is, the code will only listen for an event from the tcp stream assosciated with this particular server binding.

massa-bootstrap/src/establisher.rs
Ben-PH: Agreed. If we can go straight to a mio stream, instead of using `from_std`, that would be ideal. I'm just not so sure how to do it, given the issues described in the comment on line 40-42.

AurelienFT:
Can we close this ?

Ben-PH:
closed: work was continued in another branch, PR made.

PR #3811, opened 2023-04-11
Encapsulate the clients bootstrap steps in the type system

Ben-PH:
  * [ ] make tests compile
  * [ ] make tests pass 


massa-bootstrap/src/client_binder.rs
```diff
@@ -23,51 +25,99 @@ pub struct BootstrapClientBinder {
...
+    /// NOT cancel-safe, so does not return the binder if cancelled
+    pub fn handshake(
+        mut self,
+        version: Version,
+        duration: Duration,
+    ) -> Result<(BootstrapClientBinder, MassaTime), BootstrapError> {
+        // read error (if sent by the server)
+        self = self.check_no_error(duration)?;
+
+        // handshake
```
Ben-PH, line 62: not 100% sure about whether I've maintained the ping-calculation behavior

Ben-PH:
The basic principal here is that there is a state change between "connected, handshake incomplete" and "handshake complete".

I see this PR as a spike, as there are a lot more state-changes to consider. Furthering this spike will be easier once `bootstrap/tokio-removal` is merged in

modship:
> The basic principal here is that there is a state change between "connected, handshake incomplete" and "handshake complete".
> 
> I see this PR as a spike, as there are a lot more state-changes to consider. Furthering this spike will be easier once `bootstrap/tokio-removal` is merged in

We are currently implementing similar things with the network refactor. Don't know if you have to implement it on your side

massa-bootstrap/src/client.rs
```diff
@@ -186,79 +186,77 @@ fn stream_final_state_and_consensus(
...
-            return Err(BootstrapError::ReceivedError(error))
+            return Err((Some(client), BootstrapError::ReceivedError(error)))
         }
-        Ok(msg) => return Err(BootstrapError::UnexpectedServerMessage(msg)),
+        Ok(msg) => return Err((Some(client), BootstrapError::UnexpectedServerMessage(msg))),
     };
 
     // get the time of reception
-    let recv_time = MassaTime::now()?;
+    let recv_time = match MassaTime::now() {
```
modship, line 236: you can use 
```rust 
let recv_time = MassaTime::now()?; 
// or
let recv_time = MassaTime::now().map_err(...)?;
```

massa-bootstrap/src/client_binder.rs
```diff
@@ -23,51 +25,99 @@ pub struct BootstrapClientBinder {
     size_field_len: usize,
     remote_pubkey: PublicKey,
     duplex: TcpStream,
-    prev_message: Option<Hash>,
+    prev_message: Hash,
+    cfg: BootstrapClientConfig,
+}
+pub struct NonHSClientBinder {
```
modship, line 31: what does "NonHS" mean ?

massa-bootstrap/src/client.rs
Ben-PH: I tried it that way, but the borrow checker sees that as a move (I think it doesn't recognize the `?` that comes immediately after).

...I don't like it either. I was going to put a comment there, but that would imply needing it everywhere, which as I'm sure you noticed, would be... repetitive :(

Ben-PH:
> > The basic principal here is that there is a state change between "connected, handshake incomplete" and "handshake complete".
> > I see this PR as a spike, as there are a lot more state-changes to consider. Furthering this spike will be easier once `bootstrap/tokio-removal` is merged in
> 
> We are currently implementing similar things with the network refactor. Don't know if you have to implement it on your side

Probably not a good idea to concern ourselves too much with the gritty internals (such as with this PR) without coordinating.

massa-bootstrap/src/client.rs
modship: that's strange but ok it's not a big change

PR #3841, opened 2023-04-19
Network refactoring tester

modship:
  * [ ] make tests compile
  * [ ] make tests pass 

### DIFF
--- a/massa-node/base_config/config.toml
+++ b/massa-node/base_config/config.toml
@@ -229,6 +229,8 @@
     max_bytes_read = 20_000_000.0
     # write limitation for a connection in bytes per seconds
     max_bytes_write = 20_000_000.0
+    # number of thread tester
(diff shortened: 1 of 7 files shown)


massa-node/base_config/config.toml
```diff
@@ -175,6 +175,8 @@
     max_operations_propagation_time = 32000
     # time threshold after which endorsement are not propagated
     max_endorsements_propagation_time = 48000
+    # number of thread tester
+    thread_tester_count = 5
```
modship, line 179: maybe default thread count should be less than 5

AurelienFT: I find 5 be a good number

AurelienFT (approved):

AurelienFT:
Let's merge it as a first version to unlock the rest

PR #3858, opened 2023-04-24
Add stop command to all thread network refactoring

AurelienFT:


massa-protocol-worker-2/src/handlers/endorsement_handler/commands_propagation.rs
```diff
@@ -1,6 +1,7 @@
 use massa_storage::Storage;
 
-pub enum EndorsementHandlerCommand {
+pub enum EndorsementHandlerPropagationCommand {
```
modship, line 3: Think the previous name is better with "stop"

```suggestion
pub enum EndorsementHandlerCommand {
```

modship: Ow ok , because you have add : EndorsementHandlerRetrievalCommand

modship (approved):

massa-protocol-worker-2/src/handlers/endorsement_handler/commands_propagation.rs
AurelienFT: Yes I specified everywhere now that we have command in both threads

PR #3871, opened 2023-04-26
Add two extra saved cycles to PoS for denunciations checks

Eitu33:
No further changes need to be done than saving 2 extra cycles because:
1. We feed every cycle we get after bootstrapping in [compute_initial_draws](https://github.com/massalabs/massa/blob/testnet_22/massa-pos-exports/src/pos_final_state.rs#L195-L210)
2. Fed cycles are computed in the same order they are fed and the selector thread checks sequentiality
3. We wait for every cycle to be drawn up to C+2 (C being the latest cycle received from bootstrap) [here](https://github.com/massalabs/massa/blob/testnet_22/massa-pos-exports/src/pos_final_state.rs#L213-L215), and we now receive 6 cycles after this change, giving us the draws for C-2, C-1, C, C+1, C+2

image


damip (approved):
LGTM

PR #3872, opened 2023-04-26
Check draws before executing a denunciation

sydhds:
  * [ ] make tests compile
  * [ ] make tests pass 

### DIFF
--- a/massa-execution-worker/src/execution.rs
+++ b/massa-execution-worker/src/execution.rs
@@ -466,44 +466,33 @@ impl ExecutionState {
         match &denunciation {
             Denunciation::Endorsement(_de) => {
                 // Get selected address from selector and check
-                let selected = self.selector.get_selection(*de_slot);
-                match selected {
-                    Ok(selection) => {
-                        if let Some(address) = selection
-                            .endorsements
-                            .get(*denunciation.get_index().unwrap_or(&0) as usize)
-                        {
-                            if *address != addr_denounced {
-                                return Err(ExecutionError::IncludeDenunciationError(
-                                    "Attempt to execute a denunciation but address was not selected".to_string()
-                                ));
-                            }
-                        } else {
-                            // Panic here as this should never happen
-                            panic!("Could not get selected address for endorsements at index");
-                        }
-                    }
-                    Err(e) => {
-                        // We panic here as this should never happen
-                        panic!("Cannot get producer from selector: {}", e);
-                    }
+                let selection = self
+                    .selector
+                    .get_selection(*de_slot)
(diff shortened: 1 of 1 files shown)


massa-execution-worker/src/execution.rs
```diff
@@ -445,6 +449,57 @@ impl ExecutionState {
...
+                            .get(*denunciation.get_index().unwrap_or(&0) as usize)
+                        {
+                            if *address != addr_denounced {
+                                return Err(ExecutionError::IncludeDenunciationError(
+                                    "Attempt to execute a denunciation but address was not selected".to_string()
+                                ));
+                            }
+                        } else {
+                            // Panic here as this should never happen
+                            panic!("Could not get selected address for endorsements at index");
```
damip, line 475: if we panic here, we need to make sure that denunciation deserialization fails if index >= ENDORSEMENT_COUNT, otherwise it's an attack vector

massa-execution-worker/src/execution.rs
```diff
@@ -445,6 +449,57 @@ impl ExecutionState {
...
+                        // We panic here as this should never happen
+                        panic!("Cannot get producer from selector: {}", e);
+                    }
+                }
+            }
+        }
+
         // ignore the denunciation if it was already executed
         let de_idx = DenunciationIndex::from(denunciation);
         if context.is_denunciation_processed(&de_idx) {
```
damip, line 505: * this can be moved before the PoS check
* processed => executed

massa-execution-worker/src/execution.rs
damip: Note: the code can be simplified:

```
let selected_addr = self
    .selector
    .get_selection(*de_slot)
    .expect("could not get selection blablabla")
    .endorsements
    .get(_de.index)
    .expect("could not get selection at index blablabla");  // TODO ensure that the denunciationdeserializer prevents this from happening 
```

sydhds: I've checked it and the deserializer is bound by ENDORSEMENT_COUNT

massa-execution-worker/src/execution.rs
sydhds: Done

massa-execution-worker/src/execution.rs
sydhds: Code simplified

massa-execution-worker/src/execution.rs
damip: Will be done il a followup

damip (approved):
LGTM

PR #3880, opened 2023-04-28
Add checks for denunciations + last_start_period

sydhds:
  * [ ] make tests compile
  * [ ] make tests pass 

### DIFF
--- a/massa-execution-worker/src/execution.rs
+++ b/massa-execution-worker/src/execution.rs
@@ -421,13 +421,12 @@ impl ExecutionState {
 
         let de_slot = denunciation.get_slot();
 
-        let last_start_period = self.final_state.read().last_start_period;
-        if de_slot.period <= last_start_period {
+        if de_slot.period <= self.config.last_start_period {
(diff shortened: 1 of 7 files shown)


massa-execution-worker/src/execution.rs
```diff
@@ -419,6 +419,18 @@ impl ExecutionState {
         // acquire write access to the context
         let mut context = context_guard!(self);
 
+        let de_slot = denunciation.get_slot();
+
+        let last_start_period = self.final_state.read().last_start_period;
```
damip, line 424: last_start_period is a config parameter, it can be passed directly in the Execution config to avoid having to fetch it from final_state with a read-lock

massa-node/src/main.rs
```diff
@@ -470,6 +470,7 @@ async fn launch(
     );
 
     // launch pool controller
+    let last_start_period = args.restart_from_snapshot_at_period.unwrap_or(0);
```
damip, line 473: Problem: last_start_period can be bootstrapped !

damip (approved):
LGTM

PR #3881, opened 2023-04-28
Clean up decoding of known-len component of messages between bootstrap client/server

Ben-PH:
  * [x] make tests compile
  * [x] make tests pass 


Ben-PH (commented):
I've taken this approach instead of `impl Deserializer<ServerMessageLeader> for ServerMessageLeaderDeserializer`, as when I tried that, the type-system expectations compromises the error handling/information process without any meaningful benefit.

I've added a point to the follow up issues this relates to regarding the constants.

massa-models/src/serialization.rs
```diff
@@ -49,31 +50,28 @@ impl SerializeMinBEInt for u64 {
...
 
+/// Compute the minimal big endian deserialization size
+pub const fn u64_be_bytes_min_length(max_value: u64) -> usize {
+    size_of::<u64>() - (max_value.leading_zeros() as usize) / 8
+}
+impl DeserializeMinBEInt for u32 {
     fn from_be_bytes_min(buffer: &[u8], max_value: Self) -> Result<(Self, usize), ModelsError> {
-        let read_bytes = Self::be_bytes_min_length(max_value);
-        let skip_bytes = Self::MIN_BE_INT_BASE_SIZE - read_bytes;
+        let read_bytes = u32_be_bytes_min_length(max_value);
```
Ben-PH, line 69: const methods as part of traits is... difficult/impossible. Tried a macro, but to make it work "elegantly" would be a poor use of time. Given it's just two types, and not something that is likely to be generalised in the (near) future (from what I can tell), putting in a todo/follow-up is likely to be a distraction, though happy to be corrected.

modship (approved):

PR #3900, opened 2023-05-03
fix network

modship:
  * [ ] make tests compile
  * [ ] make tests pass 


modship:
last problem, is global dosn"t work for ipv6 -> IPv4-mapped addresses : https://doc.rust-lang.org/std/net/struct.Ipv6Addr.html#method.is_global 

so tester dosn't working

PR #3902, opened 2023-05-04
Fixes for roll slash

sydhds:
  * [ ] make tests compile
  * [ ] make tests pass 

### DIFF
--- a/massa-execution-worker/src/speculative_roll_state.rs
+++ b/massa-execution-worker/src/speculative_roll_state.rs
@@ -166,10 +166,14 @@ impl SpeculativeRollState {
 
         match self.added_changes.roll_changes.get_mut(addr) {
             None => {
-                // Address does not exist anymore, slash nothing
-                Ok(0)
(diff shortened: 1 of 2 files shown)


massa-protocol-worker/src/handlers/peer_handler/mod.rs
```diff
@@ -431,7 +431,9 @@ impl InitConnectionHandler for MassaHandshake {
                     peer_db_write
                         .index_by_newest
                         .insert(Reverse(announcement.timestamp), peer_id.clone());
-                    peer_db_write.index_by_newest.retain(|_, peer_id_stored| peer_id_stored != peer_id);
+                    peer_db_write
```
AurelienFT, line 434: Can you update your PR to not have this changes ?

sydhds: Rebased :)

AurelienFT (approved):

PR #3906, opened 2023-05-04
Add proper error result in MipStoreRaw::update_with(..) function

sydhds:
  * [ ] make tests compile
  * [ ] make tests pass 

### DIFF
--- a/massa-versioning-worker/src/versioning.rs
+++ b/massa-versioning-worker/src/versioning.rs
@@ -468,7 +468,7 @@ impl MipStore {
         lock.update_network_version_stats(slot_timestamp, network_versions);
     }
 
-    #[allow(clippy::result_unit_err)]
+    #[allow(clippy::result_large_err)]
     pub fn update_with(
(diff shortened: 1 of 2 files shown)


massa-versioning-worker/src/versioning.rs
```diff
@@ -616,7 +635,10 @@ impl MipStoreRaw {
                     } else {
                         // Something is wrong (time range not ok? / version not incr? / names?
                         // or component version not incr?)
-                        should_merge = false;
+                        should_merge = Some(UpdateWithError::Overlapping(
```
Leo-Besancon, line 638: It's not super clear from the context what Overlapping means to me. Maybe give an example to make this clearer where this applies (here or in the Error enum definition)

sydhds: Added some examples

Leo-Besancon (approved):
Perfect! Thanks for the docs improvement

PR #3907, opened 2023-05-04
Document the motive behind duplicating checks when testing

Ben-PH:
  * [ ] make tests compile
  * [ ] make tests pass 

### DIFF
--- a/massa-models/src/block_header.rs
+++ b/massa-models/src/block_header.rs
@@ -42,8 +42,8 @@ pub struct BlockHeader {
 // TODO: gh-issue #3398
 #[cfg(any(test, feature = "testing"))]
 impl BlockHeader {
-    /// This is an intentional duplication of of checking invariants. In production code,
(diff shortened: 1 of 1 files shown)


massa-models/src/block_header.rs
```diff
@@ -42,6 +42,9 @@ pub struct BlockHeader {
 // TODO: gh-issue #3398
 #[cfg(any(test, feature = "testing"))]
 impl BlockHeader {
+    /// This is an intentional duplication of of checking invariants. In production code,
+    /// these checks dispersed throughout the deserialization process. This test-only function
+    /// allows all the checks to be in one place.
```
Ben-PH, line 47:
```suggestion
    /// This is an intentional duplication of invariant checks. In production code,
    /// these checks are dispersed throughout the deserialization process. This test-only function
    /// allows all the checks to be in one place.
```

sydhds (approved):

PR #3910, opened 2023-05-05
Mockall/simple pool

Ben-PH:
  * [ ] make tests compile
  * [ ] make tests pass 


massa-pool-worker/src/tests/scenario.rs
```diff
@@ -31,93 +33,131 @@ use massa_models::slot::Slot;
...
-/// that return the full storage)
-/// The execution thread will response that no operations had been executed.
-///
-/// ## Expected results
-/// The execution controller is expected to be asked 2 times for the first interaction:
-/// - to check the already executed operations
-/// - to check the final and candidate balances of the creator address
-/// And one time for the 9 next to check the executed operations.
+/// Create a mock-execution-controller story:
+/// 1. unexpected_opse_among, returning storage
```
Ben-PH, line 66:
```suggestion
/// 1. unexpected_opse_among, returning `storage.get_op_refs()`
```

massa-pool-worker/src/tests/scenario.rs
```diff
@@ -31,93 +33,131 @@ use massa_models::slot::Slot;
...
+    /// A solution that doesn't need to use a static would be welcomed
+    /// ```rust no_run
+    /// let storage = Storage::create_root();
+    /// let mut res = Box::new(MockExecutionController::new());
+    /// res.expect_unexecuted_ops_among()
+    ///     .times(1)
+    ///     .return_once(|_, _| storage.get_op_refs().clone())
+    /// ```
+    static ref STORAGE: RwLock<Storage> = RwLock::new(Storage::create_root());
+}
```
Ben-PH, line 58: ...just use the value that the mocks need?

Ben-PH (commented):
self review

massa-pool-worker/src/tests/scenario.rs
Ben-PH: ...whichever approach is taken, it seems that a lazy-static storage is needed, as the initialization of the `HashSet<OperationId>` would need access to storage itself.

...we could use manual initialization, though...

Ben-PH: Nope... not clean

bilboquet: lazy_static may sounds like a problem as it's hard to figure out when the initialization takes place (for the app), right?
In such a case, one could fallback to old school method and add a function init, called at the beginning of main, that forces the init of every lazy_static. Not perfect, a bit manual, but deterministic.

Ben-PH: I've been working on it a bit more, and found a solution to needing something global.

The other issue is that when running a multitude of tests, it shares the same global across tests... not good.

Ben-PH: I think I garbled that a bit: "found a solution _so that we don't need a global_" should've been what I wrote...

Based on a quick reading/surface level understanding of serial_test, it will only solve the problem where  the tests are accessing the global at the same time.

The other issue, is that with this lazy-static, the global variable has its state persist across tests. If one test adds operations to the storage, the next test will also see those operations.

bilboquet: > I think I garbled that a bit: "found a solution so that we don't need a global" should've been what I wrote...

That was my guess, but better to ask :)
If you don't go static you solve all your problem at once, so I hope you solution is working. Can wait to see it!

> Based on a quick reading/surface level understanding of serial_test, it will only solve the problem where the tests are accessing the global at the same time…

That's it, it solves only one part of the problem "sharing the resource" but not the state of the resource itself, so it may be still needed to add some init/cleaning at the beginning of tests (it's only recreating the object, my taste is that's it's ok in the context of tests).

Ben-PH:
Closing in favor if #3954

PR #3921, opened 2023-05-09
Use cargo-machete to fuzzy-match unused dependencies

Ben-PH:
  * [x] make tests compile
  * [x] make tests pass 

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -908,16 +908,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "ctor"
-version = "0.1.26"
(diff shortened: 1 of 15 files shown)


Ben-PH:
cargo-machete has false positives for `serde_json` in our node and consensus worker libraries.

Ben-PH:
I don't suggest adding this to CI, as there are some issues with false positives/negatives, and can mangle the `Cargo.lock` file when using `--with-metadata`

AurelienFT (approved):
Thank you for this clean up !

PR #3928, opened 2023-05-10
Fix final slot exec broadcast

aoudiamoncef:
  * [x] make tests compile
  * [x] make tests pass 

### Summary
📢🗂️🐑

Add feature to broadcast final slot execution output to subscribers in `execution.rs`. Clone `exec_out` to avoid ownership issues.

> _`exec_out` cloned_
> _Broadcast to subscribers_
> _Cache holds the answer_

### Walkthrough
*  Add a feature to broadcast the final slot execution output to active channel subscribers ([link](https://github.com/massalabs/massa/pull/3928/files?diff=unified&w=0#diff-3c20f3c5be10cedee411058391c72ef3b17e6102f3325e0b9c98b5ca3a3edbf1L1260-R1279))

### DIFF
--- a/massa-execution-worker/src/execution.rs
+++ b/massa-execution-worker/src/execution.rs
@@ -1257,9 +1257,26 @@ impl ExecutionState {
                 // speculative execution front result matches what we want to compute
 
                 // apply the cached output and return
-                self.apply_final_execution_output(exec_out);
+                self.apply_final_execution_output(exec_out.clone());
 
                 debug!("execute_final_slot: found in cache, applied cache");
+
+                // Broadcast a final slot execution output to active channel subscribers.
+                if self.config.broadcast_enabled {
+                    let slot_exec_out = SlotExecutionOutput::FinalizedSlot(exec_out.slot);
+                    if let Err(err) = self
+                        .channels
+                        .slot_execution_output_sender
+                        .send(slot_exec_out)
+                    {
+                        trace!(
+                    "error, failed to broadcast final execution output(cached) for slot {} due to: {}",
+                    exec_out.slot,
+                    err
+                );
+                    }
+                }
+
                 return;
             } else {
                 // speculative cache mismatch


modship (approved):

massa-execution-worker/src/execution.rs
```diff
@@ -1257,9 +1257,26 @@ impl ExecutionState {
...
+                        .channels
+                        .slot_execution_output_sender
+                        .send(slot_exec_out)
+                    {
+                        trace!(
+                    "error, failed to broadcast final execution output(cached) for slot {} due to: {}",
+                    exec_out.slot,
+                    err
+                );
+                    }
```
Ben-PH, line 1277: Pure nit that jumped out to me. feel free to ignore: 

indentation seems a bit wonky. something like this, perhaps:
```suggestion
                        .send(slot_exec_out) {
                            trace!(
                                "error, failed to broadcast final execution output(cached) for slot {} due to: {}",
                                exec_out.slot,
                                err
                            );
                        }
```

this was just done manually, but I would have thought `fmt` would sort something like that out.

massa-execution-worker/src/execution.rs
```diff
@@ -1294,6 +1311,19 @@ impl ExecutionState {
 
         // Broadcast a final slot execution output to active channel subscribers.
         if self.config.broadcast_enabled {
+            let slot_exec_out = SlotExecutionOutput::ExecutedSlot(exec_out.clone());
```
AurelienFT, line 1314: After further looking at the code base this one is a duplicate because it will already be sent by the call L.1301 to `execute_slot`

aoudiamoncef: This is in case we keep the actual behaviour. I'll correct

AurelienFT (approved):

aoudiamoncef:
bors merge

PR #3938, opened 2023-05-12
Fix infinite tester con

### LINKED ISSUE #3937: Avoid launching infinite tester connections
Currently, we have an number X of testers that launch tests. The problem is that they are using a new instance of peernet to do. This makes their connexion to test launched in a different thread each time we call `try_connect()`.

The tester should now make the test directly on his thread to block. We need to remove the usage of peernet and place the code that test the peer in the same thread as the tester.

modship:
  * [ ] make tests compile
  * [ ] make tests pass 

close #3937

### DIFF
--- a/massa-protocol-worker/src/handlers/peer_handler/tester.rs
+++ b/massa-protocol-worker/src/handlers/peer_handler/tester.rs
@@ -1,5 +1,6 @@
 use std::{
     collections::HashMap,
+    io::Read,
     net::{IpAddr, SocketAddr},
     thread::JoinHandle,
     time::Duration,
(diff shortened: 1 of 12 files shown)


massa-protocol-worker/src/handlers/peer_handler/tester.rs
```diff
@@ -244,12 +400,19 @@ impl Tester {
         .spawn(move || {
             let db = peer_db.clone();
             let active_connections = active_connections.clone();
-            let config = PeerNetConfiguration::default(
+            let _config = PeerNetConfiguration::default(
```
AurelienFT, line 403: Remove this variable

massa-protocol-worker/src/handlers/peer_handler/tester.rs
```diff
@@ -376,11 +554,21 @@ impl Tester {
...
+
+                        let res = Tester::tcp_handshake(
+                            messages_handler.clone(),
+                            db.clone(),
+                            announcement_deser.clone(),
+                            VersionDeserializer::new(),
                             listener,
-                            protocol_config.timeout_connection.to_duration(),
-                            &OutConnectionConfig::Tcp(Box::new(TcpOutConnectionConfig::new(protocol_config.read_write_limit_bytes_per_second / 10, Duration::from_millis(100)))),
+                            protocol_config.version.clone(),
```
AurelienFT, line 564: Some clippy warns

AurelienFT (approved):

AurelienFT:
bors merge

PR #3963, opened 2023-05-17
Use Lsmtree instead of Monotree for Sprase Merkle Tree implementation

litchipi:
Use Lsmtree as it works better for our case


Leo-Besancon (approved):
Great, thanks a lot for your help! :D

PR #3973, opened 2023-05-23
Clique computation test

damip:
  * [ ] make tests compile
  * [ ] make tests pass 


AurelienFT (approved):
We want to have it on the version that is audited: TEST.21 also ? Because last time they asked for tests it was for this version. Otherwise this seems ok

PR #3974, opened 2023-05-23
Use schnellru crate instead of lru in massa-protocol-worker

### LINKED ISSUE #3844: Change lru crate in network
Use this crate : schnellru = "0.2.1"

litchipi:
Closes #3844

Uses `schnellru` version `0.2.1` for the Lru cache in protocol worker.
Some hacks were required to make it work as it doesn't have any `contains`, `iter_mut` and `promote` functions.
Also, some other functions have different behavior (`get` returns a `&mut` to have a immutable we have to use `peek`)

### DIFF
--- a/massa-protocol-worker/src/handlers/block_handler/propagation.rs
+++ b/massa-protocol-worker/src/handlers/block_handler/propagation.rs
@@ -81,7 +81,7 @@ impl PropagationThread {
                                     }
                                 }
                                 for peer_id in peers_connected {
-                                    if !cache_write.blocks_known_by_peer.peek(&peer_id).is_some() {
+                                    if cache_write.blocks_known_by_peer.peek(&peer_id).is_none() {
                                         //TODO: Change to detect the connection before
                                         cache_write.blocks_known_by_peer.insert(
                                             peer_id,
--- a/massa-protocol-worker/src/handlers/block_handler/propagation.rs
+++ b/massa-protocol-worker/src/handlers/block_handler/propagation.rs
@@ -1,12 +1,12 @@
-use std::{collections::VecDeque, num::NonZeroUsize, thread::JoinHandle, time::Instant};
+use std::{collections::VecDeque, thread::JoinHandle, time::Instant};
 
 use crossbeam::channel::{Receiver, Sender};
-use lru::LruCache;
 use massa_logging::massa_trace;
 use massa_models::{block_id::BlockId, prehash::PreHashSet};
 use massa_protocol_exports::PeerId;
 use massa_protocol_exports::{ProtocolConfig, ProtocolError};
 use massa_storage::Storage;
+use schnellru::{ByLength, LruMap};
 use tracing::{debug, info, warn};
 
 use crate::{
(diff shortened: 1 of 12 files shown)


litchipi:
As Aurelien mentioned you in #3844, @Eitu33 what do you think about these changes ? I don't really know the `schnellru` crate

massa-protocol-worker/src/handlers/block_handler/propagation.rs
```diff
@@ -77,18 +77,18 @@ impl PropagationThread {
                                     self.active_connections.get_peer_ids_connected();
                                 for peer_id in peers {
                                     if !peers_connected.contains(&peer_id) {
-                                        cache_write.blocks_known_by_peer.pop(&peer_id);
+                                        cache_write.blocks_known_by_peer.remove(&peer_id);
                                     }
                                 }
                                 for peer_id in peers_connected {
-                                    if !cache_write.blocks_known_by_peer.contains(&peer_id) {
+                                    if cache_write.blocks_known_by_peer.get(&peer_id).is_none() {
```
AurelienFT, line 84: maybe use peek there ?

massa-protocol-worker/src/handlers/block_handler/retrieval.rs
```diff
@@ -1105,28 +1123,46 @@ impl RetrievalThread {
...
+                                        .try_into()
                                         .expect("max_node_known_blocks_size in config must be > 0"),
-                                ),
+                                )),
                                 Instant::now(),
                             ),
                         );
                     } else {
-                        cache_write.blocks_known_by_peer.promote(&peer_id);
+                        let data = cache_write
```
AurelienFT, line 1141: [get](https://docs.rs/schnellru/0.2.1/schnellru/struct.LruMap.html#method.get) can be used to promote also

litchipi:
@AurelienFT Ready for a new review

massa-protocol-worker/src/handlers/block_handler/retrieval.rs
```diff
@@ -503,14 +502,18 @@ impl RetrievalThread {
...
+                                self.config
+                                    .max_node_known_blocks_size
+                                    .try_into()
                                     .expect("max_node_known_blocks_size in config must be > 0"),
-                            )
-                        });
+                            ))
+                        })
+                        .ok_or(())
+                        .expect("endorsements_known_by_peer limit reached");
```
AurelienFT, line 514: We shouldn't have any except because we can't save data in the cache. The cache is a good to have feature but shouldn't kill the whole handler if the cache is full. The except that was there before was on a config value set at start not being 0. I think we should remove all the except to change them to result with warning in case of error but don't stop the function

AurelienFT: Yes I just didn't place a comment each time but it apply everywhere.

AurelienFT (approved):

AurelienFT:
Nice refactor ! Thank you

PR #3975, opened 2023-05-24
upgrade jsonrpsee

aoudiamoncef:
  * [x] make tests compile
  * [x] make tests pass 

### DIFF
--- a/massa-api-exports/Cargo.toml
+++ b/massa-api-exports/Cargo.toml
@@ -8,7 +8,7 @@ edition = "2021"
 paginate = "1.1.11"
 displaydoc = "0.2"
 thiserror = "1.0"
-jsonrpsee = { git = "https://github.com/paritytech/jsonrpsee.git", rev = "ab097c8", features = ["jsonrpsee-core", "jsonrpsee-types"] }
+jsonrpsee = { version = "0.18.2", features = ["jsonrpsee-core", "jsonrpsee-types"] }
 serde = { version = "1.0", features = ["derive"] }
 strum = { version = "0.24", features = ["derive"] }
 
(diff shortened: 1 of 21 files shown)


AurelienFT:
I try to spam tx on this.

massa-api-exports/Cargo.toml
```diff
@@ -8,7 +8,7 @@ edition = "2021"
 paginate = "1.1.11"
 displaydoc = "0.2"
 thiserror = "1.0"
-jsonrpsee = { git = "https://github.com/paritytech/jsonrpsee.git", rev = "118acc3", features = ["jsonrpsee-core", "jsonrpsee-types"] }
+jsonrpsee = { git = "https://github.com/paritytech/jsonrpsee.git", rev = "ab097c8", features = ["jsonrpsee-core", "jsonrpsee-types"] }
```
AurelienFT, line 11: Why don't you just use a tagged/published version like other dependencies ?

aoudiamoncef: To get the latest bug fixes, but It seems to be a runtime issue. I'll check this quickly and be back

aoudiamoncef: In order to the latest bugfixes

AurelienFT: They didn't add much things since the last version 0.18.2 its doesn't seems that there is bugs that impact us in this version. It's better to not rely on a commit but on a tagged version in this case I think.

aoudiamoncef: done

AurelienFT (approved):

AurelienFT:
bors merge

aoudiamoncef:
> bors merge

Oups, I was waiting to see the result of using gRPC before but It's not a big deal

PR #3978, opened 2023-05-24
Initial attempt to integrate versioning

sydhds:
  * [ ] make tests compile
  * [ ] make tests pass 


massa-db/src/massa_db.rs
```diff
@@ -280,9 +345,34 @@ where
...
+                    versioning_handle,
+                    IteratorMode::From(&max_key_2, Direction::Forward),
+                );
+                iter.next();
+                iter
+            }
+        };
+
+        for (serialized_key, serialized_value) in db_iterator.flatten() {
+            if new_versioning_elements.len() < self.config.max_history_length {
```
Leo-Besancon, line 364: I use a new config arg here, it did not make sense to use `max_history_length`

Leo-Besancon (commented):
First review related to the parts I know more about. I'll check the versioning related code (extend_from_db..) later

massa-versioning/src/versioning.rs
```diff
@@ -934,6 +957,128 @@ impl MipStoreRaw {
...
+            .update_with(&store_raw_)
+            .expect("update with state cf data");
+        println!("updated len: {}, added len: {}", updated.len(), added.len());
+
+        let mut update_data: BTreeMap<MipInfo, MipState> = Default::default();
+        let versioning_handle = db.db.cf_handle(VERSIONING_CF).expect(CF_ERROR);
+        // Get data from state cf handle
+        for (ser_mip_info, ser_mip_state) in db
+            .db
+            .prefix_iterator_cf(versioning_handle, MIP_STORE_PREFIX)
```
Leo-Besancon, line 1051: You have to check that the keys starts with the prefix, and break when they do not. I know, it's weird behaviour, but the prefix_iterator keep iterating once the keys don't have the prefix ><

Issue #3967, opened 2023-05-19
Use try send instead of send in most of the channels intractions


AurelienFT:
TDB

PR #3983, opened 2023-05-25
use try_send instead of send in protocol

modship:
  * [ ] make tests compile
  * [ ] make tests pass 

close : https://github.com/massalabs/massa/issues/3967


AurelienFT (changes requested):
I already assigned the issue to myself because we made a diagram with damir about what should be send and what should be try_send.
It's not too big problem because your PR is a good start and so I will adjust from the diagram this afternoon.

AurelienFT (approved):

PR #3986, opened 2023-05-26
change MassaTime::from to MassaTime::from_millis

modship:
  * [X] make tests compile
  * [X] make tests pass 


massa-time/src/lib.rs
```diff
@@ -144,7 +144,7 @@ impl From<u64> for MassaTime {
     /// Conversion from `u64`, representing timestamp in milliseconds.
     /// ```
     /// # use massa_time::*;
-    /// let time : MassaTime = MassaTime::from(42);
+    /// let time : MassaTime = MassaTime::from_millis(42);
```
AurelienFT, line 147: Can you remove this function ? so that we are sure it's not use and will not be used anymore

modship: Maybe we should keep it and update the function to MassaTime::from_millis(val), 
Because i see now we have in massa lot of XXXXXXXX.into().

Or i delete them and update all `.into()` call 

What do you think @AurelienFT

AurelienFT: In the issue damir mentioned the Duration of the standard library they don't use From at all because it's very ambiguous so yeah the goal is to delete all the ambiguous methods.

AurelienFT:
I will fix the CI on testnet_23

massa-time/src/lib.rs
modship: removed

modship:
> I will fix the CI on testnet_23
:heart_eyes:

AurelienFT (approved):
Thank you for this rectoring

AurelienFT (approved):

PR #4002, opened 2023-05-30
Update cache sync cache.

AurelienFT:
  * [ ] make tests compile
  * [ ] make tests pass 

### DIFF
--- a/massa-protocol-worker/src/handlers/block_handler/cache.rs
+++ b/massa-protocol-worker/src/handlers/block_handler/cache.rs
@@ -47,6 +47,35 @@ impl BlockCache {
             max_known_blocks_by_peer,
         }
     }
+
+    pub fn update_cache(
+        &mut self,
+        peers_connected: HashSet<PeerId>,
+        max_known_blocks_by_peer: u32,
+    ) {
+        let peers: Vec<PeerId> = self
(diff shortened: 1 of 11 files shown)


massa-protocol-worker/src/handlers/block_handler/cache.rs
```diff
@@ -47,6 +47,35 @@ impl BlockCache {
...
         }
     }
+
+    pub fn update_cache(
+        &mut self,
+        peers_connected: HashSet<PeerId>,
+        max_known_blocks_by_peer: u32,
+    ) {
+        let peers: Vec<PeerId> = self
+            .max_known_blocks_by_peer
```
modship, line 57: max_known_blocks_by_peer is u32 no ?

AurelienFT: Fixed. I will fix my IDE also he was like "everythig is fine"

modship: me too since few days

AurelienFT: Compilation ended on my PC I fixed everything should be ok now

modship (approved):

PR #4005, opened 2023-05-31
script for package versions

modship:
  * [ ] make tests compile
  * [ ] make tests pass 

Proposal : 

if we consider that the versions >= 1 are MainNet and versions <1 are TestNet / SandBox.
Parsing the current `VERSION` in `massa-models/src/config/constants` and for now, remove `TEST` or `SAND` from it to replace by `0`. 
That give for example `0.23.0`.
Then replace in all `Cargo.toml` the package version by this version.

<details>
<summary>Deprecated</summary>

Actually the script is in build.rs file which is a pre-build script. It's fast but maybe it's "overkill" to run it at every build ? 

</details>

A new massa-package called `massa-xtask` was created with this `main.rs`
```rust
/// to use it task: cargo xtask <task_name>
/// example: cargo xtask update_package_versions to update package versions
fn main() {
    let task = env::args().nth(1);

    match task.as_deref() {
        // We can add more tasks here
        Some("update_package_versions") => update_package_versions(),
        _ => panic!("Unknown task"),
    }
}
```

We can run the task with `cargo xtask update_package_versions`. With that it's possible to add easily some other task.

Example : 

- update `VERSION` to `TEST.24.0` in `massa-models/src/config/constants.rs` 
- call `cargo xtask update_package_versions`

image

### DIFF
--- a/massa-node/Cargo.toml
+++ b/massa-node/Cargo.toml
@@ -3,14 +3,9 @@ name = "massa-node"
 version = "0.23.0"
 authors = ["Massa Labs <info@massa.net>"]
 edition = "2021"
-build = "build.rs"
 
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
 
-[build-dependencies]
-massa_models = { path = "../massa-models" }
-toml_edit = "0.19.8"
-walkdir = "2.3.3"
 
 [dependencies]
 crossbeam-channel = "0.5.6"
--- a/massa-node/Cargo.toml
+++ b/massa-node/Cargo.toml
@@ -9,6 +9,8 @@ build = "build.rs"
 
 [build-dependencies]
 massa_models = { path = "../massa-models" }
+toml_edit = "0.19.8"
+walkdir = "2.3.3"
 
 [dependencies]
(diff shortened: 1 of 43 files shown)


modship:
> We shouldn't run it on every build I think because it will trigger on the node runners build. Maybe we can use cargo script to run it whenever we want

yes i think too

massa-node/Cargo.toml
```diff
@@ -1,11 +1,12 @@
 [package]
-name = "massa-node"
-version = "0.1.0"
+name = "massa_node"
```
AurelienFT, line 2: This change and the one the client change the name of the binary also ?

modship: yes, the binary was renamed. I rollback my modification on it for now

AurelienFT: Ok yes thank you because it implies a lot for a lot of tooling etc

AurelienFT (approved):
Perfect thanks a lot !

massa-node/Cargo.toml
modship: Yes of course I had to check

PR #4007, opened 2023-05-31
Save block for the whole cycle without ops

AurelienFT:
  * [x] make tests compile
  * [x] make tests pass 

### DIFF
--- a/massa-consensus-worker/src/state/prune.rs
+++ b/massa-consensus-worker/src/state/prune.rs
@@ -24,15 +24,27 @@ impl ConsensusState {
         for a_block in self.active_index.iter() {
             if let Some(BlockStatus::Active {
                 a_block: active_block,
-                ..
-            }) = self.block_statuses.get(a_block)
+                storage,
+            }) = self.block_statuses.get_mut(a_block)
             {
                 let (_b_id, latest_final_period) =
                     self.latest_final_blocks_periods[active_block.slot.thread as usize];
+
                 if active_block.slot.period
-                    >= latest_final_period.saturating_sub(self.config.force_keep_final_periods)
+                    >= latest_final_period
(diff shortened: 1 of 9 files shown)


massa-consensus-worker/src/state/prune.rs
```diff
@@ -24,15 +24,26 @@ impl ConsensusState {
...
+
                 if active_block.slot.period
-                    >= latest_final_period.saturating_sub(self.config.force_keep_final_periods)
+                    >= latest_final_period
+                        .saturating_sub(self.config.force_keep_final_periods_without_ops)
                 {
                     retain_active.insert(*a_block);
+                    if active_block.slot.period
+                        < latest_final_period.saturating_sub(self.config.force_keep_final_periods)
+                        && !self.active_index_without_ops.contains(a_block)
```
Eitu33, line 40: I'm pretty sure we don't need an `active_index_without_ops`, because unlike `active_index` that serves a purpose in multiple consensus functions, this one is only used in prune and even has no occurrences of being used for something else than a set. Correct me if I'm wrong but it looks like this is just a waste of memory.

AurelienFT: I use it to know which block we already deleted the ops of. Otherwise each slot between 128 and 10 I would have to delete the ops again and again

Eitu33: But what I mean is that your `contains` condition is only used to feed itself, if you didn't have a `active_index_without_ops` you wouldn't need to check for a delete. It's a snake that bites its tail

Eitu33: It's keeping ids just to remove from storage against removing missing operations again and again, I guess no matter the choice the difference isn't that important.

Eitu33 (approved):

Issue #3631, opened 2023-03-02
[Meta] Versioning follow up


sydhds:
A list of idea / proposals to add to the current WIP versioning work:
* [x] Allow multiple versions to be deployed at the same time:
  * as [BIP-0009]() propose
* [x] Support new component added after main net: 
  * for now, we do not describe V0 components but what if we want to add new components (e.g. Denunciation after main net)?
  * Does this mean that we have to version MipComponent?
  * Need MipComponent deser compat with old version
  * + Detect after bootstrap if not compatible - PR: https://github.com/massalabs/massa/pull/3918
* [x] Optim MipStore with some caching to avoid querying it if it has not changed
  * Optim: avoid db writes for empty batches
    * https://github.com/massalabs/massa/pull/4169 
  * Optim: Lightweight Advance when serializing MipStore
    * https://github.com/massalabs/massa/issues/4183 - https://github.com/massalabs/massa/pull/4213
* [x] MipStoreRaw.update_with could return better error (easier to debug) - PR: https://github.com/massalabs/massa/pull/3906
* [x] Versioning + network restart
  * PR 1: https://github.com/massalabs/massa/pull/3947
  * PR 2: https://github.com/massalabs/massa/pull/3978
  * PR 3: https://github.com/massalabs/massa/pull/3981
  * PR 4: https://github.com/massalabs/massa/pull/3992
* [x] Versioning + massa client
  * Use grpc endpoint to ask for current version - and ask for version to announce/current at slot (in block header)
  * PR 1: https://github.com/massalabs/massa/pull/3957
  * PR 2: https://github.com/massalabs/massa/pull/3960
  * PR 3: https://github.com/massalabs/massa/pull/4011
* [x] Versioning + keypair generation at node startup
  * Keypair generation is done after bootstrap so versioning and thus keypair factory can be used 
* [x] Versioning grpc API stabilisation
  * https://github.com/massalabs/massa/issues/4216
  * https://github.com/massalabs/massa/pull/4238
* [ ] ~~Move addresses versioning byte so it comes before the identifier byte (requires serious changes in its serialization architecture)~~
* [x] Delete `from_bytes_without_version`
  * https://github.com/massalabs/massa/pull/4226
* [ ] ~~Remove the need for a `Defined` timestamp in `MipState` creation~~
* [x] Improve doc
  * https://github.com/massalabs/massa/pull/4008
* [x] Check for activation delay to be >= 1 cycle in MIP info? 
[...]

PR #4008, opened 2023-05-31
Use saturating_add/sub when updating mip store stats + update version…

sydhds:
…ing documentation

  * [ ] make tests compile
  * [ ] make tests pass 

### DIFF
--- a/massa-node/src/main.rs
+++ b/massa-node/src/main.rs
@@ -80,7 +80,6 @@ use massa_protocol_worker::{create_protocol_controller, start_protocol_controlle
 use massa_storage::Storage;
 use massa_time::MassaTime;
 use massa_versioning::versioning::{MipComponent, MipInfo, MipState, MipStatsConfig, MipStore};
-
 use massa_wallet::Wallet;
 use parking_lot::RwLock;
 use peernet::transports::TransportType;
--- a/massa-node/src/main.rs
+++ b/massa-node/src/main.rs
@@ -79,11 +79,14 @@ use massa_protocol_exports::{ProtocolConfig, ProtocolManager};
 use massa_protocol_worker::{create_protocol_controller, start_protocol_controller};
 use massa_storage::Storage;
 use massa_time::MassaTime;
-use massa_versioning::versioning::{MipComponent, MipInfo, MipState, MipStatsConfig, MipStore};
+use massa_versioning::{
+    mips::MIP_LIST,
+    versioning::{MipStatsConfig, MipStore},
+};
 use massa_wallet::Wallet;
 use parking_lot::RwLock;
 use peernet::transports::TransportType;
-use std::collections::{BTreeMap, HashMap};
+use std::collections::HashMap;
 use std::path::PathBuf;
 use std::sync::atomic::{AtomicUsize, Ordering};
 use std::sync::{Condvar, Mutex};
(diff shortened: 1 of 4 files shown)


Leo-Besancon (approved):
Thanks for the doc cleaning :)

massa-node/src/main.rs
```diff
@@ -80,13 +80,11 @@ use massa_protocol_exports::{ProtocolConfig, ProtocolManager};
 use massa_protocol_worker::{create_protocol_controller, start_protocol_controller};
 use massa_storage::Storage;
 use massa_time::MassaTime;
-use massa_versioning::versioning::{
-    ComponentStateTypeId, MipComponent, MipInfo, MipState, MipStatsConfig, MipStore, StateAtError,
-};
+use massa_versioning::versioning::{ComponentStateTypeId, MipStatsConfig, MipStore, StateAtError};
```
Eitu33, line 83: Idk how we're planning on deploying nodes with MIPs but removing these dependencies means we'll have to remember to put it in the deploy script if that's the way we go.

massa-versioning/src/lib.rs
```diff
@@ -9,13 +9,22 @@
...
-//! There is 2 different versions:
+//! There is 2 different 'versions':
 //! * version == Network version -> This is the network version to announce and thus to store in block header
-//! * component_version -> This is the version for the associated component and is used in VersioningFactory
+//! * component_version -> This is the version for the associated component and is used in VersioningFactory (e.g KeyPair, Address, VM)
 //!
-//! # Note on MipState
+//! # Notes on MipInfo timings and stats
+//!
+//! So in the execution module and after a block become final, we update the MipStore stats (in a blocking way) then continue one.
```
Eitu33, line 20: Then continue one? I don't understand what that means

massa-versioning/src/lib.rs
```diff
@@ -9,13 +9,22 @@
...
+//! By updating the stats, we mean sending: (Slot timestamp, Option<(current: u32, advertising: Option)>).
+//! Using the slot timestamp, ensure that the trigger (and the trigger timeout) is a consensus by all nodes
+//! (This means that the trigger and the trigger timeout are not timer based).
+//! In order to have all nodes in sync (because of node various delays), the state is se to active
```
Eitu33, line 24:
```suggestion
//! In order to have all nodes in sync (because of node various delays), the state is set to active
```

massa-versioning/src/lib.rs
```diff
@@ -9,13 +9,22 @@
...
+//! after an activation delay (duration is required to be > 1 cycle).
```
Eitu33, line 25: I there supposed to be a check on duration being above 1 cycle? Used activation delays lower than that in testing

massa-node/src/main.rs
sydhds: I've moved the MIP info declaration in versioning module + unused imports.

massa-versioning/src/lib.rs
sydhds: removed as it is misleading

massa-versioning/src/lib.rs
sydhds: done

massa-versioning/src/lib.rs
sydhds: For testing purpose, there is no limit for now. I'm adding this as a followup here: https://github.com/massalabs/massa/issues/3631

Eitu33 (approved):

PR #4010, opened 2023-06-01
Split massa-db worker and exports

Leo-Besancon:
  * [ ] make tests compile
  * [ ] make tests pass 

The goal of this PR is to avoid having RocksDB as a dependency in all the crates that interact with the DB (massa-final-state, massa-async-pool, etc.).


Leo-Besancon:
Closing, I made a new branch for easier rebase in https://github.com/massalabs/massa/pull/4045

PR #4011, opened 2023-06-01
Use keypair version to generate secret key in massa client

sydhds:
  * [ ] make tests compile
  * [ ] make tests pass 

### DIFF
--- a/massa-client/src/cmds.rs
+++ b/massa-client/src/cmds.rs
@@ -811,7 +811,8 @@ impl Command {
                             }
                         })
                         .unwrap_or(0);
-                    let key = KeyPair::generate(keypair_version as u64).unwrap();
+                    let key = KeyPair::generate(keypair_version as u64)
+                        .expect("Unable to generate key pair");
 
                     let ad = wallet.add_keypairs(vec![key])?[0];
                     if json {
--- a/massa-proto/proto/massa/api/v1/versioning.proto
+++ b/massa-proto/proto/massa/api/v1/versioning.proto
@@ -47,7 +47,7 @@ enum ComponentStateId {
   COMPONENT_STATE_ID_FAILED = 6;
 }
 
-// Same as MipComponent enum in versioning package
+// Versioning component enum
 enum MipComponent {
   MIP_COMPONENT_UNSPECIFIED = 0;
   MIP_COMPONENT_ADDRESS = 1;


massa-client/src/cmds.rs
```diff
@@ -782,27 +782,37 @@ impl Command {
...
+
+                            if is_about_keypair && state == grpc::ComponentStateId::Active {
+                                Some(entry.state_id)
+                            } else {
+                                None
                             }
                         })
                         .unwrap_or(0);
-                    println!("Should create address with version: {}", _address_version);
+                    let key = KeyPair::generate(keypair_version as u64).unwrap();
```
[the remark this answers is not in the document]
sydhds: If the version is unknown, there will be an error message like:
* "KeyPair version {} doesn't exist."

Looks good to me no?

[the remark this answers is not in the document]
Leo-Besancon: Ok sure 👍

Leo-Besancon (approved):

massa-versioning/src/grpc_mapping.rs
```diff
@@ -15,13 +15,23 @@ impl From<&ComponentStateTypeId> for grpc::ComponentStateId {
...
+impl From<&MipComponent> for grpc::MipComponent {
+    fn from(value: &MipComponent) -> Self {
+        match value {
+            MipComponent::KeyPair => grpc::MipComponent::Keypair,
+            MipComponent::Address => grpc::MipComponent::Address,
+            _ => grpc::MipComponent::Unspecified,
+        }
+    }
+}
+
```
aoudiamoncef, line 27: IMHO, if you are mirroring your Rust struct and  proto ones, just use the identifier without explicite mapping. 

The coherence is assured by the fact that we always must add new enums at the end of the enum. There is a CI in proto files to detect breaking changes

massa-proto/proto/massa/api/v1/versioning.proto
```diff
@@ -46,3 +46,10 @@ enum ComponentStateId {
   COMPONENT_STATE_ID_ACTIVE = 5;
   COMPONENT_STATE_ID_FAILED = 6;
 }
+
+// Same as MipComponent enum in versioning package
```
aoudiamoncef, line 50: In proto files, always copy/paste the doc on Rust struct. This is will be added to autogenerated documentation

massa-versioning/src/grpc_mapping.rs
aoudiamoncef: My bad, It's a new case and It's totally different than existing mappings. So, this the way. I have spoken

massa-proto/proto/massa/api/v1/versioning.proto
sydhds: Updated

aoudiamoncef (approved):
Good job 👍🏻

PR #4015, opened 2023-06-01
Remove from index when marking as invalid.

### LINKED ISSUE #4014: index not updated in consensus mark_invalid_block
Audit input: the mark_invalid_block method overwrites the state of the block to mark it as invalid. However if the block was in a different state before, it is not removed from the index of that state (eg waiting_for_dependency_index if it was in the waiting_for_dependency state before).

AurelienFT:
  * [ ] make tests compile
  * [ ] make tests pass 

Fix #4014


damip (approved):
LGTM. Maybe in the future we should encapsulate the state machine into a separate structure that only allows certain transitions and automatically manages the indexes, in order to robustify things because it's very brittle right now

PR #4016, opened 2023-06-01
Remove dependency to chrono

Leo-Besancon:
  * [ ] make tests compile
  * [ ] make tests pass 


massa-versioning/src/versioning.rs
```diff
@@ -1266,19 +1265,18 @@ mod test {
...
-            .unwrap()
-            .and_hms_opt(7, 33, 44)
-            .unwrap();
+        let now = MassaTime::now().unwrap();
+        let start = now
+            .checked_add(MassaTime::from_millis(1000 * 60 * 60))
+            .unwrap(); // In 1 hour
+        let timeout = now
+            .checked_add(MassaTime::from_millis(1000 * 60 * 60 * 2))
+            .unwrap(); // In 2 hours
```
sydhds, line 1279: I find it easier to read with real date, cannot we use another crate here?

Leo-Besancon: I don't really know, https://docs.rs/datetime/0.5.2/datetime/ seem to provide format for dates but didn't see the details. @damip do I include this crate?

Leo-Besancon: I also came across this with [time crate](https://docs.rs/time/0.3.21/time/): https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=c1b95fd188437218cdd1d97f0a771007

damip: Anything works as long as "cargo audit" doesn't tell us that the crate is unmaintained or has a dependency with a known CVE

Leo-Besancon:
I'll make a new PR because this one does not update to the latest commit, weird.

Issue #3247, opened 2022-11-24
Forbid coin transfers towards SC addresses


damip:
Make sure that normal transfer operations and ABI calls cannot credit SC addresses, and that the only way to send them coins is through function calls. This will allow filtering incoming funds and not allowing uncontrolled credits.

AurelienFT:
I don't really the need for that.

damip:
> I don't really the need for that, can you maybe explain why incoming funds from transaction is a problem ?

The SC cannot refuse them. See the discussion on discord/Dusa channel

Ben-PH:
I'll take this one, as it's a good synergy with #3475

Ben-PH:
> > I don't really the need for that, can you maybe explain why incoming funds from transaction is a problem ?
> 
> The SC cannot refuse them. See the discussion on discord/Dusa channel

@damip  is this the discussion you're referring to? https://discord.com/channels/761151090673844244/973847775060054026/1029769416549871708

Are there any other convos/informations that I can refer to for better context?

Ben-PH:
agreed.

damip:
It's the right discord convo.

* for the `Operation::Transaction` operation type: the operation fails if the `sender` or `destination` is an SC address
* for ABI coin transfers, the transfer fails if the destination is an SC address, unless the sender has write access to it

Ben-PH:
Waiting on versioning to get started.

PR #4027, opened 2023-06-02
Prevent coin transfer to SC addresses

damip:
Solves https://github.com/massalabs/massa/issues/3247

  * [ ] make tests compile
  * [ ] make tests pass 

### DIFF
--- a/massa-execution-worker/src/context.rs
+++ b/massa-execution-worker/src/context.rs
@@ -455,14 +455,6 @@ impl ExecutionContext {
 
     /// Creates a new smart contract address with initial bytecode, and returns this address
     pub fn create_new_sc_address(&mut self, bytecode: Bytecode) -> Result<Address, ExecutionError> {
-        // TODO: collision problem:
-        //  prefix addresses to know if they are SCs or normal,
-        //  otherwise people can already create new accounts by sending coins to the right hash
-        //  they won't have ownership over it but this can still be unexpected
-        //  to have initial extra coins when an address is created
-        //  It may also induce that for read-only calls.
-        //  https://github.com/massalabs/massa/issues/2331
-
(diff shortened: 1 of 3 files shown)


massa-execution-worker/src/context.rs
```diff
@@ -475,20 +467,48 @@ impl ExecutionContext {
...
+        let mut nonce = 0u64;
+        let mut address;
+
+        // Loop over nonces until we find an address that doesn't exist in the speculative ledger.
+        // Note that this loop is here for robustness, and should not be looping because
+        // even through the SC addresses are predictable, nobody can create them beforehand because:
+        // - their category is "SC" and not "USER" so they can't be derived from a public key
+        // - sending tokens to the target SC address to create it by funding is not allowed because transactions towards SC addresses are not allowed
+        loop {
+            address = self.address_factory.create(
```
AurelienFT, line 489: In the first iteration you don't add the nonce to the hash and so it will be complicated to document and explain that if nonce is 0 we don't add one. Maybe it could be better to just add the nonce at the first iteration to have an unify behaviour ?

massa-execution-worker/src/execution.rs
```diff
@@ -686,7 +686,7 @@ impl ExecutionState {
 
         // send `roll_price` * `roll_count` coins from the sender to the recipient
```
AurelienFT, line 687: If you take to modify the first comment I made it would be nice to change this source code comment which is totally wrong even if it's not related to this PR.

massa-execution-worker/src/context.rs
damip: done

AurelienFT (approved):
Ok for me. I think you can merge it.

PR #4028, opened 2023-06-02
Remove FactoryStrategy::Latest variant

sydhds:
  * [ ] make tests compile
  * [ ] make tests pass 


Eitu33 (commented):
Are we sure that the only versioned objects that must be created on timestamps are API keypairs and SC addresses? Trying to think of others but can't find

Eitu33 (approved):

sydhds:
> Are we sure that the only versioned objects that must be created on timestamps are API keypairs and SC addresses? Trying to think of others but can't find

As I've removed the FactoryStrategy::Latest and also the Option, we should be ok :)

Issue #2404, opened 2022-03-10
Improve logs when node looses internet connectivity 


sebastien-forestier:
Here is a typical loss of internet connectivity. I added logs to the node to better understand what's happening: logs for the reception of blocks and headers, and logs giving the number of active peers.

```
[2m2022-03-10T10:13:02.071596Z[0m [32m INFO[0m [2mmassa_graph::block_graph[0m[2m:[0m received header 1uSjQHtzwco7U9EYGTgWJNFPU884CRNr2haRJdk4zYJqFDU8 for slot (period: 35823, thread: 21)
[2m2022-03-10T10:13:02.084129Z[0m [32m INFO[0m [2mmassa_graph::block_graph[0m[2m:[0m received header ecUeBhB55BVhdXQp1wELSoPW78osXKHwsHhedrdt4iJXXiELY for slot (period: 35823, thread: 20)
[2m2022-03-10T10:13:02.090100Z[0m [32m INFO[0m [2mmassa_graph::block_graph[0m[2m:[0m received header 2n15q7TgdGuqAGJaKpCG34X6iCgks7c1J9qzhHfUUbTr7r5c19 for slot (period: 35823, thread: 22)
[2m2022-03-10T10:13:02.096205Z[0m [32m INFO[0m [2mmassa_graph::block_graph[0m[2m:[0m received header 23scTRyiZHbbrGC9tXWPge14EoMrVzaguTfc2NA83Dxx68jTCV for slot (period: 35823, thread: 23)
[2m2022-03-10T10:13:02.124083Z[0m [32m INFO[0m [2mmassa_graph::block_graph[0m[2m:[0m received block 21dKQnikzwBKSmiKuShR2mv9TFfKtzv5izBR7JamAEZG9oKzBG for slot (period: 35823, thread: 12)
[2m2022-03-10T10:13:02.137956Z[0m [32m INFO[0m [2mmassa_graph::block_graph[0m[2m:[0m received header VWFRXMKSL9T3G3Bbt2TrfoQoeVTQVT5PQo6DPwB4a2EJgoETi for slot (period: 35823, thread: 24)
[2m2022-03-10T10:13:02.199550Z[0m [32m INFO[0m [2mmassa_graph::block_graph[0m[2m:[0m received header 267SPZQHB9CUqZX5wbnRfibceHuKTUjeAf6s1KMTUUptbAApmH for slot (period: 35823, thread: 26)
[...]

PR #4029, opened 2023-06-02
Internal metrics

### LINKED ISSUE #4006: Create internal monitoring infrastructure
## Context

We want to have some statistics on the internal structures of massa. We can use prometheus exporters to save the content of the internal data and follow their evolution across the time

## TODO

Here is two first things that we can monitor :

- [ ] Create a `massa-channel` module that wrap the crossbeam channel but also export the size of the channel often (when receiving for example or at a certain interval)
- [ ] Monitor the `massa-storage` module to have his number of blocks, ops and endorsements stored displayed often (when reading or at a certain interval)

It would be nice to setup a grafana architecture to view those stats. Maybe we can take example from https://github.com/massalabs/sre-scripts/pull/2

modship:
  * [x] make tests compile
  * [x] make tests pass 

close #4006 #2404 

TODO : 
 - [x] Create `massa-metrics` package
 - [x] Create `massa-channel` package
 - [x] Add new features **metrics** // REMOVED
 - [x] Start metrics server 
 - [x] Expose metrics in http (for prometheus on `/metrics`)
 - [x] Set listener addr from config
 - [x] tick delay from config
 - [x] Add crossbeam channel wrapper which expose for each channel : 
 	- [X] actual_len 
 	- [X] total_receive 
- [x] Unregister when all ref to the channel are dropped 
 - [x] Counter :
 	- [x] bytes receive / send for peer manager
 	- [x] bytes receive / send for each endpoint 
 	- [x] block graph counter
 	- [x] block graph ms (sum of delta in ms between block inclusion in graph and block slot)
- [x] Gauge : 
     - [x] total blocks in storage
	 - [x] total operations in storage
	 - [X] total endorsements in storage
	- [x] final_cursor_period
	- [x] final_cursor_thread 		
	- [x] active_cursor_period
	- [x] active_cursor_thread 
	- [x] consensus period for each thread
	- [x] block cache (checked_headers_size, blocks_know_by_peer)
	- [x] operation cache (checked_operations, checked_operations_prefix, ops_know_by_peers)
	- [x] consensus state (active_index, active_index_without_ops, incoming_index, discarded_index, block_statuses)
	- [x] endorsement cache (checked_endorsements, know_by_peer)
- [x] Connection metrics
	- [x] IN
	- [x] OUT 
	- [x] bandwidth  (peernet https://github.com/massalabs/PeerNet/pull/56)	
		- [x] total sent / receive by manager
		- [x] total sent / receive for each endpoint  
- [ ] Warning!
	- [x] when the node has no connections with another peer
	- [x] no data sent / receive since x seconds
- [x]  Dashboard

Fix: 

- [x] node panic when error "AlreadyReg" when we try to create a channel that was previously create
- [x] "metrics" feature not trigger in massa-channel


massa-metrics/src/lib.rs
```diff
@@ -82,14 +98,30 @@ pub struct MassaMetrics {
 
     final_cursor_thread: IntGauge,
     final_cursor_period: IntGauge,
+
+    // peer bandwidth (bytes sent, bytes received)
+    // todo remove Arc<RwLock<...>> ?
+    peers_bandwidth: Arc<RwLock<HashMap<String, (IntCounter, IntCounter)>>>,
```
modship, line 104: Actually i can remove this Arc<RwLock<>> because only the connectivity thread will edit this HashMap

damip:
@modship  alsolet'as avoid magic constants in the code (eg. the scraping interval of 5 seconds) and make it a config file parameter

modship:
> @modship let's avoid magic constants in the code (eg. the scraping interval of 5 seconds) and make it a config file parameter

If there are still hard-coded values, it's an oversight on my part because I created the configuration.
https://github.com/massalabs/massa/pull/4029/files#diff-5275c26ef00940f5cb65a70bbaac1ffe9e40f1fd639b1bec4e7ae9f131972184R224

massa-metrics/src/lib.rs
```diff
@@ -405,6 +415,80 @@ impl MassaMetrics {
...
+
+    pub fn inc_operations_final_counter(&self, diff: u64) {
+        self.operations_final_counter.inc_by(diff);
+    }
+
+    /// Update the bandwidth metrics for all peers
+    /// HashMap<peer_id, (tx, rx)>
+    pub fn update_peers_tx_rx(&self, data: HashMap<String, (u64, u64)>) {
+        if self.enabled {
+            // #[cfg(not(feature = "testing"))]
```
AurelienFT, line 435: dead code

massa-protocol-worker/src/connectivity.rs
```diff
@@ -98,7 +99,7 @@ pub(crate) fn start_connectivity_thread(
             std::thread::sleep(Duration::from_millis(100));
 
             // Create cache outside of the op handler because it could be used by other handlers
-            let total_in_slots = config.peers_categories.values().map(|v| v.max_in_connections_post_handshake).sum::<usize>() + config.default_category_info.max_in_connections_post_handshake + 1;
+            let total_in_slots = config.peers_categories.values().map(|v| v.max_in_connections_post_handshake).sum::<usize>() + config.default_category_info.max_in_connections_post_handshake +1 ;
```
AurelienFT, line 102: remove this change

AurelienFT (approved):
Small changes but the PR is very useful and we already tried it a lot. For a first version of metrics it's really interesting

Ben-PH:
#4148 Might want to look into this.

PR #4032, opened 2023-06-05
Initial fix for ledger convertion from testnet22

Leo-Besancon:
  * [ ] make tests compile
  * [ ] make tests pass 


Leo-Besancon:
This is not needed if we don't keep the ledger of testnet22 for testnet23. As the Address format changed inbetween, it's probably easier to just restart from a clean ledger.

Issue #3444, opened 2023-01-19
Have the same way to create and encode wallet between Thyra and Massa-client


SebastienValla:
We changed the way to encode our thyra wallet to be iso with massa-client following this ticket : 
https://github.com/massalabs/thyra-plugin-wallet/issues/29
Please check the comments on the PR : 
https://github.com/massalabs/thyra-plugin-wallet/pull/30

We need to decide what we do for the salt length and for the file format (you use .dat we use json)

AurelienFT:
I've answered you about the salt size on the other PR.

About the encoding json vs our proper. I don't know why it wasn't JSON at start I think only @damip is the last person remaining from this period.

damip:
> I've answered you about the salt size on the other PR.
> 
> About the encoding json vs our proper. I don't know why it wasn't JSON at start I think only @damip is the last person remaining from this period.

Nothing prevents from encoding the plaintexy contents as json, but the whole thing then needs to be encrypted, so either the encrypted version itself is encapsulated as string in a json (what would be the point ?), or just left as a binary (.dat)

AurelienFT:
To add more precision (cc @SebastienValla) currently our dat file is a JSON object encrypted. The representation of the object when decrypted is : 
```
{
   "Addr": {
       "secret_key": "xxx",
       "public_key": "xxx"
  },
  "Addr2": ...
}
```

bilboquet:
(cc @SebastienValla )
Having read the discussion and both related issues in Thyra, I feel like I'm lacking some context. (I'm new to the subject).
Can someone confirm that on core side we only need to change the size of the salt from 12 to 16 bytes?
Side question: how to test with Thyra?

SebastienValla:
@bilboquet  we need to change the salt to 16 bytes and if possible have the following output file format  (.json) 
{
    "version": 0,
    "nickname": "seb",
    "address": "A16eDqJHyZWQfhNq7xUbyqMMMjb1iKfqQy9p91uTCqY1xkUTC7k",
    "keyPairs": [
        {
            "PrivateKey": "ZnS0859HSmkG1aFaF0+N8FAhPElYdbaxSD/KdwlH+Wwxh99J6ppjTBWFpaLCfCig9HM1UJX0Sz4eXrwVzHSe6X9SevCrBynKqn8qFprm3vI=",
            "PublicKey": "EtOWslwg46zJEPk/uLhGhMUs9V9N2mpWnPk0gmmWLsc=",
            "Salt": [
[...]

PR #4044, opened 2023-06-06
Wallet format

AurelienFT:
  * [x] make tests compile
  * [x] make tests pass 

Fix https://github.com/massalabs/massa/issues/3444

PR that make the same format of wallet in python in our deployment tooling : https://github.com/massalabs/private-deployment/pull/16 the wallet is well read by this PR


bilboquet (approved):
Only a comment about comments to fix :)

bilboquet (approved):
:+1:

massa-cipher/src/constants.rs
```diff
@@ -6,17 +6,14 @@
...
 /// A nonce is a single-use value which enables securely encrypting multiple messages under the same key.
 /// Nonces need not be random: a counter can be used so long as the values are never repeated under the same key.
 pub const NONCE_SIZE: usize = 12;
 
 /// `PBKDF2` salt size.
-pub const SALT_SIZE: usize = 12;
+pub const SALT_SIZE: usize = 16;
 
 /// `PBKDF2` hash parameters.
 pub const HASH_PARAMS: Params = Params {
```
damip, line 19: number of rounds does not follow spec: 600000

massa-cipher/src/encrypt.rs
```diff
@@ -10,14 +10,19 @@ use pbkdf2::password_hash::Salt;
...
+}
 
 /// Encryption function using AES-GCM cipher.
 ///
 /// Read `lib.rs` module documentation for more information.
-pub fn encrypt(password: &str, data: &[u8]) -> Result<Vec<u8>, CipherError> {
+pub fn encrypt(password: &str, data: &[u8]) -> Result<CipherData, CipherError> {
     // generate the PBKDF2 salt
     let raw_salt: String = thread_rng()
         .sample_iter(&Alphanumeric)
```
damip, line 28: does not follow spec: must be a sequence of bytes, not only alphanumeric which has way less entropy

damip: note that the lib expects us to base64-encode the bytes to generate the salt

damip:
To be tested:
* [ ] save a wallet with this code and try to load it with massa station
* [ ] save a wallet with massa station and try to load it with this code

damip (changes requested):
Some critical security considerations from the spec are not applied

massa-cipher/src/decrypt.rs
```diff
@@ -11,32 +11,17 @@ use pbkdf2::{
...
     // parse PBKDF2 salt
-    let salt_data = rest.get(..SALT_SIZE).ok_or_else(|| {
-        CipherError::DecryptionError(
-            "wallet file truncated: salt missing or incomplete".to_string(),
-        )
-    })?;
-    let salt = SaltString::new(std::str::from_utf8(salt_data)?)
+    let salt_data = data.salt;
+    let salt = SaltString::new(std::str::from_utf8(&salt_data)?)
         .map_err(|e| CipherError::DecryptionError(e.to_string()))?;
```
damip, line 25: Note: I think we can use

```rust
let salt = SaltString::generate(&mut rng);
```

massa-cipher/src/encrypt.rs
AurelienFT: The problem is that the crate we use only accept &str and so utf8 chain of char which will not be the case if we use total random bytes

massa-cipher/src/constants.rs
AurelienFT: Done

AurelienFT:
> To be tested before we merge:
> 
> * [ ]  save a wallet with this code and try to load it with massa station => compare keypairs
> * [ ]  save a wallet with massa station and try to load it with this code => compare keypairs

Will test when massa station will be up to date in testnet 24

massa-cipher/src/encrypt.rs
damip: I think that our crate expects us to provide `base64(random_bytes)`. Can you try and see if encryption on thyra + decryption in node works (and the opposite) works ?

Eitu33:
@damip 
So after testing today with latest buildnet version, saving a massa-station wallet and importing it in the node massa-client produces this error:
```
Error: `MassaCipher` error: `Utf8` error: invalid utf-8 sequence of 1 bytes from index 0
```

It comes from `massa-cipher/src/decrypt.rs` at the salt parsing step:
```rust
let salt = SaltString::new(std::str::from_utf8(&salt_data)?)
    .map_err(|e| CipherError::DecryptionError(e.to_string()))?;
```
It expects an utf8 string and the massa-station wallet salt is composed of random bytes.

In the crate we use the salt implementation states the following:
```rust
/// Recommended length of a salt: 16-bytes.
///
/// This recommendation comes from the [PHC string format specification]:
///
/// > The role of salts is to achieve uniqueness. A *random* salt is fine
/// > for that as long as its length is sufficient; a 16-byte salt would
/// > work well (by definition, UUID are very good salts, and they encode
/// > over exactly 16 bytes). 16 bytes encode as 22 characters in B64.
```

Generating 16 bytes and encoding them as base64 solve two issues:
1. Fix the issue mentioned in https://github.com/massalabs/massa/pull/4044#discussion_r1226645490
2. Fix the incompatibility issue between massa-station and massa-cipher

But this requires us to make a change in the massa-station wallet encryption, adding a base64 encoding after the 16 random bytes generation.

We have to make a choice here:
1. Either keep what we have and add a base64 encoding step for the salt on both sides
2.

gregLibert:
> Note: in the standard, we must not justify this requirement by "the rust lib needs it" but by the real reason behing it: to adhere to the [PHC string format specification](https://github.com/P-H-C/phc-string-format/blob/master/phc-sf-spec.md#function-duties) recommended best practices.
> 
> More info here: https://docs.rs/password-hash/latest/password_hash/struct.Salt.html :
> 
> > The role of salts is to achieve uniqueness. A random salt is fine for that as long as its length is sufficient; a 16-byte salt would work well (by definition, UUID are very good salts, and they encode over exactly 16 bytes). 16 bytes encode as 22 characters in B64. Functions should disallow salt values that are too small for security (4 bytes should be viewed as an absolute minimum).

The standard we wrote is already following the good practice of `a 16-byte salt would work well`, see https://github.com/massalabs/massa-standards/blob/main/wallet/file-format.md#serialization.

damip:
> > Note: in the standard, we must not justify this requirement by "the rust lib needs it" but by the real reason behing it: to adhere to the [PHC string format specification](https://github.com/P-H-C/phc-string-format/blob/master/phc-sf-spec.md#function-duties) recommended best practices.
> > More info here: https://docs.rs/password-hash/latest/password_hash/struct.Salt.html :
> > > The role of salts is to achieve uniqueness. A random salt is fine for that as long as its length is sufficient; a 16-byte salt would work well (by definition, UUID are very good salts, and they encode over exactly 16 bytes). 16 bytes encode as 22 characters in B64. Functions should disallow salt values that are too small for security (4 bytes should be viewed as an absolute minimum).
> 
> The standard we wrote is already following the good practice of `a 16-byte salt would work well`, see https://github.com/massalabs/massa-standards/blob/main/wallet/file-format.md#serialization.

If Salt is a 16-byte vector of random bytes, the safety of `base64_encode(Salt)` is at least as good as that of `Salt` itself and can easily be supported in the Go wallet library. In the case of the rust library however, an extra layer of PHC-compliance is required and recommends the use base64-encoded salts.

We therefore have 3 options:
* implement different standards in Go vs Rust which will fragment the ecosystem
* update the standard to use base64, which makes it compatible with rust but needs a small update on the Go library to add the bs64 encoding
* fork the rust cryptographic library to remove the requirement for PHC compliance and maintain the fork

I like option 2 the most

litchipi:
Ready to merge:
- This PR
- https://github.com/massalabs/station-massa-wallet/pull/596
- https://github.com/massalabs/massa-standards/pull/111
All at once, when this PR is reviewed and approved, and more tests are performed.

Test passing right now:
- Generating a wallet using the library in  `station-massa-wallet` (`wallet.Generate` then `wallet.Persist`), move it to `massa-clients/wallets`, start the client and run `wallet_info` to load the file

litchipi:
@damip @bilboquet Ready for review before merging (with all the other PRs)

litchipi:
Adapted the code to follow the new standard (unpadded base64)

>  Extra tests to make sure:
> 
>     load the wallet on both sides and ensure that the private AND public keys are the same
>     would be important to try the other way around as well: wallet generated by massa/massa-wallet and loaded by Station + perform the same checks above
> 

Tested these cases, it works well now :+1:

damip:
If @AurelienFT confirms it's good to me :) congrats !

litchipi:
This PR now gets the salt from the raw binary data `[u8; SALT_SIZE]` exported as `pub type Salt` from `massa-cipher`.
It converts it to the way we want before performing encryption / decryption operations.

No more changes are needed in the standard and in the massa-station side.
If you are OK with it @AurelienFT @damip , I'll close the other PRs once this PR is accepted and merged

damip:
So now it works end-to-end between the node and massa-station without changes to the standard ?
If so, all good for me :)

litchipi:
Yes it does

damip (approved):

PR #4049, opened 2023-06-06
Improve error messages in Massa client

### LINKED ISSUE #4042: Improve gRPC error message in Massa client
Currently, we display a strange error message which tells us that the gRPC API is not activated. But it should be a warning ⚠️ and with the related commands which could fails if not enabled.

aoudiamoncef:
  * [x] make tests compile
  * [x] make tests pass 

resolves #4042


AurelienFT (approved):

aoudiamoncef:
Bors merge

PR #4054, opened 2023-06-07
Make MassaDbLsmtree and RawMassaDB fields pub

Leo-Besancon:
  * [ ] make tests compile
  * [ ] make tests pass 

In short, this PR is needed in order to build a Wrapped DB in the following PR: https://github.com/massalabs/massa-ledger-editor/pull/1.

**Additional info:**
We have to build this wrapper because RocksDB refuses to open an existing DB if we don't put every Column Family of that DB.
As we removed LEDGER_CF, we can't open testnet22 ledger with the current massa-db struct.

### DIFF
--- a/massa-db-worker/src/massa_db.rs
+++ b/massa-db-worker/src/massa_db.rs
@@ -44,21 +44,21 @@ pub struct RawMassaDB<
     /// The rocksdb instance
     pub db: Arc<DB>,
     /// configuration for the `RawMassaDB`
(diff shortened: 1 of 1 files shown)


litchipi:
Arg sorry ! I didn't meant to merge it :sob:

Leo-Besancon:
> Arg sorry ! I didn't meant to merge it 😭

No worries, as long as you did not have any changes to suggest ;)

PR #4053, opened 2023-06-07
Clean bootstrap state and versioning cursors

Leo-Besancon:
* [ ] document all added functions
* [ ] try in sandbox /simulation/labnet
* [ ] unit tests on the added/changed features
  * [ ] make tests compile
  * [ ] make tests pass 
* [ ] add logs allowing easy debugging in case the changes caused problems
* [ ] if the API has changed, update the API specification

Still need to test in labnet.

PR #4056, opened 2023-06-08
Clean bootstrap cursors

Leo-Besancon:
  * [X] make tests compile
  * [X] make tests pass 

Same as https://github.com/massalabs/massa/pull/4053 but targeting main


AurelienFT:
Merged in #4060

PR #4071, opened 2023-06-13
Data robustness during restart

Leo-Besancon:
  * [ ] make tests compile
  * [ ] make tests pass 

### DIFF
--- a/massa-final-state/src/final_state.rs
+++ b/massa-final-state/src/final_state.rs
@@ -290,18 +290,20 @@ impl FinalState {
         current_slot: Slot,
         end_slot: Slot,
     ) -> Result<(), FinalStateError> {
-        let latest_snapshot_cycle =
-            self.pos_state
-                .cycle_history_cache
-                .pop_back()
-                .ok_or(FinalStateError::SnapshotError(String::from(
-                    "Invalid cycle_history",
-                )))?;
+        let latest_snapshot_cycle = self.pos_state.cycle_history_cache.back().cloned().ok_or(
+            FinalStateError::SnapshotError(String::from("Invalid cycle_history")),
+        )?;
 
         let latest_snapshot_cycle_info = self.pos_state.get_cycle_info(latest_snapshot_cycle.0);
 
         let mut batch = DBBatch::new();
 
+        self.pos_state
+            .cycle_history_cache
+            .pop_back()
+            .ok_or(FinalStateError::SnapshotError(String::from(
+                "Invalid cycle_history",
+            )))?;
         self.pos_state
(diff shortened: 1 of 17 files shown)


Leo-Besancon:
@aoudiamoncef, if you think it's relevant to add a context info that we are in downtime on the API response of get_stakers, please guide me on how to do so ;)

aoudiamoncef:
@Leo-Besanconwe have two ways to bring it to Massa, using a field in the response or a status codes. 

JsonRPC https://www.jsonrpc.org/specification
gRPC https://grpc.github.io/grpc/core/md_doc_statuscodes.html

I'm voting for a status field in the response which is explicit

massa-final-state/src/final_state.rs
```diff
@@ -287,18 +287,20 @@ impl FinalState {
...
     ) -> Result<(), FinalStateError> {
-        let latest_snapshot_cycle =
-            self.pos_state
-                .cycle_history_cache
-                .pop_back()
-                .ok_or(FinalStateError::SnapshotError(String::from(
-                    "Invalid cycle_history",
-                )))?;
+        let latest_snapshot_cycle = self.pos_state.cycle_history_cache.back().cloned().ok_or(
+            FinalStateError::SnapshotError(String::from("Invalid cycle_history")),
```
aoudiamoncef, line 291: Don't hesitate to add more détails in error messages, this is very helpful

massa-pos-exports/src/pos_final_state.rs
```diff
@@ -711,6 +711,10 @@ impl PoSFinalState {
     pub fn get_all_roll_counts(&self, cycle: u64) -> BTreeMap<Address, u64> {
         let db = self.db.read();
 
+        if self.get_cycle_index(cycle).is_none() {
+            panic!("Cycle {} not in history", cycle)
```
aoudiamoncef, line 715: Is It mandatory to kill the node ?

aoudiamoncef (approved):

massa-pos-exports/src/pos_final_state.rs
Leo-Besancon: Not really mandatory, but we should never query this info on a cycle that is not on disk. If it's the case, I would not trust the state our node is in, so we should panic to be safe and avoid unexpected consequences of that IMO.

aoudiamoncef:
@Leo-Besancon please don't forget to propagate the changes to `GetLargestStaker` in gRPC API

aoudiamoncef (approved):
LGTM 👍🏻

PR #4074, opened 2023-06-13
Prevent set_bytecode on a UserAddress

Leo-Besancon:
  * [ ] make tests compile
  * [ ] make tests pass 


Leo-Besancon:
@AurelienFT, the previous test I replaced is irrelevent now I assume?

AurelienFT:
Yes ! better now

AurelienFT (approved):

Issue #4043, opened 2023-06-06
Address string parsing incorrect


damip:
In string deserialization

```rust
        let err = Err(ModelsError::AddressParseError(s.to_string()));

        // Handle the prefix ("A{U|S}")
        let mut chars = s.chars();
        let Some(ADDRESS_PREFIX) = chars.next() else {
            return err;
        };
        let Some(pref) = chars.next() else {
            return err;
        };

        let res = match pref {
            'U' => Address::User(UserAddress::from_str_without_prefixed_type(&s[2..])?),
            'S' => Address::SC(SCAddress::from_str_without_prefixed_type(&s[2..])?),
            _ => return err,
        };
```

This is dangerous and could be used by an attacker to crash the node.

Explanation:
```rust
// the following code panics
fn main() {
 let s: &str = "🫱💣";
 println!("s[1..] = {}", &s[1..]);
}
```
the code panics because there is a fundamental difference between characters and bytes: utf-8 characters are composed of a variable number of bytes. You can get the indices of characters using https://doc.rust-lang.org/std/primitive.str.html#method.char_indices 
The crash happens because `&str` always needs to be a valid string. And by doing `s[1..]` we are cutting the first character 🫱 (4 bytes) in the middle and it results in an invalid string.

To remember: you should basically never use the `s[i]`notation for strings unless you know exactly what you are doing !

So we can do the following instead:
```rust
        let err = Err(ModelsError::AddressParseError(s.to_string()));

        // Handle the prefix ("A{U|S}")
        let mut chars = s.chars();
        let Some(ADDRESS_PREFIX) = chars.next() else {
            return err;
        };
        let Some(pref) = chars.next() else {
            return err;
        };

        let res = match pref {
            'U' => Address::User(UserAddress::from_str_without_prefixed_type(chars.as_str())?),
            'S' => Address::SC(SCAddress::from_str_without_prefixed_type(chars.as_str())?),
            _ => return err,
        };
```

TODO: similar problems for other strucutres like public key etc... and maybe other parts using strings

Ben-PH:
this ripgrep ~~will~~ should find all instances of array-indexing in our codebase, excluding the ones in test-files (exclusion done by the glob)

`rg '\[\d+\]' -g '*.rs' --glob '!*test*'`

From a quick, casual search, i couldn't find a lint that warns of array-indexing.

This might be something worth putting into the style-guide. @damip Do you want to lead/open the discussion? happy to open a discussion myself and link back to here otherwise.

[...]

PR #4076, opened 2023-06-13
Fix address FromStr implementation

Leo-Besancon:
  * [ ] make tests compile
  * [ ] make tests pass 

I Checked the implementations for other structs (basically looked at FromStr implementations and the code around it. Couldn't find any other instance of this problem.
Additionnaly, manually reviewing all the results from @Ben-PH 's regex (https://github.com/massalabs/massa/issues/4043#issuecomment-1578758649) did not reveal anything.

### DIFF
--- a/massa-models/src/address.rs
+++ b/massa-models/src/address.rs
@@ -233,8 +233,8 @@ impl FromStr for Address {
         };
 
         let res = match pref {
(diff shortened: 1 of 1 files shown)


Ben-PH (approved):
playground link to show equivalence: https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=fef86bfc770967843b72f1288a347cd0

PR #4084, opened 2023-06-14
Clean event polling bootstrap - Windows routability

Leo-Besancon:
  * [ ] make tests compile
  * [ ] make tests pass 

The goal of this PR was to clean up some event polling for Bootstrap listener, in order to resolve routability issues on Windows.
It is currently running on one testnet bootstrap server and seems to be doing fine (no signs of regressions), but I'll keep monitoring what's happening.

### DIFF
--- a/massa-bootstrap/src/listener.rs
+++ b/massa-bootstrap/src/listener.rs
@@ -3,7 +3,7 @@ use std::net::{SocketAddr, TcpListener, TcpStream};
 use mio::net::TcpListener as MioTcpListener;
 
 use mio::{Events, Interest, Poll, Token, Waker};
-use tracing::info;
+use tracing::{debug, info};
 
 use crate::error::BootstrapError;
 use crate::server::BSEventPoller;
@@ -43,7 +43,7 @@ impl BootstrapTcpListener {
         if addr.is_ipv6() {
             socket.set_only_v6(false)?;
         }
-        //socket.set_nonblocking(true)?;
+        socket.set_nonblocking(true)?;
         socket.bind(&(*addr).into())?;
 
         // Number of connections to queue, set to the hardcoded value used by tokio
@@ -54,7 +54,7 @@ impl BootstrapTcpListener {
 
         let mut mio_server =
             MioTcpListener::from_std(server.try_clone().expect("Unable to clone server socket"));
-            
+
         let poll = Poll::new()?;
 
         // wake up the poll when we want to stop the listener
@@ -64,7 +64,7 @@ impl BootstrapTcpListener {
             .register(&mut mio_server, NEW_CONNECTION, Interest::READABLE)?;
 
         // TODO use config for capacity ?
-        let events = Events::with_capacity(1);
+        let events = Events::with_capacity(32);
         Ok((
             waker,
             BootstrapTcpListener {
@@ -98,13 +98,10 @@ impl BSEventPoller for BootstrapTcpListener {
             }
         }
 
-        // FIXME: We should not have to re-register the listener after each poll
-        self.poll
-            .registry()
-            .reregister(&mut self._mio_server, NEW_CONNECTION, Interest::READABLE)?;
+        if self._mio_server.accept().is_ok() {
+            debug!("Mio server still had bootstrap connection data to read");
+        }
 
-        // There could be more than one connection ready, but we want to re-check for the stop
-        // signal after processing each connection.
         Ok(results)
     }
 }
--- a/massa-bootstrap/src/listener.rs
+++ b/massa-bootstrap/src/listener.rs
@@ -44,6 +44,7 @@ impl BootstrapTcpListener {
             socket.set_only_v6(false)?;
         }
         // This is needed for the mio-polling system, which depends on the socket being non-blocking.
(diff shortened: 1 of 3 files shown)


massa-bootstrap/src/listener.rs
```diff
@@ -43,6 +43,7 @@ impl BootstrapTcpListener {
         if addr.is_ipv6() {
             socket.set_only_v6(false)?;
         }
+        socket.set_nonblocking(true)?;
```
Leo-Besancon, line 46: @Ben-PH Can this introduce unexpected behaviours?
Thinking of something like failing with a WouldBlock would be raised as an error somewhere where we would not want it to.

Ben-PH: This is needed for the mio-polling system, which depends on the socket being non-blocking.

massa-bootstrap/src/listener.rs
```diff
@@ -77,19 +78,33 @@ impl BootstrapTcpListener {
...
-            return Ok(PollEvent::Stop);
+            return Ok(vec![PollEvent::Stop]);
         }
 
-        // Ther could be more than one connection ready, but we want to re-check for the stop
-        // signal after processing each connection.
-        Ok(PollEvent::NewConnection(
-            self.server.accept().map_err(BootstrapError::from)?,
-        ))
+        let mut results = Vec::new();
```
Ben-PH, line 89: nit: use `with_capacity` here.

massa-bootstrap/src/listener.rs
```diff
@@ -77,19 +78,33 @@ impl BootstrapTcpListener {
...
+                    results.push(PollEvent::NewConnection(self.server.accept()?));
+                }
+                _ => unreachable!(),
+            }
+        }
+
+        // We need to have an accept() error with WouldBlock, otherwise polling may not raise any new events.
+        // See https://users.rust-lang.org/t/why-mio-poll-only-receives-the-very-first-event/87501
+        if self._mio_server.accept().is_ok() {
+            debug!("Mio server still had bootstrap connection data to read");
```
Ben-PH, line 105: What are your thoughts on escalating this to error?

massa-bootstrap/src/listener.rs
Ben-PH: ...my understanding might be outdated, so mught be worth trying to prove me wrong here: See if we need this line at all.

Ben-PH (approved):
Looks good to me. Nice work.

massa-bootstrap/src/listener.rs
Ben-PH: I like it.

how about this:

```rust
while let Ok(conn) = self._mio_server.accept() {
    warn!("...");
    results.push(...);
}
```
...My thinking being is from the comment in the linked conversation:

> Therefore, you must keep calling accept until it fails with a WouldBlock error.

...and we won't be throwing away the connection either.

Thoughts?

Leo-Besancon: The problem with that all the other code handles std::net::TcpStream, and mio_server.accept() yields mio::net::TcpStream.
It may be the cleanest way to do it (I don't really like that we have both an std listener and a mio listener on the same socket), but it's probably a bigger (and maybe unwanted for other reasons) refactor.

Ben-PH: ...it was a bit of a neccisary hack. 

I can see why you chose not to go that way now.

massa-bootstrap/src/listener.rs
```diff
@@ -85,11 +87,25 @@ impl BSEventPoller for BootstrapTcpListener {
...
+            match event.token() {
+                NEW_CONNECTION => {
+                    results.push(self.server.accept()?);
+                }
+                _ => unreachable!(),
+            }
+        }
+
+        // We need to have an accept() error with WouldBlock, otherwise polling may not raise any new events.
+        // See https://users.rust-lang.org/t/why-mio-poll-only-receives-the-very-first-event/87501
```
Ben-PH, line 103: Consider providing context about why we cannot add it to the results vec, and what changes might be implemented to improve things here.

massa-bootstrap/src/listener.rs
```diff
@@ -85,11 +87,25 @@ impl BSEventPoller for BootstrapTcpListener {
...
+        while self._mio_server.accept().is_ok() {
+            warn!("Mio server still had bootstrap connection data to read");
```
Ben-PH, line 105: might be fruitful to include the remote address in here. If someone reports a bug, we might be able to connect it to this warning.

massa-bootstrap/src/listener.rs
```diff
@@ -43,6 +43,8 @@ impl BootstrapTcpListener {
         if addr.is_ipv6() {
             socket.set_only_v6(false)?;
         }
+        // This is needed for the mio-polling system, which depends on the socket being non-blocking.
```
Ben-PH, line 46: You mention in discord a more refined explination of the pathology around not setting non-blocking. If you feel that might be useful context, consider providing the extra detail.

Ben-PH (approved):
Looking good. My comments are for consideration, and don't block a merge.

AurelienFT (approved):
approve to unlock merge

AurelienFT:
bors merge

PR #4085, opened 2023-06-14
Panics if we detect we are double staking

Leo-Besancon:
  * [ ] make tests compile
  * [ ] make tests pass 

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3061,6 +3061,7 @@ dependencies = [
  "massa_storage",
  "massa_time",
  "massa_versioning",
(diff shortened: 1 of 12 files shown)


damip (changes requested):
This shouldn't be checked in protocol.
We should only check it at double stake slashing:
if here https://github.com/massalabs/massa/blob/main/massa-execution-worker/src/execution.rs#L433 one of our addresses gets slashed, we panic.

This check should be enabled by a config boolean parameter

AurelienFT:
Close in favor of #4132

PR #4086, opened 2023-06-14
Bootstrap/no opt waker

Ben-PH:
blocked pending #4084 

implements many of the changes from #3964, reverting the cond-var uses.

  * [x] make tests compile
  * [x] make tests pass 


modship (approved):
that's better without the Option for the stopper

PR #4090, opened 2023-06-15
Include xor versioning

Leo-Besancon:
  * [ ] make tests compile
  * [ ] make tests pass 


Leo-Besancon:
We need to test this on labnet and have the following behaviour:
- Before the MipInfo relative to the FinalStateHashKind becomes active, nodes with and without this PR should stay sync
- Afterwards, they are no longer sync
- Afterwards, the nodes with the MipInfo can execute more operations while staying sync

Leo-Besancon:
/!\ The lsm_tree initialization will fail massa_db::new() if we already used XOR, I'll work on a quick fix for that

damip:
See https://github.com/massalabs/massa/discussions/3852#discussioncomment-6188158 to make it secure

AurelienFT (approved):

AurelienFT:
bors merge

AurelienFT:
bors merge

PR #4093, opened 2023-06-16
Correct amount parsing

damip:
This PR corrects the parsing of Amounts (adds "exact" to string parsing).
It also adds a non-panicking from_mantissa_scale

  * [x] make tests compile
  * [x] make tests pass 

### DIFF
--- a/massa-models/src/amount.rs
+++ b/massa-models/src/amount.rs
@@ -33,21 +33,53 @@ impl Amount {
         Self(0)
     }
 
-    /// Create an Amount from the form `mantissa / (10^scale)`
-    /// Panics on any error.
-    /// Used for constant initialization.
+    /// Convert to decimal
+    fn to_decimal(self) -> Decimal {
+        Decimal::from_u64(self.0)
+            .unwrap() // will never panic
+            .checked_div(AMOUNT_DECIMAL_FACTOR.into()) // will never panic
+            .unwrap() // will never panic
+    }
+
+    /// Create an Amount from a Decimal
+    fn from_decimal(dec: Decimal) -> Result<Self, ModelsError> {
+        let res = dec
+            .checked_mul(AMOUNT_DECIMAL_FACTOR.into())
+            .ok_or_else(|| ModelsError::AmountParseError("amount is too large".to_string()))?;
+        if res.is_sign_negative() {
+            return Err(ModelsError::AmountParseError(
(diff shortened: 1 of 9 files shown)


massa-models/src/amount.rs
```diff
@@ -33,21 +33,53 @@ impl Amount {
...
+        let res = res.to_u64().ok_or_else(|| {
+            ModelsError::AmountParseError(
+                "amount is too large to be represented as u64".to_string(),
+            )
+        })?;
+        Ok(Amount(res))
+    }
+
+    /// Create an Amount from the form `mantissa / (10^scale)` in a const way.
+    /// WARNING: Panics on any error.
```
bilboquet, line 69: I think we could use https://crates.io/crates/static_assertions in this function to replace both assert and panic and catch errors at compile time.
Maybe as a follow up or separated issue?

damip: I didn't want to go for more dependencies but it could be a nice followup

massa-versioning/src/versioning.rs
```diff
@@ -899,7 +899,7 @@ impl MipStoreRaw {
 
             let vote_ratio_ = 100.0 * network_version_count / block_count_considered;
 
-            let vote_ratio = Amount::from_mantissa_scale(vote_ratio_.round() as u64, 0);
+            let vote_ratio = Amount::const_init(vote_ratio_.round() as u64, 0);
```
bilboquet, line 902: Just curious, vote_ration is not const sont const_init will be called at runtime, so why replacing from_mantissa_scale by const_init?

damip: Good point. Do you mind replacing all the runtime calls back to from_mantissa_scale to avoid panics ?

massa-models/src/amount.rs
bilboquet: Just for information I added https://github.com/massalabs/massa/issues/4095 to keep that in mind

massa-versioning/src/versioning.rs
bilboquet: ok, gonna do that :)

bilboquet: Won't change this one, I've added #4099 to track it.

bilboquet (approved):
Ok for me
rebased and squashed
If you're ok we can merge

PR #4094, opened 2023-06-16
Compute hash at the end of bootstrap

Leo-Besancon:
  * [ ] make tests compile
  * [ ] make tests pass 

This PR fixes an issue where we compute the lsm_tree updates on bootstrap even if the xor version is active. If there are a lot of operations, we may never catch up to them and bootstrap will fail.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -99,7 +99,7 @@ jobs:
         os:
           - ubuntu-latest
           - macOS-latest
-          - windows-latest
+          # - windows-latest
     steps:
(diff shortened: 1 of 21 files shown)


damip (commented):
LGTM but as a temporary fix, we will remove this in testnet 24. Let's not forget to open a followup issue

Leo-Besancon:
> LGTM but as a temporary fix, we will remove this in testnet 24. Let's not forget to open a followup issue

https://github.com/massalabs/massa/issues/4097

AurelienFT (approved):

PR #4096, opened 2023-06-16
ci: add custom OS matrix

aoudiamoncef:
  * [ ] make tests compile
  * [ ] make tests pass 


aoudiamoncef:
See https://docs.github.com/en/actions/managing-workflow-runs/manually-running-a-workflow

PR #4113, opened 2023-06-19
Bootstrap/fix event loop exit

Ben-PH:
  * [x] make tests compile
  * [x] make tests pass 


AurelienFT (approved):
Trust you on the test

Ben-PH:
> Trust you on the test

That's a tough one. This should at least be better, but setting up regression tests for this will be non-trivial, I think.

In any case, this should be a step in the right direction.

PR #4116, opened 2023-06-19
Allow local IPs, use same config flag for pre and post max_in_connections

### LINKED ISSUE #4035: Add acceptance of local addresses in categories
In categories we can remove the pre and post handshake nb_connections as it's not used anymore. We also need to add a boolean on the peer category setup to accept local addresses to connect to us.

litchipi:
Closes #4035 
PR in PeerNet: https://github.com/massalabs/PeerNet/pull/58

  * [ ] make tests compile
  * [ ] make tests pass 


AurelienFT:
Can you make your PR point to testnet_24 please

litchipi:
@modship @damip Assigning you for review as Aurelian is not available

modship (approved):
solve clippy and it's ok for me

massa-node/base_config/config.toml
```diff
@@ -210,10 +210,10 @@
     # Nb max in connections that we accept
     max_in_connections = 100
     # Peer default category limits
-    default_category_info = { target_out_connections = 10, max_in_connections_per_ip = 2, max_in_connections_pre_handshake = 70, max_in_connections_post_handshake = 15}
+    default_category_info = { target_out_connections = 10, max_in_connections_per_ip = 2, max_in_connections = 15}
```
AurelienFT, line 213: It would be nice to take the boolean about the possibility to have local address to connect to us directly in the config there.

AurelienFT (changes requested):
Thank you for the clean up. I think there is other places where we use is_global that should be changed like [there](https://github.com/massalabs/massa/blob/b640980d15ce46b6dd25a29656e1c01158c04a8e/massa-protocol-worker/src/handlers/peer_handler/models.rs#L133) and in the tester of peer_handler. Search for all the `is_global()` call in the code and you should find everywhere you need to modify

litchipi:
Actually the other `is_global` are skipped on purpose

In `massa-protocol-worker/src/handlers/peer_handler/models.rs`, in the function `get_rand_peers_to_send`, it makes no sense to send a local node to a remote peer, so the `is_global` filters well the peers here

In `massa-protocol-worker/src/handlers/peer_handler/tester.rs`, it doesn't make sense to test the local peer as it doesn't rely on the network (it's loopback)

What is the expected behaviour you were expecting in these cases ?

AurelienFT (approved):
Ok for me can you just test in labnet before merge plz

litchipi:
Seams to work like a charm :smiley:

Issue #3932, opened 2023-05-11
Bootstrap servers seems to stop after an error


AurelienFT:
One of our bootstrap server have this logs : 
```
2023-05-11T09:29:49.082634Z  INFO massa_bootstrap::server: bootstrapped peer [::ffff:173.212.193.122]:53088
2023-05-11T09:29:53.437678Z  INFO massa_bootstrap::server: bootstrapped peer [::ffff:184.174.34.65]:47658
2023-05-11T09:29:54.890168Z  INFO massa_bootstrap::server: bootstrapped peer [::ffff:173.212.194.120]:45618
2023-05-11T09:29:55.038266Z  INFO massa_bootstrap::server: bootstrapped peer [::ffff:75.119.137.36]:34142
2023-05-11T09:29:55.720178Z ERROR massa_bootstrap::server: bootstrap listener error: all io errors except for Timedout, and would-block (unix error when timed out)
2023-05-11T09:29:56.392776Z  INFO massa_bootstrap::server: bootstrapped peer [::ffff:194.60.201.37]:34336
2023-05-11T09:30:08.320367Z  INFO massa_bootstrap::server: bootstrapped peer [::ffff:194.163.151.185]:40196
```
After this it didn't bootstrapped any new peers since 20 minutes. So I think the error has stopped the bootstrap server but it shouldn't. It should just stop the bootstrap with this.
@Ben-PH Can you investigate this in prio as it seems to make our bootstrap servers go to stuck.

AurelienFT:
Ok I think I known what is it related to. It's because there is no file descriptors open because of the network module. Will upgrade the fd count on our servers. Just need a more informative message so maybe a duplicate a #3927 . In this case you can close it.

Ben-PH:
fix: #4113

PR #4119, opened 2023-06-19
Add regression test for #3932

Ben-PH:
The fix for #3932 is #4113 

This PR verifies the fix is correct, and provides a regression test in case the bug is re-introduced.

  * [x] make tests compile
  * [x] make tests pass 

### DIFF
--- a/massa-bootstrap/src/tests/scenarios.rs
+++ b/massa-bootstrap/src/tests/scenarios.rs
@@ -12,7 +12,7 @@ use crate::{
     client::MockBSConnector, get_state, server::MockBSEventPoller, start_bootstrap_server,
     tests::tools::get_bootstrap_config,
 };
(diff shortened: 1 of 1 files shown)


massa-bootstrap/src/tests/scenarios.rs
```diff
@@ -555,6 +555,129 @@ fn test_bootstrap_server() {
...
+    // // setup final states
+    let final_state_server = Arc::new(RwLock::new(get_random_final_state_bootstrap(
+        pos_server.unwrap(),
+        final_state_local_config.clone(),
+        db_server.clone(),
+    )));
+
+    // setup final state mocks.
+
+    let (mock_bs_listener, _mock_remote_connector) = accept_err_accept_stop_mocks();
```
Ben-PH, line 646: clean me

Ben-PH (commented):
self review

modship (approved):

PR #4127, opened 2023-06-20
Remove tokio custom runtimes

aoudiamoncef:
  * [x] make tests compile
  * [x] make tests pass 

KISS PR


aoudiamoncef:
@AurelienFT @Ben-PH 

I did some debugging sessions and I found that instead of spawning one custom runtime, we have 2...with the client, we got 3. 

With this PR, we are using one tokio runtime

Ben-PH:
From what I can see, this aims to reduce the number of runtimes of the node process (which is seperate from the client process), by having propogate operations not run on its _own_ runtime?

I'm having trouble deriving the context in which this runtime is created. Best guess is it's used in quite a few areas? both within the node, and the client processes?

Issue #4106, opened 2023-06-19
Check constants


damip:
In constants.rs:
* [x] remove `pub const ADDRESS_SIZE_BYTES: usize = 32;` 
* [x] set BLOCK_REWARD = 1.02  (mantissa = 102, scale = 2)
* [x] choose MAX_ASYNC_GAS wisely
* [x] at this line `pub const MAX_GAS_PER_BLOCK: u64 = u32::MAX as u64;`
  * [x] make sure we support u64 bit gas everywhere (also in runtime)
  * [x] make sure that in the future when computer perf increases, we can increase `MAX_GAS_PER_BLOCK` without hitting a limit
 * [x] TODO : adjust LRU sizes cf  https://github.com/massalabs/massa/issues/4129  in on_operations_announcements_received , insert_checked_operation
 * [x] Storage costs

AurelienFT:
Could be nice to group with : https://github.com/massalabs/massa/issues/3739 and https://github.com/massalabs/massa/issues/2863

AurelienFT:
Done in #4133

PR #4133, opened 2023-06-21
Fix constants

AurelienFT:
  * [ ] make tests compile
  * [ ] make tests pass 

Partial fix of #4106


damip (approved):
Test needs fixing: 

```
thread 'tests::scenarios_mandatories::tests::local_execution' panicked at 'assertion failed: Amount::from_str(\"299_979\").unwrap() < amount &&\n    amount < Amount::from_str(\"299_980\").unwrap()', massa-execution-worker/src/tests/scenarios_mandatories.rs:854:9
```

Otherwise LGTM

PR #4137, opened 2023-06-21
Use XOF hash in the DB

Leo-Besancon:
  * [ ] make tests compile
  * [ ] make tests pass 

TODO: Check every instance of get_db_hash() calls, we probably want to replaced them with the xof alternative.

### DIFF
--- a/massa-db-exports/src/controller.rs
+++ b/massa-db-exports/src/controller.rs
@@ -1,5 +1,5 @@
 use crate::{DBBatch, Key, MassaDBError, StreamBatch, Value};
-use massa_hash::{Hash, HashXof, XOF_HASH_SIZE_BYTES};
+use massa_hash::{HashXof, HASH_XOF_SIZE_BYTES};
 use massa_models::{error::ModelsError, slot::Slot, streaming_step::StreamingStep};
 use parking_lot::RwLock;
 use std::{fmt::Debug, sync::Arc};
@@ -52,11 +52,8 @@ pub trait MassaDBController: Send + Sync + Debug {
         prefix: &[u8],
     ) -> Box<dyn Iterator<Item = (Key, Value)> + '_>;
 
-    /// Get the current state hash of the database
-    fn get_db_hash(&self) -> Hash;
-
     /// Get the current extended state hash of the database
-    fn get_xof_db_hash(&self) -> HashXof<XOF_HASH_SIZE_BYTES>;
+    fn get_xof_db_hash(&self) -> HashXof<HASH_XOF_SIZE_BYTES>;
 
     /// Flushes the underlying db.
     fn flush(&self) -> Result<(), MassaDBError>;
--- a/massa-db-exports/src/controller.rs
+++ b/massa-db-exports/src/controller.rs
@@ -1,5 +1,5 @@
 use crate::{DBBatch, Key, MassaDBError, StreamBatch, Value};
-use massa_hash::Hash;
+use massa_hash::{Hash, HashXof};
 use massa_models::{error::ModelsError, slot::Slot, streaming_step::StreamingStep};
 use parking_lot::RwLock;
 use std::{fmt::Debug, sync::Arc};
@@ -54,6 +54,9 @@ pub trait MassaDBController: Send + Sync + Debug {
 
     /// Get the current state hash of the database
     fn get_db_hash(&self) -> Hash;
+
+    /// Get the current extended state hash of the database
+    fn get_xof_db_hash(&self) -> HashXof;
 
     /// Flushes the underlying db.
     fn flush(&self) -> Result<(), MassaDBError>;
--- a/massa-db-exports/src/controller.rs
+++ b/massa-db-exports/src/controller.rs
@@ -1,5 +1,5 @@
 use crate::{DBBatch, Key, MassaDBError, StreamBatch, Value};
-use massa_hash::{Hash, HashXof};
+use massa_hash::{Hash, HashXof, XOF_HASH_SIZE_BYTES};
 use massa_models::{error::ModelsError, slot::Slot, streaming_step::StreamingStep};
 use parking_lot::RwLock;
 use std::{fmt::Debug, sync::Arc};
@@ -56,7 +56,7 @@ pub trait MassaDBController: Send + Sync + Debug {
     fn get_db_hash(&self) -> Hash;
 
     /// Get the current extended state hash of the database
-    fn get_xof_db_hash(&self) -> HashXof;
+    fn get_xof_db_hash(&self) -> HashXof<XOF_HASH_SIZE_BYTES>;
 
     /// Flushes the underlying db.
     fn flush(&self) -> Result<(), MassaDBError>;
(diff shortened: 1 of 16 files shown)


massa-hash/src/hash_xof.rs
```diff
@@ -0,0 +1,57 @@
...
+use crate::{Hash, XOF_HASH_SIZE_BYTES};
+
+/// Extended Hash
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub struct HashXof(pub [u8; XOF_HASH_SIZE_BYTES]);
+
+impl HashXof {
+    /// Truncate the extended hash to a regular hash
+    pub fn clip_to_massa_hash(&self) -> Hash {
+        Hash::from_bytes(&self.0[..32].try_into().unwrap())
```
damip, line 12: when is this useful ?

massa-hash/src/hash_xof.rs
```diff
@@ -0,0 +1,57 @@
+use std::ops::{BitXor, BitXorAssign};
+
+use crate::{Hash, XOF_HASH_SIZE_BYTES};
+
+/// Extended Hash
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub struct HashXof(pub [u8; XOF_HASH_SIZE_BYTES]);
+
+impl HashXof {
```
damip, line 9: if we go for genericity, we should have the size be a const template parameter :)

massa-hash/src/hash_xof.rs
```diff
@@ -0,0 +1,57 @@
...
+        let mut hash = [0u8; XOF_HASH_SIZE_BYTES];
+        let mut output_reader = hasher.finalize_xof();
+        output_reader.fill(&mut hash);
+        HashXof(hash)
+    }
+
+    /// Compute from key and value
+    pub fn compute_from_kv(key: &[u8], value: &[u8]) -> HashXof {
+        let mut hasher = blake3::Hasher::new();
+        hasher.update(key);
```
damip, line 28:
```
hasher.update(key.len().to_be_bytes());
hasher.update(key);
hasher.update(value);
```

So that that Hash("abc", "def") != Hash("ab", "cdef")

massa-hash/src/hash_xof.rs
```diff
@@ -0,0 +1,57 @@
...
+        hasher.update(value);
+        let mut hash = [0u8; XOF_HASH_SIZE_BYTES];
+        let mut output_reader = hasher.finalize_xof();
+        output_reader.fill(&mut hash);
+        HashXof(hash)
+    }
+}
+
+impl BitXorAssign for HashXof {
```
damip, line 37: I think we should drop BitXorAssign for normal hashes (unsafe) and only keep it for HashXof

massa-db-exports/src/controller.rs
```diff
@@ -55,6 +55,9 @@ pub trait MassaDBController: Send + Sync + Debug {
     /// Get the current state hash of the database
     fn get_db_hash(&self) -> Hash;
```
damip, line 56: this will be removed ?

massa-hash/src/hash_xof.rs
damip: https://practice.rs/generics-traits/const-generics.html

massa-hash/src/hash_xof.rs
Leo-Besancon: That is a very good point.

massa-hash/src/hash_xof.rs
Leo-Besancon: Yes, I only kept it initially because we still use them in some tests I think, but it should be cleaned out.

massa-db-exports/src/controller.rs
Leo-Besancon: Yes

massa-hash/src/hash_xof.rs
AurelienFT: Removed.

massa-hash/src/hash_xof.rs
AurelienFT: Changed.

massa-hash/src/hash_xof.rs
AurelienFT: Removed.

massa-db-exports/src/controller.rs
AurelienFT: Done.

AurelienFT:
Can't ask for a review to @Leo-Besancon because he is the PR creator. I invoke him with this comment.

massa-hash/src/hash_xof.rs
```diff
@@ -0,0 +1,157 @@
...
+    }
+}
+
+impl<const SIZE: usize> BitXorAssign for HashXof<SIZE> {
+    fn bitxor_assign(&mut self, rhs: Self) {
+        *self = *self ^ rhs;
+    }
+}
+
+impl<const SIZE: usize> BitXor for HashXof<SIZE> {
```
Leo-Besancon, line 66: Maybe explain / refer to the discussion here and on the BitXorAssign the security considerations for the SIZE const?

massa-pos-exports/src/cycle_info.rs
```diff
@@ -61,18 +61,22 @@ impl CycleInfoHashComputer {
         Hash::compute_from(&buffer)
     }
 
-    fn compute_roll_entry_hash(&self, address: &Address, roll_count: u64) -> Hash {
+    fn compute_roll_entry_hash(
+        &self,
+        address: &Address,
+        roll_count: u64,
+    ) -> HashXof<HASH_XOF_SIZE_BYTES> {
```
Leo-Besancon, line 68: I think we should either add a comment and make it clear that this hash is never included in the db hash. I think we may be able to remove entirely the Hash tracker from the cycle info struct.
I'm not sure if it's the same with deferred credits as they are used speculatively, but I think we can remove it also.

Leo-Besancon (commented):
I can't approve officially this PR, but I do so with this comment :).

massa-pos-exports/src/cycle_info.rs
damip: hashes should not be computed at speculative time, only when finalizing, so yes we should be able to remove all hash trackers except the automatic one in the db

massa-hash/src/hash_xof.rs
AurelienFT: I don't know which discussion you speak about for the xor operator

massa-pos-exports/src/cycle_info.rs
AurelienFT: done

massa-hash/src/hash_xof.rs
Leo-Besancon: https://github.com/massalabs/massa/discussions/3852#discussioncomment-6188158

AurelienFT: Added.

PR #4145, opened 2023-06-23
Reduce memory usage

damip:
## Contents

* [x] reduce the number of force-kept final periods to 5 (1min20s)
* [x] in protocol, when retrieving block ops, they were added to `Storage` but not to `stored_operations` so they were not cleaned by `clear_storage` => refactor this to not use `self.storage`, to make better use of Storage instances in general, and move the responsibility of keeping ops temporarily alive during propagation to the op propagation handler instead of the retrievers
* [x] box heavy elements in Storage to avoid huge allocations on HashMap growth
* [x] cap the number of operations kept for propagation
* [x] reduce the endorsement propagation time to 2 periods
* [x] make operation batch announcement more robust
* [x] achieve stable 3GB RAM usage at 4ktxps on testnet

## Checklist

  * [x] make tests compile
  * [x] make tests pass 
* [ ] after merging in main, merge main to branch testnet24

### DIFF
--- a/massa-metrics/src/lib.rs
+++ b/massa-metrics/src/lib.rs
@@ -56,8 +56,6 @@ pub struct MassaMetrics {
     active_in_connections: IntGauge,
     active_out_connections: IntGauge,
 
-    retrieval_thread_stored_operations_sum: IntGauge,
-
     // block_cache
     block_cache_checked_headers_size: IntGauge,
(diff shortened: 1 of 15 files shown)


AurelienFT (approved):
How the ops storage wasn't growing with the leak on block handler. Anyway. Nice found and if tested on the testnet we can merge it.

AurelienFT:
bors merge

PR #4147, opened 2023-06-23
Refactor/launch loop ctrl

Ben-PH:
The launch loop control can lead to a particular bug, where if a resync is performed, there is no ctrl-c handler active. This will cause ctrl-c to be unresponsive during launch  following a resync

This PR does several things.

- Resolves unresponsive launch process following resync.
- Single, one-off ctrl-c handler initialization.
- Uses Condvar: A single `Arc<(Mutex<bool>, Condvar)>`, cloned as appropriate, serve the role of both ends of a signal channel.
- Reduces the use of async code.
- This arc can now be used globally. Where a system want's to be sleep and/or be aware of the system being directed to shutdown, a read/conditional sleep of this Arc is all that's needed.

Should a sigint/shutdown occur during the bootstrap retry-wait, the system will invoke `std::sys::exit(0)` without a graceful shutdown. This is a legacy of bootstraps async days:

```rust
    // bootstrap
    let bootstrap_state = tokio::select! {
        _ = &mut stop_signal => {
            info!("interrupt signal received in bootstrap loop");
            process::exit(0);
        },
        res = get_state(/*...*/) => match res {
            Ok(vals) => vals,
            Err(err) => panic!("critical error detected in the bootstrap process: {}", err)
        }
    };
```
but now, get_state will can detect the interupt directly while waiting for the next retry:

```rust
    let bootstrap_state = match get_state(/*...*/) {
        Ok(vals) => vals,
        Err(BootstrapError::Interupted(msg)) => {
            info!("{}", msg);
            process::exit(0);
        }
        Err(err) => panic!("critical error detected in the bootstrap process: {}", err),
    };
```

  * [x] make tests compile
  * [x] make tests pass 

### DIFF
--- a/massa-node/src/main.rs
+++ b/massa-node/src/main.rs
@@ -11,7 +11,7 @@ use crate::operation_injector::start_operation_injector;
 use crate::settings::SETTINGS;
 
 use crossbeam_channel::TryRecvError;
-use ctrlc as _;
+// use ctrlc as _;
 use dialoguer::Password;
 use massa_api::{ApiServer, ApiV2, Private, Public, RpcServer, StopHandle, API};
 use massa_api_exports::config::APIConfig;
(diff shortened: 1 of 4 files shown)


massa-node/src/main.rs
```diff
@@ -11,7 +11,7 @@ use crate::operation_injector::start_operation_injector;
 use crate::settings::SETTINGS;
 
 use crossbeam_channel::TryRecvError;
-use ctrlc as _;
+// use ctrlc as _;
```
Ben-PH, line 14:
```suggestion
```

Ben-PH (commented):
self-review

modship (approved):

Issue #4148, opened 2023-06-23
massa metrics listener causes panic on resync.


Ben-PH:
Caught while testing #4147 in labnet. Panic cleared out once when I deactivated the massa-metric system.

To recreate the panic: (#4150 adds this behind a feature-gate)

before the launch loop: 

```rust
    let start = Instant::now();
    let mut resynced = false;
```

...and in the launch loop

```rust
            if !resynced && start.elapsed() > Duration::from_secs(10) {
                resynced = true;
                warn!("in response to a desynchronization, the node is going to bootstrap again");
                break true;
            }
        };
        stop(
            consensus_event_receiver,
            Managers {

```

log:

```
2023-06-23T16:39:35.477808Z  INFO massa_protocol_worker::connectivity: Trying to connect to addr 51.75.131.129:31244
2023-06-23T16:39:35.655763Z  INFO massa_protocol_worker::handlers::peer_handler: Peer connected: PeerId { public_key: P12sNdL7YwSawpnJrk9XCWDjKbgfNamAobp62AX5qfkgpBkGh2wC }
2023-06-23T16:39:39.191882Z  WARN massa_node: in response to a desynchronization, the node is going to bootstrap again
2023-06-23T16:39:39.191989Z  INFO massa_bootstrap::server: ip white/blacklist updater exited cleanly
2023-06-23T16:39:39.192101Z  INFO massa_node: Start stopping API's: gRPC, EXPERIMENTAL, PUBLIC, PRIVATE
2023-06-23T16:39:39.192155Z  INFO massa_grpc::server: gRPC API stop signal sent successfully
2023-06-23T16:39:39.192174Z  INFO massa_api: API stop signal sent successfully
2023-06-23T16:39:39.192246Z  INFO massa_node: API | EXPERIMENTAL JsonRPC | stopped
2023-06-23T16:39:39.192261Z  INFO massa_api: API stop signal sent successfully
2023-06-23T16:39:39.192358Z  INFO massa_node: API | PUBLIC JsonRPC | stopped
2023-06-23T16:39:39.192372Z  INFO massa_api: API stop signal sent successfully
2023-06-23T16:39:39.192441Z  INFO massa_node: API | PRIVATE JsonRPC | stopped
2023-06-23T16:39:39.192455Z  INFO massa_factory_worker::manager: stopping factory...
2023-06-23T16:39:39.192731Z  INFO massa_factory_worker::manager: factory stopped
[...]

PR #4150, opened 2023-06-26
Feature/node resync check flag

Ben-PH:
Adds a feature flag that forces a "resync" of a node after a short wait. Follow-up to #4148 

  * [ ] make tests compile
  * [ ] make tests pass 

### DIFF
--- a/massa-node/src/main.rs
+++ b/massa-node/src/main.rs
@@ -1195,6 +1195,9 @@ async fn run(args: Args) -> anyhow::Result<()> {
     // })
     // .expect("Error setting Ctrl-C handler");
 
+    #[cfg(feature = "resync_check")]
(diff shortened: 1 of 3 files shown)


modship (approved):

massa-node/src/main.rs
```diff
@@ -1205,6 +1205,9 @@ async fn run(args: Args) -> anyhow::Result<()> {
     })
     .expect("Error setting Ctrl-C handler");
 
+    #[cfg(feature = "resync_check")]
+    let mut resync_check = Some(std::time::Instant::now() + std::time::Duration::from_secs(10));
```
modship, line 1209: can you explain with comment in the code why we need to resync_check

PR #4151, opened 2023-06-26
Prevent block and endorsements to be created if no connections

### LINKED ISSUE #4052: Disable block and endorsement production when we have 0 connections
Disable block and endorsement production when our node is not connected to any other nodes. This prevents our node from finalizing its own blocks when it loses connection. Also emit warnings to say that a block/endorsement production was skipped for lack of connectivity

This should be a config parameter.
It should be disabled for nodes that run as initial bootstrap (they can be initially alone in the network) as well as sandbox

AurelienFT:
  * [x] make tests compile
  * [x] make tests pass 

Fix #4052 

PR in deployment script to disable : https://github.com/massalabs/private-deployment/pull/21

### DIFF
--- a/massa-factory-exports/src/config.rs
+++ b/massa-factory-exports/src/config.rs
@@ -9,31 +9,24 @@ use massa_time::MassaTime;
 pub struct FactoryConfig {
     /// number of threads
     pub thread_count: u8,
-
     /// genesis timestamp
     pub genesis_timestamp: MassaTime,
-
     /// period duration
     pub t0: MassaTime,
-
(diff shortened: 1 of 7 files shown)


damip:
Make sure this is disabled for:
- our bootstrap nodes
- in sandbox mode
- in tests that require production with zero connections

damip (approved):
LGTM

Make sure this is disabled for:
* our bootstrap nodes
* on buildnet nodes
* in sandbox mode
* in tests that require production with zero connections

Issue #4092, opened 2023-06-15
Secure xor hashes


AurelienFT:
After further study with Seb, there is a way to make the XOR secure against the algorithm we mentioned.
The paper states an attack complexity of `O(2^'2sqrt(n))` where n is the number of bits of the hash.

So if we want a 128 bit security (the same as the one of 256bit hash collisions) we need to increase the hash size to `n=4096bits` which is **512 bytes**. (that's a big hash !). This can be done with the blake3 crate using XOF extensions, see the "extended output" example in https://docs.rs/blake3/latest/blake3/#examples

_Originally posted by @damip in https://github.com/massalabs/massa/discussions/3852#discussioncomment-6188158_

damip:
We should remove the XOR operator we added to MassaHash

Leo-Besancon:
Adapting the code to use extended hash for the db xor seems relatively straightforward, but I was thinking about what we print in the node. Printing 512 bytes of hash every slot on finalize() seems too much, but I assume everything else (such as printing the first and last bytes, but cutting out the middle) would decrease the security?

damip:
> Adapting the code to use extended hash for the db xor seems relatively straightforward, but I was thinking about what we print in the node. Printing 512 bytes of hash every slot on finalize() seems too much, but I assume everything else (such as printing the first and last bytes, but cutting out the middle) would decrease the security?

This is what printing the whole thing would look like:
```
[...]

PR #4155, opened 2023-06-26
Fix display of xof.

AurelienFT:
  * [x] make tests compile
  * [x] make tests pass 

Fix : https://github.com/massalabs/massa/issues/4092#issuecomment-1600953171


Leo-Besancon (approved):
It may be confusing to default to this behaviour.. but on the other hand I don't see a basic case where it would be a problem either.

Issue #4018, opened 2023-06-01
Consensus - Encapsulate the state machine into a separate structure


AurelienFT:
Maybe in the future we should encapsulate the state machine into a separate structure that only allows certain transitions and automatically manages the indexes, in order to robustify things because it's very brittle right now

_Originally posted by @damip in https://github.com/massalabs/massa/pull/4015#pullrequestreview-1455490724_

damip:
Upgrading to P1 because of external audit timings dependency

AurelienFT:
Done in https://github.com/massalabs/massa/pull/4162

Issue #4107, opened 2023-06-19
Add consensus tests


AurelienFT:
During the audit of consensus our external auditor needed some unit test and tools to create new. They were auditing the testnet_21 version so we added some test in a special branch for them : `add_test_audit`. They took inspiration from this and created a bunch of new tests.

We need to port the test we added in 21 to 24 and also port the one from the auditor.

## TODO:

- [x] Port the consensus test of branch `add_test_audit`
- [x] Port the test of the external auditor available [here](https://github.com/massalabs/massa/files/11787854/tests.zip)

AurelienFT:
Done in https://github.com/massalabs/massa/pull/4162

PR #4162, opened 2023-06-27
Create state machine consensus & add tests

AurelienFT:
  * [x] make tests compile
  * [x] make tests pass 

Fix https://github.com/massalabs/massa/issues/4018 and https://github.com/massalabs/massa/issues/4107

### DIFF
--- a/massa-consensus-worker/src/state/blocks_state.rs
+++ b/massa-consensus-worker/src/state/blocks_state.rs
@@ -1,9 +1,6 @@
 use std::collections::hash_map::Entry;
 
-use massa_consensus_exports::{
-    block_status::{BlockStatus, HeaderOrBlock},
-    error::ConsensusError,
-};
+use massa_consensus_exports::{block_status::BlockStatus, error::ConsensusError};
 use massa_models::{
     block_id::BlockId,
     prehash::{PreHashMap, PreHashSet},
@@ -11,13 +8,14 @@ use massa_models::{
 };
 use tracing::warn;
 
+#[derive(Debug, Clone)]
 pub struct BlocksState {
     /// Every block we know about
     block_statuses: PreHashMap<BlockId, BlockStatus>,
     /// Ids of incoming blocks/headers
     incoming_index: PreHashSet<BlockId>,
     /// Used to limit the number of waiting and discarded blocks
-    pub sequence_counter: u64,
+    sequence_counter: u64,
     /// ids of waiting for slot blocks/headers
     waiting_for_slot_index: PreHashSet<BlockId>,
     /// ids of waiting for dependencies blocks/headers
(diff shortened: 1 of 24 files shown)


AurelienFT:
@damip That is my first attempt. There is still things that I don't like especially now that I update instead of delete/readd in `block_statuses` I have some `clone` to do because elements are still in the hashmap and I fought with the borrow checker. I can check to remove these `clone` after we validate the way I took to avoid loosing time. Tried in labnet works well.

AurelienFT:
Tests are passing on my side will try to improve them but doesn't block review @damip

massa-consensus-worker/src/state/blocks_state.rs
```diff
@@ -0,0 +1,289 @@
...
+    pub fn discarded_blocks(&self) -> &PreHashSet<BlockId> {
+        &self.discarded_index
+    }
+
+    pub fn active_blocks(&self) -> &PreHashSet<BlockId> {
+        &self.active_index
+    }
+
+    // Internal function to update the indexes
+    fn update_indexes(
```
damip, line 78: maybe this could be changed into: 
```
transition_block_state(&mut self, block_id: &BlockId, new_status: Option<&BlockStatus>) {
   // here we automatically go check the old state (if present), verify that the transition is allowed, and perform the transition, together with updating the indexes 
}


```

AurelienFT: Yes it seems the same as directly change the index in `update_block_state` I will try

damip: I'm not a big fan of the cloning of Active blocks because it clones the Storage of the block, taking locks on block storage, endorsement storage and operation storage, and taking 5000 references to operations. Then it happens again when the cloned object is dropped

damip: So if we imagine coming back to the "removal" pattern, we can do:

```rust
pub fn transition_map(&mut self, block_id: &BlockId, Fn map_fn(Option<BlockState>)->Option<BlockState>) {
   // removes the block from block_statuses, getting its current state in the process and updating indexes
   // calls map_fn that consomes the old state, and returns the new state
   // add the new state (if any) to block_statuses and updating indexes
 }
```

AurelienFT: I have updated the only clone that is left is : https://github.com/massalabs/massa/blob/c6949ca9f1a045a1428424eedb5c7379f534f465/massa-consensus-worker/src/state/process.rs#L158 and it's a pain to remove

AurelienFT: Removed. I should be clean now. Labnet test running.

damip (approved):
Still some async tests to port but overall LGTM. Good job !!

PR #4164, opened 2023-06-27
Fix MipInfo Ord + write MIP store info when we are a bootstap server

sydhds:
  * [ ] make tests compile
  * [ ] make tests pass 


Leo-Besancon (approved):
Good job finding these!

Eitu33:
I've ran the test that fails in the mac CI and it runs correctly in local I can't seem to reproduce it

Eitu33 (approved):

PR #4170, opened 2023-06-28
Change the interface to sync with ABI changes

Leo-Besancon:
  * [ ] make tests compile
  * [ ] make tests pass 

### DIFF
--- a/massa-execution-worker/Cargo.toml
+++ b/massa-execution-worker/Cargo.toml
@@ -31,6 +31,7 @@ massa_models = { path = "../massa-models" }
 massa_storage = { path = "../massa-storage" }
 massa_hash = { path = "../massa-hash" }
 massa-sc-runtime = { git = "https://github.com/massalabs/massa-sc-runtime", branch = "feature/Improve_ABI_types_in_wasmv1" }
+massa-proto-rs = { git = "https://github.com/massalabs/massa-proto-rs", rev = "8873f96"}
 massa_metrics = { path = "../massa-metrics" }
 massa_module_cache = { path = "../massa-module-cache" }
 massa_signature = { path = "../massa-signature" }
--- a/massa-execution-worker/src/interface_impl.rs
+++ b/massa-execution-worker/src/interface_impl.rs
@@ -303,7 +303,10 @@ impl Interface for InterfaceImpl {
         let (mantissa, scale) = self
             .amount_to_mantissa_scale(amount.to_raw())
             .unwrap_or_default();
-        let native_amount = NativeAmount { mantissa, scale };
+        let native_amount = NativeAmount {
+            mandatory_mantissa: Some(mantissa),
+            mandatory_scale: Some(scale),
+        };
 
         Ok(native_amount)
     }
@@ -915,7 +918,14 @@ impl Interface for InterfaceImpl {
         from_address: Option<String>,
     ) -> Result<()> {
         let to_address = Address::from_str(&to_address)?;
-        let amount = Amount::from_mantissa_scale(raw_amount.mantissa, raw_amount.scale)?;
+        let amount = Amount::from_mantissa_scale(
+            raw_amount
+                .mandatory_mantissa
+                .ok_or(anyhow!("No mantissa provided"))?,
+            raw_amount
+                .mandatory_scale
+                .ok_or(anyhow!("No scale provided"))?,
+        )?;
         let mut context = context_guard!(self);
         let from_address = match from_address {
             Some(from_address) => Address::from_str(&from_address)?,
--- a/massa-execution-worker/src/interface_impl.rs
+++ b/massa-execution-worker/src/interface_impl.rs
@@ -15,8 +15,10 @@ use massa_models::config::MAX_DATASTORE_KEY_LENGTH;
 use massa_models::{
     address::Address, amount::Amount, slot::Slot, timeslots::get_block_slot_timestamp,
 };
+use massa_proto_rs::massa::model::v1::NativeAmount;
 use massa_sc_runtime::RuntimeModule;
 use massa_sc_runtime::{Interface, InterfaceClone};
+
 use parking_lot::Mutex;
 use rand::Rng;
 use sha2::{Digest, Sha256};
@@ -140,6 +142,18 @@ impl InterfaceClone for InterfaceImpl {
     }
 }
 
+/// Helper function to get the address from the option given as argument to some ABIs
+/// Fallback to the current context address if not provided.
+fn get_address_from_opt_or_context(
+    context: &ExecutionContext,
+    option_address_string: Option<String>,
+) -> Result<Address> {
+    match option_address_string {
+        Some(address_string) => Address::from_str(&address_string).map_err(|e| e.into()),
+        None => context.get_current_address().map_err(|e| e.into()),
+    }
+}
+
 /// Implementation of the Interface trait providing functions for massa-sc-runtime to call
 /// in order to interact with the execution context during bytecode execution.
 /// See the massa-sc-runtime crate for a functional description of the trait and its methods.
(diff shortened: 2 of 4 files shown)


Leo-Besancon:
The build fails after rebase because of grpc mapping to models structs.

massa-execution-exports/Cargo.toml
```diff
@@ -27,7 +26,8 @@ massa_final_state = { path = "../massa-final-state" }
 massa_ledger_exports = { path = "../massa-ledger-exports", optional = true }
 massa_module_cache = { path = "../massa-module-cache" }
 massa_versioning = { path = "../massa-versioning" }
-massa-sc-runtime = { git = "https://github.com/massalabs/massa-sc-runtime", branch = "feature/Improve_ABI_types_in_wasmv1" }
+massa-proto-rs = { git = "https://github.com/massalabs/massa-proto-rs", branch = "feature/Improve_ABI_types_in_wasmv1_leo", features = ["tonic"] }
```
bilboquet, line 29: ;)

massa-execution-worker/Cargo.toml
```diff
@@ -28,7 +28,8 @@ massa_execution_exports = { path = "../massa-execution-exports" }
 massa_models = { path = "../massa-models" }
 massa_storage = { path = "../massa-storage" }
 massa_hash = { path = "../massa-hash" }
-massa-sc-runtime = { git = "https://github.com/massalabs/massa-sc-runtime", branch = "feature/Improve_ABI_types_in_wasmv1" }
+massa-sc-runtime = { git = "https://github.com/massalabs/massa-sc-runtime", branch = "feature/Improve_ABI_types_in_wasmv1_leo" }
```
bilboquet, line 31: ;)

massa-execution-worker/src/interface_impl.rs
```diff
@@ -332,6 +365,18 @@ impl Interface for InterfaceImpl {
         }
     }
 
+    fn raw_get_data_wasmv1(&self, key: &[u8], address: Option<String>) -> Result<Vec<u8>> {
+        let context = context_guard!(self);
+        let address = match address {
```
bilboquet, line 370: we can probably have an helper function here

massa-module-cache/Cargo.toml
```diff
@@ -18,7 +18,7 @@ rocksdb = "0.20"
 massa_hash = { path = "../massa-hash" }
 massa_models = { path = "../massa-models" }
 massa_serialization = { path = "../massa-serialization" }
-massa-sc-runtime = { git = "https://github.com/massalabs/massa-sc-runtime", branch = "feature/Improve_ABI_types_in_wasmv1", features = [
+massa-sc-runtime = { git = "https://github.com/massalabs/massa-sc-runtime", branch = "feature/Improve_ABI_types_in_wasmv1_leo", features = [
```
bilboquet, line 21: ;)

massa-execution-worker/src/interface_impl.rs
```diff
@@ -315,6 +332,22 @@ impl Interface for InterfaceImpl {
...
 
+    fn get_keys_wasmv1(&self, prefix: &[u8], address: Option<String>) -> Result<BTreeSet<Vec<u8>>> {
+        let context = context_guard!(self);
+        let address = match address {
+            Some(address) => massa_models::address::Address::from_str(&address)?,
+            None => context.get_current_address()?,
+        };
+        match (context.get_keys(&address), prefix) {
+            (Some(mut value), prefix) if !prefix.is_empty() => {
+                value.retain(|key| key.iter().zip(prefix.iter()).all(|(k, p)| k == p));
```
Eitu33, line 343: That's not what its supposed to do is it? Unless I have misunderstood the prefix behaviour. Because here you're not treating it as a prefix since you match the entire key with it.

massa-execution-worker/src/interface_impl.rs
```diff
@@ -512,6 +613,46 @@ impl Interface for InterfaceImpl {
...
+        // compute prefix range
+        let prefix_range = if !prefix.is_empty() {
+            // compute end of prefix range
+            let n_keep = prefix
+                .iter()
+                .enumerate()
+                .rev()
+                .find_map(|(i, v)| if v < &255 { Some(i + 1) } else { None })
+                .unwrap_or(0);
+            let mut prefix_end = prefix[..n_keep].to_vec();
```
Eitu33, line 626: This seems a bit too complicated for what it does, probably a better way to do it but I don't care so much as long as it works.

Maybe something like:
```rust
let mut prefix_end = prefix.to_vec();
while let Some(255) = end_range.last() {
    end_range.pop();
}
```
I know iterators are cheap but you're chaining 3 of them for a pretty simple operation. Its also easier to read.

massa-execution-worker/src/interface_impl.rs
```diff
@@ -512,6 +613,46 @@ impl Interface for InterfaceImpl {
         Ok(keys)
     }
 
+    fn get_op_keys_wasmv1(&self, prefix: &[u8]) -> Result<Vec<Vec<u8>>> {
+        // compute prefix range
+        let prefix_range = if !prefix.is_empty() {
```
bilboquet, line 618: maybe a bit simpler (please double check it does the same)
```rust
let prefix_range = if prefix.is_empty() {
    (std::ops::Bound::Unbounded, std::ops::Bound::Unbounded)
} else {
    let n_keep = prefix
        .iter()
        .enumerate()
        .rev()
        .find(|(_, v)| v >= &&255)
        .map(|(i, _)| i + 1)
        .unwrap_or(0);
    let mut prefix_end = prefix[..n_keep].to_vec();
    prefix_end.last_mut().map(|v| *v += 1);
    (
        std::ops::Bound::Included(prefix.to_vec()),
        prefix_end
            .is_empty()
            .then(|| std::ops::Bound::Unbounded)
            .unwrap_or(std::ops::Bound::Excluded(prefix_end)),
    )
};
```

massa-module-cache/Cargo.toml
bilboquet, line 21: ;)

massa-execution-worker/src/interface_impl.rs
Leo-Besancon: The zip operation yields None if the prefix.iter() yields None, so it only checks that the key and prefix are equal for the length of the prefix.

massa-async-pool/Cargo.toml
```diff
@@ -19,7 +19,7 @@ massa_serialization = { path = "../massa-serialization" }
 massa_signature = { path = "../massa-signature" }
 massa_db_exports = { path = "../massa-db-exports" }
 massa_time = { path = "../massa-time" }
-massa-proto-rs = { git = "https://github.com/massalabs/massa-proto-rs", rev = "18ec02f", features = ["tonic"] }
+massa-proto-rs = { git = "https://github.com/massalabs/massa-proto-rs", branch = "feature/Improve_ABI_types_in_wasmv1", features = ["tonic"] }
```
bilboquet, line 22: if you want to merge into testnet_24
you must point to a commit sha, not to a branch id that can change

massa-execution-worker/src/interface_impl.rs
Leo-Besancon: It is better indeed!

massa-async-pool/Cargo.toml
```diff
@@ -19,7 +19,7 @@ massa_serialization = { path = "../massa-serialization" }
 massa_signature = { path = "../massa-signature" }
 massa_db_exports = { path = "../massa-db-exports" }
 massa_time = { path = "../massa-time" }
-massa-proto-rs = { git = "https://github.com/massalabs/massa-proto-rs", rev = "18ec02f", features = ["tonic"] }
+massa-proto-rs = { git = "https://github.com/massalabs/massa-proto-rs", rev = "ba569c8", features = ["tonic"] }
```
bilboquet, line 22: I feel like we must merge massa-proto#feature/Improve_ABI_types_in_wasmv1 into massa-proto#main in order for this commit sha to be correct

PR #4177, opened 2023-06-30
Use Ratio instead of Amount for versioning threshold

sydhds:
  * [ ] make tests compile
  * [ ] make tests pass 

### DIFF
--- a/massa-models/src/config/constants.rs
+++ b/massa-models/src/config/constants.rs
@@ -85,10 +85,6 @@ lazy_static::lazy_static! {
         .parse()
         .unwrap()
     };
-
-    /// Versioning
-    /// Threshold to accept a new versioning
-    pub static ref VERSIONING_THRESHOLD_TRANSITION_ACCEPTED: Ratio<u64> = Ratio::new(75, 100);
 }
 
 /// Helper function to parse args for lazy_static evaluations
@@ -300,6 +296,8 @@ pub const MAX_LISTENERS_PER_PEER: u64 = 100;
 //
 // Constants used in versioning
 //
+/// Threshold to accept a new versioning
+pub const VERSIONING_THRESHOLD_TRANSITION_ACCEPTED: Ratio<u64> = Ratio::new_raw(75, 100);
 /// Block count to process in MipStoreStats (for state change threshold)
 pub const MIP_STORE_STATS_BLOCK_CONSIDERED: usize = 1000;
 
--- a/massa-models/src/config/constants.rs
+++ b/massa-models/src/config/constants.rs
@@ -86,6 +86,9 @@ lazy_static::lazy_static! {
         .unwrap()
     };
 
+    /// Versioning
+    /// Threshold to accept a new versioning
+    pub static ref VERSIONING_THRESHOLD_TRANSITION_ACCEPTED: Ratio<u64> = Ratio::new(75, 100);
 }
 
 /// Helper function to parse args for lazy_static evaluations
@@ -297,8 +300,6 @@ pub const MAX_LISTENERS_PER_PEER: u64 = 100;
 //
 // Constants used in versioning
 //
-/// Threshold to accept a new versioning
-pub const VERSIONING_THRESHOLD_TRANSITION_ACCEPTED: Amount = Amount::const_init(75, 0);
 /// Block count to process in MipStoreStats (for state change threshold)
 pub const MIP_STORE_STATS_BLOCK_CONSIDERED: usize = 1000;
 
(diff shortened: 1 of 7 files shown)


massa-versioning/src/versioning.rs
```diff
@@ -921,12 +922,14 @@ impl MipStoreRaw {
...
-                .unwrap_or(&0) as f32;
-            let block_count_considered = self.stats.config.block_count_considered as f32;
-
-            let vote_ratio_ = 100.0 * network_version_count / block_count_considered;
-
-            let vote_ratio = Amount::const_init(vote_ratio_.round() as u64, 0);
+                .unwrap_or(&0); // as f32;
+
+            let vote_ratio = Ratio::new(100, 1);
+            vote_ratio.mul(Ratio::new(network_version_count, 1));
```
Leo-Besancon, line 928: I would directly use the method `from_integer` to construct these. It probably is also possible to multiply and divided by an integer directly.

massa-serialization/src/lib.rs
```diff
@@ -392,3 +395,87 @@ impl Deserializer<bool> for BoolDeserializer {
         })(buffer)
     }
 }
+
+/// Serializer for Ratio
```
damip, line 399: why do we need to serialize ratios ? I don't think we should be sending them through the wire

massa-versioning/src/versioning.rs
```diff
@@ -921,12 +922,14 @@ impl MipStoreRaw {
...
                 .get(&mi.version)
-                .unwrap_or(&0) as f32;
-            let block_count_considered = self.stats.config.block_count_considered as f32;
-
-            let vote_ratio_ = 100.0 * network_version_count / block_count_considered;
-
-            let vote_ratio = Amount::const_init(vote_ratio_.round() as u64, 0);
+                .unwrap_or(&0); // as f32;
+
+            let vote_ratio = Ratio::new(100, 1);
```
damip, line 927: we should not be multiplying everything by 100 here, it is already a ratio

massa-versioning/src/versioning.rs
```diff
@@ -921,12 +922,14 @@ impl MipStoreRaw {
                 .stats
                 .network_version_counters
                 .get(&mi.version)
-                .unwrap_or(&0) as f32;
-            let block_count_considered = self.stats.config.block_count_considered as f32;
-
-            let vote_ratio_ = 100.0 * network_version_count / block_count_considered;
-
-            let vote_ratio = Amount::const_init(vote_ratio_.round() as u64, 0);
+                .unwrap_or(&0); // as f32;
```
damip, line 925: "as f32" should be removed, floats are not deterministic

massa-models/src/config/constants.rs
```diff
@@ -297,8 +300,6 @@ pub const MAX_LISTENERS_PER_PEER: u64 = 100;
 //
 // Constants used in versioning
 //
-/// Threshold to accept a new versioning
-pub const VERSIONING_THRESHOLD_TRANSITION_ACCEPTED: Amount = Amount::const_init(75, 0);
```
damip, line 301: use Ratio::new_raw which is const

massa-versioning/src/versioning.rs
sydhds: fixed

massa-versioning/src/versioning.rs
sydhds: fixed

massa-versioning/src/versioning.rs
sydhds: Old comment, removed

massa-versioning/src/versioning.rs
```diff
@@ -921,12 +922,14 @@ impl MipStoreRaw {
...
+
+            let vote_ratio = Ratio::new(100, 1);
+            vote_ratio.mul(Ratio::new(network_version_count, 1));
+            vote_ratio.div(Ratio::new(
+                u64::try_from(self.stats.config.block_count_considered).unwrap_or(0),
+                1,
+            ));
```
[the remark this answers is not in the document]
sydhds: Fixed

massa-models/src/config/constants.rs
sydhds: Done

Leo-Besancon (approved):
Much better now :)

massa-serialization/src/lib.rs
damip: Let's open a followup on this

sydhds: https://github.com/massalabs/massa/issues/4183

damip:
@sydhds there are merge conflicts

PR #4178, opened 2023-06-30
Harden block handler (Versioning)

sydhds:
  * [ ] make tests compile
  * [ ] make tests pass 


massa-protocol-worker/src/handlers/block_handler/retrieval.rs
```diff
@@ -463,6 +463,23 @@ impl RetrievalThread {
                 received: header.content.current_version,
             })
         } else {
+            // announced version == 0 -> No announce
+            if header.content.announced_version != 0 {
+                if header.content.announced_version <= version {
```
Eitu33, line 468:
```suggestion
                if header.content.announced_version < version {
```

Can't the announced version be the same as the current version? I'm pretty sure it can be as long as no new version has been pushed, so probably `<` instead of `<=`.

massa-protocol-worker/src/handlers/block_handler/retrieval.rs
```diff
@@ -463,6 +463,23 @@ impl RetrievalThread {
...
+                        local: version,
+                        announced_received: header.content.announced_version,
+                    });
+                }
+
+                let version_to_announce = self.mip_store.get_network_version_to_announce();
+                if header.content.announced_version != version_to_announce {
+                    warn!("Please update your node if you want to support this update (network version: {})", 
+                        header.content.announced_version);
+                }
```
Eitu33, line 480: Not a fan of putting out a warning for version support in the logs of the node that receives an outdated header...

massa-protocol-worker/src/handlers/block_handler/retrieval.rs
damip: I'm not a fan of special cases that need to be remembered "remember that announcing 0 means that there is no announcement" because it **will** cause errors. Indeed we should either announce the same version as the current version if we don't want to announce, or make the announcement an Option and set it to None

sydhds: It should not be allowed to have network versions like (1, 1). You can either have:
* (0, 1) -> announced a new network version: 1
* (1, 0) -> Version 1 is the current network version, 0 means nothing to announce
* (1, 2) -> Version 1 is the current network version, 2 is the announced network version

massa-protocol-worker/src/handlers/block_handler/retrieval.rs
Eitu33: But here you're using `!=` which means that the receiving node can be up to date, receive an outdated header, and still ask the up to date node to upgrade for support. What I'm saying is have a `header.content.announced_version > version_to_announce` check, and for the case where `header.content.announced_version < version_to_announce` I imagine the creator node will eventually get the warning when an up to date node sends him a block.

massa-protocol-worker/src/handlers/block_handler/retrieval.rs
Eitu33: I don't really mind about the solution we choose as long as its a logical behaviour

massa-protocol-worker/src/handlers/block_handler/retrieval.rs
sydhds: fixed

Eitu33 (approved):

massa-protocol-worker/src/handlers/block_handler/retrieval.rs
Eitu33: I agree with the implementation and will resolve this, if any of the other participants want to reopen this discussion please feel free

massa-versioning/src/versioning.rs
```diff
@@ -878,9 +873,10 @@ impl MipStoreRaw {
     fn update_network_version_stats(
         &mut self,
         slot_timestamp: MassaTime,
-        network_versions: Option<(u32, u32)>,
+        network_versions: Option<(u32, Option<u32>)>,
     ) {
-        if let Some((_current_network_version, announced_network_version)) = network_versions {
+        if let Some((_current_network_version, Some(announced_network_version))) = network_versions
+        {
```
Leo-Besancon, line 879: I'm going a bit fast on my review, but if announced_network_version is None, I don't think you're updating the stats? We still want to!

Leo-Besancon (changes requested):

massa-versioning/src/versioning.rs
sydhds: Done

Leo-Besancon (approved):

massa-protocol-worker/src/handlers/block_handler/retrieval.rs
damip: One more comment: we must not tell the node runner to update their node every time they receive a block announcing something unknown. The warning should only appear following the counters: eg. if more than 30% of the network is announcing something we don't know, then warn the user. Otherwise attackers can cause a lot of noise

massa-protocol-worker/src/handlers/block_handler/retrieval.rs
```diff
@@ -456,13 +456,39 @@ impl RetrievalThread {
...
+                        local: current_version,
+                        announced_received: announced_version,
+                    });
+                }
+
+                match self.mip_store.get_network_version_to_announce() {
+                    Some(version_to_announce) => {
+                        // Unknown/new announced version -> user might need to upgrade
+                        if announced_version > version_to_announce {
+                            warn!("Please update your node if you want to support this update (new network version: {})", announced_version);
```
damip, line 480: we must not tell the node runner to update their node every time they receive a block announcing something unknown. The warning should only appear following the counters: eg. if more than 30% of the network is announcing something we don't know, then warn the user. Otherwise attackers can cause a lot of noise

massa-versioning/src/versioning.rs
```diff
@@ -908,11 +908,32 @@ impl MipStoreRaw {
...
                     let entry_value = e.get_mut();
                     *entry_value = entry_value.saturating_sub(1);
+                    network_version_count = *entry_value;
                     if *entry_value == 0 {
                         self.stats.network_version_counters.remove(&removed_version);
                     }
                 }
             }
+
+            if announced_network_version != 0 {
```
damip, line 918: didn't we remove the "0" special case ?

massa-versioning/src/versioning.rs
```diff
@@ -908,11 +908,32 @@ impl MipStoreRaw {
...
+                let vote_ratio = Ratio::new(
+                    network_version_count,
+                    self.stats.config.block_count_considered as u64,
+                );
+
+                if vote_ratio > MIP_STATS_RATIO_WARN_ANNOUNCED_VERSION {
```
damip, line 924: `MIP_STATS_RATIO_WARN_ANNOUNCED_VERSION` should be a config file parameter and not a constant

massa-versioning/src/versioning.rs
```diff
@@ -908,11 +908,32 @@ impl MipStoreRaw {
...
+                    let last_key_value = self.store.last_key_value();
+                    if let Some((mi, _ms)) = last_key_value {
+                        if announced_network_version > mi.version {
+                            // Vote ratio is > 30%
+                            // announced version is not known (not in MIP store)
+                            // announced version is > to the last known network version in MIP store
+                            // -> Warn the user to update
+                            warn!("Please update your node if you want to support this update (new network version: {})", announced_network_version);
```
damip, line 932: we can make the message a bit more informative: `XXX our of YYY last blocks advertised that they are willing to transition to version ZZZ. You should update your node if you wish to move to that version.`

Leo-Besancon (approved):
I'm good with this version.

massa-versioning/src/versioning.rs
```diff
@@ -986,6 +1007,8 @@ impl MipStoreRaw {
         version
     }
 
+    // fn contains_network_version(&self, current_network_version: u32, announced)
```
damip, line 1010: remove comment

massa-versioning/src/versioning.rs
sydhds: We did in block header but we use it in mip store stats, when we received None we process it like it is nerwork version 0, this simplifies the code

damip: what if someone announces Some(0) ?

Leo-Besancon: Some(0) should be semantically the same as announcing None, because we can never roll back versions. (We can always introduce a new version that rolls back previous changes, but we can't go from active version of 2 to 0.

massa-versioning/src/versioning.rs
sydhds: Done

massa-versioning/src/versioning.rs
sydhds: Done

massa-versioning/src/versioning.rs
sydhds: Done

damip (approved):
LGTM

Leo-Besancon (approved):

PR #4188, opened 2023-06-30
make state fingerprint available

damip:
  * [x] make tests compile
  * [x] make tests pass 

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2822,6 +2822,7 @@ dependencies = [
  "massa_db_exports",
  "massa_db_worker",
  "massa_executed_ops",
+ "massa_hash",
  "massa_ledger_exports",
  "massa_ledger_worker",
  "massa_models",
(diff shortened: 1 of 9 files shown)


Leo-Besancon:
The code looks pretty straightforward to me. However, we have to be careful about timings. Because we do not have a "final_state_hash" field on the final_state, two nodes may show a different fingerprint if atomicity is not verified during a slot.
I think it should not cause a problem in most cases, but it could.

For example: we edit the final_state_hash_snapshot of a cycle after finalizing a slot, which changes the hash.

damip:
@Leo-Besancon writes to the final state happen only under an execution write lock. And query_state read-locks it. So it should be ok. Do you confirm ?

Leo-Besancon:
> @Leo-Besancon writes to the final state happen only under an execution write lock. And query_state read-locks it. So it should be ok. Do you confirm ?

Yes it makes sence, as the versioning batches are also written on finalization now. Maybe we should add a comment in the fingerprint function that this is the expected behaviour for it to work (only one write, with a lock, per slot).

Leo-Besancon (approved):

PR #4193, opened 2023-07-03
Use new version of PeerNet with the `stream_limiter` set

litchipi:
Uses the changes from https://github.com/massalabs/PeerNet/pull/55 inside the `massa_protocol_{worker,exports}`

  * [ ] make tests compile
  * [ ] make tests pass 

### DIFF
--- a/massa-node/src/operation_injector.rs
+++ b/massa-node/src/operation_injector.rs
@@ -83,6 +83,7 @@ pub fn start_operation_injector(
         use rand::Rng;
         let mut rng = rand::thread_rng();
         loop {
+            let now = std::time::Instant::now();
             let mut storage = storage.clone_without_refs();
             let txps = nb_op / 32;
             let final_slot = get_closest_slot_to_timestamp(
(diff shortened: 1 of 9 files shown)


massa-node/src/operation_injector.rs
```diff
@@ -113,7 +114,7 @@ pub fn start_operation_injector(
             protocol_controller
                 .propagate_operations(storage.clone())
                 .unwrap();
-            std::thread::sleep(Duration::from_secs(1));
+            std::thread::sleep(Duration::from_secs(1).saturating_sub(now.elapsed()));
```
litchipi, line 117: This is a fix for the operation injector, I'll take it out of this PR and create a new one especially for this

AurelienFT: You can let it there it's ok

litchipi: PR: #4194

Leo-Besancon (approved):
Congrats!

AurelienFT (approved):

litchipi:
I just fixed tests in `massa-bootstrap`. Because some `Limiter` were created in advance, they had the time to stack a bit of tokens before the read, so it took less time than it was expected.
To solve this, I added a `std::thread::sleep` to stack to the maximum the Limiters, then assert they take longer than expected (but lower the expectation as we stacked up)

Issue #3728, opened 2023-03-27
Remove FinalState::reset() (drop the object and create a new one)


bilboquet:
We want to recreate everything on bootstrap (with reset we may forget something)

_Originally posted by @damip in https://github.com/massalabs/massa/pull/3489#discussion_r1143561709_
            
Some comments are handled in #3698
For FinalState is a bit more tricky, hence this followup issue

damip:
Also applies to ExecutedDenunciations

Ben-PH:
Original comment: https://github.com/massalabs/massa/pull/3489#discussion_r1141961674

PR #4197, opened 2023-07-03
incremental hash

damip:
## Todo
* [x] define the trail hash
* [x] take into account the trail hash in Execution outputs
* [x] use the trail hash for PRNG seeding in Execution
* [x] write the trail hash to the final state (compatible with bootstrap, included in XOF hash)
* [x] read the trail hash from the final state
* [x] initialize the trail hash in the final state to Hash::zero() on genesis

## Checklist

  * [x] make tests compile
  * [ ] make tests pass 


massa-final-state/src/final_state.rs
```diff
@@ -25,6 +27,10 @@ use tracing::{debug, info, warn};
 use massa_models::config::PERIODS_BETWEEN_BACKUPS;
 use massa_models::timeslots::get_block_slot_timestamp;
 
+// TODO when creating a final ledger at genesis, initialize the execution hash trail to Hash::zero()
+
+// TODO after bootstrap, check that the execution hash trail is present and in correct format
```
Leo-Besancon, line 32: This would be done in the is_db_valid() method at the end of this file.
If the db iterator encounters your added (serialized_key, serialized_value), it should not return false.
(And you can also add a separate check that the key exists in the DB, although I never check this currently).

Leo-Besancon (commented):
Another thing to handle is the reset of the final state.
You should delete your prefix at two places:
FinalState::reset()
FinalState::new(), if the reset_final_state argument is set to  true

You can call the .delete_prefix() function on the db to do that.

massa-final-state/src/final_state.rs
damip: I think that things could be simplified a bit and better encapsulated:
* remove the is_db_valid function completely
* after the bootstrap of the db (or its initialization or loading from disk) is done, create final_state by passing it the db and letting it initialize all its subsystems in its constructor: 
  * pos_state = PosState::try_from_db(&db).map_err(||e| format!("could not init PosState in FinalState: {}"), e)?;
  * async_pool = AsyncPool::try_from_db(&db).map_err(||e| format!("could not AsyncPool state in FinalState: {}"), e)?;
  * ... each of those constructors initializes the subsystem but also checks the db validity on the prefix subset(s) they manage

That way we have each module take responsibility for checking its own parts of the ledger properly and we reduce the error surface when adding new elements to the db (eg. forgetting to add them to is_db_valid)

damip:
> Another thing to handle is the reset of the final state. You should delete your prefix at two places: FinalState::reset() FinalState::new(), if the reset_final_state argument is set to true
> 
> You can call the .delete_prefix() function on the db to do that.

okay :)

Improvement: remove those concepts completely by unifying https://github.com/massalabs/massa/issues/3728 and https://github.com/massalabs/massa/issues/4234 => final_state is initialized from a db, and when re-attempting a bootstrap, the final_state object is just re-created

massa-final-state/src/final_state.rs
Leo-Besancon: I agree things could be improved!
I didn't want to change too much things at one time, but I agree avoiding initialization of the final state before the bootstrap would be a better workflow.

Here is the issue: https://github.com/massalabs/massa/issues/4234#issue-1794866153

massa-final-state/src/final_state.rs
```diff
@@ -25,6 +27,10 @@ use tracing::{debug, info, warn};
 use massa_models::config::PERIODS_BETWEEN_BACKUPS;
 use massa_models::timeslots::get_block_slot_timestamp;
 
+// TODO when creating a final ledger at genesis, initialize the execution hash trail to Hash::zero()
```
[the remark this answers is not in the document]
damip: We should use a tool to generate an initial DB on disk and just let the system load it from there, like it would for a restart to avoid having this weird special case. Can we open a followup ?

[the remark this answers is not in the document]
damip: > The easiest to integrate with your current code is IMO to do it directly in the load_initial_ledger() method in ledger_db.rs : just add the key/value to the batch there.

Until we do the followup I mentioned, is there a way to do it without contaminating either low level DB or the unrelated Ledger module ?

massa-final-state/src/final_state.rs
damip: Thanks https://github.com/massalabs/massa/issues/4234

massa-final-state/src/final_state.rs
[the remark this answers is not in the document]
Leo-Besancon: You could call another function here in massa-bootstrap>src>client.rs, that handles this, but it would have to be in final_state or in the db.
```rust
if !bootstrap_config.keep_ledger {
    // load ledger from initial ledger file
    final_state_guard
        .ledger
        .load_initial_ledger()
        .map_err(|err| {
            BootstrapError::GeneralError(format!(
            "could not load initial ledger: {}",
            err
        ))
    })?;
}
```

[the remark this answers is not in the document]
Leo-Besancon: > We should use a tool to generate an initial DB on disk and just let the system load it from there, like it would for a restart to avoid having this weird special case. Can we open a followup ?

https://github.com/massalabs/massa/issues/4237

massa-final-state/src/final_state.rs
```diff
@@ -117,6 +121,10 @@ impl FinalState {
         };
 
         if reset_final_state {
+            // delete the execution trail hash
+            self.db
```
Leo-Besancon, line 125: final_state.db

massa-final-state/src/final_state.rs
```diff
@@ -666,23 +710,24 @@ impl FinalState {
...
-                && !serialized_key.starts_with(EXECUTED_DENUNCIATIONS_PREFIX.as_bytes())
-                && !serialized_key.starts_with(LEDGER_PREFIX.as_bytes())
-                && !serialized_key.starts_with(MIP_STORE_PREFIX.as_bytes())
-            {
-                warn!(
-                    "Key/value does not correspond to any prefix: serialized_key: {:?}, serialized_value: {:?}",
-                    serialized_key, serialized_value
-                );
+            let execution_trail_hash_serialized: [u8] =
+                match db.get(EXECUTION_TRAIL_HASH_PREFIX.as_bytes()) {
```
Leo-Besancon, line 716: Should be ok with get_cf() and a call to .expect() to unwrap potential DB read errors.

massa-final-state/src/final_state.rs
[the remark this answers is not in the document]
damip: > You could call another function here in massa-bootstrap>src>client.rs, that handles this, but it would have to be in final_state or in the db.
> 
> ```rust
> if !bootstrap_config.keep_ledger {
>     // load ledger from initial ledger file
>     final_state_guard
>         .ledger
>         .load_initial_ledger()
>         .map_err(|err| {
>             BootstrapError::GeneralError(format!(
>             "could not load initial ledger: {}",
>             err
>         ))
>     })?;
> }
> ```

Thanks for the feedback. Checking it out.

I guess that keep_ledger should be renamed keep_state (?)

A general observation: without having worked on it previously, it is essentially impossible to add something to the state without making many more or less silent mistakes. I hope that the refactorings will help make it less error-prone :)

massa-final-state/src/final_state.rs
damip: corrected

massa-final-state/src/final_state.rs
damip: corrected, without expect to really have a true is_valid function :)

massa-final-state/src/final_state.rs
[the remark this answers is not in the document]
Leo-Besancon: Sure, keep_ledger could probably become keep_state or keep_db.

Aside from the refactor, I think we could improve the documentation too.

damip:
@Leo-Besancon the bootstrap test fails. Any idea ?

Leo-Besancon (approved):
Looks like everything is taken care of.
Sidenote: We must not forget to init the execution trail hash in the db if we want to convert from a previous version (e.g. if we want to keep the testnet24 db for testnet25, or when converting the buildnet ledger)

massa-hash/src/hash.rs
```diff
@@ -70,6 +70,11 @@ impl std::fmt::Debug for Hash {
 }
 
 impl Hash {
+    /// Creates a hash full of zeros bytes.
+    pub fn zero() -> Self {
+        Hash(blake3::Hash::from([0; HASH_SIZE_BYTES]))
```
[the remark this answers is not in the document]
damip: it's fundamentally different: [000000000000000] vs hash([00000000000000000000])

damip:
thanks @Leo-Besancon for the review and corrections !

PR #4199, opened 2023-07-03
NativeAmount arithmetic implementation

bilboquet:
  * [X] make tests compile
  * [X] make tests pass 

### DIFF
--- a/massa-execution-worker/src/interface_impl.rs
+++ b/massa-execution-worker/src/interface_impl.rs
@@ -142,25 +142,16 @@ impl InterfaceClone for InterfaceImpl {
 
 /// Helper function that creates an amount from a NativeAmount
 fn amount_from_native_amount(amount: &NativeAmount) -> Result<Amount> {
-    let mantissa = amount
-        .mandatory_mantissa
-        .ok_or_else(|| anyhow!("Missing mantissa".to_string()))?;
-    let scale = amount
-        .mandatory_scale
-        .ok_or_else(|| anyhow!("Missing scale".to_string()))?;
-    let amount =
-        Amount::from_mantissa_scale(mantissa, scale).map_err(|err| anyhow!(format!("{}", err)))?;
+    let amount = Amount::from_mantissa_scale(amount.mantissa, amount.scale)
+        .map_err(|err| anyhow!(format!("{}", err)))?;
 
     Ok(amount)
 }
 
 /// Helper function that creates a NativeAmount from the amount internal representation
-fn amount_to_native_amount(amount: &Amount) -> Result<NativeAmount> {
+fn amount_to_native_amount(amount: &Amount) -> NativeAmount {
     let (mantissa, scale) = amount.to_mantissa_scale();
-    Ok(NativeAmount {
-        mandatory_mantissa: Some(mantissa),
-        mandatory_scale: Some(scale),
-    })
+    NativeAmount { mantissa, scale }
 }
 
 /// Helper function to get the address from the option given as argument to some ABIs
@@ -316,7 +307,7 @@ impl Interface for InterfaceImpl {
     fn get_balance_wasmv1(
         &self,
         address: Option<String>,
-    ) -> Result<massa_proto_rs::massa::model::v1::NativeAmount> {
+    ) -> Result<NativeAmount> {
         let context = context_guard!(self);
         let address = get_address_from_opt_or_context(&context, address)?;
 
@@ -921,14 +912,7 @@ impl Interface for InterfaceImpl {
         from_address: Option<String>,
     ) -> Result<()> {
         let to_address = Address::from_str(&to_address)?;
-        let amount = Amount::from_mantissa_scale(
-            raw_amount
-                .mandatory_mantissa
-                .ok_or(anyhow!("No mantissa provided"))?,
-            raw_amount
-                .mandatory_scale
-                .ok_or(anyhow!("No scale provided"))?,
-        )?;
+        let amount = Amount::from_mantissa_scale(raw_amount.mantissa, raw_amount.scale)?;
         let mut context = context_guard!(self);
         let from_address = match from_address {
             Some(from_address) => Address::from_str(&from_address)?,
@@ -1203,9 +1187,7 @@ impl Interface for InterfaceImpl {
     /// Returns a NativeAmount from a string
     fn native_amount_from_str_wasmv1(&self, amount: &str) -> Result<NativeAmount> {
         let amount = Amount::from_str(amount).map_err(|err| anyhow!(format!("{}", err)))?;
-        let amount = amount_to_native_amount(&amount).map_err(|err| anyhow!(format!("{}", err)))?;
-
-        Ok(amount)
+        Ok(amount_to_native_amount(&amount))
     }
 
     /// Returns a string from a NativeAmount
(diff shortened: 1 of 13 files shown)


bilboquet:
I'm going to add tests, maybe you wanna have a look during this time et maybe we can merge this code so we car run some SC in sandbox quickly and tests will come before the merge into main/testnet_24

massa-models/src/amount.rs
```diff
@@ -106,6 +107,25 @@ impl Amount {
         Amount::from_decimal(res)
     }
 
+    /// Creates an amount from a NativeAmount
+    pub fn from_native_amount(amount: &NativeAmount) -> Result<Self, ModelsError> {
```
damip, line 111: To avoid the low-level Amount to depend on proto-generated stuff, I would recommend doing the conversion through mantissa/scale where it is necessary

massa-execution-worker/src/interface_impl.rs
```diff
@@ -143,6 +143,29 @@ impl InterfaceClone for InterfaceImpl {
     }
 }
 
+/// Helper function that creates an amount from a NativeAmount
+fn from_native_amount(amount: &NativeAmount) -> Result<Amount> {
```
damip, line 147: let's call it amount_from_native_amount :)

massa-execution-worker/src/interface_impl.rs
```diff
@@ -143,6 +143,29 @@ impl InterfaceClone for InterfaceImpl {
...
+        .mandatory_scale
+        .ok_or_else(|| anyhow!("Missing scale".to_string()))?;
+    let amount =
+        Amount::from_mantissa_scale(mantissa, scale).map_err(|err| anyhow!(format!("{}", err)))?;
+
+    Ok(amount)
+}
+
+/// Helper function that creates a NativeAmount from the amount internal representation
+fn to_native_amount(amount: &Amount) -> Result<NativeAmount> {
```
damip, line 161: amount_to_native_amount

massa-models/src/amount.rs
bilboquet: done

massa-execution-worker/src/interface_impl.rs
bilboquet: done

massa-execution-worker/src/interface_impl.rs
bilboquet: done

massa-models/src/amount.rs
```diff
@@ -106,6 +108,11 @@ impl Amount {
         Amount::from_decimal(res)
     }
 
+    /// Return the representation of the amount as a mantissa and a scale.
+    pub fn to_mantissa_scale(&self) -> (u64, u32) {
```
bilboquet, line 112: @aoudiamoncef 
I also needed this function so i copied/pasted what you did in the issue.
I'm not sure about the file where you wanted to put the constants ?

Leo-Besancon (approved):
I'm not sure I understand 100% the logic of the two div functions, but otherwise it seems like a good base to me!

bilboquet:
> I'm not sure I understand 100% the logic of the two div functions, but otherwise it seems like a good base to me!

To be honest it's a bit of a "best effort" for the moment.
The use case for 2 operations is legit, one may want to split some amounts in X (hence the div by a u64), and one may want to know how many times some amounts are in some other amount (hence the div by a native amount).
I'm not that proud of the current implementation… I hope to clean it later

Leo-Besancon:
> > I'm not sure I understand 100% the logic of the two div functions, but otherwise it seems like a good base to me!
> 
> To be honest it's a bit of a "best effort" for the moment. The use case for 2 operations is legit, one may want to split some amounts in X (hence the div by a u64), and one may want to know how many times some amounts are in some other amount (hence the div by a native amount). I'm not that proud of the current implementation… I hope to clean it later

What I don't understand is how the quotient could be an "Amount". For me the quotient would be a u64 in both cases. But you probably thought about this more than me :)

massa-execution-worker/src/interface_impl.rs
```diff
@@ -1175,6 +1198,131 @@ impl Interface for InterfaceImpl {
...
+    fn mul_native_amount_wasmv1(&self, amount: &NativeAmount, factor: u64) -> Result<NativeAmount> {
+        let amount = amount_from_native_amount(amount)?;
+        let mul = amount.saturating_mul_u64(factor);
+        let mul = amount_to_native_amount(&mul)
+            .map_err(|err| anyhow!(format!("Couldn't mul native amount: {}", err)))?;
+        Ok(mul)
+    }
+
+    /// Divides a native amount by a divisor, return an error if the divisor is 0.
+    fn div_rem_native_amount_wasmv1(
```
damip, line 1265: I think it would be better to add this to the Amount structure:

```rust
pub fn checked_rem(&self, other: &Amount) -> Option<Amount> {
    Amount(self.0.checked_rem(other.0)?)
}

pub fn checked_rem_u64(&self, other: u64) -> Option<Amount> {
    Amount(self.0.checked_rem(other)?)
}

```

massa-execution-worker/src/interface_impl.rs
```diff
@@ -143,6 +143,29 @@ impl InterfaceClone for InterfaceImpl {
...
+        .mandatory_scale
+        .ok_or_else(|| anyhow!("Missing scale".to_string()))?;
+    let amount =
+        Amount::from_mantissa_scale(mantissa, scale).map_err(|err| anyhow!(format!("{}", err)))?;
+
+    Ok(amount)
+}
+
+/// Helper function that creates a NativeAmount from the amount internal representation
+fn amount_to_native_amount(amount: &Amount) -> Result<NativeAmount> {
```
damip, line 161: this shouldn't return a Result: it cannot fail

massa-execution-worker/src/interface_impl.rs
bilboquet: many thanks, will use it after I've synced with some dependencies

massa-execution-worker/src/interface_impl.rs
```diff
@@ -143,6 +143,29 @@ impl InterfaceClone for InterfaceImpl {
...
+    let (mantissa, scale) = amount.to_mantissa_scale();
+    Ok(NativeAmount {
+        mandatory_mantissa: Some(mantissa),
+        mandatory_scale: Some(scale),
```
damip, line 165: todo remove "mandatory"

bilboquet:
> > > I'm not sure I understand 100% the logic of the two div functions, but otherwise it seems like a good base to me!
> > 
> > 
> > To be honest it's a bit of a "best effort" for the moment. The use case for 2 operations is legit, one may want to split some amounts in X (hence the div by a u64), and one may want to know how many times some amounts are in some other amount (hence the div by a native amount). I'm not that proud of the current implementation… I hope to clean it later
> 
> What I don't understand is how the quotient could be an "Amount". For me the quotient would be a u64 in both cases. But you probably thought about this more than me :)

if you have an amount, you split it into two part, you have 2 amounts, right?

massa-execution-worker/src/interface_impl.rs
bilboquet: done

Leo-Besancon:
> > > > I'm not sure I understand 100% the logic of the two div functions, but otherwise it seems like a good base to me!
> > > 
> > > 
> > > To be honest it's a bit of a "best effort" for the moment. The use case for 2 operations is legit, one may want to split some amounts in X (hence the div by a u64), and one may want to know how many times some amounts are in some other amount (hence the div by a native amount). I'm not that proud of the current implementation… I hope to clean it later
> > 
> > 
> > What I don't understand is how the quotient could be an "Amount". For me the quotient would be a u64 in both cases. But you probably thought about this more than me :)
> 
> if you have an amount, you split it into two part, you have 2 amounts, right?

Ok I tried on an example and I get it now :)

massa-execution-worker/src/interface_impl.rs
bilboquet: done

Leo-Besancon (approved):

PR #4208, opened 2023-07-04
Add more unit test for Mip store update_with method

sydhds:
  * [ ] make tests compile
  * [ ] make tests pass 

### DIFF
--- a/massa-versioning/src/versioning.rs
+++ b/massa-versioning/src/versioning.rs
@@ -1897,11 +1897,27 @@ mod test {
                 MipStoreRaw::try_from(([(mi_1.clone(), ms_1.clone())], mip_stats_cfg.clone()))
                     .unwrap();
             let mut mi_2_2 = mi_2.clone();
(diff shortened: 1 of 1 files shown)


massa-versioning/src/versioning.rs
```diff
@@ -1793,104 +1793,164 @@ mod test {
...
 
             // MIP-0003 in vs_raw_1 & vs_raw_2 has != components
             assert_matches!(
-                vs_raw_1.update_with(&vs_raw_2),
+                store_1.update_with(&store_2),
+                Err(UpdateWithError::Overlapping(..))
+            );
+        }
+
+        // case 3: trying to downgrade network version
```
Leo-Besancon, line 1894: You test with equal versions, so the "downgrade" comment is not perfect

Leo-Besancon (approved):
Nice cleanup and tests

massa-versioning/src/versioning.rs
sydhds: Done

PR #4211, opened 2023-07-04
Homogenize state machine transition condition

sydhds:
  * [ ] make tests compile
  * [ ] make tests pass 


Leo-Besancon (approved):
Is it the only place you found a different condition?

PR #4213, opened 2023-07-04
Feature/advance lw

sydhds:
  * [ ] make tests compile
  * [ ] make tests pass 


massa-versioning/src/versioning.rs
```diff
@@ -179,30 +179,55 @@ pub struct Advance {
...
+            && self.now == other.now
+            && self.activation_delay == other.activation_delay
+    }
+}
+
+impl Eq for Advance {}
+
+// A Lightweight version of 'Advance' (used in MipState history)
+#[derive(Clone, Debug)]
+pub struct AdvanceLW {
```
Leo-Besancon, line 196: It's a bit heavy to create a new struct just for this cleanup IMO. If you can find a way to just use the now it would be great ;)

sydhds: Try to use only now but it breaks a lot of things. I think this is the best solution AFAIK

Leo-Besancon (approved):

PR #4214, opened 2023-07-05
Check for min activation delay in MIP info

sydhds:
  * [ ] make tests compile
  * [ ] make tests pass 

### DIFF
--- a/massa-versioning/src/lib.rs
+++ b/massa-versioning/src/lib.rs
@@ -29,7 +29,7 @@
 //!
 //! At the slot when the activation happens, need to make sure that everyone knows it should happen,
 //! so we need to make sure that everyone has seen as final the slot that triggered the locked-in state,
-//! and the worst-case delay required for a slot to become is the definition of a cycle.
+//! and the worst-case delay required for a slot to become final is the definition of a cycle.
 //!
 //! The activation delay counts how long we wait to activate after the vote threshold was reached
 //! and we entered into locked-in state. During that delay, and after it, the number of blocks considered


Leo-Besancon (approved):

massa-models/src/config/constants.rs
```diff
@@ -300,6 +300,8 @@ pub const MAX_LISTENERS_PER_PEER: u64 = 100;
 pub const VERSIONING_THRESHOLD_TRANSITION_ACCEPTED: Ratio<u64> = Ratio::new_raw(75, 100);
 /// Block count to process in MipStoreStats (for state change threshold)
 pub const MIP_STORE_STATS_BLOCK_CONSIDERED: usize = 1000;
+/// Minimum value allowed for activation delay (in MIP info)
+pub const VERSIONING_ACTIVATION_DELAY_MIN: MassaTime = T0.saturating_mul(PERIODS_PER_CYCLE);
```
Eitu33, line 304: Any particular reason for the min activation delay being 1 cycle? Also seeing `MIP_STORE_STATS_BLOCK_CONSIDERED` reminds me that we should probably be checking the coherence between the amount of block we consider for a version change and time given for update in the MIP. So we can be sure to never receive an MIP with a voting period that's below 1000 blocks * 0.5s (actually a bit more because its blocks not slots)

damip: > Any particular reason for the min activation delay being 1 cycle?

At the slot when the activation happens, need to make sure that everyone knows it should happen, so we need to make sure that everyone has seen as final the slot that triggered the locked-in state, and the worst-case delay required for a slot to become is the definition of a cycle.

The activation delay counts how long we wait to activate after the vote threshold was reached and we entered into locked-in state. During that delay, and after it, the number of blocks considered for the vote is not relevant. The only reason why we should consider a sufficient number of votes is to get a reasonable p-value on the vote itself:
- if we consider only 5 votes, the probability that a 30%-stake was selected to vote at least 75% of the times is 3%: https://www.wolframalpha.com/input?i=Probability%5Bx+%3E%3D+4%2C+x+BinomialDistribution%5B5%2C+0.3%5D%5D
- if we consider 1000 votes that probability falls to 1e-149: https://www.wolframalpha.com/input?i=Probability%5Bx+%3E%3D+750%2C+x+BinomialDistribution%5B1000%2C+0.3%5D%5D

As for the vote timeout delay, if it is set too low, it will always timeout and never become active, but that's ok, it's a choice of the creator of the MIP, not an error per se.

massa-versioning/src/lib.rs
```diff
@@ -25,6 +25,20 @@
 //! In order to have all nodes in sync (because of node various delays), the state is set to active
 //! after an activation delay (duration is required to be > 1 cycle).
 //!
+//! About activation delay:
+//!
+//! At the slot when the activation happens, need to make sure that everyone knows it should happen,
+//! so we need to make sure that everyone has seen as final the slot that triggered the locked-in state,
+//! and the worst-case delay required for a slot to become is the definition of a cycle.
```
Eitu33, line 32:
```suggestion
//! and the worst-case delay required for a slot to become final is the definition of a cycle.
```

Eitu33 (approved):

PR #4215, opened 2023-07-05
Lower numbers of in connections in config

AurelienFT:


damip (approved):
good catch

AurelienFT:
bors merge

PR #4223, opened 2023-07-05
Implement bs58 to/from bytes

bilboquet:
Implement bs58 to/from bytes
Implement comparison functions

*[ ] document all added functions
  * [ ] make tests compile
  * [ ] make tests pass 

### DIFF
--- a/massa-execution-worker/src/interface_impl.rs
+++ b/massa-execution-worker/src/interface_impl.rs
@@ -325,6 +325,7 @@ impl Interface for InterfaceImpl {
     /// The raw representation (no decimal factor) of the balance of the address,
     /// or zero if the address is not found in the ledger.
     fn get_balance_wasmv1(&self, address: Option<String>) -> Result<NativeAmount> {
+    fn get_balance_wasmv1(&self, address: Option<String>) -> Result<NativeAmount> {
         let context = context_guard!(self);
         let address = get_address_from_opt_or_context(&context, address)?;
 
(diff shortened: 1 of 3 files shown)


massa-execution-worker/src/interface_impl.rs
```diff
@@ -1435,6 +1436,79 @@ impl Interface for InterfaceImpl {
...
+        };
+
+        Ok(res)
+    }
+
+    fn compare_sig_wasmv1(&self, left: &str, right: &str) -> Result<ComparisonResult> {
+        let left = Signature::from_str(left)?;
+        let right = Signature::from_str(right)?;
+        todo!()
+        // let res = match left.cmp(&right) {
```
bilboquet, line 1504: @damip 
Our Signature does not have comparison function and the underlying ed25529_dalek::Signature does not implement Ord or PartilOrd.
So I'm wondering how to implement comparison, any tips?

damip: Hey indeed I don't think we need order nor comparison on Signature for now !

bilboquet: fine, I'm gonna drop this abi :)

Leo-Besancon (approved):
Approving the other implementations!

massa-execution-worker/src/interface_impl.rs
bilboquet: done

PR #4239, opened 2023-07-10
Feature/ledger editor versioning 1

sydhds:
  * [ ] make tests compile
  * [ ] make tests pass 


Leo-Besancon (approved):
I don't know where, but can you update / produce some brief documentation about how the MIP_Store is handled for initialization (interactions with network restart / bootstrap / get_mip_list())?

Eitu33 (approved):

PR #4243, opened 2023-07-11
Protocol improvements

modship:
- We need to stop iteration if the peer was disconnected when sending msg

- ~~`network_controller.try_connect` use only `peer_db_read.index_by_newest` but we should add TRUSTED peer of `peer_db_read.peers` (filter on state and peer which is not actually connected)~~

:warning: most connection attempts return handshake timeout error when trying to get message length :warning:

⚠️  ~~after a while sometimes i have :  `Failed to send AskForOperations message to peer...`~~ ⚠️   #4248 

Peernet fix : #64

Peer priorisation : #4260


modship:
When a remote peer tries to connect, it waits for a response for 5 seconds (default). 
On our side, we send the connection request from the peer to the tester thread via a channel.

```rust
            PeerManagementMessage::NewPeerConnected((peer_id, listeners)) => {
                                    debug!("Received peer message: NewPeerConnected from {}", peer_id);
                                    if let Err(e) = test_sender.try_send((peer_id, listeners)) {
                                        debug!("error when sending msg to peer tester : {}", e);
                                    }
                                }
```
The issue is that the messages are not being processed quickly enough to respond to the remote peer within a reasonable time frame.

Once the channel is full, the incoming connection messages are not being processed.

Perhaps we should consider splitting the "tester"  and the incoming connection.

massa-protocol-worker/src/connectivity.rs
```diff
@@ -259,59 +261,82 @@ pub(crate) fn start_connectivity_thread(
...
-                                }
-                                if let Some(peer_info) = peer_db_read.peers.get(peer_id).and_then(|peer| {
-                                    if peer.state == PeerState::Trusted {
-                                        Some(peer.clone())
-                                    } else {
-                                        None
-                                    }
-                                }) {
-                                    if peer_info.last_announce.listeners.is_empty() {
+                                let peer_db_read = peer_db.read();
```
litchipi, line 264: There's an indentation here that makes it a big diff where there isn't need for one
Maybe try formatting the code ?

massa-protocol-worker/src/handlers/peer_handler/tester.rs
```diff
@@ -220,10 +235,18 @@ impl Tester {
 
             // if handshake failed, we set the peer state to HandshakeFailed
             if res.is_err() {
+                println!("Handshake failed in tester");
```
litchipi, line 238: Remove this print

massa-protocol-worker/src/tests/mock_network.rs
```diff
@@ -98,6 +98,10 @@ impl ActiveConnectionsTrait for SharedMockActiveConnections {
     fn get_peers_connections_bandwidth(&self) -> HashMap<String, (u64, u64)> {
         HashMap::new()
     }
+
+    fn get_peer_ids_connection_queue(&self) -> HashSet<std::net::SocketAddr> {
+        HashSet::new()
```
litchipi, line 103: Could be done in a default implementation of the trait directly

litchipi (changes requested):

litchipi (approved):

massa-protocol-worker/src/handlers/peer_handler/tester.rs
```diff
@@ -303,6 +327,14 @@ impl Tester {
                                     let db = db.clone();
                                     // receive new listener to test
                                     for (addr, _) in listener.1.iter() {
+                                        if peers_in_test.read().unwrap().contains(addr) {
```
AurelienFT, line 330: We use the parking lot RwLock everywhere in the code it would be nice to keep the standard (and they don't use unwrap !)

modship: done :)

AurelienFT (approved):

AurelienFT:
> Should we merge #4261 before merging ? Or should I make it point to `testnet_25` once this is merged ?

I think it's better to merge this step byt step and so first this PR then the peer prio

PR #4250, opened 2023-07-13
Sync with massa-sc-runtime#287

bilboquet:


Leo-Besancon (approved):

bilboquet:
Thinking twice about this commit, I think this is useless (aka previous commit did not broke old abis because the function from the interface was renamed at all points of use).
@Leo-Besancon @Eitu33 
what do you think?

Leo-Besancon:
>  previous commit did not broke old abis because the function from the interface was renamed at all points of use

I think it's fine with me with this point in mind :)

PR #4261, opened 2023-07-20
Implement peer priorization

### LINKED ISSUE #4260: Peer out connection priorisation.
When seeking to reach target out connection, a node has to choose between a lot of different peers to connect to.

The priorisation of some peers over others has to follow some criteria:
- Date of last failure to connect to the peer (most recent failure = less prio)
- Date of last success to connect to the peer (most recent success = more prio)
- Date of last successfull test of the peer (most recent success = more prio)
- Otherwise, random

Some metadata about the peer's success / failure of connection  / test has to be added in memory, and the priorisation process has to be implemented in `massa-protocol-worker/src/connectivity`

litchipi:
Closes #4260
Useful for #4243 

- Uses `try_connect_history` to store metadata about peer connections.
- Adds new `ConnectionMetadata` struct to store metadata about the peers, and implement `PartialOrd` on this struct to be able to sort a vector of peers.
- Before attempting to connect to new peers, sort all the addresses we wish to connect to
- Adds unit tests about the sorting behaviour of the metadata

### DIFF
--- a/massa-protocol-worker/src/connectivity.rs
+++ b/massa-protocol-worker/src/connectivity.rs
@@ -8,7 +8,6 @@ use massa_pool_exports::PoolController;
 use massa_pos_exports::SelectorController;
 use massa_protocol_exports::{PeerCategoryInfo, PeerId, ProtocolConfig, ProtocolError};
 use massa_storage::Storage;
-use massa_time::MassaTime;
 use massa_versioning::versioning::MipStore;
 use parking_lot::RwLock;
 use peernet::peer::PeerConnectionType;
@@ -18,6 +17,7 @@ use std::{collections::HashMap, net::IpAddr};
 use std::{thread::JoinHandle, time::Duration};
 use tracing::{info, warn};
 
+use crate::handlers::peer_handler::models::ConnectionMetadata;
 use crate::{
     handlers::peer_handler::models::{InitialPeers, PeerState, SharedPeerDB},
     worker::ProtocolChannels,
@@ -258,7 +258,9 @@ pub(crate) fn start_connectivity_thread(
                         let mut slot_default_category = config.default_category_info.target_out_connections.saturating_sub(peers_connected.iter().filter(|(_, peer)| {
                             peer.1 == PeerConnectionType::OUT && peer.2.is_none()
                         }).count());
-                        let mut addresses_to_connect: Vec<SocketAddr> = Vec::new();
+
+                        // Get all the addresses we can cannot to, without any filter or priorisation done yet
+                        let mut addresses_can_connect  = Vec::new();
                         {
                             let peer_db_read = peer_db.read();
                             for (_, peer_id) in &peer_db_read.index_by_newest {
@@ -284,9 +286,13 @@ pub(crate) fn start_connectivity_thread(
                                             continue;
                                         }
 
+                                        let connection_metadata = match peer_db_read.try_connect_history.get(addr) {
+                                            Some(e) => e.clone(),
+                                            None => ConnectionMetadata::default(),
+                                        };
                                         // check if the peer last connect attempt has not been too recent
-                                        if let Some(last_try) = peer_db_read.try_connect_history.get(addr) {
-                                            let last_try = last_try.estimate_instant().expect("Time went backward");
+                                        if let ConnectionMetadata { last_try: Some(lt), .. } = connection_metadata {
+                                            let last_try = lt.estimate_instant().expect("Time went backward");
                                             if last_try.elapsed() < config.try_connection_timer_same_peer.to_duration() {
                                                 continue;
                                             }
@@ -310,35 +316,58 @@ pub(crate) fn start_connectivity_thread(
                                             continue;
                                         }
 
-                                        if let Some(category) = category_found {
-                                            for (name, category_infos) in &mut slots_per_category {
-                                                if name == category && category_infos > &mut 0 {
-                                                        addresses_to_connect.push(*addr);
-                                                        *category_infos -= 1;
-                                                }
-                                            }
-                                        } else if slot_default_category > 0 &&  !addresses_to_connect.contains(addr) {
-                                            addresses_to_connect.push(*addr);
-                                            slot_default_category -= 1;
-                                        }
+                                        addresses_can_connect.push((*addr, connection_metadata, category_found));
+                                    }
+                                }
+                            }
+                        }
 
-                                            // IF all slots are filled, stop
-                                        if slot_default_category == 0 && slots_per_category.iter().all(|(_, slots)| *slots == 0) {
-                                            break;
+                        // Sort addresses using the metadata
+                        addresses_can_connect.sort_by(|a, b| a.1.cmp(&b.1));
+
+                        let mut addresses_to_connect = vec![];
+                        for (addr, _, category) in addresses_can_connect.iter() {
+                            match category {
+                                Some(cat) => {
+                                    for (name, category_infos) in &mut slots_per_category {
+                                        if name == *cat && category_infos > &mut 0 {
+                                            addresses_to_connect.push(*addr);
+                                            *category_infos -= 1;
                                         }
                                     }
                                 }
+                                None if !addresses_to_connect.contains(&addr) => {
+                                    slot_default_category -= 1;
+                                    addresses_to_connect.push(*addr);
+                                }
+                                _ => continue,
+                            }
+
+                            // IF all slots are filled, stop
+                            if slot_default_category == 0 && slots_per_category.iter().all(|(_, slots)| *slots == 0) {
+                                break;
                             }
                         }
 
                         for addr in addresses_to_connect {
                             info!("Trying to connect to addr {}", addr);
 
-                            peer_db.write().try_connect_history.insert(addr, MassaTime::now().unwrap());
+                            let conn_res = network_controller.try_connect(addr, config.timeout_connection.to_duration());
+                            {
+                                let mut peer_db_write = peer_db.write();
+                                let addr_conn_history = peer_db_write.try_connect_history.remove(&addr);    // Remove it to own it, then re-insert it modified
 
-                            // We only manage TCP for now
-                            if let Err(err) = network_controller.try_connect(addr, config.timeout_connection.to_duration()) {
-                                warn!("Failed to connect to peer {:?}: {:?}", addr, err);
+                                let new_addr_conn_history = match (addr_conn_history, &conn_res) {
+                                    (Some(md), Ok(_)) => md.new_success(),
+                                    (None, Ok(_)) => ConnectionMetadata::default().new_success(),
+                                    (Some(md), Err(_)) => md.new_failure(),
+                                    (None, Err(_)) => ConnectionMetadata::default().new_failure(),
+                                };
+
+                                peer_db_write.try_connect_history.insert(addr, new_addr_conn_history);
+                                if let Err(err) = conn_res {
+                                    warn!("Failed to connect to peer {:?}: {:?}", addr, err);
+                                }
                             }
                         }
                     }
--- a/massa-protocol-worker/src/handlers/peer_handler/models.rs
+++ b/massa-protocol-worker/src/handlers/peer_handler/models.rs
@@ -4,7 +4,7 @@ use massa_time::MassaTime;
 use parking_lot::RwLock;
 use peernet::transports::TransportType;
 use rand::seq::SliceRandom;
-use std::cmp::Reverse;
+use std::cmp::{Reverse, Ordering};
 use std::collections::BTreeSet;
 use std::time::Duration;
 use std::{collections::HashMap, net::SocketAddr, sync::Arc};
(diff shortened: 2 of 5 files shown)


litchipi:
@modship @AurelienFT You can already review the PR, I just need to test and fix some cases about the sorting algorithm.
(For example `Some VS None` cases in metadata), but the rest won't move

massa-protocol-worker/src/connectivity.rs
```diff
@@ -284,9 +286,13 @@ pub(crate) fn start_connectivity_thread(
                                             continue;
                                         }
 
+                                        let connection_metadata = match peer_db_read.try_connect_history.get(addr) {
```
modship, line 289: i think you can use .unwrap_or(ConnectionMetadata::default())

massa-protocol-worker/src/connectivity.rs
```diff
@@ -310,35 +314,58 @@ pub(crate) fn start_connectivity_thread(
...
                         }
 
                         for addr in addresses_to_connect {
                             info!("Trying to connect to addr {}", addr);
 
-                            peer_db.write().try_connect_history.insert(addr, MassaTime::now().unwrap());
+                            let conn_res = network_controller.try_connect(addr, config.timeout_connection.to_duration());
+                            {
+                                let mut peer_db_write = peer_db.write();
+                                let addr_conn_history = peer_db_write.try_connect_history.remove(&addr);    // Remove it to own it, then re-insert it modified
```
modship, line 356: why not get_mut() ? or try_connect_history.entry(XXX).and_modify(XXX)

litchipi: I'd honestly be open for a better way to handle this tbh

litchipi:
@damip Added you to the review process, but as you are already heavily loaded, don't bother reading the code itself
Just if you can validate the assumptions I made about the prioritisation, that I wrote as comments in the `tests/peer_priorization.rs` file.
Also, I didn't yet implemented the random choice in case we can get a clear priority, for now I just return them as `Equal` and let `sort()` do the rest

litchipi:
Now that I think of it, `index_by_newest` was used for a first draft of priorisation, we don't need it anymore.
The process is becoming:
- Iter through peers in no particular order
- If peer passes the filtering process, add it to `addresses_can_connect`
- Sort the `addresses_can_connect` by priorization
- Iter through the `addresses_can_connect` to get the peers to connect (with quotas from categories), add the results to `addresses_to_connect`
- Iter through `addresses_to_connect` and try to connect to it

massa-protocol-worker/src/connectivity.rs
```diff
@@ -283,9 +279,11 @@ pub(crate) fn start_connectivity_thread(
                                             continue;
                                         }
```
modship, line 280: we can remove unwrap() on next()

litchipi:
I changed the loop to support multiple listeners:
- I `for` loop on all the listeners
- If any of the check doesn't pass, go to next listener ( `continue` )
- If everything is alright, add the address to the list and break the loop over the listeners

This way it'll only get 1 listener addr, but still iter through all possible listeners to check if it can

massa-protocol-worker/src/connectivity.rs
litchipi: Removed using the change I described below

massa-protocol-worker/src/handlers/peer_handler/models.rs
```diff
@@ -16,15 +16,114 @@ const THREE_DAYS_MS: u64 = 3 * 24 * 60 * 60 * 1_000_000;
...
+                ..self
+            },
+            2 => ConnectionMetadata {
+                last_try: data,
+                ..self
+            },
+            _ => unreachable!("connection metadata data_type not recognized: {data_type}"),
+        }
+    }
+    pub fn new_failure(self) -> ConnectionMetadata {
```
modship, line 100: you can take a mutable ref and update self instead of return new ConnectionMetadata

massa-protocol-worker/src/handlers/peer_handler/tester.rs
```diff
@@ -234,20 +224,30 @@ impl Tester {
...
+                        .entry(peer_id)
+                        .and_modify(|info| {
+                            info.state = super::PeerState::HandshakeFailed;
+                        })
+                        .or_insert(PeerInfo {
+                            last_announce: None,
+                            state: super::PeerState::HandshakeFailed,
+                        });
+                } else {
+                    let new_md = if let Some(md) = peer_db_write.try_connect_history.remove(&addr) {
```
modship, line 242: maybe we should separate the attempts made by the testers from the reel connection attempts. 
For the testers, we should prioritize using the tested_addresses map.

litchipi: Here, we only update the metadata associated with this peer, that will be used later for connection attempt

The tester is completely separated from the connection attempts, the only link is the metadata `last_try` that shows when is the last time the peer was  tested for connectivity using the tester.

massa-protocol-worker/src/connectivity.rs
```diff
@@ -310,35 +314,58 @@ pub(crate) fn start_connectivity_thread(
...
+                            if slot_default_category == 0 && slots_per_category.iter().all(|(_, slots)| *slots == 0) {
+                                break;
                             }
                         }
 
                         for addr in addresses_to_connect {
                             info!("Trying to connect to addr {}", addr);
 
-                            peer_db.write().try_connect_history.insert(addr, MassaTime::now().unwrap());
+                            let conn_res = network_controller.try_connect(addr, config.timeout_connection.to_duration());
```
modship: "try_connect()" spawns a thread in PeerNet, so we don't have the real result here

massa-protocol-worker/src/handlers/peer_handler/tester.rs
modship: But for me `try_connect` from `connectivity.rs` is different from the connection made by testers and it would be nice to make a difference here.
We have information of when the peer was last tested with `tested_addresses`.
With that we have the last time when the peer was tested by tester and the last time of try_connect was called for the peer.

massa-protocol-worker/src/connectivity.rs
modship: yeah :/ need to check but i think we have Ok if the thread was correctly spawned and an error in the other case

massa-protocol-worker/src/handlers/peer_handler/tester.rs
litchipi: Actually, we don't really care of the last attempt of `try_connect` as we have already the time for the possible outcomes, but I see what you mean, it's overlapping

massa-protocol-worker/src/handlers/peer_handler/models.rs
```diff
@@ -16,15 +16,107 @@ const THREE_DAYS_MS: u64 = 3 * 24 * 60 * 60 * 1_000_000;
...
+                ..self
+            },
+            2 => ConnectionMetadata {
+                last_try: data,
+                ..self
+            },
+            _ => unreachable!("connection metadata data_type not recognized: {data_type}"),
+        }
+    }
+    pub fn new_failure(&mut self) {
```
modship, line 100: with the modification (&mut) you don't return a new ConnectionMetadata object. So you can call it `failure()` instead of `new...`

massa-protocol-worker/src/handlers/peer_handler/tester.rs
litchipi: The thing is that I want to have a struct `ConnectionMetadata` containing all the data required to perform priorization, and removing this piece of data would forbid me to implement `PartialOrd` on the structs fields...

I think the best is still to let these overlap a little bit, one is used only for testers (`tested_addresses`), and one only for peer priorization (`try_connect_history`)

modship: ok i see. 

You could write like this instead of remove/insert : 
```rust
         peer_db_write
                        .try_connect_history
                        .entry(addr)
                        .and_modify(|md| {
                            md.new_try();
                        })
                        .or_insert(ConnectionMetadata {
                            last_try: Some(MassaTime::now().unwrap()),
                            ..Default::default()
                        });
                        ```
 And why update ConnectionMetadata of the peer only if res.is_ok() ?
 if res_is_err() we should also set new_try to MD, no ?

File: massa-protocol-worker/src/connectivity.rs (hunk 6 above)
Comment (litchipi) [reply massa-protocol-worker/src/connectivity.rs]: Ok one thing I can do, is only do a `md.failure()` if the `try_connect` thread fails to start, and in the `MassaHandshake::perform_handshake` function, I call `md.success()` or `md.failure()` depending on how the handshake is going

Comment (modship) [reply massa-protocol-worker/src/connectivity.rs]: yes, for now i think it is the ~~only~~ right way

File: massa-protocol-worker/src/handlers/peer_handler/tester.rs (hunk 5 above)
Comment (litchipi) [reply massa-protocol-worker/src/handlers/peer_handler/tester.rs]: In the metadata:
- `last_success` (updated with `success()`) is when we successfully connected to a peer
- `last_failure` (updated with `failure()`) is when we failed to connect to a peer after it was tested as active
- `last_test` (updated with `test()`) is when we test a peer, whatever the result of this test is

We don't need the time of the last connection attempt as it will be updated in either `last_success` or `last_failure` depending on if it was successful or not

Comment (litchipi) [reply massa-protocol-worker/src/handlers/peer_handler/tester.rs]: Sorry I see what you said now, I'll change this

**Comment by litchipi** [comment]
@modship Changed a lot of things, reviewed it and I think it's ready for a second round of review :-)

File: massa-protocol-worker/src/connectivity.rs [hunk 8]
```diff
@@ -257,87 +257,103 @@ pub(crate) fn start_connectivity_thread(
...
                             peer.1 == PeerConnectionType::OUT && peer.2.is_none()
                         }).count());
-                        let mut addresses_to_connect: Vec<SocketAddr> = Vec::new();
+
+                        // Get all the addresses we can cannot to, without any filter or priorisation done yet
+                        let mut addresses_can_connect  = Vec::new();
                         {
                             let peer_db_read = peer_db.read();
-                            for (_, peer_id) in &peer_db_read.index_by_newest {
+                            for (peer_id, peer_info) in &peer_db_read.peers {
```
modship, line 265: to improve this part, maybe it's possible to calculate the number of connections needed before this loop and avoid to iter over all peer_db_read.peers. 
No need to iter 1000 times if we want only one connection

litchipi: Yes but as we need to prioritize peers, in the case we need only `n` connections we have to consider all the peers and take the `n` best peers.
If you break the loop earlier, you won't see all the peers, and may miss some with higher priority

modship: Yes, you are right

modship (approved):

massa-protocol-worker/src/connectivity.rs
```diff
@@ -257,87 +257,103 @@ pub(crate) fn start_connectivity_thread(
                         let mut slot_default_category = config.default_category_info.target_out_connections.saturating_sub(peers_connected.iter().filter(|(_, peer)| {
                             peer.1 == PeerConnectionType::OUT && peer.2.is_none()
                         }).count());
-                        let mut addresses_to_connect: Vec<SocketAddr> = Vec::new();
+
+                        // Get all the addresses we can cannot to, without any filter or priorisation done yet
```
damip, line 261:
```suggestion
                        // Get all the addresses we can connect to, without any filter or prioritization done yet
```

litchipi:
@damip Using this hyper-iterative algorithm, I measured these performances.
Note that I measured for `N` peers means that `N` peers are inside the database, and `N` peers are selected for priorisation. In reality, for 10000 active peers, only a really rare "peak" event would require to priorize all of them at the same time.
But if that happens, here are the estimate performance hit I measured:
```
1000:    ~ 3 ms
5000:    ~ 10.5 ms
10000:    ~ 17 ms
30000:    ~ 63 ms
```
The time is measured from before the `for peer in peer_db.peers` loop to after all the `try_connect` are done.
I skipped a number of operations in order to make this test, you can look at the code in the branch `peer_prio_perftest`.

litchipi:
What I propose is to merge it from now as I don't really think it's that impactful on the perf (even if happening every 250ms), but open an issue to optimize this later.
I could see a possible solution where we use a `BTreeMap` to store the peers using the metadata (ordered by prio) as indexes somehow, and then iter a single time over them to fill the connection slots, but we should investigate this further

damip:
> What I propose is to merge it from now as I don't really think it's that impactful on the perf (even if happening every 250ms), but open an issue to optimize this later. I could see a possible solution where we use a `BTreeMap` to store the peers using the metadata (ordered by prio) as indexes somehow, and then iter a single time over them to fill the connection slots, but we should investigate this further

Ok let's merge, but it is undeniably heavy (permanently using more than 5% of a core just for choosing to whom to connect) so optimizing it would be a nice followup

litchipi:
I'll add a debug log on the loop so we can track on a future testnet how long does this take in a real scenario.
It's difficult to predict how heavy it'll be as the iterations can be cut short for various reasons. Maybe it'll never be in the scenario of a high CPU load except for special cases like initialization.

AurelienFT (approved):
More clear prioritizing. Good job !

AurelienFT:
Merging it to deal with conflicts now.

PR #4276, opened 2023-07-25
Remove local struct hashes

Leo-Besancon:
  * [ ] make tests compile
  * [ ] make tests pass 

We need to update the AsyncMessage proto files too to remove the hash from the grpc mapping file.


Eitu33:
Is there an initial issue to have context for this?

Eitu33 (approved):

PR #4277, opened 2023-07-25
Protocol: Add periodic unban of every peers

Leo-Besancon:
  * [ ] make tests compile
  * [ ] make tests pass 

### DIFF
--- a/massa-protocol-worker/src/handlers/peer_handler/models.rs
+++ b/massa-protocol-worker/src/handlers/peer_handler/models.rs
@@ -72,8 +72,8 @@ impl PeerDB {
     }
 
     pub fn unban_peer(&mut self, peer_id: &PeerId) {
-        if self.peers.contains_key(peer_id) {
-            self.peers.remove(peer_id);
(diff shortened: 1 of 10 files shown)


massa-protocol-worker/src/handlers/peer_handler/models.rs
```diff
@@ -72,8 +72,8 @@ impl PeerDB {
     }
 
     pub fn unban_peer(&mut self, peer_id: &PeerId) {
-        if self.peers.contains_key(peer_id) {
-            self.peers.remove(peer_id);
+        if let Some(peer) = self.peers.get_mut(peer_id) {
+            peer.state = PeerState::Trusted;
```
modship, line 76: don't know if the state should be Trusted.  
Trusted means the peer has been tested successfully.

Leo-Besancon: Changed to HandshakeFailed like you proposed.

modship (approved):
:ok_hand: :ok_hand:

PR #4284, opened 2023-07-26
update consensus rule

damip:
  * [ ] make tests compile
  * [ ] make tests pass 


damip:
@AurelienFT already reviewed by Leo

AurelienFT (approved):

PR #4285, opened 2023-07-27
Add max event size check

Leo-Besancon:
  * [ ] make tests compile
  * [ ] make tests pass 

### DIFF
--- a/massa-execution-worker/src/interface_impl.rs
+++ b/massa-execution-worker/src/interface_impl.rs
@@ -1055,6 +1055,10 @@ impl Interface for InterfaceImpl {
     ///
     /// [DeprecatedByNewRuntime] Replaced by `get_current_slot`
     fn generate_event(&self, data: String) -> Result<()> {
+        if data.len() > self.config.max_event_size {
+            bail!("Event data size is too large");
+        };
+
         let mut context = context_guard!(self);
         let event = context.event_create(data, false);
         context.event_emit(event);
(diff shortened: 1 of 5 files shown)


massa-execution-worker/src/interface_impl.rs
```diff
@@ -1066,8 +1070,11 @@ impl Interface for InterfaceImpl {
     /// # Arguments:
     /// data: the bytes_array data that is the payload of the event
     fn generate_event_wasmv1(&self, data: Vec<u8>) -> Result<()> {
-        let data_str = String::from_utf8(data.clone()).unwrap_or(format!("{:?}", data));
+        if data.len() > self.config.max_event_size {
```
bilboquet, line 1073: I bet it should be used in many other places.
So I'd make it an helper function :
```rust
fn check (len: usize) -> Result<()> {}
// or
fn check_len (data: &[u8]) -> Result<()> {}
```
so it can be called 
`check(data.len())?;`

massa-models/src/config/constants.rs
```diff
@@ -236,6 +236,8 @@ pub const POOL_CONTROLLER_DENUNCIATIONS_CHANNEL_SIZE: usize = 1024;
 pub const MAX_GAS_PER_BLOCK: u64 = u32::MAX as u64;
 /// Maximum of GAS allowed for asynchronous messages execution on one slot
 pub const MAX_ASYNC_GAS: u64 = 1_000_000_000;
+/// Maximum event size in bytes
+pub const MAX_EVENT_SIZE: usize = 102_400;
```
bilboquet, line 240: I propose 
```rust
pub const MAX_EVENT_SIZE: usize = 100*1024;
```
probably a personal taste (a limitation?) but I find it far easier to read (100KB)

Eitu33 (approved):

Eitu33:
Nothing else to say

massa-execution-worker/src/interface_impl.rs
```diff
@@ -1066,8 +1070,11 @@ impl Interface for InterfaceImpl {
     /// # Arguments:
     /// data: the bytes_array data that is the payload of the event
     fn generate_event_wasmv1(&self, data: Vec<u8>) -> Result<()> {
```
bilboquet, line 1072: we should pass data as &[u8] here, we're probably doing a useless copy

AurelienFT: Opened a follow up issue #4298

massa-execution-worker/src/interface_impl.rs
AurelienFT: The helper function would also need the limit passed as parameter which is nearly the same as making the check directly in the function.

PR #4291, opened 2023-07-27
Merge main into testnet_25

damip:
  * [ ] make tests compile
  * [ ] make tests pass 


litchipi:
We have a first merge done in the branch `merge_testnet_25_in_main`, but some conflicts are still there because the latest `massa-sc-runtime` version isn't aligned with the version of `massa-proto-rs` we are using.

aoudiamoncef:
@bilboquet there is a new commit on the main branch of massa-proto-rs 

We should be sure to use https://github.com/massalabs/massa-proto-rs/commit/32179e291bb1ad521d362390318d7dad80c04026

PR #4304, opened 2023-07-31
Add server self signed certificate generation

aoudiamoncef:
  * [x] make tests compile
  * [x] make tests pass 


aoudiamoncef:
Hi @modship @Leo-Besancon 

We should make a small decisions here:

- Which cryptographic hash functions to use ? (IMHO SHA256)
- Should we generate one client self signed certificate ?
- Should we enable TLS by default if we enable generating self signed certificates ?

massa-grpc/src/config.rs
```diff
@@ -21,8 +21,15 @@ pub struct GrpcConfig {
     pub enable_reflection: bool,
     /// whether to enable TLS
     pub enable_tls: bool,
+
```
modship, line 24: remove this line

modship (dismissed):

massa-grpc/src/config.rs
aoudiamoncef: done

aoudiamoncef:
Hi @damip 

The main goal of this PR is to introduce the minimaliste way to generate self signed certificates. 

If both tls and generate self signed certificates are enabled -> generate CA/Server certs
and If mTLS is enabled -> generate CA/Server certs/Client certs

massa-grpc/src/config.rs
```diff
@@ -23,6 +23,10 @@ pub struct GrpcConfig {
     pub enable_tls: bool,
     /// whether to enable mTLS (requires `enable_tls` to be true)
     pub enable_mtls: bool,
+    /// whether to generate a self-signed certificate if none is provided(ignored if `enable_tls` is false)
+    pub generate_self_signed_certificates: bool,
+    /// Subject Alternative Names is an extension in X.509 certificates that allows a certificate to specify additional subject identifiers. It is used to support alternative names for a subject, other than its primary Common Name (CN), which is typically used to represent the primary domain name.
+    pub subject_alt_names: Vec<String>,
```
Leo-Besancon, line 29: I assume this will remain empty for most nodes? Is it needed if you want to access it with both the IP:port and a domain name?
Maybe add an example use case here, if it's not too long to explain!

massa-grpc/src/server.rs
```diff
@@ -192,15 +195,25 @@ where
...
             .expect("error, failed to read server certificat");
         let key = std::fs::read_to_string(config.server_private_key_path.clone())
             .expect("error, failed to read server private key");
-        let server_identity = Identity::from_pem(cert, key);
 
+        let server_identity = Identity::from_pem(cert, key);
         let tls = ServerTlsConfig::new().identity(server_identity);
 
         if config.enable_mtls {
+            //TODO add client certificate generation ? yes if not exists
```
Leo-Besancon, line 216: Should this be done in this PR?

massa-node/base_config/config.toml
```diff
@@ -102,9 +112,13 @@
         # whether to enable gRPC reflection(introspection)
         enable_reflection = true
         # whether to enable TLS
-        enable_tls = false
+        enable_tls = true
```
aoudiamoncef, line 115: Should be disabled

aoudiamoncef:
@Leo-Besancon this should be good to merge

Leo-Besancon (approved):
Tested it on my machine, self generation of certificates works, as well as importing it in the Postman workspace.

aoudiamoncef:
bors merge

PR #4307, opened 2023-08-01
Update peernet

litchipi:
Update Peernet dependency to profit from the fix https://github.com/massalabs/PeerNet/pull/65


AurelienFT:
It's already in https://github.com/massalabs/massa/pull/4306 and testnet_25 has already been merged

PR #4320, opened 2023-08-03
Adds a limitation on the stream_limiter

### LINKED ISSUE #4222: Reinstate stream limiter
- Re-add stream limiter in bootstrap and protocol
- Uncomment check in test bootstrap_bandwidth

litchipi:
Close #4222 

Adds a new option `rate_limit` in the config

Set the default for the rate limit to 2 Mo/secs (OK in labnet at 10k txps)

Passes the config value to PeerNet configuration


massa-node/base_config/config.toml
```diff
@@ -279,6 +279,8 @@
     max_in_connections = 250
     # Cooldown before testing again old peer
     test_oldest_peer_cooldown = 720000
+    # Rate limitation on the data streams (per second)
+    rate_limit = 5_242_880    # 5 Mo / secs
```
damip, line 283: we should use a different limit for the bootstrap, probably something like 10_000_000

damip:
@litchipi can you try this on the current running testnet to see if your node has trouble following the flow ?

litchipi:
No troubles when applied over TEST.25.2, wasn't able to get IN connections (but got the same behavior on `TEST.25.2`)
I'm open to increase / reduce the limits if you think it's better, just wrote arbitrary values

damip:
> No troubles when applied over TEST.25.2, wasn't able to get IN connections (but got the same behavior on `TEST.25.2`) I'm open to increase / reduce the limits if you think it's better, just wrote arbitrary values

can you try to sync with someone for them to try to bootstrap from your node while running on testnet ?

litchipi:
@damip We had to increase the bootstrap limit because it was taking too long (~1m10) to bootstrap.
We made 2 tests on the labnet:
- One under heavy load (10k txps), the first blocks can take time to come (around a minute), and it can happen sometimes that the node has to be restarted because the initialization process doesn't succeeds.
- One under moderate load (5k txps), the bootstrap now takes ~15 secs, and the blocks arrive very soon after, everything seams to work just fine.

I don't know yet if the problems encountered under heavy load is because of the stream limiter or if it would happen without it.
Either cases, do you think this is an acceptable result under such conditions ?

damip:
I think we need to:
* check whether the problem happens without the stream limiter as well
  * if so, we need to find why this happens
  * if not, it means that either the limit is too low or that there is an issue with the stream limiter itself and this needs to be corrected
* take some margin for the protocol peers: if the bootstrap takes time, then the nodes end up having to download a lot of blocks through Protocol, just after bootstrap =>  let's put 10M

litchipi:
Just made more tests, it seams that using 20Mo/secs for bootstrap and 5Mo/secs for the protocol worker works the same as with no limiter under heavy load (10k txps)
Also just doubled checked that everything is alright on the testnet.

massa-node/base_config/config.toml
```diff
@@ -307,6 +307,8 @@
     max_in_connections = 250
     # Cooldown before testing again old peer
     test_oldest_peer_cooldown = 720000
+    # Rate limitation on the data streams (per second)
+    rate_limit = 2097152    # 2 Mo / secs
```
damip, line 311: Mo is in French (o = octet), use byte "B" (1 byte = 8 bits).

Also, those are not megabytes (1 MB = 1000000 B) but mebibytes (1 MiB = 1048576 B).

litchipi:
@damip  Yes, thanks for noticing, I changed the comments, and also raised the base value for `rate_limit` in the protocol.

This is because I tested it on `TEST.25.2` and encountered no real difficulty, but it seams like the traffic is around 1.5MiB/secs, so kinda close to the limit.
I raised the limit to 5MiB/secs in order to give a large margin, but that would still be useful to throttle some kind of spam problem

Tell me if you think it's alright

damip (approved):
LGTM :)

Issue #4333, opened 2023-08-10
Regression on Bootstrap event polling


Leo-Besancon:
It seems that https://github.com/massalabs/massa/pull/4319, and more specifically https://github.com/massalabs/massa/pull/4319/commits/ad7ddc5b8c5063cec0183f19af7ed5ad6af2bb99, introduced a regression on a fix introduced in https://github.com/massalabs/massa/pull/4084. The result is a lack of routability on windows (which seems to be the only platform affected).

We should re-introduce looping on `mio_server` accept (and not only on the main listener), in order to be sure to correctly poll Mio's events.
More info here: https://users.rust-lang.org/t/why-mio-poll-only-receives-the-very-first-event/87501
> Mio is edge triggered. You only get a notification when readiness changes from NOT READY to READY, and it only changes from READY to NOT READY when you get a WouldBlock error. Therefore, you must keep calling accept until it fails with a WouldBlock error.

Side note: we should also make sure if the same fix is needed in Peernet!

Reported in: https://discord.com/channels/828270821042159636/1137667605751136327

damip:
@Leo-Besancon I think this is done ?

Leo-Besancon:
This is done, but not merged yet. We have to target testnet26 and make sure the dependencies are lined up!

Leo-Besancon:
Closed by https://github.com/massalabs/massa/pull/4343

PR #4334, opened 2023-08-10
Fix bootstrap event polling - windows routability

Leo-Besancon:
  * [ ] make tests compile
  * [ ] make tests pass 

Fixes https://github.com/massalabs/massa/issues/4333


massa-bootstrap/src/listener.rs
```diff
@@ -111,6 +111,17 @@ impl BootstrapTcpListener {
             }
         }
 
+        // We need to have an accept() error with WouldBlock, otherwise polling may not raise any new events.
+        // See https://users.rust-lang.org/t/why-mio-poll-only-receives-the-very-first-event/87501
```
damip, line 115: this discussion concludes that we need to implement exactly what is implemented above no ?

Leo-Besancon: No, the loop above does that for the std::net::TcpListener, we need it for the mio_server :)

damip: what happens if we call accept above on mio_server instead of server ?

Leo-Besancon: Then we retrieve mio::net::TcpStream instead of std ones, and I don't think we can easily convert from one to the other (although I didn't look into it much, there is a from_std method but not to_std)

modship: > what happens if we call accept above on mio_server instead of server ?

that return a TcpStream of mio, but we need a std TcpStream

massa-bootstrap/src/listener.rs
```diff
@@ -34,8 +34,30 @@ impl BootstrapTcpListener {
...
+        socket.set_nonblocking(true)?;
+        socket.bind(&(*addr).into())?;
+
+        // Number of connections to queue, set to the hardcoded value used by tokio
+        socket.listen(1024)?;
+
+        info!("Starting bootstrap listener on {}", &addr);
+        let std_server: std::net::TcpListener = socket.into();
+
+        let mut server = TcpListener::from_std(std_server.try_clone().expect("Unable to clone server socket"));
```
damip: instead of `let mut server = TcpListener::from_std(std_server.try_clone().expect("Unable to clone server socket"));` why not just `let mut server = TcpListener::from_std(std_server);` ?

massa-bootstrap/src/listener.rs
```diff
@@ -48,26 +46,29 @@ impl BootstrapTcpListener {
...
         // Number of connections to queue, set to the hardcoded value used by tokio
         socket.listen(1024)?;
 
         info!("Starting bootstrap listener on {}", &addr);
-        let server: TcpListener = socket.into();
+        let std_server: std::net::TcpListener = socket.into();
 
-        let mut mio_server =
-            MioTcpListener::from_std(server.try_clone().expect("Unable to clone server socket"));
+        let mut server = TcpListener::from_std(
```
damip, line 59: instead of `let mut server = TcpListener::from_std(std_server.try_clone().expect("Unable to clone server socket"));` why not just `let mut server = TcpListener::from_std(std_server);` ?

I want to avoid keeping useless references from different libs to the same object at the same time

Leo-Besancon: Yes, the clone is no longer needed as we don't keep the std_server! Thanks for pointing it out

Leo-Besancon:
The current fix (with the peernet changes too), seems to fix well the initial problem with Windows routability.

Leo-Besancon:
Closed in favor of #4343 (based on testnet_26 instead of main)

PR #4337, opened 2023-08-14
Limit number of backups to keep

Leo-Besancon:
  * [ ] make tests compile
  * [ ] make tests pass 

### DIFF
--- a/massa-db-worker/src/massa_db.rs
+++ b/massa-db-worker/src/massa_db.rs
@@ -6,6 +6,7 @@ use massa_db_exports::{
 };
 use massa_hash::{HashXof, HASH_XOF_SIZE_BYTES};
 use massa_models::{
+    config::MAX_BACKUPS_TO_KEEP,
     error::ModelsError,
     slot::{Slot, SlotDeserializer, SlotSerializer},
     streaming_step::StreamingStep,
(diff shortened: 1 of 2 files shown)


massa-db-worker/src/massa_db.rs
```diff
@@ -620,9 +621,43 @@ impl MassaDBController for RawMassaDB<Slot, SlotSerializer, SlotDeserializer> {
...
+                let Ok(thread) = vec[2].parse::<u8>() else {
+                    continue;
+                };
+                let backup_slot = Slot::new(period, thread);
+                previous_backups.insert(backup_slot, backup_path);
+            }
+        }
+
+        // Remove the oldest backup if we have too many
+        while MAX_BACKUPS_TO_KEEP.is_some()
```
litchipi, line 652: I'd be better to have a check on `MAX_BACKUPS_TO_KEEP.is_none()` at the beginning of the function, and only prune old backups if we have a `Some` here

massa-db-worker/src/massa_db.rs
```diff
@@ -620,9 +621,43 @@ impl MassaDBController for RawMassaDB<Slot, SlotSerializer, SlotDeserializer> {
...
+            && previous_backups.len() >= MAX_BACKUPS_TO_KEEP.unwrap()
+        {
+            if let Some((_, oldest_backup_path)) = previous_backups.first_key_value() {
```
litchipi, line 655: You could directly use `if let Some((_, oldest_backup_path)) = previous_backups.pop_first()` here, avoiding to call it later

massa-models/src/config/constants.rs
```diff
@@ -131,8 +131,10 @@ pub const MAX_ASYNC_MESSAGE_DATA: u64 = 1_000_000;
 pub const OPERATION_VALIDITY_PERIODS: u64 = 10;
 /// cycle duration in periods
 pub const PERIODS_PER_CYCLE: u64 = 128;
-/// cycle duration in periods
-pub const PERIODS_BETWEEN_BACKUPS: u64 = 128;
+/// Number of periods between two backups
+pub const PERIODS_BETWEEN_BACKUPS: u64 = 12800; // 100 cycles
```
litchipi, line 135: Instead of setting a const value and add a comment, you could directly do `PERIODS_PER_CYCLE * 100` or something of the sort. (just sugar, up to you)

massa-db-worker/src/massa_db.rs
Leo-Besancon: Refactored it

massa-db-worker/src/massa_db.rs
Leo-Besancon: True!

litchipi (approved):

PR #4342, opened 2023-08-15
Reduce memory usage

damip:
## Problem

Memory usage was too high because we were keeping all the operation-block indexes in the storage for a large history of blocks in consensus.

## TODO

* [x] drop storage blocks in consensus when they are not necessary anymore and only keep a consensus-local version of the header
* [x] do not keep the descendants list in final blocks

## Checklist

  * [x] make tests compile
  * [x] make tests pass 

## Test results on testnet at 5ktxps

Screenshot from 2023-08-17 14-59-45


massa-execution-exports/src/types.rs
```diff
@@ -18,8 +18,18 @@ use massa_models::{
...
 use massa_pos_exports::ProductionStats;
+use massa_storage::Storage;
 use std::collections::{BTreeMap, BTreeSet};
 
+/// Metadata needed to execute the block
+#[derive(Clone, Debug)]
+pub struct ExecutionBlockMetadata {
+    /// Address of the creator of the parent in the same thread
+    pub same_thread_parent_creator: Option<Address>,
+    /// Storage referncing the block and its contents
```
sydhds, line 29: referencing?

massa-consensus-exports/src/block_status.rs
```diff
@@ -58,6 +63,45 @@ impl From<&BlockStatus> for BlockStatusId {
...
+#[derive(Debug, Clone)]
+pub enum StorageOrBlock {
+    /// Keep a full storage with operations
+    Storage(Storage),
+    /// Keep only the block header and list of ops (but not the ops)
+    Block(Box<SecureShareBlock>),
+}
+
+impl StorageOrBlock {
+    /// Return a clone of the underlying block
```
sydhds, line 77: Maybe add more doc about when this is need; not obvious because of CLone & clone_block...

massa-consensus-exports/src/block_status.rs
```diff
@@ -58,6 +63,45 @@ impl From<&BlockStatus> for BlockStatusId {
...
+                .get(block_id)
+                .expect("block absent from its own storage")
+                .clone(),
+            StorageOrBlock::Block(block) => *block.clone(),
+        }
+    }
+
+    /// Convert any StorageOrBlock variant into a StorageOrBlock::Block variant.
+    /// This effectively drops the operations of the block.
+    pub fn strip_to_block(&mut self, block_id: &BlockId) {
```
sydhds, line 91: maybe rename to into_variant_block ?

massa-consensus-worker/src/state/graph.rs
```diff
@@ -43,6 +63,9 @@ impl ConsensusState {
             if let Some(BlockStatus::Active { a_block: ab, .. }) =
                 self.blocks_state.get_mut(&ancestor_h)
             {
+                if ab.is_final {
```
sydhds, line 66: As it seems this was previously missed, maybe add a comment here

massa-consensus-exports/src/block_status.rs
damip: it does not have the typical "into" behavior that consumes `self` and produces a new `Self`, here it mutates `&mut self`

massa-consensus-exports/src/block_status.rs
damip: done

massa-execution-exports/src/types.rs
damip: done

massa-consensus-worker/src/state/graph.rs
damip: done

sydhds (approved):

PR #4351, opened 2023-08-18
Update crate ed25519-dalek from 1.0 to 2.0

sydhds:
  * [ ] make tests compile
  * [ ] make tests pass 


damip:
@sydhds nice !!
Can you ensure retro-compatibility ? eg. try to verify and sign in both directions between an "old" and a "new" wallet ?

sydhds:
> @sydhds nice !! Can you ensure retro-compatibility ? eg. try to verify and sign in both directions between an "old" and a "new" wallet ?

I've been testing in the following way:
* Checkout branch with ed25519-dalek v2.0.0
* Generate a new address and sign a message using massa-client
* Checkout branch with ed25519-dalek v1.0.0
* Relaunch massa-client, and reuse the previous wallet to sign the same message
* Compare both signatures

then done the same process again but with reversed versions (start with v1 then checkout v2).

All good!

Cargo.toml
```diff
@@ -123,7 +123,8 @@ ctrlc = "3.2.5"
 dialoguer = "0.10"
 directories = "4.0"
 displaydoc = "0.2"
-ed25519-dalek = "=1.0"
+# ed25519-dalek = "=1.0"
+ed25519-dalek = { version = "2.0", features = ["rand_core", "zeroize"] }
```
damip, line 127: the version should be fixed "=2.0" like before

massa-signature/src/signature_impl.rs
```diff
@@ -25,16 +25,8 @@ use transition::Versioned;
 #[allow(missing_docs)]
 /// versioned KeyPair used for signature and decryption
 #[transition::versioned(versions("0", "1"))]
```
damip, line 27: We should drop the fake versions we created for testing

Cargo.toml
sydhds: Done

massa-signature/src/signature_impl.rs
sydhds: Removed :)

damip (approved):

PR #4353, opened 2023-08-18
peers_in_test into PeerDB

modship:
  * [ ] make tests compile
  * [ ] make tests pass 

### DIFF
--- a/massa-protocol-worker/src/handlers/peer_handler/tester.rs
+++ b/massa-protocol-worker/src/handlers/peer_handler/tester.rs
@@ -341,11 +341,9 @@ impl Tester {
                                     let db = db.clone();
                                     // receive new listener to test
                                     for (addr, _) in listener.1.iter() {
-                                        if db.read().peers_in_test.contains(addr) {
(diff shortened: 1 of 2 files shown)


massa-protocol-worker/src/handlers/peer_handler/tester.rs
```diff
@@ -348,12 +341,11 @@ impl Tester {
...
                                     // receive new listener to test
                                     for (addr, _) in listener.1.iter() {
-                                        if peers_in_test.read().contains(addr) {
+                                        if db.read().peers_in_test.contains(addr) {
                                             continue;
                                         }
                                         {
-                                            let mut peers_in_test = peers_in_test.write();
-                                            peers_in_test.insert(*addr);
+                                            db.write().peers_in_test.insert(*addr);
```
damip, line 348: this whole thing can be replaced more reliably with 

```rust
if !db.write().peers_in_test.insert(*addr) {
    continue;
}
```

damip (approved):

PR #4356, opened 2023-08-21
cargo update

modship:
  * [x] make tests compile
  * [x] make tests pass 

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -77,9 +77,9 @@ dependencies = [
 
 [[package]]
 name = "aho-corasick"
-version = "1.0.3"
+version = "1.0.4"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "86b8f9420f797f2d9e935edf629310eb938a0d839f984e25327f3c7eed22300c"
+checksum = "6748e8def348ed4d14996fa801f4122cd763fff530258cdc03f64b25f89d3a5a"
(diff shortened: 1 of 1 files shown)


modship:
> this seems to have been done on main, I guess we should also merge main into testnet_26

Yes, we need to merge main into testnet_26 because the build fails after cargo update (there's a missing update for massa-proto).

damip:
> > this seems to have been done on main, I guess we should also merge main into testnet_26
> 
> Yes, we need to merge main into testnet_26 because the build fails after cargo update (there's a missing update for massa-proto).

@aoudiamoncef can you sync with this ?

modship:
> > > this seems to have been done on main, I guess we should also merge main into testnet_26
> > 
> > 
> > Yes, we need to merge main into testnet_26 because the build fails after cargo update (there's a missing update for massa-proto).
> 
> @aoudiamoncef can you sync with this ?

@damip https://github.com/massalabs/massa/pull/4358

damip:
https://github.com/massalabs/massa/pull/4358 is done, so this can be closed ?

modship:
> #4358 is done, so this can be closed ?

If TN26 will be merged into main, yes

damip:
> > #4358 is done, so this can be closed ?
> 
> If TN26 will be merged into main, yes

it will, on release

PR #4362, opened 2023-08-22
remove bound_map feature

modship:
  * [ ] make tests compile
  * [ ] make tests pass 


rust-toolchain.toml
```diff
@@ -1,2 +1,2 @@
 [toolchain]
-channel = "nightly-2023-07-01"
+channel = "stable-2023-08-03"
```
AurelienFT, line 2: Don't change this now all the others features are not removed yet

modship: done

AurelienFT (approved):

PR #4363, opened 2023-08-22
remove variant_count feature

modship:
  * [ ] make tests compile
  * [ ] make tests pass 

### DIFF
--- a/massa-versioning/src/versioning.rs
+++ b/massa-versioning/src/versioning.rs
@@ -111,7 +111,7 @@ impl Eq for MipInfo {}
 
 machine!(
     /// State machine for a Versioning component that tracks the deployment state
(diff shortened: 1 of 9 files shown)


massa-versioning/src/versioning.rs
```diff
@@ -108,7 +111,7 @@ impl Eq for MipInfo {}
 
 machine!(
     /// State machine for a Versioning component that tracks the deployment state
-    #[derive(Clone, Copy, Debug, PartialEq)]
+    #[derive(Clone, Copy, Debug, PartialEq, VariantCount)]
     pub(crate) enum ComponentState {
```
modship, line 115: For now, it doesn't work with this enum.

PR #4373, opened 2023-08-25
Add op_exec_status + buildnet

aoudiamoncef:
  * [x] make tests compile
  * [x] make tests pass 


damip:
@aoudiamoncef can you rebase this on testnet26 ?

PR #4375, opened 2023-08-27
Unify seeds in execution

damip:
  * [x] make tests compile
  * [x] make tests pass 


massa-execution-worker/src/context.rs
```diff
@@ -460,47 +460,38 @@ impl ExecutionContext {
         )
         .expect("could not compute current slot timestamp");
 
-        // create a seed from the current slot
-        let mut data: Vec<u8> = self.slot.to_bytes_key().to_vec();
-        // add the index of the created address within this context to the seed
-        data.extend(self.created_addr_index.to_be_bytes());
-        // add a flag on whether we are in read-only mode or not to the seed
-        // this prevents read-only contexts from shadowing existing addresses
```
Leo-Besancon, line 468: > // this prevents read-only contexts from shadowing existing addresses

Here IIRC the read-only flag is included in the execution_trail_hash and that's why it's not needed anymore, right?

Leo-Besancon (commented):
Just one comment to make sure I understand correctly before approval.

massa-execution-worker/src/context.rs
damip: exactly :)

Leo-Besancon (approved):

PR #4380, opened 2023-09-01
fix links README

qdrn:


AurelienFT (approved):

AurelienFT:
I merge t on testnet_26 because it will not add much delay as we will release it soon but the CI is fix and better there.

PR #4393, opened 2023-09-06
remove staking lock

### LINKED ISSUE #4389: Node stuck
**Context:**
- Labnet network with 10k txps
- My local node have 20 staking address (Previous network and test)

When i try to remove multiple staking address with massa-client `node_stop_staking XXX...` my node stuck and doesn't receive any block for a while.
Then it unblocks and works normally

modship:
  * [ ] make tests compile
  * [ ] make tests pass 


modship:
the problem must be the same with `add_keypairs()`

modship:
this solution works but we need to call manually save() with read_lock after remove_addresses

damip:
can you rebase it on main ?

modship:
> can you rebase it on main ?

done :)

damip (approved):

damip:
bors merge

Issue #4264, opened 2023-07-21
Emit an event when an execution succeeds


damip:
Currently we emit an event when an operation execution fails.

We should also emit an event when it succeeds, and summarize the state changes it has caused, as well as its return value and used gas if applicable (eg. if it is a smart contract call)

PR #4400, opened 2023-09-11
emit event when an execution succeeds

modship:
  * [ ] make tests compile
  * [ ] make tests pass 

#4264


damip:
thanks for the update, let's block it for now, until we prioritize things post-testnet

PR #4411, opened 2023-09-19
Add checks ops

### LINKED ISSUE #4392: invalid gasLimit
**Context**
When using "send_operations" endpoint with a big value for maxgas, the operation is submitted successfully, operation ID is returned but the transaction is not processed.

What are the constraints behind this?

If the maxGas value is invalid, it would be nice to return an error.
If its more complicated than that,  another way would be to emit en error event which could be retrieved when listening smart contract events in context of SC deployment or SC call.

I know it is already possible to check the operation status of an operation using OperationInfo but returning directy will make a smoother developer experience!

AurelienFT:
  * [x] make tests compile
  * [x] make tests pass 

Fix #4392

### DIFF
--- a/massa-grpc/src/stream/send_operations.rs
+++ b/massa-grpc/src/stream/send_operations.rs
@@ -3,11 +3,13 @@
 use crate::error::{match_for_io_error, GrpcError};
 use crate::server::MassaPublicGrpc;
 use futures_util::StreamExt;
-use massa_models::operation::{OperationDeserializer, SecureShareOperation};
+use massa_models::operation::{OperationDeserializer, OperationType, SecureShareOperation};
 use massa_models::secure_share::SecureShareDeserializer;
+use massa_models::timeslots::get_latest_block_slot_at_timestamp;
 use massa_proto_rs::massa::api::v1 as grpc_api;
 use massa_proto_rs::massa::model::v1 as grpc_model;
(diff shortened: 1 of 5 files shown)


massa-grpc/src/stream/send_operations.rs
```diff
@@ -54,6 +56,29 @@ pub(crate) async fn send_operations(
...
+                            config.thread_count,
+                            config.t0,
+                            config.genesis_timestamp,
+                            now,
+                        ) else {
+                            report_error(
+                                tx.clone(),
+                                tonic::Code::InvalidArgument,
+                                "too many operations per message".to_owned(),
+                            )
```
Leo-Besancon, line 78: I don't understand the error message here and above

Leo-Besancon (dismissed):
Approved with a small comment

massa-grpc/src/stream/send_operations.rs
AurelienFT: fixed. Shitty copy:paste sorry

Leo-Besancon (dismissed):

aoudiamoncef (approved):
LGTM 👍🏻

Leo-Besancon (approved):

AurelienFT:
bors merge

PR #4413, opened 2023-09-19
fix asking for new password

### LINKED ISSUE #4405: Client asks for new password and confirmation at each opening
Client asks for new password and confirmation at each opening.

If the wallet already exists, it should just ask once for the existing password, not to "create a new password".

AurelienFT:
Fix #4405


damip (approved):

AurelienFT:
bors merge

PR #4421, opened 2023-09-21
Add tests and rename handler and data

AurelienFT:
## Important description

This PR improves readability of the `AsyncMessage` structure, fix potential panics.

### Details : 

- Fix potential panic behaviour : Denom value of 0 was accepted in deserialisation
- Fix problem function name in async message could be only u8::MAX size but the limit of length function is u16::MAX
- Unify data limits of async message that represent the parameters of the function call with the limits of parameters of a normal call. The values was differents.
- Wrong limit used for handler name in SetOrKeep serialisation
- Renamed handler -> function
- Renamed Data - > functions params
- Remove from bootstrap async pool variables unused
- Change excluded to included in  max size of trigger datastore key to match the behaviour of the ledger datastore key limit

### Follow up that needs to be done
In GRPC :
- Rename handler -> function
- Rename data -> function params

On a more general note, I think we should make a changelog for the mainnet so that we could add breaking changes (but useful to be done now otherwise it will be a bigger mess) like this change and #4409 for example. @Leo-Besancon and @damip I let you decide on that.

The `messages.rs` file now have 90% of line coverage and all min/max limit on serializer/deserializer are tested


massa-async-pool/src/config.rs
```diff
@@ -7,8 +7,10 @@
 pub struct AsyncPoolConfig {
     /// max number of messages in the pool
     pub max_length: u64,
-    /// max async message data (for bootstrap limits)
-    pub max_async_message_data: u64,
+    /// max handler length
+    pub max_function_length: u16,
+    /// max handler params length (for bootstrap limits)
```
Leo-Besancon, line 12: The "(for bootstrap limits)" is probably no longer relevant here.

massa-async-pool/src/lib.rs
```diff
@@ -31,7 +31,7 @@
 //!
 //! ## How to send a message during bytecode execution
 //!
-//! * messages are sent using an ABI: `send_message(target_address, target_handler, validity_start, validity_end, max_gas, fee, coins, data: JSON string) -> Result<(), ABIReturnError>`. Note that data has a configuration defined `max_async_message_data_size`.
+//! * messages are sent using an ABI: `send_message(target_address, target_handler, validity_start, validity_end, max_gas, fee, coins, data) -> Result<(), ABIReturnError>`.
```
Leo-Besancon, line 34: Also rename handler and data here

massa-async-pool/src/message.rs
```diff
@@ -96,7 +97,10 @@ impl Serializer<AsyncMessageId> for AsyncMessageIdSerializer {
 #[derive(Clone)]
 pub struct AsyncMessageIdDeserializer {
     slot_deserializer: SlotDeserializer,
-    u64_deserializer: U64VarIntDeserializer,
+    // TODO: Denom should be separate because it can't be 0
```
Leo-Besancon, line 100: Remove this todo

massa-async-pool/src/pool.rs
```diff
@@ -955,11 +958,11 @@ impl AsyncPool {
         }
 
         // Data
-        if let SetOrKeep::Set(data) = message_update.data {
+        if let SetOrKeep::Set(params) = message_update.function_params {
```
Leo-Besancon, line 961: Rename (or remove if useless) the comments Handler and Data

massa-async-pool/src/test_exports/bootstrap.rs
```diff
@@ -40,101 +31,3 @@ pub fn get_random_message(fee: Option<Amount>, thread_count: u8) -> AsyncMessage
...
-            // We reset the current values
-            current_id = Some(message_id_1);
-            current_count = 1;
-            current_message_1.clear();
-            current_message_1.extend(val1.1.iter());
-            current_message_2.clear();
-            current_message_2.extend(val2.1.iter());
-        }
-    }
-}
```
Leo-Besancon, line 140: Was this dead_code?

AurelienFT: Yes

massa-async-pool/src/lib.rs
AurelienFT: Done

massa-async-pool/src/config.rs
AurelienFT: Done

massa-async-pool/src/pool.rs
AurelienFT: Done and also in others places in this file.

Leo-Besancon (approved):
Thanks for the small corrections, everything looks good!

AurelienFT:
bors merge

PR #4432, opened 2023-09-28
Fix GetSelectorDraws response

### LINKED ISSUE #4429: GRPC Api : get_selector_draws
- Slot range request parameters is overwritten in the code

https://github.com/massalabs/massa/blob/fa9bd21ceac04da1819d3afef31bc0563ddd7f6f/massa-grpc/src/public.rs#L602
https://github.com/massalabs/massa/blob/fa9bd21ceac04da1819d3afef31bc0563ddd7f6f/massa-grpc/src/public.rs#L603

- get_selector_draws should only return the slots that match the filter, currently, all slots are sent.
The unit test in https://github.com/massalabs/massa/pull/4423 should pass.

aoudiamoncef:
  * [x] make tests compile
  * [x] make tests pass 

Fix #4429

### DIFF
--- a/massa-grpc/src/public.rs
+++ b/massa-grpc/src/public.rs
@@ -624,27 +624,19 @@ pub(crate) fn get_selector_draws(
             .unwrap_or_default()
             .into_iter()
             .map(|(v_slot, v_sel)| {
-                let block_producer: Option<String> = match restrict_to_addresses {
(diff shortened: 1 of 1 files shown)


modship (approved):
The slotDraw vector is still not filtered.

aoudiamoncef:
bors merge

PR #4436, opened 2023-09-29
Use is_terminal instead of atty crate + update rust_decimal to 1.32 (…

sydhds:
…+ disable default features)

  * [ ] make tests compile
  * [ ] make tests pass 


.github/workflows/ci.yml
```diff
@@ -96,6 +96,17 @@ jobs:
           token: ${{ secrets.GITHUB_TOKEN }}
           args: -- -A clippy::uninlined-format-args
 
+  security:
+    if: github.ref != 'refs/heads/staging'
+    needs: sanity
+    runs-on: ubuntu-latest
+    continue-on-error: true
```
Leo-Besancon, line 103: If I understand correctly, the continue-on-error would mean a dependancy audit warning would not block the CI, right? It would be treated as a success for `bors merge`?

massa-client/src/main.rs
```diff
@@ -49,17 +51,12 @@ struct Args {
     #[structopt(name = "PARAMETERS")]
     parameters: Vec<String>,
     /// Path of wallet folder
-    #[structopt(
-        short = "w",
-        long = "wallet",
-        parse(from_os_str),
-        default_value = "wallets/"
-    )]
+    #[structopt(short = 'w', long = "wallet", default_value = "wallets/")]
```
Leo-Besancon, line 54: Why is there still references to structopt? Did clap integrate these macros?

Leo-Besancon (commented):
Thanks for fixing these :). Just two comments to be sure of the behaviour

massa-client/src/main.rs
sydhds: Removed :)

.github/workflows/ci.yml
sydhds: From what I read, yes.

sydhds:
> Great, if you have tested that there are no behaviour changes between using clap and structopt, I approve!

There is only one change when massa-node is compiled with feature op-spammer, the short command for number-operations is 'n' (was 'nb-op' before).

Leo-Besancon (approved):

Cargo.toml
```diff
@@ -172,6 +172,7 @@ sha3 = "=0.10"
 socket2 = "0.5"
 stream_limiter = "3.2"
 structopt = "0.3"
+clap = { version = "4.4.6", features = ["derive", "cargo"] }
```
aoudiamoncef, line 175: 4.4 to comply to the general deps versioning

aoudiamoncef (approved):
LGTM

Cargo.toml
sydhds: Done :)

sydhds:
bors merge

Leo-Besancon (approved):

sydhds:
bors merge

PR #4437, opened 2023-09-29
Remove unused dependencies

aoudiamoncef:
  * [x] make tests compile
  * [x] make tests pass 


massa-consensus-exports/Cargo.toml
```diff
@@ -8,17 +8,14 @@ edition = "2021"
...
 displaydoc = {workspace = true}
 nom = {workspace = true}
 serde = {workspace = true, "features" = ["derive"]}
 serde_json = {workspace = true}   # BOM UPGRADE     Revert to "1.0" if problem
 thiserror = {workspace = true}
-jsonrpsee = {workspace = true, "features" = ["server"]}
 tokio = {workspace = true, "features" = ["sync"]}
 mockall = {workspace = true, "optional" = true}   # BOM UPGRADE     Revert to {"version": "0.11.4", "optional": true} if problem
 massa_channel = {workspace = true}
-massa_hash = {workspace = true}
```
Leo-Besancon, line 21: I think there are check comments associated (see at the bottom of the diff page on github).
`massa_hash` is used in the doc comments examples (https://github.com/massalabs/massa/blob/main/massa-consensus-exports/src/export_active_block.rs#L144), so I think we can put it as [dev-dependencies] instead?

Can you check for similar behaviours?

Leo-Besancon (changes requested):
We need to be sure the dependancies are not used with a feature combination / in comment examples before merging this one. Can you check?
Thanks for digging everything up :).

massa-consensus-exports/Cargo.toml
aoudiamoncef: I'm agree, we have to check all combinations.

aoudiamoncef: It seems that the deps analyzer don't recognize the doctests for example

Leo-Besancon (approved):
Nice job :)

aoudiamoncef:
bors merge

PR #4444, opened 2023-10-03
Fix deadlock issue in `GetBlocks`

### LINKED ISSUE #4438: Node API thread crash - Testnet 26
**Describe the bug**
When I run the indexer updated to version 26 I make the node crash.
The only new feature I am using is the QueryState request in gRPC. Here is the query inside the request in question, with the Operation ID field replaced by a dummy value:

ExecutionQueryRequestItem{
			RequestItem: ExecutionQueryRequestItem_OpExecutionStatusFinal{
				OpExecutionStatusFinal: OpExecutionStatusFinal{OperationId: "An Operation ID"},
			},
})

I am sending a lot of these queries in one request. I don't have the exact number right now but I can log it if required.

**To Reproduce**
DM me

**Expected behavior**
The node shouldn't crash when I send requests to it

aoudiamoncef:
  * [ ] make tests compile
  * [ ] make tests pass 

fix #4438

### DIFF
--- a/massa-grpc/src/public.rs
+++ b/massa-grpc/src/public.rs
@@ -484,33 +484,34 @@ pub(crate) fn get_operations(
         })
         .collect::<Result<_, _>>()?;
 
-    let read_blocks = grpc.storage.read_blocks();
-    let read_ops = grpc.storage.read_operations();
-
-    // Get the operations and the list of blocks that contain them from storage
-    let storage_info: Vec<(&SecureShareOperation, HashSet<BlockId>)> = operation_ids
-        .into_iter()
-        .filter_map(|ope_id| {
-            read_ops.get(&ope_id).map(|secure_share| {
-                let block_ids = read_blocks
-                    .get_blocks_by_operation(&ope_id)
-                    .map(|hashset| hashset.iter().cloned().collect::<HashSet<BlockId>>())
-                    .unwrap_or_default();
-
-                (secure_share, block_ids)
+    let storage_info: Vec<(SecureShareOperation, PreHashSet<BlockId>)> = {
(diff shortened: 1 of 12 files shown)


massa-grpc/src/public.rs
```diff
@@ -482,33 +484,34 @@ pub(crate) fn get_operations(
...
-            read_ops.get(&ope_id).map(|secure_share| {
-                let block_ids = read_blocks
-                    .get_blocks_by_operation(&ope_id)
-                    .map(|hashset| hashset.iter().cloned().collect::<HashSet<BlockId>>())
-                    .unwrap_or_default();
-
-                (secure_share, block_ids)
+    let storage_info: Vec<(SecureShareOperation, PreHashSet<BlockId>)> = {
+        let read_blocks = grpc.storage.read_blocks();
+        let read_ops = grpc.storage.read_operations();
```
damip, line 489: never take the two at the same time !

start by locking the ops, list them, then unlock, then lock blocks and add get_blocks_by_operation information

aoudiamoncef: We should do the same in the old API

damip: yes, can you make the changes directly in this PR ?

aoudiamoncef: In JsonRPC API, we have a different behaviour, we took the lock for every operation, release it. 

What's more efficient ?

damip (dismissed):

damip (approved):

aoudiamoncef:
Bors merge

modship (approved):

PR #4447, opened 2023-10-04
Add 4 unit tests for final state (line coverage ~ 80%)

sydhds:
  * [ ] make tests compile
  * [ ] make tests pass 

### DIFF
--- a/massa-final-state/src/final_state.rs
+++ b/massa-final-state/src/final_state.rs
@@ -1045,6 +1045,7 @@ mod test {
 
         assert_eq!(fstate.get_slot(), initial_slot);
 
+        // This should also fail because there is no initial cycle (required by POS state)
         let res = fstate._finalize(ok_next_slot, changes.clone());
 
         assert!(res.is_err());
--- a/massa-final-state/src/final_state.rs
+++ b/massa-final-state/src/final_state.rs
@@ -7,12 +7,13 @@
 
 use crate::{config::FinalStateConfig, error::FinalStateError, state_changes::StateChanges};
 
+use anyhow::{anyhow, Result as AnyResult};
 use massa_async_pool::AsyncPool;
 use massa_db_exports::EXECUTION_TRAIL_HASH_PREFIX;
 use massa_db_exports::{
     DBBatch, MassaIteratorMode, ShareableMassaDBController, ASYNC_POOL_PREFIX,
-    CHANGE_ID_DESER_ERROR, CYCLE_HISTORY_PREFIX, DEFERRED_CREDITS_PREFIX,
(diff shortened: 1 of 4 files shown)


massa-final-state/src/final_state.rs
```diff
@@ -579,17 +580,22 @@ impl FinalState {
     ///
     /// Panics if the new slot is not the one coming just after the current one.
     pub fn finalize(&mut self, slot: Slot, changes: StateChanges) {
-        let cur_slot = self.db.read().get_change_id().expect(CHANGE_ID_DESER_ERROR);
+        self._finalize(slot, changes).unwrap()
```
Leo-Besancon, line 583: Won't this unwrap interfere with error handling? (e.g. if the pos_state apply changes throws an error, would that be picked up?)

massa-final-state/src/final_state.rs
```diff
@@ -811,3 +829,460 @@ impl FinalState {
...
+            .err()
+            .unwrap()
+            .to_string()
+            .contains("while the current slot is"));
+
+        assert_eq!(fstate.get_slot(), initial_slot);
+
+        let res = fstate._finalize(ok_next_slot, changes.clone());
+
+        assert!(res.is_err());
```
Leo-Besancon, line 1050: Here, with "OK" data, we should still get a POS error? Seems a bit weird for me, maybe rename the test to `test_final_state_finalize_slot_check()` ?

Leo-Besancon (commented):
Nice coverage improvements! I went quickly over this PR, it looks fine to me. Just check my comments and ensure the error handling stays the same with a running node!

massa-final-state/src/final_state.rs
sydhds: _finalize returns an error on pos_state.apply_changes and the previous code did an expect so I would say that it's ok no?

massa-final-state/src/final_state.rs
Leo-Besancon: It's mainly just a readability issue for me. I understand why it errors out, but you should explain that in comments and/or rename the test so that it is clear that it should not finalize successfully.

massa-final-state/src/final_state.rs
Leo-Besancon: Yes, I don't know why but I thought the error message would not show up :)

massa-final-state/src/final_state.rs
sydhds: I've added a small comment so a future reader can easily understand why it returns an error :)

Leo-Besancon (approved):
thanks for the explanation on my questions ;)

sydhds:
bors merge

PR #4452, opened 2023-10-05
Add initial unit tests for massa db

sydhds:
  * [ ] make tests compile
  * [ ] make tests pass 

### DIFF
--- a/massa-db-worker/src/lib.rs
+++ b/massa-db-worker/src/lib.rs
@@ -1,3 +1,51 @@
+// Copyright (c) 2022 MASSA LABS <info@massa.net>
+
+//! # General description
+//!
+//! MassaDB is a wrapper around:
+//! * A RocksDB database (on Disk)
+//! * Some caches (on RAM).
+//! * a config
+//!
+//! # RocksDB
+//!
+//! RocksDB stores keys and values, which are arbitrarily-sized byte streams (aka vec<u8> or &[u8]).
+//! It supports both point lookups and range scans.
+//!
+//! For MassaDB, we use 3 rocksdb column:
+//! * state: all data used to compute the db hash (or final state hash)
+//! * versioning: partial MIP store data see Versioning doc section: "MipStore and Final state hash"
+//! * metadata: final state hash + slot
+//!
+//! Note that data is stored with a prefix (see constants.rs in massa-db-exports).
+//! For instance, a ledger update, will be stored (in column: 'state') as:
+//! * key: LEDGER_PREFIX+Address+[...] (serialized as bytes)
+//! * value: Amount (serialized as bytes)
+//!
+//! # DB hash
+//!
+//! Whenever something is stored on the column 'state', the db hash is updating using Xor
+//! (For more detail: HashXof).
+//! This hash is often referred as 'final state hash'.
+//!
+//! # Caches
+//!
+//! A cache of db changes is kept in memory allowing to easily stream it
+//! (by streaming, we means: sending it to another massa node (aka bootstrap))
+//! There is 2 separate caches: one for 'state' and one for 'versioning'
+//!
+//! These caches is stored as a key, value: slot -> insertion_data|deletion_data.
+//!
+//! # Streaming steps
+//!
+//! 1- Streaming always starts by reading data from disk (from rocksdb). Data is send by chunk.
+//! 2- As this process can be slow (e.g. db size is high, network is slow), updates can happen simultaneously
+//!    (as new slots are processed by the node, data can be updated)
+//!    thus the (DB) caches are updated so as the streaming process is ongoing, we can easily only send
+//!    the updates (by querying only the cache)
+//! 3- Even after this process is finished (and as other things like consensus data are streamed),
+//!    we can send the updates
+
 mod massa_db;
 
 pub use crate::massa_db::*;
--- a/massa-db-worker/src/lib.rs
+++ b/massa-db-worker/src/lib.rs
@@ -13,7 +13,7 @@
 //! It supports both point lookups and range scans.
 //!
 //! For MassaDB, we use 3 rocksdb column:
-//! * state: all data used to compute the db hash (or final state hash)
+//! * state: all data for (async pool, executed ops/de, ledger ...) and used to compute the db hash
 //! * versioning: partial MIP store data see Versioning doc section: "MipStore and Final state hash"
 //! * metadata: final state hash + slot
 //!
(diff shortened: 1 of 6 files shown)


massa-db-worker/src/massa_db.rs
```diff
@@ -837,3 +850,803 @@ impl MassaDBController for RawMassaDB<Slot, SlotSerializer, SlotDeserializer> {
         self.recompute_db_hash()
```
Leo-Besancon, line 850: IIRC, recompute_db_hash() should not be used in the codebase (this statement should be checked).
It was introducted when computing the hash took too much time with LSM trees.

Could you check that, and either remove it (in this PR or another) or explain in comments?

massa-db-worker/src/massa_db.rs
```diff
@@ -837,3 +850,803 @@ impl MassaDBController for RawMassaDB<Slot, SlotSerializer, SlotDeserializer> {
...
+        // assert!(dump_column(db_.clone(), "versioning").is_empty());
+        assert!(dump_column(db.clone(), "versioning").is_empty());
+    }
+
+    #[test]
+    fn test_basics_2() {
+        // 1- Init a db + check initial hash
+        // 2- Add some data
+        // 3- Read it
+        // 4- Remove data
```
Leo-Besancon, line 1010: The difference here with the previous test is the delete with prefix? If so, describe it in this comment

massa-db-worker/src/massa_db.rs
```diff
@@ -837,3 +850,803 @@ impl MassaDBController for RawMassaDB<Slot, SlotSerializer, SlotDeserializer> {
...
+        let batch_key_1 = vec![1, 2, 3];
+        let batch_value_1 = vec![4, 5, 6];
+        let batch = DBBatch::from([(batch_key_1.clone(), Some(batch_value_1.clone()))]);
+        let versioning_batch = DBBatch::from([(vec![10, 20, 30], Some(vec![127, 128, 254, 255]))]);
+        let slot_1 = Slot::new(1, 0);
+        let mut guard = db.write();
+        guard.write_batch(batch.clone(), versioning_batch.clone(), Some(slot_1));
+        drop(guard);
+
+        // Add data 2 (at slot 1)
```
Leo-Besancon, line 1316: you insert at slot 2 here i think

massa-db-worker/src/massa_db.rs
```diff
@@ -837,3 +850,803 @@ impl MassaDBController for RawMassaDB<Slot, SlotSerializer, SlotDeserializer> {
...
+        // Stream using StreamingStep::Finished
+
+        let last_state_step: StreamingStep<Vec<u8>> =
+            StreamingStep::Finished(Some(batch_key_2.clone()));
+        let stream_batch_ = db
+            .read()
+            .get_batch_to_stream(&last_state_step, Some(slot_1));
+        let stream_batch = stream_batch_.unwrap();
+
+        // FIXME?: new_elements is empty, everything is on updates_on_previous_elements
```
Leo-Besancon, line 1632: Ah, I think we should add a comment somewhere (in lib.rs or in StreamBatch definition), indeed here "new" keys will be streamed in the updates field.
Updates represent "everything that changed since the last change_id streamed, up to a certain key".
This definition also applies to keys that were not in the DB beforehand.

Leo-Besancon (commented):
I think the tests made here are pretty good!

I don't know what else could be tested, apart from basic things delegated to RocksDB (.flush, etc.), that probably is not necessary.

 We just have to make sure that these tests and the bootstrap tests made by @litchipi cover well the "limit" between the two modules, without overlaping too much.

massa-db-worker/src/massa_db.rs
sydhds: Fixed

massa-db-worker/src/massa_db.rs
sydhds: I would do it in another PR so we keep this PR focused on unit tests :)

Leo-Besancon (dismissed):

massa-db-worker/src/lib.rs
```diff
@@ -1,3 +1,51 @@
...
+//!
+//! For MassaDB, we use 3 rocksdb column:
+//! * state: all data used to compute the db hash (or final state hash)
+//! * versioning: partial MIP store data see Versioning doc section: "MipStore and Final state hash"
+//! * metadata: final state hash + slot
+//!
+//! Note that data is stored with a prefix (see constants.rs in massa-db-exports).
+//! For instance, a ledger update, will be stored (in column: 'state') as:
+//! * key: LEDGER_PREFIX+Address+[...] (serialized as bytes)
+//! * value: Amount (serialized as bytes)
```
Eitu33, line 23: Think it would be nice to explain that final state information is all stored at the same level with prefixes working as a way to individual parts of data together. Here you talk about prefixes but don't explain why they are there.

massa-db-worker/src/lib.rs
```diff
@@ -1,3 +1,51 @@
...
+//!
+//! # Streaming steps
+//!
+//! 1- Streaming always starts by reading data from disk (from rocksdb). Data is send by chunk.
+//! 2- As this process can be slow (e.g. db size is high, network is slow), updates can happen simultaneously
+//!    (as new slots are processed by the node, data can be updated)
+//!    thus the (DB) caches are updated so as the streaming process is ongoing, we can easily only send
+//!    the updates (by querying only the cache)
+//! 3- Even after this process is finished (and as other things like consensus data are streamed),
+//!    we can send the updates
```
Eitu33, line 47: Seems weird to me that you ask the reader to refer to other doc / files for information that is actually related to the DB module but then explain bootstrap steps here .

massa-db-worker/src/lib.rs
```diff
@@ -1,3 +1,51 @@
...
+//! * Some caches (on RAM).
+//! * a config
+//!
+//! # RocksDB
+//!
+//! RocksDB stores keys and values, which are arbitrarily-sized byte streams (aka vec<u8> or &[u8]).
+//! It supports both point lookups and range scans.
+//!
+//! For MassaDB, we use 3 rocksdb column:
+//! * state: all data used to compute the db hash (or final state hash)
```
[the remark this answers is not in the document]
sydhds: Updated

massa-db-worker/src/lib.rs
sydhds: I've added some doc on hash with xor with a small example

sydhds:
@damir if you can have a quick look at the doc :)

Leo-Besancon (approved):

sydhds:
bors merge

Issue #4430, opened 2023-09-27
Bootstrap


Leo-Besancon:

litchipi:
If we start a labnet without keeping the ledger, just a normal one without any operations on the branch `test_bootstrap_big`, the bootstrap process will work but the hashes of `final_state` will be different.

It seams that some changes I added in the branch `test_bootstrap_big` is responsible for the hash difference

litchipi:
I figured out why the `final state` hash are different when the node generates a brand new ledger, it works now.
However, it's still different when a ledger is generated using `massa-ledger-editor`

litchipi:
The hash mismatch occurs when launching a labnet on `main` with crafted ledger.
After I tried with smaller keys / values in datastore (length 9000), it still does it, so I don't think it's a matter of serialization.
@Leo-Besancon Could it be because of [this piece of code](https://github.com/massalabs/massa/pull/4420/files#diff-535b75052259fc1e2f15ad89ecd55326f762d8a118074d573377d09116533761R425) we use when crafting the ledger ?
How could I modify the `massa-ledger-editor` in order to craft a ledger that doesn't require this check bypass ?

Leo-Besancon:
> The hash mismatch occurs when launching a labnet on `main` with crafted ledger. After I tried with smaller keys / values in datastore (length 9000), it still does it, so I don't think it's a matter of serialization. @Leo-Besancon Could it be because of [this piece of code](https://github.com/massalabs/massa/pull/4420/files#diff-535b75052259fc1e2f15ad89ecd55326f762d8a118074d573377d09116533761R425) we use when crafting the ledger ? How could I modify the `massa-ledger-editor` in order to craft a ledger that doesn't require this check bypass ?

Nice research! I think the easiest to ensure that all necessary keys are found would be to start from a working ledger (built normally by the node), and then only add your crafted keys.
This way, the trail_hash, cycle_history and so one would be in your crafted ledger, and it's less likely to cause a problem.

litchipi:
That's actually what I already do
- A while ago I started a node normally, and copied the storage into a local folder `A`.
- To craft a new ledger, I copy the content of the folder `A` to a new folder `B`
- Using `massa-ledger-editor`, I add data to the ledger stored in ledger `B`
- Before I launch the labnet, I copy the folder `B` to the storage of the massa-node.
- 
(Note that most of what I listed above is automated so I don't forget a step while doing it)

[...]

PR #4420, opened 2023-09-20
Allow big ledger bootstrapping

litchipi:
# Big ledger bootstrap

Changes in the code allowing the node to bootstrap from a big ledger.
Some changes on this PR may not be useful and has to be stripped, if you see some left please tell me.

## Tests for bootstrap using big ledger

- 111591403520 bytes  ( 104G ) :     1h 32 mins 53 secs
- 53717592215 bytes ( 52G ) :      46 mins 30 secs

It bootstraps to an average bytes per second of `19638597 bytes`, and seams to be increasing linearly.

- 200 G  ->  3h 2m 15s
- 500 G  ->  7h 35m 37s
- 800 G  ->  12h 9m 0s
- 1 To -> 15h 33m 7s
- 2 To -> 1j 7h 6m 14s
- 5 To -> 3j 5h 45m 36s
- 8 To -> 5j 4h 24m 58s
- 10 To -> 6j 11h 31m 12s

## Limits to discuss

Some configuration about the limits to set for the deserialization, timeouts, etc ... has to be discussed.
As they have been problematic during the bootstrap of 100G, they have to be adapted to the max size of the packets we can support (see https://github.com/massalabs/massa/issues/4406), but also about how long it takes to stream these packets on the network.

## Testing setup used
- Using the branch `fill_ledger` of [massa-ledger-editor](https://github.com/massalabs/massa-ledger-editor), fill an **existing** ledger to the wanted size
- Start a labnet with the necessary changes in `labnet.toml` (in addition to the existing info):
```
commit = "test_bootstrap_big"
genesis_timestamp_mins = 15

[servers.labnet0]
initial_rolls = 100
bootstrap_server = true
exec_args = [ "--keep-ledger" ]
add_script_pre_launch = [ "rsync --delete -r /path/to/edited/ledger ./storage/ledger" ]

[servers.labnet1]
initial_rolls = 0
bootstrap_server = false
add_script_pre_launch = [
  "sleep $(( 15 * 60 ))",
]
```
This ensures that `labnet1` boostraps to `labnet0`, and `labnet0` uses the edited ledger.
- Watch logs of `labnet1` till you see blocks arriving.

Issue #4406, opened 2023-09-14
Bootstrap fails when we push the ledger to its storage limits


litchipi:
**Describe the bug**
When inserting data (specifically a **bytecode** associated with an address) to the limit of the ledger's capability, the serialization of the ledger will succeed (bootstrap server will be OK), but the deserialization will fail.

The failure will come from the deserialization of:
```
BootstrapServerMessage::state_part::new_elements
```
The serialization process expects that the length of the new_elements `BTreeMap` is added before the data itself, and is formatted using a `u64 varint`, in this case, this is what fails to deserialize.

**Concerns**
This is a **High security problem** as an attacker could make any future bootstrapping process fail by simply storing a big enough bytecode, creating a DoS over the whole network.

It's an asymmetry between the serialization and deserialization process.

**To Reproduce**
Start a labnet using a custom ledger containing an entry for an address whose bytecode length is `> 9_999_990`

**Expected behavior**
Suppose to accept bytecode with length `> 10_000_000`
A check should detect that the data is too long

litchipi:
@AurelienFT If  you have a list on the issue board for security issues, this one could go into it. If you don't have this list, consider making one, that'd be useful for later :smiley:

AurelienFT:
Done !

PR #4453, opened 2023-10-06
Bootstrap improvements

litchipi:
Take most of the work done in #4420 to merge it in `main`

Replace the `MAX_BOOTSTRAP_NEW_ELEMENTS` with a size limit `MAX_BOOTSTRAP_NEW_ELEMENTS_SIZE` so that voluminous messages (because of big datastore values or bytecodes) can be transferred without going through the limit.

Limits on the timeouts, and message size had to be raised in order to allow messages crafted to be very voluminous to be transferred without any problem.
They may not have to be reduced down, but high enough to allow a big ledger to bootstrap successfully.
This will have to be done after all the issues documented in #4430 and #4406


massa-models/src/config/constants.rs
```diff
@@ -192,7 +192,7 @@ pub const MAX_RNG_SEED_LENGTH: u32 = PERIODS_PER_CYCLE.saturating_mul(THREAD_COU
 //
 
 /// Max message size for bootstrap
-pub const MAX_BOOTSTRAP_MESSAGE_SIZE: u32 = 1048576000;
+pub const MAX_BOOTSTRAP_MESSAGE_SIZE: u32 = u32::MAX; //1048576000;
```
Leo-Besancon, line 195: Will you be able to reduce this limit after more thourough testing on sizes you're doing currently?

massa-node/base_config/config.toml
```diff
@@ -367,17 +367,17 @@
     # [optional] port on which to listen for incoming bootstrap requests. You may need to change this to "0.0.0.0:port" if IPv6 is disabled system-wide.
     bind = "[::]:31245"
     # timeout to establish a bootstrap connection
-    connect_timeout = 15000
+    connect_timeout = 150000000000
     # timeout for providing the bootstrap to a connection
-    bootstrap_timeout = 1200000
+    bootstrap_timeout = 120000000000000
```
Leo-Besancon, line 372: I think we can't merge with those limits for connect_timeout and bootstrap_timeout, is it possible to narrow down the acceptable values and explain how those new limits are computed?

massa-models/src/config/constants.rs
```diff
@@ -211,7 +211,7 @@ pub const MAX_BOOTSTRAP_ASYNC_POOL_CHANGES: u64 = 100_000;
 /// Max bytes in final states parts
 pub const MAX_BOOTSTRAP_FINAL_STATE_PARTS_SIZE: u64 = 1_000_000_000;
 /// Max bytes in final states parts
-pub const MAX_BOOTSTRAPPED_NEW_ELEMENTS: u64 = 500;
+pub const MAX_BOOTSTRAPPED_NEW_ELEMENTS_SIZE: u64 = 100_000_000;
```
Leo-Besancon, line 214: I saw that in main you use the constant above (MAX_BOOTSTRAP_FINAL_STATE_PARTS_SIZE).
Do we still need both constants?

Leo-Besancon (commented):
Nice general improvements (<3 for the value -> key in serializer!).

I understand this PR let's you test more easily, but I think it's better if you can reduce the limits a bit to merge to main (even if you raise them x100, it's still better than a x10000 IMO :)).

massa-models/src/config/constants.rs
litchipi: One is used in `get_batch_to_stream`, the other in `get_versioning_batch_to_stream`

massa-node/base_config/config.toml
Leo-Besancon: Does the bootstrap timeout limit corresponds roughly to 6j 11h 31m 12s?

litchipi: No, I just added a bunch of zero
For now we could put a timeout of 1 day, that corresponds to `86400000` milliseconds.
I'll add the `update bootstrap timeout` on the list of issues regarding the bootstrapping process

litchipi: And I'll reduce the other timeouts as well

Leo-Besancon: Perfect, this seems good to me

massa-models/src/config/constants.rs
Leo-Besancon:
```        state_new_elements_length_deserializer: U64VarIntDeserializer::new(
                Included(0),
                Included(args.max_versioning_elements_size),
```
For me there are still some incoherence there! We probably should have a state_new_elements and a versioning_new_element deserializer, with each deserializer taking its own limit, no?

massa-models/src/config/constants.rs
litchipi: I made a test with the max sizes everywhere in the message generation.
Hope I didn't miss anything that could go beyond the values I used (even if I uses constants defined in the file)
The new version got an updated value for the max message size

massa-models/src/config/constants.rs
litchipi: Yes indeed, it's a bit confused in the code I thought it was just the naming but no, I'll check this out

Leo-Besancon: Resolving because it's no big deal, I still see a very small inconcistancy on the config. You pass around `MAX_BOOTSTRAP_VERSIONING_ELEMENTS_SIZE` in the config, but query `MAX_BOOTSTRAP_FINAL_STATE_PARTS_SIZE` directly (unless I'm mistaken, we should pass either both or none in the config?).

Leo-Besancon (dismissed):
Thanks! Looks good to me.

litchipi:
bors merge

massa-db-worker/src/massa_db.rs
```diff
@@ -287,7 +290,8 @@ where
             };
 
             for (serialized_key, serialized_value) in db_iterator.flatten() {
-                if new_elements.len() < self.config.max_new_elements {
+                new_elements_size += serialized_value.len();
+                if new_elements_size < self.config.max_final_state_elements_size {
```
Leo-Besancon, line 294: versioning here, no?

massa-node/base_config/config.toml
```diff
@@ -367,17 +367,17 @@
     # [optional] port on which to listen for incoming bootstrap requests. You may need to change this to "0.0.0.0:port" if IPv6 is disabled system-wide.
     bind = "[::]:31245"
     # timeout to establish a bootstrap connection
-    connect_timeout = 15000
+    connect_timeout = 300000 # 5 mins
```
Leo-Besancon, line 370: Just so I'm clear on this, why did we have to increase this? Is this timeout checkout for each message received? If so, 5 mins seem a bit high, no? (someone could idle for almost 5 mins and then send 1 message and idle again..)

litchipi: Yeah that's just a leftover from previous tests, I hit it sometimes when the server was really busy which is why I left it, but I guess it's OK to restore it back

Leo-Besancon: Ok understood, IMO we basically should approach this as "if the server is too busy and we hit this timeout, then either we have to accept less clients, or we should calibrate it better with the test-framework."

Leo-Besancon (dismissed):
LGTM now :)

litchipi:
bors merge

PR #4461, opened 2023-10-09
add better tests protocol

AurelienFT:
This PR add : 
- Tests on deserializers/serializers that measures lengths.
- Tests on handshake
- Remove the messages handlers fields from the `MassaHandshake` which was a duplicate

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -100,7 +100,7 @@ massa_wallet = { path = "./massa-wallet" }
 # Massa projects dependencies
 massa-proto-rs = {git = "https://github.com/massalabs/massa-proto-rs", "rev" = "b45cccca3fb09aa2bbee08aade080127521e39b8"}
 massa-sc-runtime = {git = "https://github.com/massalabs/massa-sc-runtime", "rev" = "4883fe075afca4dd4b7346ed7dff2d506323f9c7"}
-peernet = { git = "https://github.com/massalabs/PeerNet", "rev" = "d65f620fdf064c813c2abc582fbcc7c28969b579" }
+peernet = { git = "https://github.com/massalabs/PeerNet", "rev" = "8b0b27fbd720fcf0b120e5177402c5f66ed40ae7" }
 
 # Common dependencies
 transition = { git = "https://github.com/massalabs/transition.git", "rev" = "93fa3bf82f9f5ff421c78536879b7fd1b948ca75" }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -100,7 +100,7 @@ massa_wallet = { path = "./massa-wallet" }
 # Massa projects dependencies
 massa-proto-rs = {git = "https://github.com/massalabs/massa-proto-rs", "rev" = "b45cccca3fb09aa2bbee08aade080127521e39b8"}
 massa-sc-runtime = {git = "https://github.com/massalabs/massa-sc-runtime", "rev" = "4883fe075afca4dd4b7346ed7dff2d506323f9c7"}
-peernet = { git = "https://github.com/massalabs/PeerNet", "branch" = "main" }
+peernet = { git = "https://github.com/massalabs/PeerNet", "rev" = "5553ea96bbf1a3df2ae9cea7ea03c7ee2a4116cd" }
 
 # Common dependencies
 transition = { git = "https://github.com/massalabs/transition.git", "rev" = "93fa3bf82f9f5ff421c78536879b7fd1b948ca75" }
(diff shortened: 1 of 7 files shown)


massa-protocol-worker/src/handlers/block_handler/messages.rs
```diff
@@ -355,3 +355,160 @@ impl Deserializer<BlockMessage> for BlockMessageDeserializer {
...
+            block_id: BlockId::from_str("B12DvrcQkzF1Wi8BVoNfc4n93CD3E2qhCNe7nVhnEQGWHZ24fEmg")
+                .unwrap(),
+            block_info: super::BlockInfoReply::Operations(vec![]),
+        };
+        let mut buffer2 = Vec::new();
+        serializer.serialize(&message2, &mut buffer2).unwrap();
+        let (rest2, deserialized_message2) = deserializer
+            .deserialize::<DeserializeError>(&buffer2)
+            .unwrap();
+        assert_eq!(rest2.len(), 0);
```
modship, line 420: assert!(rest2.is_empty()) :see_no_evil:

modship (dismissed):

massa-protocol-worker/src/handlers/block_handler/messages.rs
AurelienFT: Fixed.

modship (dismissed):

massa-protocol-worker/src/handlers/block_handler/messages.rs
```diff
@@ -355,3 +355,160 @@ impl Deserializer<BlockMessage> for BlockMessageDeserializer {
...
+        match (deserialized_message2, message2) {
```
sydhds, line 421: Dumb question but why assert_eq! is not used here?

massa-protocol-worker/src/handlers/peer_handler/mod.rs
```diff
@@ -630,3 +624,80 @@ impl InitConnectionHandler<PeerId, Context, MessagesHandler> for MassaHandshake
...
+    use massa_signature::KeyPair;
+    use parking_lot::RwLock;
+    use peernet::{peer::InitConnectionHandler, transports::endpoint::Endpoint};
+
+    use crate::{context::Context, messages::MessagesHandler};
+
+    use super::models::PeerDB;
+
+    #[test]
+    fn test_handshake_working_behaviour() {
```
sydhds, line 644: Can we add more tests with invalid handshake + handshake that takes too much time here + handshake with close connection in the middle ?

sydhds (changes requested):

massa-protocol-worker/src/handlers/block_handler/messages.rs
modship: assert!(rest2.is_empty()) should be use here

AurelienFT: It's not a dumb question, it's me the dumb man 😆

massa-protocol-worker/src/handlers/peer_handler/mod.rs
AurelienFT: I have added invliad handshake and disconnect. "Takes too much time" isn't testable because it's managed by the timeout on the socket directly in peernet

Cargo.toml
```diff
@@ -100,7 +100,7 @@ massa_wallet = { path = "./massa-wallet" }
 # Massa projects dependencies
 massa-proto-rs = {git = "https://github.com/massalabs/massa-proto-rs", "rev" = "b45cccca3fb09aa2bbee08aade080127521e39b8"}
 massa-sc-runtime = {git = "https://github.com/massalabs/massa-sc-runtime", "rev" = "4883fe075afca4dd4b7346ed7dff2d506323f9c7"}
-peernet = { git = "https://github.com/massalabs/PeerNet", "branch" = "main" }
+peernet = { git = "https://github.com/massalabs/PeerNet", "rev" = "1d10f55ea883e685691ff527c2466f733ba3e1b2" }
```
modship, line 103: target main ?

modship (approved):

Cargo.toml
AurelienFT: No I don't think it's a great idea because if we push a new version and our IDE sync up with it and it has breaking change it will make the compilation fail. I see the rev kinda like a version number we just don't have version on peernet yet but if we had a 0.1.0 we will not point to main but to this version

sydhds (approved):

AurelienFT:
bors merge

AurelienFT:
bors merge

PR #4468, opened 2023-10-11
Add tests serialization

AurelienFT:
massa-serialization has 93% coverage on this PR.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3273,6 +3273,7 @@ dependencies = [
  "displaydoc",
  "nom",
  "num",
+ "paste",
  "thiserror",
  "unsigned-varint 0.7.2",
 ]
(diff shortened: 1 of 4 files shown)


sydhds (commented):
Looks good but I see that only u64 varint is tested here. Coud we use parametric test and test u16 / u32 varint too?

AurelienFT:
> Looks good but I see that only u64 varint is tested here. Coud we use parametric test and test u16 / u32 varint too?

As they are all generating using the same code I didn't think it was useful but can do. I don't know anything about parametric test will try otherwise I just make more tests

AurelienFT:
@sydhds I have made a macro to make parameterized test. I re-ask the review. Ty for the suggestion it was cool to do.

sydhds (dismissed):

Leo-Besancon (dismissed):
If you can add a quick comment above the macro to explain what it tests for each type, it would be better than reading through the macro, but the macro isn't horrible so it's good to be merged as is.

AurelienFT:
> If you can add a quick comment above the macro to explain what it tests for each type, it would be better than reading through the macro, but the macro isn't horrible so it's good to be merged as is.

Thank you. Done.

Leo-Besancon (approved):
Thanks :)

AurelienFT:
bors merge

PR #4470, opened 2023-10-11
Add test code coverage + checksums

aoudiamoncef:
  * [ ] make tests compile
  * [ ] make tests pass 


.github/workflows/ci.yml
```diff
@@ -201,19 +201,18 @@ jobs:
           repo-token: ${{ secrets.GITHUB_TOKEN }}
       - uses: actions-rs/cargo@v1
         with:
-          command: install
-          args: cargo-nextest
-      - uses: actions-rs/cargo@v1
-        with:
-          command: nextest
-          args: run --features testing --retries 10
```
AurelienFT, line 209: if you change this nearly 80% of our CI will fail, some tests are too sensible for the CI and so fail often. You can see it with the tests on this PR.

AurelienFT: The tools you are using takes more than the current space we have in the runners

aoudiamoncef: Yep, we need more powerful runners, I'll mark this PR as blocked

aoudiamoncef:
@damip we'll upload the coverage report at the end of testing phase. This PR will be merge when we got a better runners.

AurelienFT:
bors check

aoudiamoncef:
@AurelienFT I did a small change to this PR. We have to run the test coverage locally as It's very heavy. Then it'll be pushed automatically. 

I added release note files checksum which is very important for the mainet binaries

PR #4490, opened 2023-10-18
Remove not necessary clone in consensus

AurelienFT:
The communications with channels should only be done in the worker. The controller should only gather data from the state. By removing the communications with channels in the controller to place them to the worker we avoid making a clone of all the channels.

### DIFF
--- a/massa-consensus-worker/src/state/process_commands.rs
+++ b/massa-consensus-worker/src/state/process_commands.rs
@@ -5,7 +5,9 @@ use massa_consensus_exports::{
     error::ConsensusError,
 };
 use massa_logging::massa_trace;
-use massa_models::{block_header::SecuredHeader, block_id::BlockId, slot::Slot};
+use massa_models::{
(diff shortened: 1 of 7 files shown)


massa-consensus-worker/src/state/process_commands.rs
```diff
@@ -33,6 +35,11 @@ impl ConsensusState {
             return Ok(());
         }
 
+        let de_p = DenunciationPrecursor::from(&header);
```
sydhds, line 38: Not sure to understand why we add a DenunciationPrecursor here.

Would you mind to explain it and maybe add some doc here?

sydhds: Miss the removal in the controller sorry :-)

sydhds (approved):

damip (approved):
nice cleanup

AurelienFT:
bors merge

PR #4496, opened 2023-10-20
Move coins to FunctionCall only

aoudiamoncef:
  * [x] make tests compile
  * [x] make tests pass 


Leo-Besancon (approved):
This seems to do the job!

aoudiamoncef:
bors merge

Issue #4483, opened 2023-10-16
SC execution gas costs


Eitu33:
## Problem summary

We have multiple issues concerning gas costs and how we handle them in operations and ABIs related to smart-contracts.

## Current and target behaviours

|                          | CURRENT              | TARGET                      | CHANGED  |
|--------------------------|----------------------|-----------------------------|  :---:   |
| **CallSC (OP)**          |                      |                             |          |
| compilation              | NOT PAID             | NOT PAID                    |          |
| VM instantiation         | NOT PAID             | PAID IN MAX_GAS (UNIFIED CONSTANT)  | *|
| module instantiation     | PAID IN MAX_GAS OR COMPILATION_GAS | PAID IN MAX_GAS       | *|
| execution                | PAID IN MAX_GAS      | PAID IN MAX_GAS             |          | 
| **ExecuteSC (OP)**       |                      |                             |          |
| compilation              | PAID IN MAX_GAS      | PAID IN MAX_GAS             |          |
| VM instantiation         | NOT PAID             | PAID IN MAX_GAS (UNIFIED CONSTANT)  | *|
| module instantiation     | PAID IN MAX_GAS OR COMPILATION_GAS | PAID IN MAX_GAS       | *|
| execution                | PAID IN MAX_GAS      | PAID IN MAX_GAS             |          |
| **Call (ABI)**           |                      |                             |          |
| compilation              | NOT PAID             | NOT PAID                    |          |
| VM instantiation         | PAID IN ABI GAS      | PAID IN CALL GAS (UNIFIED CONSTANT) | *|
| module instantiation     | PAID IN ABI GAS      | PAID IN CALL GAS            |         *|
| execution                | PAID IN CALL GAS     | PAID IN CALL GAS            |          |
| base gas cost of the ABI | PAID IN ABI GAS      | PAID IN ABI GAS             |          |
| **LocalExecution (ABI)** |                      |                             |          |
| compilation              | NOT PAID             | PAID IN CALL GAS            |         *|
| VM instantiation         | PAID IN ABI GAS      | PAID IN CALL GAS (UNIFIED CONSTANT) | *|
| module instantiation     | PAID IN ABI GAS      | PAID IN CALL GAS            |         *|
| execution                | PAID IN CALL GAS     | PAID IN CALL GAS            |          |
| base gas cost of the ABI | PAID IN ABI GAS      | PAID IN ABI GAS             |          |
| **CreateSC (ABI)**       |                      |                             |          |
[...]

PR #4497, opened 2023-10-24
Instrument massa-execution-worker to print slot execution time (for f…

bilboquet:
…unctonal tests)

Signed-off-by: Jean-François <jfm@laposte.net>* [ ] document all added functions
  * [ ] make tests compile
  * [ ] make tests pass 


massa-models/src/operation.rs
```diff
@@ -972,13 +970,13 @@ impl SecureShareOperation {
...
+    pub fn get_gas_usage(&self, gas_usage: u64) -> u64 {
         match &self.content.op {
             OperationType::ExecuteSC { max_gas, .. } => *max_gas,
             OperationType::CallSC { max_gas, .. } => *max_gas,
-            OperationType::RollBuy { .. } => 0,
-            OperationType::RollSell { .. } => 0,
-            OperationType::Transaction { .. } => 0,
+            OperationType::RollBuy { .. } => gas_usage,
+            OperationType::RollSell { .. } => gas_usage,
+            OperationType::Transaction { .. } => gas_usage,
```
Leo-Besancon, line 979: Maybe the designed changed in the meantime, but I though that the gas_usage would also be added for the ExecuteSC and CallCS, no?

Because there are still some things we do when including SC operations that would not be taken into accout in the (user declared) max_gas.

bilboquet: You're right!!! I've addressed, so far, only the basic operations.
Partly because:
1. I was (am) not that satisfied by the design (how the `gas_usage` in injected from it's constant definition to its point of use). But doing better would involve too big refactor.
2. I have to check if adding `gas_usage` does not conflict with what Thomas is doing #4483

Leo-Besancon: Ok it makes sence, thanks!

PR #4505, opened 2023-10-30
Fix smart-contracts gas payments

Eitu33:
closes https://github.com/massalabs/massa/issues/4483


massa-module-cache/src/controller.rs
```diff
@@ -122,12 +132,20 @@ impl ModuleCache {
...
 
-    /// Load a cached module for execution and check its validity for execution
+    /// Load a cached module for execution and check its validity for execution.
+    /// Also checks that the provided execution gas is enough to pay for the instance creation cost.
     pub fn load_module(
         &mut self,
         bytecode: &[u8],
         execution_gas: u64,
-    ) -> Result<RuntimeModule, CacheError> {
+    ) -> Result<(RuntimeModule, u64), CacheError> {
```
Eitu33, line 141: `load_module` ended up not changing anything to the input gas that is later returned but this does still provide multiple things:
* Uniformity with `load_tmp_module` which actually subs costs.
* Simplifies the dev usage after using a cache load. Before the used gas for runtime execution was different in every context, now it's only the return value after loading.
* As module cache functions are always called before any execution (even arbitrary ones) adding pre-execution gas costs becomes a simple addition

massa-execution-worker/src/context.rs
```diff
@@ -459,6 +459,16 @@ impl ExecutionContext {
 
     /// Creates a new smart contract address with initial bytecode, and returns this address
     pub fn create_new_sc_address(&mut self, bytecode: Bytecode) -> Result<Address, ExecutionError> {
+        // pay the Cranelift compilation cost
+        let cost = self.config.gas_costs.cl_compilation_cost;
+        if let Some(remaining) = self.max_gas.checked_sub(cost) {
```
Eitu33, line 464: If you wonder why the costs deduction have been placed where they are, the though process was (in no specific order) that they must be in a place where:
* they are in lowest function possible
* its logical for them to located in
* it's hard for a developper to make a mistake leading to erroneous cost handling
* a developper does not have to constantly think about them when making a change to execution

Eitu33: This is outdated as I realised just now that the context `max_gas` value actually never held anything. Moved the CL compilation payment back to ABI costs and will update the doc accordingly. The statement still holds for the cache payments.

massa-module-cache/src/controller.rs
Eitu33: For ref: https://github.com/massalabs/massa/pull/4505#discussion_r1376135529

massa-execution-worker/src/context.rs
Eitu33: I am not happy with the fact that CL compilation are still paid in ABI costs (lack of better alternatives), this is something that we will have to remember when making modifications.

massa-execution-worker/src/interface_impl.rs
```diff
@@ -278,10 +277,23 @@ impl Interface for InterfaceImpl {
     ///
     /// # Returns
     /// A `massa-sc-runtime` compiled module
```
damip, line 279: document the return value

massa-execution-worker/src/interface_impl.rs
```diff
@@ -278,10 +277,23 @@ impl Interface for InterfaceImpl {
...
-    fn get_module(&self, bytecode: &[u8], limit: u64) -> Result<RuntimeModule> {
+    fn get_module(&self, bytecode: &[u8], limit: u64) -> Result<(RuntimeModule, u64)> {
```
damip, line 280: limit -> gas_limit

massa-execution-worker/src/interface_impl.rs
```diff
@@ -278,10 +277,23 @@ impl Interface for InterfaceImpl {
...
-        let module = context.module_cache.write().load_module(bytecode, limit)?;
-        Ok(module)
+        let (module, remaining_gas) = context.module_cache.write().load_module(bytecode, limit)?;
+        Ok((module, remaining_gas))
+    }
+
+    /// Compile and return a temporary module
+    ///
+    /// # Returns
+    /// A `massa-sc-runtime` compiled module
```
damip, line 289: document return value

massa-execution-worker/src/interface_impl.rs
```diff
@@ -278,10 +277,23 @@ impl Interface for InterfaceImpl {
...
+    fn get_tmp_module(&self, bytecode: &[u8], limit: u64) -> Result<(RuntimeModule, u64)> {
```
damip, line 290: limit -> gas_limit

massa-execution-worker/src/context.rs
```diff
@@ -919,6 +919,16 @@ impl ExecutionContext {
         address: &Address,
         bytecode: Bytecode,
     ) -> Result<(), ExecutionError> {
+        // pay the Cranelift compilation cost
+        let cost = self.config.gas_costs.cl_compilation_cost;
+        if let Some(remaining) = self.max_gas.checked_sub(cost) {
+            self.max_gas = remaining
```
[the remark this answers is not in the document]
Eitu33: Outdated code, removed it

massa-module-cache/src/controller.rs
```diff
@@ -55,16 +58,21 @@ impl ModuleCache {
                 warn!("compilation of module {} failed with: {}", hash, e);
                 ModuleInfo::Invalid
             }
-        }
+        };
+        let end = Instant::now();
+        let elapsed = end.duration_since(start).as_micros();
+        debug!("TIMER: cl compilation took {} μs", elapsed);
```
damip, line 64: remove

massa-module-cache/src/controller.rs
```diff
@@ -55,16 +58,21 @@ impl ModuleCache {
...
     pub fn save_module(&mut self, bytecode: &[u8]) {
         let hash = Hash::compute_from(bytecode);
-        if let Some(hd_module_info) =
-            self.hd_cache
-                .get(hash, self.cfg.compilation_gas, self.cfg.gas_costs.clone())
-        {
+        if let Some(hd_module_info) = self.hd_cache.get(
+            hash,
+            self.cfg.gas_costs.max_instance_cost,
+            self.cfg.gas_costs.clone(),
```
damip, line 74: can we pass by reference to avoid cloning the whole thing ?

massa-module-cache/src/controller.rs
```diff
@@ -80,6 +88,7 @@ impl ModuleCache {
 
     /// Set the initialization cost of a cached module
     pub fn set_init_cost(&mut self, bytecode: &[u8], init_cost: u64) {
+        warn!("set_init_cost {}", init_cost);
```
damip, line 91: remove

massa-module-cache/src/controller.rs
```diff
@@ -122,12 +132,20 @@ impl ModuleCache {
         }
     }
 
-    /// Load a cached module for execution and check its validity for execution
+    /// Load a cached module for execution and check its validity for execution.
+    /// Also checks that the provided execution gas is enough to pay for the instance creation cost.
     pub fn load_module(
```
damip, line 137: document return values

damip:
One small thing that seems a bit inconsistent to me: 
* as a web3 dev, when I perform a call using a CallSC op => in the max_gas I need to fill, I don't need to take compilation into account but only the cost of "code execution"
* as a web3 dev, when I perform an execution using an ExecuteSC op on the same wasm => in the max_gas I need to fill, I need to take into account compilation costs. So it will require me to put a higher value in the "max_gas" field when using ExecuteSC than when using CallSC, even for the same code

This is because in the case of ExecuteSC, the compilation gas is taken from the max_gas provided by the sender.

Intuitively, I would expect it to be more consistent the following way:
* the ExecuteSC has a [base gas cost](https://github.com/massalabs/massa/pull/4510) that is higher than other ops because it would include SP compilation costs. That way, as a user, if I execute the same code using ExecuteSC or CallSC, the max_gas I need to fill will be the same.

@Eitu33 what do you think ?

massa-module-cache/gas-table.md
```diff
@@ -0,0 +1,61 @@
...
+
+| TYPE                     | PAYMENT                             |
+| ------------------------ | ----------------------------------- |
+| **CallSC (OP)**          |                                     |
+| compilation              | NOT PAID                            |
+| VM instantiation         | PAID IN MAX_GAS (UNIFIED CONSTANT)  |
+| module instantiation     | PAID IN MAX_GAS                     |
+| execution                | PAID IN MAX_GAS                     |
+| **ExecuteSC (OP)**       |                                     |
+| compilation              | PAID IN MAX_GAS                     |
```
damip, line 19: wouldn't it be more consistent if it is paid in the [base op cost](https://github.com/massalabs/massa/pull/4510) instead ?

massa-module-cache/gas-table.md
```diff
@@ -0,0 +1,61 @@
...
+| module instantiation     | PAID IN MAX_GAS                     |
+| execution                | PAID IN MAX_GAS                     |
+| **Call (ABI)**           |                                     |
+| compilation              | NOT PAID                            |
+| VM instantiation         | PAID IN CALL GAS (UNIFIED CONSTANT) |
+| module instantiation     | PAID IN CALL GAS                    |
+| execution                | PAID IN CALL GAS                    |
+| base gas cost of the ABI | PAID IN ABI GAS                     |
+| **LocalExecution (ABI)** |                                     |
+| compilation              | PAID IN CALL GAS                    |
```
damip, line 30: Wouldn't it be more consistent if this part is paid in the ABI gas ? note that it is equivalent in practice

massa-module-cache/src/controller.rs
Eitu33: I agree that this would be a nice change, but hd_cache requires the object because module deserializer requires it, which is later required by the engine init etc.

I'll open an issue on this & can take care of it later but it's a lot of changes in runtime & node for that to happen.

massa-module-cache/src/controller.rs
Eitu33: You want to remove this one but not the SP compilation timer? Or did you miss it and want both removed?

massa-module-cache/src/controller.rs
Eitu33: Done

Eitu33: Done

massa-module-cache/gas-table.md
Eitu33: Yes it would but that means we need to explicitly pay what need to be payed for anything that uses tmp modules. Whether its in the ABI costs or execution.rs we would have to never forget about it, with the current system anything that wants to use a tmp module automatically pays and checks compil & instance cost, since there are no cases where we want to do one but not the other. I agree that changing the ABI costs would make it more consistent when looking at the system from outside but inside it actually makes it more complex and prone to error. I have no strong opinion on this if you prefer one or the other let me know I'll change it just wanted you to be aware of that.

massa-module-cache/gas-table.md
Eitu33: Same as https://github.com/massalabs/massa/pull/4505#discussion_r1381893641, that means having different types of behaviours for every node operation using tmp modules when we only have 1 atm

Eitu33: Again no strong opinion, can change it if you want

Eitu33: Also this is related to your comment above: https://github.com/massalabs/massa/pull/4505#issuecomment-1792619101

This is also a good point, I'm okay with both will let you have the final say

massa-module-cache/src/controller.rs
damip: oh :) both if the calibration is done

massa-module-cache/gas-table.md
damip: Okay I understand your point. Then let's keep it as is

massa-module-cache/gas-table.md
damip: I tend to prioritize user-facing consistency over code factorization

massa-module-cache/gas-table.md
damip: After further thought on your previous comments, if we move the compilation base cost of Execute SC towards the constant operation cost, this should be changed as well to move towards the ABI cost

Eitu33: So we keep it as is yes?

Eitu33: Or do we move the compilation cost to OP and ABI costs respectively?

damip: I would vote for moving it to OP/ABI to prioritize user-facing consistency over code factorization

massa-module-cache/src/controller.rs
```diff
@@ -43,7 +43,7 @@ impl ModuleCache {
     fn compile_cached(&mut self, bytecode: &[u8], hash: Hash) -> ModuleInfo {
         match RuntimeModule::new(
             bytecode,
-            self.cfg.compilation_gas,
+            self.cfg.gas_costs.max_instance_cost,
```
bilboquet, line 46: this parameter is part of `gas_costs` and is always passed along with it up to its point of use.
Do we want to drop it?

massa-module-cache/src/controller.rs
```diff
@@ -61,10 +61,11 @@ impl ModuleCache {
     /// Save a new or an already existing module in the cache
     pub fn save_module(&mut self, bytecode: &[u8]) {
         let hash = Hash::compute_from(bytecode);
-        if let Some(hd_module_info) =
-            self.hd_cache
-                .get(hash, self.cfg.compilation_gas, self.cfg.gas_costs.clone())
-        {
+        if let Some(hd_module_info) = self.hd_cache.get(
+            hash,
+            self.cfg.gas_costs.max_instance_cost,
```
bilboquet, line 66: same as https://github.com/massalabs/massa/pull/4505/files#r1383040875

massa-module-cache/src/controller.rs
```diff
@@ -144,21 +161,37 @@ impl ModuleCache {
...
+        remaining
+            .checked_sub(self.cfg.gas_costs.max_instance_cost)
+            .ok_or(CacheError::LoadError(
+                "Provided max gas is below the default instance creation cost".to_string(),
+            ))?;
+
+        let module = RuntimeModule::new(
             bytecode,
-            limit,
+            remaining,
```
bilboquet, line 191: In previous uses of `RuntimeModule::new()` the value used for the `limit` parameter is `self.cfg.gas_costs.max_instance_cost`.
Here we have already subtracted this value to the parameter we are providing. It sounds weird to me.
I agree that `load_tmp_module` should return `remaining` (as it's computed here) but I think that we have to call `new()` with `self.cfg.gas_costs.max_instance_cost` (hence the comment above, the limit param can be dropped)

massa-module-cache/gas-table.md
Eitu33: Ok

Eitu33:
Test related update: https://github.com/massalabs/massa-unit-tests-src/pull/37

massa-module-cache/src/controller.rs
```diff
@@ -144,21 +148,25 @@ impl ModuleCache {
...
-    ) -> Result<RuntimeModule, CacheError> {
+    ) -> Result<(RuntimeModule, u64), CacheError> {
         debug!("load_tmp_module");
-        Ok(RuntimeModule::new(
-            bytecode,
-            limit,
-            self.cfg.gas_costs.clone(),
-            Compiler::SP,
-        )?)
+        limit
```
bilboquet, line 164: shouldn't this be 
`let limit = limit.checked_sub()`
(i think that the gas is not paid anymore here)

bilboquet (changes requested):

massa-module-cache/src/controller.rs
```diff
@@ -122,12 +116,22 @@ impl ModuleCache {
...
+    /// Returns the module and the remaining gas after loading.
     pub fn load_module(
         &mut self,
         bytecode: &[u8],
         execution_gas: u64,
-    ) -> Result<RuntimeModule, CacheError> {
+    ) -> Result<(RuntimeModule, u64), CacheError> {
+        // TODO: interesting but unimportant optim
+        // remove max_instance_cost hard check if module is cached and has a delta
+        execution_gas
```
bilboquet, line 130: same as below
should probably be `execution_gas = execution_gas.checked_sub()` else it seems the gas is not paid

massa-module-cache/src/controller.rs
Eitu33: It is paid, this is only a check, not supposed to actually take anything from the limit. Basically what happens now is:
1. Check that given gas is above max_instance_cost
2. Module compilation & deserialization is always done with max_instance_cost as input, which is supposed to be enough for the lock-out gas used to initialize. And it needs more the module will be considered invalid as we don't want to accept SC with insane instance cost (being above 2.1 mil is crazy). So we don't take limit input for compil & deser.
3. Call execution (run_function & run_main) with given gas, runtime refills the module with the correct amount as you can change the internal metering values after the instance step is done. So; compil payed by ABI costs / OP costs, VM & instance gas amount hard checked by cache, VM & instance + execution payed by runtime

Eitu33:
Linking this here as it's important information: https://github.com/massalabs/massa/pull/4505#discussion_r1384540339

massa-module-cache/src/controller.rs
Eitu33: Added doc on it

bilboquet: I was missing the very last part `VM & instance + execution payed by runtime`. Many thanks for the clarification.

bilboquet (dismissed):

damip:
Security issue:

```
    pub fn get_gas_usage(&self, base_operation_gas_cost: u64, sp_compilation_cost: u64) -> u64 {
        match &self.content.op {
            OperationType::ExecuteSC { max_gas, .. } => sp_compilation_cost + *max_gas,
            OperationType::RollBuy { .. } => 0,
            OperationType::RollSell { .. } => 0,
            OperationType::Transaction { .. } => 0,
        }
        .saturating_add(base_operation_gas_cost)
    }
```

Use `max_gas.saturating_add(sp_compilation_cost)` because max_gas is declared by the user and can overflow

damip (approved):

bilboquet (approved):

PR #4508, opened 2023-10-31
Update naming slo sequencer

### LINKED ISSUE #3919: Internal Audit: Massa-Execution-Worker slot_sequencer
In [`slot_sequencer`](https://github.com/massalabs/massa/blob/main/massa-execution-worker/src/slot_sequencer.rs) of [`Massa-Execution-Worker`](https://github.com/massalabs/massa/tree/main/massa-execution-worker):
- [x] Renaming legacy `css_final` to consensus_final (CF).
- [x] Renaming legacy `sce_final` to execution_final (EF).
- [x] Protection against having multiple instances of slot sequencer. - This is not needed since we do not create multiple instances, however, a protection is nice to have incase we can do it without adding much complexities.
- [x] Maybe `init()` can be called from `new()` since it is called only once. It is logicial to expect instance created from new() has empty queue so calling `init()` here makes more sense rather than `update()`. Parameters maybe checked to see if this is a bootstrap. In reality, after release of mainnet, it will always be bootstrap. - This is optional as it requires changes in the upstream.

Optimisation possible : 

Whenever the `update()` function is called, a new sequence is created. Maybe, is there a way to update the existing sequence to reduce the overhead.

The rest of the code seems well-written and well-commented. :)

AurelienFT:


damip (approved):

AurelienFT:
bors merge

AurelienFT:
bors merge

PR #4512, opened 2023-11-03
Mock Final State/Improve tests execution

AurelienFT:


massa-execution-worker/src/tests/scenarios_mandatories.rs
```diff
@@ -986,188 +901,164 @@ fn send_and_receive_transaction() {
...
+        .times(1)
+        .with(predicate::eq(Slot::new(3, 0)), predicate::always())
+        .returning(move |_, changes| {
+            let amount = changes
+                .ledger_changes
+                .get_balance_or_else(&address, || None)
+                .unwrap();
+            assert_eq!(
+                amount,
+                Amount::from_mantissa_scale(100, 0)
```
Eitu33, line 1040: Maybe specify that this is coming from the boilerplate, or set the default boilerplate value to 0

Eitu33 (approved):

massa-final-state/src/final_state.rs
```diff
@@ -821,13 +651,233 @@ impl FinalState {
         Ok(())
     }
 
-    /// Initialize the execution trail hash to zero.
-    pub fn init_execution_trail_hash_to_batch(&mut self, batch: &mut DBBatch) {
+    /// Initializes a `FinalState` from a snapshot. Currently, we do not use the final_state from the ledger,
+    /// we just create a new one. This will be changed in the follow-up.
```
Leo-Besancon, line 655: Realized that this part is outdated, can you remove it real quick? :)

massa-final-state/src/controller_trait.rs
```diff
@@ -0,0 +1,93 @@
...
+    fn get_pos_state(&self) -> &PoSFinalState;
+
+    /// Get pos state mut
+    fn get_pos_state_mut(&mut self) -> &mut PoSFinalState;
+
+    /// check if an operation is in the executed ops
+    fn executed_ops_contains(&self, op_id: &OperationId) -> bool;
+
+    /// Get the executed status ops
+    fn get_ops_exec_status(&self, batch: &[OperationId]) -> Vec<Option<bool>>;
```
Leo-Besancon, line 68: Why do you leak executed_ops related functions here? Why can't you do the same getter as with AsyncPool or PosFinalState ?

Leo-Besancon (approved):

massa-final-state/src/controller_trait.rs
AurelienFT: Because I believe how Async Pool and PoS is done it's badly done you shouldn't have a &ConcreteType in a trait it's a short term solution. The FInalState should just expose his methods and if necessary call another module (or make the module out of the final state). I made it correctly for the executed because there was only two methods and to test it

massa-final-state/src/final_state.rs
AurelienFT: DOne

PR #4514, opened 2023-11-06
Increase bootstrap read/write timeouts to 3 mins

Leo-Besancon:
When bootstraping max sized bootstrap message, it is easy to hit the timeouts.

1Gio with a timeout of 3 mins means we need at least ~5Mo/s connection to run a node.

  * [ ] make tests compile
  * [ ] make tests pass 


AurelienFT (approved):

Leo-Besancon:
bors merge

litchipi (approved):

PR #4535, opened 2023-11-14
Remove compatibility layer in tracing

### LINKED ISSUE #4478: Replace compatibility layer in tracing
In the context of the `tracing` crate in Rust, `tracing::debug` and `tracing::log::debug` refer to different macros used for emitting log events. Here's the difference between them:

1. **`tracing::debug!`**: This macro is provided directly by the `tracing` crate. It allows you to emit a log event with the `debug` level. You can use it like this:

   ```rust
   tracing::debug!("This is a debug log message");
   ```

   With this macro, you are using the `tracing` crate's native logging capabilities directly.

2. **`tracing::log::debug!`**: This macro is part of the `tracing-log` crate, which is a compatibility layer between the `log` crate and the `tracing` ecosystem. If you want to bridge the gap between traditional `log`-based logging and `tracing`, you can use `tracing-log` and its macros. With `tracing-log`, you can continue using the familiar `log` macros while leveraging `tracing`'s structured and contextual logging capabilities.

   ```rust
   tracing_log::log::debug!("This is a debug log message using tracing-log");
   ```

   Using `tracing-log::log::debug!` allows you to integrate `log`-style logging into your `tracing` instrumentation. This can be useful if you have existing code or libraries using the `log` crate and want to transition gradually to `tracing`.

In summary, if you are working exclusively with `tracing` and want to take advantage of its full capabilities, you can use `tracing::debug!`. If you are integrating `tracing` into an existing codebase that heavily uses the `log` crate, or if you want to bridge the gap between `log`-style logging and `tracing`, you can use `tracing-log::log::debug!`.

aoudiamoncef:
  * [ ] make tests compile
  * [ ] make tests pass 

closes #4478

### DIFF
--- a/massa-consensus-worker/src/controller.rs
+++ b/massa-consensus-worker/src/controller.rs
@@ -20,7 +20,7 @@ use massa_models::{
 use massa_storage::Storage;
 use parking_lot::RwLock;
 use std::sync::Arc;
-use tracing::log::{debug, trace, warn};
+use tracing::{debug, trace, warn};
 
 use crate::{commands::ConsensusCommand, state::ConsensusState};
 
--- a/massa-consensus-worker/src/manager.rs
+++ b/massa-consensus-worker/src/manager.rs
@@ -1,7 +1,7 @@
 use massa_channel::sender::MassaSender;
 use massa_consensus_exports::ConsensusManager;
 use std::thread::JoinHandle;
-use tracing::log::info;
+use tracing::info;
 
 use crate::commands::ConsensusCommand;
 
--- a/massa-consensus-worker/src/state/process.rs
+++ b/massa-consensus-worker/src/state/process.rs
@@ -22,7 +22,7 @@ use massa_models::{
 use massa_signature::PublicKey;
 use massa_storage::Storage;
 use massa_time::MassaTime;
-use tracing::log::{debug, info};
+use tracing::{debug, info};
 
 use crate::state::{
     clique_computation::compute_max_cliques,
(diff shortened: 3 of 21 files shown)


Leo-Besancon (approved):

AurelienFT:
A question about this changes, is it still possible to filter the logs with RUST_LOG standard log variable in this PR ?

aoudiamoncef:
I'll check this before merging

IMHO, we should work on an easy way to enable and disable dependencies logs inside Massa. I did this with @modship to debug an issue in jsonrpsee.

aoudiamoncef:
@AurelienFT could you test the behavior on main branch without this changes ? I get a weird result, I would like to get a confirmation.

AurelienFT:
> @AurelienFT could you test the behavior on main branch without this changes ? I get a weird result, I would like to get a confirmation.

When launching the node, we override the default value that could be in the env variable with our config.toml but in case of the test for example I use this standard variable (https://github.com/massalabs/massa/blob/5f56515208ecd9a8f66ce2c5b8c4f10dc9e5a6c2/massa-test-framework/src/lib.rs#L36) and so in the test that use the new framework we can choose level of print we want when we run our tests using the rust standard way.

aoudiamoncef:
@AurelienFT can you confirm that It's working as expected in main branch ? (launch in debug with sandbox) 

It's cool to choose, if we could inject a list of logger to enable with an env variable, It'll be awesome

AurelienFT:
I just answered you above. The behaviour on the node will work as we are not using the env variable but a config file. In the tests it's not the case. I just want to be sure that your changes doesn't prevents us to see the logs in the tests.

aoudiamoncef:
@AurelienFT after testing on `buildnet_wip` branch, It works

AurelienFT:
And on this branch it works also?

Leo-Besancon (approved):
I confirm that `RUST_LOG` var env is taken into account both with `tracing::` and `tracing::log::`

PR #4537, opened 2023-11-14
Fix bootstrap deser

Leo-Besancon:
  * [ ] make tests compile
  * [ ] make tests pass 

### DIFF
--- a/massa-bootstrap/src/messages.rs
+++ b/massa-bootstrap/src/messages.rs
@@ -29,7 +29,7 @@ use massa_serialization::{
 
 use massa_time::{MassaTime, MassaTimeDeserializer, MassaTimeSerializer};
 use nom::error::context;
-use nom::multi::{length_count, length_data};
+use nom::multi::{length_count, length_data, length_value};
 use nom::sequence::tuple;
 use nom::Parser;
 use nom::{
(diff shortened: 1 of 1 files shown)


massa-bootstrap/src/tests/messages.rs
```diff
@@ -178,14 +178,6 @@ fn test_serialize_error_cases_servermsg() {
...
-            // TODO    Remove once limits for deser of final_state_parts is by bytes, not length
-            continue;
-        }
-        if n == 8 {
-            // TODO    Remove once limits for deser of versionning_parts is by bytes, not length
-            continue;
-        }
         println!("Testing {n} faulty case");
         let mut bytes = Vec::new();
         let msg = BootstrapServerMessage::generate_faulty(&mut rng, n);
```
sydhds, line 183: Using n with just a number is difficult to read. Maybe use an enum here? So it would be easier to read and understand why we skip some of them for now...

massa-bootstrap/src/messages.rs
```diff
@@ -207,58 +206,72 @@ impl Serializer<BootstrapServerMessage> for BootstrapServerMessageSerializer {
...
+                        .serialize(value, &mut state_new_element_buffer)?;
+                }
                 self.u64_serializer.serialize(
-                    &(state_part
-                        .new_elements
+                    &state_new_element_buffer
                         .len()
                         .try_into()
-                        .expect("Overflow of new_elements len")),
+                        .expect("Overflow of state new_elements len"),
```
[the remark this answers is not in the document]
Leo-Besancon: I don't understand what you mean here. We can't know the length of `state_new_element_buffer` and catch an overflow until we serialize it above?

massa-bootstrap/src/tests/messages.rs
Leo-Besancon: Thanks, it's better indeed, done.

modship (approved):

PR #4552, opened 2023-11-21
Add versioninig key & value checks

sydhds:
  * [ ] make tests compile
  * [ ] make tests pass 

### DIFF
--- a/massa-final-state/src/final_state.rs
+++ b/massa-final-state/src/final_state.rs
@@ -10,12 +10,12 @@ use crate::{config::FinalStateConfig, error::FinalStateError, state_changes::Sta
 
 use anyhow::{anyhow, Result as AnyResult};
 use massa_async_pool::AsyncPool;
-use massa_db_exports::EXECUTION_TRAIL_HASH_PREFIX;
 use massa_db_exports::{
     DBBatch, MassaIteratorMode, ShareableMassaDBController, ASYNC_POOL_PREFIX,
     CYCLE_HISTORY_PREFIX, DEFERRED_CREDITS_PREFIX, EXECUTED_DENUNCIATIONS_PREFIX,
     EXECUTED_OPS_PREFIX, LEDGER_PREFIX, MIP_STORE_PREFIX, STATE_CF,
 };
+use massa_db_exports::{EXECUTION_TRAIL_HASH_PREFIX, MIP_STORE_STATS_PREFIX, VERSIONING_CF};
 use massa_executed_ops::ExecutedDenunciations;
 use massa_executed_ops::ExecutedOps;
 use massa_hash::Hash;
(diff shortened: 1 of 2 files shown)


massa-final-state/src/final_state.rs
```diff
@@ -635,7 +635,16 @@ impl FinalState {
...
             } else if serialized_key.starts_with(MIP_STORE_PREFIX.as_bytes()) {
-                // TODO: check MIP_STORE_PREFIX
+                if !self
+                    .mip_store
+                    .is_key_value_valid(&serialized_key, &serialized_value)
+                {
+                    warn!("Wrong key/value for MIP Store");
+                    return Err(anyhow!(
+                        "Wrong key/value for MIP Store serialized_key: {:?}, serialized_value: {:?}", 
+                        serialized_key, serialized_value
```
bilboquet, line 645: I don't know how big `serialized_key` and `serialized_value` can be.
If they can be bigger than a "limit", maybe we want to use `.get(0..limit)` to return only their beginning

massa-final-state/src/final_state.rs
```diff
@@ -648,6 +657,32 @@ impl FinalState {
...
+        {
+            if serialized_key.starts_with(MIP_STORE_PREFIX.as_bytes())
+                || serialized_key.starts_with(MIP_STORE_STATS_PREFIX.as_bytes())
+            {
+                if !self
+                    .mip_store
+                    .is_key_value_valid(&serialized_key, &serialized_value)
+                {
+                    warn!("Wrong key/value for MIP Store");
+                    return Err(anyhow!(
```
bilboquet, line 671: same as above

massa-final-state/src/final_state.rs
```diff
@@ -648,6 +657,32 @@ impl FinalState {
...
+                        "Wrong key/value for MIP Store serialized_key: {:?}, serialized_value: {:?}",
+                        serialized_key, serialized_value
+                    ));
+                }
+            } else {
+                warn!(
+                    "Key/value does not correspond to any prefix: serialized_key: {:?}, serialized_value: {:?}",
+                    serialized_key, serialized_value
+                );
+                return Err(anyhow!(""));
```
bilboquet, line 681: why an empty message?

massa-versioning/src/versioning.rs
```diff
@@ -1296,6 +1309,61 @@ impl MipStoreRaw {
         Ok(())
     }
 
+    // Final state
+    pub fn is_key_value_valid(&self, serialized_key: &[u8], serialized_value: &[u8]) -> bool {
+        self._is_key_value_valid(serialized_key, serialized_value)
+            .is_ok()
+    }
+
+    pub fn _is_key_value_valid(
```
bilboquet, line 1318: I'd vote for a different name `verify_key_value_validity` or maybe `check_key_value`
(I don't like much to have that two functions that differ only by a `_` in their name)

sydhds: this is the very same function and _is_key_value_valid only exists for and if in unit tests we'd want to check for a particular error...

massa-final-state/src/final_state.rs
sydhds: Done

massa-final-state/src/final_state.rs
sydhds: Fixed

massa-final-state/src/final_state.rs
bilboquet: fine

massa-versioning/src/versioning.rs
bilboquet: So shouldn't it be scoped for test non `pub`?
With `#[cfg(feature = "test-exports")]` the intent would be clearer.

massa-final-state/src/final_state.rs
bilboquet: I'm still seeing `return Err(anyhow!(""));`
What am I missing?

bilboquet (approved):
Last opened discussion is not critical.
So I approve.

Leo-Besancon (approved):
LGTM!

PR #4554, opened 2023-11-21
Feature/feature/versioning ser der harden 2

sydhds:
  * [ ] make tests compile
  * [ ] make tests pass 

### DIFF
--- a/massa-versioning/src/versioning_ser_der.rs
+++ b/massa-versioning/src/versioning_ser_der.rs
@@ -943,9 +943,7 @@ mod test {
 
         // A MIP info with a very long name
         let mi_2 = MipInfo {
-            name: std::iter::repeat("a")
-                .take(MIP_INFO_NAME_MAX_LEN as usize + 1)
-                .collect::<String>(),
+            name: "a".repeat(MIP_INFO_NAME_MAX_LEN as usize + 1),
             version: 2,
             components: BTreeMap::from([(MipComponent::Address, 1)]),
             start: MassaTime::from_millis(2),
--- a/massa-versioning/src/versioning_ser_der.rs
+++ b/massa-versioning/src/versioning_ser_der.rs
@@ -59,10 +59,7 @@ impl Serializer<MipInfo> for MipInfoSerializer {
         // name
         let name_len_ = value.name.len();
         let name_len = u32::try_from(name_len_).map_err(|_| {
-            SerializeError::GeneralError(format!(
-                "Cannot convert to name_len: {} to u32",
-                name_len_
-            ))
+            SerializeError::GeneralError(format!("Cannot convert name_len ({}) to u32", name_len_))
         })?;
         self.u32_serializer.serialize(&name_len, buffer)?;
         buffer.extend(value.name.as_bytes());
@@ -73,7 +70,7 @@ impl Serializer<MipInfo> for MipInfoSerializer {
         let components_len_ = value.components.len();
         let components_len = u32::try_from(components_len_).map_err(|_| {
             SerializeError::GeneralError(format!(
-                "Cannot convert to component_len: {} to u32",
+                "Cannot convert component_len ({}) to u32",
                 name_len_
             ))
         })?;
--- a/massa-versioning/src/versioning_ser_der.rs
+++ b/massa-versioning/src/versioning_ser_der.rs
@@ -111,11 +111,11 @@ impl MipInfoDeserializer {
             u32_deserializer: U32VarIntDeserializer::new(Included(0), Excluded(u32::MAX)),
             name_len_deserializer: U32VarIntDeserializer::new(
                 Included(0),
-                Excluded(MIP_INFO_NAME_MAX_LEN),
+                Included(MIP_INFO_NAME_MAX_LEN),
             ),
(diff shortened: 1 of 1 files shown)


massa-versioning/src/versioning_ser_der.rs
```diff
@@ -58,12 +58,6 @@ impl Serializer<MipInfo> for MipInfoSerializer {
...
         let name_len_ = value.name.len();
-        if name_len_ > MIP_INFO_NAME_MAX_LEN as usize {
-            return Err(SerializeError::StringTooBig(format!(
-                "MIP info name len is {}, max: {}",
-                name_len_, MIP_INFO_NAME_MAX_LEN
-            )));
-        }
         let name_len = u32::try_from(name_len_).map_err(|_| {
             SerializeError::GeneralError(format!(
                 "Cannot convert to name_len: {} to u32",
```
bilboquet, line 63: Isn't it "Cannot convert **from** name_len: {} to u32"?

massa-versioning/src/versioning_ser_der.rs
```diff
@@ -77,12 +71,6 @@ impl Serializer<MipInfo> for MipInfoSerializer {
...
         let components_len_ = value.components.len();
-        if components_len_ > MIP_INFO_COMPONENTS_MAX_ENTRIES as usize {
-            return Err(SerializeError::NumberTooBig(format!(
-                "MIP info cannot have more than {} components, got: {}",
-                MIP_STORE_MAX_ENTRIES, components_len_
-            )));
-        }
         let components_len = u32::try_from(components_len_).map_err(|_| {
             SerializeError::GeneralError(format!(
                 "Cannot convert to component_len: {} to u32",
```
bilboquet, line 76: same as above `to` -> `from` (I think)

bilboquet: And the parameter to the `format!` should be `compenets_len` (not `name_len_` as it is currently.

sydhds: Fixed

massa-versioning/src/versioning_ser_der.rs
```diff
@@ -77,15 +68,9 @@ impl Serializer<MipInfo> for MipInfoSerializer {
...
-            return Err(SerializeError::NumberTooBig(format!(
-                "MIP info cannot have more than {} components, got: {}",
-                MIP_STORE_MAX_ENTRIES, components_len_
-            )));
-        }
         let components_len = u32::try_from(components_len_).map_err(|_| {
             SerializeError::GeneralError(format!(
-                "Cannot convert to component_len: {} to u32",
+                "Cannot convert component_len ({}) to u32",
                 name_len_
```
bilboquet, line 74:
```suggestion
                components_len_
```

bilboquet (dismissed):
looks good to me (just take the suggestion into account)

massa-versioning/src/versioning_ser_der.rs
```diff
@@ -123,10 +109,14 @@ impl MipInfoDeserializer {
...
         Self {
             u32_deserializer: U32VarIntDeserializer::new(Included(0), Excluded(u32::MAX)),
-            len_deserializer: U32VarIntDeserializer::new(
+            name_len_deserializer: U32VarIntDeserializer::new(
                 Included(0),
                 Excluded(MIP_INFO_NAME_MAX_LEN),
             ),
+            components_len_deserializer: U32VarIntDeserializer::new(
+                Included(0),
+                Excluded(MIP_INFO_COMPONENTS_MAX_ENTRIES),
```
Leo-Besancon, line 118: I think in the codebase we almost always use Included() on MAX constants, is there a reason to exclude the max value here (and above)?

massa-versioning/src/versioning_ser_der.rs
```diff
@@ -905,40 +895,120 @@ mod test {
...
-        let mut buf = Vec::new();
-        let state_ser = ComponentStateSerializer::new();
-        state_ser.serialize(&st_1, &mut buf).unwrap();
+        {
+            let mut buf = Vec::new();
+            let mip_info_ser = MipInfoSerializer::new();
+            mip_info_ser.serialize(&mi_1, &mut buf).unwrap();
+
+            let mut mip_info_der = MipInfoDeserializer::new();
+            // Allow only a max of 2 components per MIP info
```
Leo-Besancon, line 933: Here 2 is excluded? We should change the comment

Leo-Besancon (commented):
Two small comments, otherwise all good!

massa-versioning/src/versioning_ser_der.rs
sydhds: Because the constant name is max entries and we start at 0 so I would say it makes sense to use Excluded here no?

massa-versioning/src/versioning_ser_der.rs
sydhds: Changed

massa-versioning/src/versioning_ser_der.rs
sydhds: Fixed

Leo-Besancon (approved):

Issue #4555, opened 2023-11-21
execute_read_only_bytecode Internal error


Thykof:
**Describe the bug**

On buildnetv27, the execute_read_only_bytecode call fails with 

```json
{
    "jsonrpc": "2.0",
    "error": {
        "code": -32603,
        "message": "Internal error"
    },
    "id": 1
}
```

**To Reproduce**

```
curl --location 'https://buildnet.massa.net/api/v2' \
--header 'Content-Type: application/json' \
--data '{
    "jsonrpc": "2.0",
    "id": 1,
    "method": "execute_read_only_bytecode",
    "params": [[{"max_gas":3980167295,"coins":"1.9253","fee":"0","address":"AU1wN8rn4SkwYSTDF3dHFY4U28KtsqKL1NnEjDZhHnHEy6cEQm53","bytecode":[0,97,115,109,1,009,97,112],"operation_datastore":[0]}]]
}'
```

*I truncated the bytecode.*

**Expected behavior**

If the call has an error, I should receive a clear error message.

bilboquet:
Just curious, does a regular call succeed?

Thykof:
Yes the regular call succeed.

PR #4570, opened 2023-11-29
fix internal error serde

modship:
  * [ ] make tests compile
  * [ ] make tests pass 

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2680,6 +2680,8 @@ dependencies = [
  "parking_lot",
  "rand",
  "serde",
+ "serde_json",
+ "serde_with",
  "tempfile",
 ]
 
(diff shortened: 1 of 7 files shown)


Leo-Besancon:
@modship can you rebase on `pre_mainnet_wip` branch please? So we can create a 27.2 from this directly

bilboquet:
I've tested this branch, and it solves the problem described in #4555

Leo-Besancon (approved):
Nice solve, this seems hard to catch! Approved once rebased on pre_mainnet_wip.
Do you think there is potentially other cases where something like this may happen?

aoudiamoncef:
We have to reuse existing tests like we did here and add Serde usecase to be sure

modship:
> @modship can you rebase on `pre_mainnet_wip` branch please? So we can create a 27.2 from this directly

done

PR #4582, opened 2023-12-08
Test 27 6 wip

Leo-Besancon:
  * [ ] make tests compile
  * [ ] make tests pass 


damip:
@Leo-Besancon Let's also do the following to proactively try to fix the regression Dusa observed:
* update the toolchain to 1.74.1 on all related projects (massa, massa sc runtime, as ffi bindings) as some bugs were fixed: https://blog.rust-lang.org/2023/12/07/Rust-1.74.1.html
* do a `cargo update` on massa, massa sc runtime and as ffi bindings

This should increase our chances of fixing the unexpected regression

damip:
@Leo-Besancon Since keys are rotated, we should update the securenet faucet

Leo-Besancon:
> @Leo-Besancon Since keys are rotated, we should update the securenet faucet

Faucet was updated on wednesday ;). Indead it did not work before as it tried to send coins with an address not in its wallet.

PR #4588, opened 2023-12-12
Feature/chainid

sydhds:
  * [ ] make tests compile
  * [ ] make tests pass 


massa-models/src/config/constants.rs
```diff
@@ -84,6 +84,21 @@ lazy_static::lazy_static! {
...
+        // MASS (SecureNet):         77658383
+        // MASS275 (SecureNet 27.5): 77658383505553
+        // MASL (Labnet):            77658376
+        // SANDBOX (Sandbox):        77
+        if cfg!(feature = "sandbox") {
+            77
+        } else {
+            77658383
+        }
+    };
```
Leo-Besancon, line 101: Can you make an issue in private-deployment to add the chain id config for the various networks we deploy?

massa-models/src/operation.rs
```diff
@@ -412,7 +414,18 @@ impl std::fmt::Display for Operation {
 /// signed operation
 pub type SecureShareOperation = SecureShare<Operation, OperationId>;
 
-impl SecureShareContent for Operation {}
+impl SecureShareContent for Operation {
+    fn compute_hash(&self, content_serialized: &[u8], content_creator_pub_key: &PublicKey) -> Hash {
+        let mut hash_data = Vec::new();
+        // Note: Add chain id before content hash in order to avoid replay attacks,
+        //       otherwise someone can copy an operation from testnet and execute it on main net
+        // Note 2: This makes the OperationId unique (per chain id)
```
Leo-Besancon, line 422: If I understand correctly, with how this is implemented, we cannot know what ChainId an operation is from. Would we ever want that? (I don't think so, but want to make sure).

Leo-Besancon (commented):
Nice work! 👌

massa-api/src/public.rs
```diff
@@ -33,7 +33,7 @@ use massa_models::{
     block_id::BlockId,
     clique::Clique,
     composite::PubkeySig,
-    config::CompactConfig,
+    config::{CompactConfig, CHAINID},
```
damip, line 36: let's not import constants directly but pass them through `APIConfig` for example (see how we do for other constants like THREAD_COUNT or VERSION)

massa-grpc/src/public.rs
```diff
@@ -15,7 +15,7 @@ use massa_models::address::Address;
 use massa_models::amount::Amount;
 use massa_models::block::{Block, BlockGraphStatus};
 use massa_models::block_id::BlockId;
-use massa_models::config::CompactConfig;
+use massa_models::config::{CompactConfig, CHAINID};
```
damip, line 18: here as well, pass it through the config

massa-models/src/config/constants.rs
sydhds: private-deploymenent PR aready there :)

massa-models/src/operation.rs
sydhds: Only way is for the signer but otherwise no. @damip any comments on this point?

massa-api/src/public.rs
sydhds: Fixed

massa-execution-worker/src/interface_impl.rs
```diff
@@ -11,7 +11,7 @@ use massa_async_pool::{AsyncMessage, AsyncMessageTrigger};
 use massa_execution_exports::ExecutionConfig;
 use massa_execution_exports::ExecutionStackElement;
 use massa_models::bytecode::Bytecode;
-use massa_models::config::MAX_DATASTORE_KEY_LENGTH;
```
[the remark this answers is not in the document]
sydhds: Fixed

massa-grpc/src/public.rs
sydhds: Fixed

massa-models/src/operation.rs
damip: The use case [like in ethereum ](https://github.com/ethereum/EIPs/blob/master/EIPS/eip-155.md) is only to verify that the tx is in its expected network for safety. Otherwise we would have added a field "chain_id" in the Operation structure itself but that would have increased the operation size, network load and reduces throughput

massa-models/src/config/constants.rs
```diff
@@ -84,6 +84,21 @@ lazy_static::lazy_static! {
         .parse()
         .unwrap()
     };
+    /// node chain id (to avoid replay attacks)
+    pub static ref CHAINID: u64 = {
+        // MASSA..
+        // MASM (MainNet):           77658377
+        // MASB (BuildNet):          77658366
+        // MASS (SecureNet):         77658383
+        // MASS275 (SecureNet 27.5): 77658383505553
```
damip, line 93: maybe remove this one for now, it can create uncertainty on which chainid to use for a given version

sydhds: Removed

damip:
@sydhds can you resolve conflicts before review ?

massa-models/src/secure_share.rs
```diff
@@ -97,7 +98,12 @@ where
     }
 
     /// Compute hash
-    fn compute_hash(&self, content_serialized: &[u8], content_creator_pub_key: &PublicKey) -> Hash {
+    fn compute_hash(
+        &self,
+        content_serialized: &[u8],
+        content_creator_pub_key: &PublicKey,
+        _chain_id: u64,
```
modship, line 105: not used here ?

modship (approved):

damip:
this PR still contains many unrelated changes

massa-models/src/config/constants.rs
```diff
@@ -87,14 +87,22 @@ lazy_static::lazy_static! {
...
-        // MASB (BuildNet):          77658366
         // MASS (SecureNet):         77658383
+        // MASB (BuildNet):          77658366
         // MASL (Labnet):            77658376
         // SANDBOX (Sandbox):        77
-        if cfg!(feature = "sandbox") {
-            77
-        } else {
-            77658383
+        match VERSION.to_string() {
```
damip, line 94: this check is cleaner: `VERSION.instance == "SAND"`

sydhds: instance is private. Not sure this is a good idea to make it pub ?

damip: I think adding a public Version::get_instance_name() function is a good idea:
* prevents accidental mutability that would happen if public
* allows users to not assume anything about the internal or serialized representation of the Version

damip (approved):

damip (approved):

PR #4591, opened 2023-12-14
Add context to error message when gas is lower that intance creation cost

bilboquet:
  * [ ] make tests compile
  * [ ] make tests pass 


Leo-Besancon (approved):
Approved if you don't forget to change to latest commit id :)

PR #4594, opened 2023-12-15
Use chain id everywhere (round 1)

sydhds:
  * [ ] make tests compile
  * [ ] make tests pass 

### DIFF
--- a/massa-client/base_config/config.toml
+++ b/massa-client/base_config/config.toml
@@ -9,6 +9,8 @@ private_port = 33034
 public_port = 33035
 grpc_public_port = 33037
 grpc_private_port = 33038
+# Chain id for node in sandbox mode, please update to match the target node chain id
+chain_id = 77
 
 [client]
     # maximum size in bytes of a request. Defaults to 50MB
--- a/massa-client/base_config/config.toml
+++ b/massa-client/base_config/config.toml
@@ -9,8 +9,8 @@ private_port = 33034
 public_port = 33035
 grpc_public_port = 33037
 grpc_private_port = 33038
-# Chain id for node in sandbox mode, please update to match the target node chain id
-chain_id = 77
+# Chain id for SecureNet, please update to match the target node chain id
+chain_id = 77658383
 
 [client]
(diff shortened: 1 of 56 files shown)


massa-client/base_config/config.toml
```diff
@@ -9,6 +9,8 @@ private_port = 33034
 public_port = 33035
 grpc_public_port = 33037
 grpc_private_port = 33038
+# Chain id for node in sandbox mode, please update to match the target node chain id
+chain_id = 77
```
damip, line 13: by default we should set it to securenet for now, so that people can just download the client and use it for securenet

massa-models/src/secure_share.rs
```diff
@@ -163,7 +170,8 @@ where
             serialized_data[..serialized_data.len() - rest.len()].to_vec()
         };
         let creator_address = Address::from_public_key(&creator_public_key);
-        let hash = Self::compute_hash(&content, &content_serialized, &creator_public_key);
+        // TODO
```
damip, line 173: remove TODO

massa-wallet/src/lib.rs
```diff
@@ -207,7 +207,11 @@ impl Wallet {
         let sender_keypair = self
             .find_associated_keypair(&address)
             .ok_or_else(|| WalletError::MissingKeyError(address))?;
-        Ok(Operation::new_verifiable(content, OperationSerializer::new(), sender_keypair).unwrap())
+        // TODO: chain id from config
```
damip, line 210: process this  TODO

massa-client/base_config/config.toml
sydhds: Done

massa-models/src/secure_share.rs
sydhds: Removed

massa-wallet/src/lib.rs
sydhds: Fixed

damip (approved):

PR #4596, opened 2023-12-15
Add cross OS checksums

aoudiamoncef:
* [x]  no breaking changes
* [x]  could fail by nature and not affect the release process
* [x]  could be simply relaunched
* [x]  uses Github artifacts
* [x]  uses official upload/download Github Actions
* [x]  tested locally (scripts)
* [x]  tested on a tweaked [fork of massa](https://github.com/aoudiamoncef/massa/actions/runs/7227988447/job/19696638265)


AurelienFT (approved):

AurelienFT:
@Leo-Besancon can you check if it ok for you also ?

Leo-Besancon:
Hi, for me it was ok yes (tested on a fork).
Even if as much as possible was done to test this, not sure if we want to risk having a failed CD release though.

AurelienFT:
> Hi @AurelienFT , for me it was ok yes (tested on a fork). Even if as much as possible was done to test this, not sure if we want to risk having a failed CD release though.

If we tested it, then it should be okay + if the CD fail we just revert this PR it will not slow down us a lot compare to the whole time it takes to set up and launch a release

PR #4600, opened 2023-12-21
Fix #4573: readonly call gas computation

### LINKED ISSUE #4573: Read_only gas_used value cannot be used to estimate the gas needed when the contract called init new calls
When we return the gas cost of a read_only_execution, we check against the max_instance_cost in order to make it easier to have a ggas estimation.
 
```rust
let exact_cost = req.max_gas.saturating_sub(exec_response.remaining_gas);
Ok(ReadOnlyExecutionOutput {
  out: execution_output,
  // return max_instance_cost if the exact cost is below
  // users can paste the estimated amount into a real call
  // without having to worry about underlying limits
  gas_cost: if exact_cost > self.config.gas_costs.max_instance_cost {
    exact_cost
  } else {
    self.config.gas_costs.max_instance_cost
}
```

The problem is that this check is only made at the base level call_sc, but not for abi calls inside.
We need to return a mix of gas_used, which may not be easily done.

Example: We make a call with max_gas of 2_400_000.
> Call_0_0 (real init cost: 100_000) - Available gas after: 2_300_000
	> Call_1_0 (real init cost: 100_000) - Available gas after: 2_200_000
	> Call_1_1 (real init cost: 100_000) - Available gas after: 2_100_000
	> Call_1_2 (real init cost: 100_000) - Available gas after: 2_000_000
	> Call_1_3 (real init cost: 100_000) - Fails because <= 2_100_000.

Here, we will return a gas_used of 2_100_000, but a real call if 2_100_000 will fail immediately.
We should return a gas estimation of 2_500_000 to ensure everything passes.

Tasks:
- [x] #4586
- [ ] Fix computation of `max_gas` see: https://github.com/massalabs/massa/issues/4573#issuecomment-1833595505

bilboquet:
  * [x] make tests compile
  * [x] make tests pass 

fixes #4573
* [x] depends on massalabs/massa-sc-runtime#325
* [x] depends on massalabs/massa-functional-tests#16

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -106,7 +106,7 @@ massa_wallet = { path = "./massa-wallet" }
 
 # Massa projects dependencies
 massa-proto-rs = { git = "https://github.com/massalabs/massa-proto-rs", "rev" = "84678fb77cf6d06d85d55e2c20502908ff292e61" }
-massa-sc-runtime = { git = "https://github.com/massalabs/massa-sc-runtime", "rev" = "d1dce679951ce1d1dad0471b2ac6b8ef12b44e9d" }
+massa-sc-runtime = { git = "https://github.com/massalabs/massa-sc-runtime", "rev" = "bbab9f329eee08d1f54c595214634a03824e9343" }
 peernet = { git = "https://github.com/massalabs/PeerNet", "branch" = "main" }
 
 # Common dependencies
--- a/massa-execution-worker/src/execution.rs
+++ b/massa-execution-worker/src/execution.rs
@@ -801,7 +801,8 @@ impl ExecutionState {
             // Set the call stack to a single element:
             // * the execution will happen in the context of the address of the operation's sender
             // * the context will give the operation's sender write access to its own ledger entry
-            // This needs to be defined before anything can fail, so that the emitted event contains the right stack
+            // This needs to be defined before anything can fail, so that the emitted event
+            // contains the right stack
             context.stack = vec![ExecutionStackElement {
                 address: sender_addr,
                 coins: Amount::zero(),
@@ -811,7 +812,7 @@ impl ExecutionState {
         };
 
         // load the tmp module
-        let (module, remaining_gas) = self
+        let module = self
             .module_cache
             .read()
             .load_tmp_module(bytecode, *max_gas)?;
@@ -819,7 +820,7 @@ impl ExecutionState {
         massa_sc_runtime::run_main(
             &*self.execution_interface,
             module,
-            remaining_gas,
+            *max_gas,
             self.config.gas_costs.clone(),
         )
         .map_err(|error| ExecutionError::VMError {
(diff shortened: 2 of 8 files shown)


Cargo.toml
```diff
@@ -106,7 +106,8 @@ massa_wallet = { path = "./massa-wallet" }
 
 # Massa projects dependencies
 massa-proto-rs = { git = "https://github.com/massalabs/massa-proto-rs", "rev" = "30b91d705b704287662b0e42457ece442005fa46" }
-massa-sc-runtime = { git = "https://github.com/massalabs/massa-sc-runtime", "rev" = "0bef3a9c0013e61b0c2b18105238649ac74007c9"  }
+#massa-sc-runtime = { git = "https://github.com/massalabs/massa-sc-runtime", "branch" = "4573-read_only-gas" }
```
bilboquet, line 109: Change ref when sc-runtime is merged

massa-execution-worker/src/execution.rs
```diff
@@ -1507,17 +1508,37 @@ impl ExecutionState {
...
         let execution_output = context_guard!(self).settle_slot(None);
-        let exact_cost = req.max_gas.saturating_sub(exec_response.remaining_gas);
+        let gas_used_until_last_subexecution = (context_guard!(self))
+            .gas_remaining_before_subexecution
+            .unwrap_or_default(); // 0 if no subexecutions
+        let exact_exec_cost = req.max_gas.saturating_sub(exec_response.remaining_gas);
+        // compute a gas cost, estimating the gas of the last SC call to be max_instance_cost
+        let corrected_cost = req
+            .max_gas
+            .saturating_sub(gas_used_until_last_subexecution)
```
damip, line 1518: to me it feels like it should be:

```
corrected_cost = gas_used_until_last_subexecution + max_instance_cost
```

what do you think ?

massa-execution-worker/src/execution.rs
```diff
@@ -1507,17 +1508,37 @@ impl ExecutionState {
...
+            .saturating_add(self.config.gas_costs.max_instance_cost);
+        // keep the max of the two so the last SC call has at least max_instance_cost of gas
```
damip, line 1520: I also have a doubt about "launch_cost", you might need to take a look whether it should or not be counted here

massa-execution-worker/src/execution.rs
damip: I think there might be a confusion between "remaining" and "used" gas until last subexecution

massa-execution-worker/src/execution.rs
```diff
@@ -1508,29 +1508,26 @@ impl ExecutionState {
...
-            .saturating_sub(gas_used_until_last_subexecution)
-            .saturating_add(self.config.gas_costs.max_instance_cost);
+        let exact_exec_cost = req.max_gas.saturating_sub(exec_response.remaining_gas);
+
+        let corrected_cost = match (context_guard!(self)).gas_remaining_before_subexecution {
+            Some(gas_remaining) => req
+                .max_gas
+                .saturating_sub(gas_remaining) // gas used until last subexecution
+                .saturating_add(self.config.gas_costs.max_instance_cost),
+            None => 0u64,
```
damip, line 1519:
```suggestion
            None => self.config.gas_costs.max_instance_cost,
```

massa-execution-worker/src/execution.rs
bilboquet: I forgot to rename.
Should be clearer now.
The case where `gas_remaining_before_subexecution` is None should also be fixed. It yields 0 so the `exact_exec_cost` will be selected by the max()

massa-execution-worker/src/execution.rs
bilboquet: done thanks

massa-execution-worker/src/execution.rs
bilboquet: clarified

damip (approved):

PR #4603, opened 2023-12-27
fix wallet export version

damip:
* Station Wallet had a requirement that was not present in the standard: it required wallet nicknames to be less than 32 characters long
  * this was the cause of some compliant wallets to be rejected by Station Wallet
  * this requirement is to be removed from Station Wallet to stop rejecting some standard-compliant wallets
  * we also need to ensure that long nicknames are truncated in the UX
* contrary to the standard, Station Wallet was cyphering not only the ed25519 secret key but the concatenation (secret_key, public_key)
  * this can cause standard-compliant wallets to be rejected by Station Wallet
  * it also causes Station-generated wallets to be non-compliant
  * the encoding in Station Wallet is to be changed to comply with the standard when saving the wallet file
  * tolerance is to be added both to Station Wallet and massa-client when loading wallet files to support both encoding formats to not break existing wallets incorrectly generated by Station
* contrary to the standard, massa-client was setting the version number to "0" despite saving the wallet in otherwise standard-compliant "version 1" format,
  * this was the cause of some massa-client wallets to be rightfully rejected by Station Wallet
  * the massa-client was set to set the version to "1" when saving a wallet
  * both massa-client and Station Wallet are to be set to interpret version "0" wallets as version "1" wallets provided that the encoded key is the right length to reject old (true version 0) wallets while ensuring that wallets generated with the broken v27 massa-client remain compatible with the latest versions of both wallets

Station side: https://github.com/massalabs/station-massa-wallet/issues/858

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -156,9 +156,9 @@ dependencies = [
 
 [[package]]
 name = "anyhow"
-version = "1.0.76"
+version = "1.0.77"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "59d2a3357dde987206219e78ecfbbb6e8dad06cbb65292758d3270e6254f7355"
+checksum = "c9d19de80eff169429ac1e9f48fffb163916b448a44e8e046186232046d9e1f9"
 
 [[package]]
 name = "arrayref"
@@ -254,7 +254,7 @@ checksum = "16e62a023e7c117e27523144c5d2459f4397fcc3cab0085af8e2224f643a0193"
 dependencies = [
  "proc-macro2 1.0.71",
  "quote 1.0.33",
- "syn 2.0.42",
+ "syn 2.0.43",
 ]
 
 [[package]]
@@ -265,7 +265,7 @@ checksum = "fdf6721fb0140e4f897002dd086c06f6c27775df19cfe1fccb21181a48fd2c98"
 dependencies = [
  "proc-macro2 1.0.71",
  "quote 1.0.33",
- "syn 2.0.42",
+ "syn 2.0.43",
 ]
 
 [[package]]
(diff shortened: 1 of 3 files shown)


sajjadiza:
AU128waeDayJk4nU86KYvqcyKjJ5tk8N58bJYQ5GEnDXtWqSJX9Dp

sajjadiza:
AU128waeDayJk4nU86KYvqcyKjJ5tk8N58bJYQ5GEnDXtWqSJX9Dp

Leo-Besancon (commented):
I have not tested the changes with wallets of the various formats, but the current implementation looks good to me.
Using the [transition crate](https://github.com/massalabs/transition) (similarly to how we handle the version in other components, such as KeyPair, PublicKey and Address) would be best IMO to make future wallet versions more straightforward to implement.

damip:
> I have not tested the changes with wallets of the various formats, but the current implementation looks good to me. Using the [transition crate](https://github.com/massalabs/transition) (similarly to how we handle the version in other components, such as KeyPair, PublicKey and Address) would be best IMO to make future wallet versions more straightforward to implement.

Sure, that could be a nice followup after launch

Leo-Besancon (approved):
Tested:
- [x] A wallet generated by massa station (version 1) could be opened with massa-client
- [x] A wallet generated by massa station (with its version manually set to 0) could be opened with massa-client
- [x] Old wallets ("false" version 0) genetated by massa-client could still be opened by massa-client
- [x] New wallets can be opened by massa-client

Not tested:
- To open any wallet with massa-station
- To open a "true version 0" wallet with massa-client

damip:
Important tests remaining:
* attempt to sign and emit a labnet operation from Station using a massa-client generated wallet
* attempt to sign and emit a labnet operation from massa-wallet using a Station generated wallet
Note: those tests make sense only with the fixed version of Station: https://github.com/massalabs/station-massa-wallet/pull/862

PR #4612, opened 2024-01-04
Mainnet genesis

Leo-Besancon:
  * [x] make tests compile
  * [x] make tests pass 

TODO:
- [x] Add initial files
  - [x] deferred_credits (/!\ no longer from submodule)
  - [x] initial_ledger (/!\ no longer from submodule)
  - [x] initial_rolls (/!\ no longer from submodule)
  - [x] initial_peers
  - [x] bootstrap whitelist 
- [x] Settle the Version
  - [x] Node Version constant, currently `MAIN.1.0`
  - [x] Package versions, currently `1.1.0`, but should probably be `1.0.0` 
- [x] Make sure the initial files are up to date and final! 
- [x] Check the pre-genesis warning and maybe panics ? for absence of pre-genesis network communication


massa-node/base_config/bootstrap_whitelist.json
```diff
@@ -1,4 +1,2 @@
 [
```
modship, line 1: don't remember if have a empty array is a problem, maybe we should delete the file ?

damip: I think deleting it has a different meaning

README.md
```diff
@@ -34,15 +34,11 @@ Here is a list of tools to easily build applications on the Massa blockchain:
...
+## Become a node runner
 
 With decentralization as a core value, we've gone to great lengths to lower the barrier of entry for community
-participation in our community testnet, and we
-[invite you to join in](https://docs.massa.net/docs/node/install), and register for the
-testnet participation reward program.
-
-Your participation will help improve decentralization, usability, and network reliability leading up to, and following,
-the launch of mainnet.
+participation in our community testnet, and we [invite you to join in](https://docs.massa.net/docs/node/home) by
```
AurelienFT, line 40:
```suggestion
participation in our mainnet, and we [invite you to join in](https://docs.massa.net/docs/node/home) by
```

maybe ?

massa-node/base_config/bootstrap_whitelist.json
Leo-Besancon: Indeed, deleting the file would be "I allow everyone to bootstrap on me".
Having an empty array means "I allow noone to bootstrap on me", which I think is the wanted behaviour?

damip: we could try to deploy before tagging in order to test

AurelienFT: empty whitelist == nobody can bootstrap. No whitelist, everyone can. Personally I think it's nice if we keep one of our IP in this file so that we will be able to ping nodes to have some kind of topology of the network

damip: > Indeed, deleting the file would be "I allow everyone to bootstrap on me". Having an empty array means "I allow noone to bootstrap on me", which I think is the wanted behaviour?

yes we want "nobody to bootstrap on me"

AurelienFT: I didnt had the full conversation when I wrote : https://github.com/massalabs/massa/pull/4612#discussion_r1442632919 but still relevant

damip: hmm see @AurelienFT 's comment below, it's also a good idea, assuming people would keep that port open and accessible

AurelienFT: > hmm see @AurelienFT 's comment below, it's also a good idea, assuming people would keep that port open and accessible

Not everyone but could help us in many cases in the future I think

damip: yeah good idea.
Maybe let's whitelist all of our bootstrap server IPs, both IPV4 and ipv6

Leo-Besancon: Done in a87836b484a7d12e8055e8ec0beeb5a1e7454d3a

sydhds (approved):

AurelienFT (approved):

damip (approved):

modship (approved):

PR #4617, opened 2024-01-08
Fix protocol handler retrieval thread CPU usage when no blocks are pr…

sydhds:
…ocessed

  * [ ] make tests compile
  * [ ] make tests pass 


bilboquet (approved):
Solves the problem on my side.
Tracing the tokio loop from which this function is called shows that branches from the select are taken each second.

massa-protocol-worker/src/handlers/block_handler/retrieval.rs
```diff
@@ -933,6 +933,9 @@ impl RetrievalThread {
             .expect("could not compute next block retrieval timer tick");
 
         if self.asked_blocks.is_empty() && self.block_wishlist.is_empty() {
+            // Note: in mainnet and before ICO, no blocks are processed but the timer need to be updated
```
damip, line 936: not ICO but genesis, and it is not specific to mainnet

damip: need => needs

damip (changes requested):
Nice catch !
TODO: fix comment, shorten all thread names to fit

AurelienFT:
> @sydhds @bilboquet @AurelienFT Does this require an urgent 1.1 deployment ? or does it have no impact at genesis ?

No it should stop when blocks will be produced. Could be cool to verify with a labnet

modship (approved):

Floyz:
> > @sydhds @bilboquet @AurelienFT Does this require an urgent 1.1 deployment ? or does it have no impact at genesis ?
> 
> No it should stop when blocks will be produced. Could be cool to verify with a labnet

It has impact on node runners hardware tho, if you want people to have node running before genesis it would be great to publish a new release.
I stopped mine due to this bug, consuming one full CPU for nothing and impacting other services on a shared hardware. ( + CPU temp going high )

massa-protocol-worker/src/handlers/block_handler/retrieval.rs
sydhds: done

massa-node/src/survey.rs
```diff
@@ -12,6 +12,10 @@ use tracing::info;
 // use std::time::Duration;
 use tracing::warn;
 
+// massa-survey
+const THREAD_NAME: &str = "massa-survey";
```
damip, line 16: clippy

sydhds: fixed

massa-node/src/survey.rs
```diff
@@ -49,13 +49,16 @@ impl MassaSurvey {
         if massa_metrics.is_enabled() {
             #[cfg(all(not(feature = "sandbox"), not(test)))]
             {
+                // massa-survey
+                const THREAD_NAME: &str = "massa-survey";
```
bilboquet, line 53:
```suggestion
                const THREAD_NAME: &str = "massa-survey";
```
?

bilboquet: can't apply because of #cfg()

damip (approved):
Good !

Issue #4101, opened 2023-06-17
Node hangs while recomputing final_state caches


Leo-Besancon:
During some cycles, the function get_cycle_history_cycles() in pos_final_state, used to recompute the pos caches after bootstrap, never seems to return.
The while let loop seems to never stop (for instance I printed the found cycles and it went 509, 510, 511, 512, 513, 514, 515, 509...).

This is because the code for get_cycle_history_cycles() assumes the serialized keys order increase with the cycle. It is not true, e.g. the key for cycle 509 is greater than the key for cycle 513 in lexicography order.

PR #4619, opened 2024-01-09
Fix DB serialization order for Ord types

Leo-Besancon:
An error prone behaviour in our code is that when serializing numbers (or slots) for the keys of the DB, we use varint serializer. As a result, the ordering of such numbers is not garantied in the DB's ordering (for instance when the number is around a power of two).

This is e.g. related to the following issue and associated PR: https://github.com/massalabs/massa/issues/4101 and https://github.com/massalabs/massa/pull/4104

In this PR, we will replace every instance of such serialization (and deserialization).

TODO:
- [x] List every DB key and check whether it is impacted 
- [x] Fix serialization of the keys in the DB
- [x] Fix deserialiazation of the keys in the DB (also in the is_db_key_value_valid) check function!)

Potentially impacted files list. For each file, list the potentially impacted keys:
- [x] https://github.com/massalabs/massa/blob/main/massa-async-pool/src/pool.rs @Leo-Besancon
  - recompute_message_info_cache iterates over the DB (https://github.com/massalabs/massa/blob/main/massa-async-pool/src/pool.rs#L228)
  - fetch_message(message_id) iterates over all the keys with the message_id as a prefix.
  - is_key_value_valid(message_id, message) deserializes the message_id key
  - put_entry(&self, message_id: &AsyncMessageId, message: AsyncMessage, batch: &mut DBBatch) serializes the message_id
  - Same for update_entry(&self, message_id: &AsyncMessageId, message_update: AsyncMessageUpdate, batch: &mut DBBatch)
  - Same for delete_entry
  - For a given (async_message_id, async_message), it is serialized in the DB with:
    - ASYNC_POOL_PREFIX + AsyncMessageIdSerializer::serialize(message_id) + MESSAGE_PART_IDENT
- [x] https://github.com/massalabs/massa/blob/main/massa-db-worker/src/massa_db.rs @AurelienFT 
  - https://github.com/massalabs/massa/blob/main/massa-db-worker/src/massa_db.rs#L174 => We iter but is not a problem if we don’t have the right order
  - https://github.com/massalabs/massa/blob/main/massa-db-worker/src/massa_db.rs#L181C34-L181C34 => Shouldn't be an issue as we are not caring about the key value.
  - https://github.com/massalabs/massa/blob/main/massa-db-worker/src/massa_db.rs#L303 => We iter but is not a problem if we don’t have the right order
  - https://github.com/massalabs/massa/blob/main/massa-db-worker/src/massa_db.rs#L326 =>  Shouldn't be an issue as we are not caring about the key value.
  - https://github.com/massalabs/massa/blob/main/massa-db-worker/src/massa_db.rs#L371 =>  Shouldn't be an issue as we are not caring about the key value.
  - https://github.com/massalabs/massa/blob/main/massa-db-worker/src/massa_db.rs#L706 => Iter  Shouldn't be an issue as we are not caring about the key value.
- [x] https://github.com/massalabs/massa/blob/main/massa-executed-ops/src/executed_ops.rs @AurelienFT => => No problem for me as keys contains only prefix + operation Id which doesn’t have a variant and are not meant to be sorted
- [x] https://github.com/massalabs/massa/blob/main/massa-executed-ops/src/executed_denunciations.rs @AurelienFT
  - https://github.com/massalabs/massa/blob/main/massa-executed-ops/src/executed_denunciations.rs#L63 => Iter but shouldn’t be a problem because it iter on the whole content of denunciations
  - Serialisation in order is needed broken but not assumed anywhere in this file so it’s not a breaking problem.
- [x] https://github.com/massalabs/massa/blob/main/massa-final-state/src/final_state.rs @Eitu33
- `pos_state.create_new_cycle_from_last`: function from PoS, looks like a potential issue but haven’t looked all the way as I was not assigned to that file
lines are:
L234
L303
L351
L380
- `pos_state.apply_changes_to_batch`: takes slot as input and uses its serialized version when extending deferred credits
side note: other apply_changes functions in finalize taking slot as input are only using it for pruning purposes
lines are:
L450
- when checking db validity, iterates over the whole database and deserializes values to check validity, order should have no impact here
L556-688
- deletes:
reset values using constant prefixes
L128-135
L848-858
delete a cycle info with prefix, prefix is built with cycle number
L224
L280
L398

- [x] https://github.com/massalabs/massa/blob/main/massa-ledger-worker/src/ledger_db.rs @Eitu33 
- key_serializer_db & key_deserializer_db usage:
takes care only of ledger keys, which isnt the problem here (version byte + addr serialize + key type)
side note: to see all of them you can just query “serializer_db” the file
- iterators based on prefixes (again only using ledger key formatting)
L191
L463
L490 (test_export)
L526 (test_export)
- delete_prefix with base prefix input (basically just a clear)
L214

- [x] https://github.com/massalabs/massa/blob/main/massa-module-cache/src/hd_cache.rs @Leo-Besancon
  - Is not impacted (only Hash::to_bytes as keys)
- [x] https://github.com/massalabs/massa/blob/main/massa-pos-exports/src/pos_final_state.rs @Leo-Besancon
  - Every key in CYCLE_HISTORY_PREFIX starts with the cycle_prefix serialized as a Varint.
    - complete_key
    - rng_seed_key
    - final_state_hash_snapshot_key
    - roll_count_prefix
    - roll_count_key
    - prod_stats_prefix
    - upper_limit_prefix
    - prod_stats_fail_key
    - prod_stats_success_key
  - Deferred_credits starts with the slot serialized as a Varint of period and varint of thread
- [x] https://github.com/massalabs/massa/blob/main/massa-versioning/src/versioning.rs @sydhds 
  * Every key have MIP Info serialized with some varint (version & component version)
    * when we init from the db (e.g. from disk) we load everything, store the udpates in a BTreeMap (thus sorting the MIP Info) then call the update_with functions. This can become a problem if we have way too much MIP info stored in db (> 100 000) ? - Should be discussed..

Impacted keys:
- [x] Critical:
  - [x] All the keys that are prefixed by a cycle (pos_final_state.rs)
  - [x] Deferred credits key (contains a slot - pos_final_state.rs)
- [x] Non critical instances that we won't change in this PR (to avoid too many unnecessary changes that could introduce other problems): 
  - [x] In the AsyncPool, the keys in the DB are AsyncMessageId that are Ord and partly encoded as Varint. We do not (currently) use this Ord, but may in the future).
  - [x] Executed Denunciations are indexed by DenunciationIndexed that are Ord and partly encoded as Varint.. We do not (currently) use this Ord, but may in the future)
  - [x] MIP Info related keys, but we don't iter on them in the DB, we query all of them at start. 

# OTHER FACTORS

* [x] make sure network ban times are 1 day
* [x] update version to MAIN.2.0
* [x] update package versions
* [x] add validated Genesis files
  * [x] wait for validation
  * [x] update files
  * [x] update readme


damip:
For slots that are used in a key: use Slot::to_bytes_key() and Slot::from_bytes_key()

For integers, use to_be_bytes() and from_be_bytes()

AurelienFT:
My part is done.

damip:
My independent analysis (I went through everything except versioning/MIP):

https://github.com/massalabs/massa/pull/4620/files

Eitu33:
final state and ledger db done

Leo-Besancon:
I checked every change made in https://github.com/massalabs/massa/pull/4620/files , this PR seems to take into account the same ser/deser.

AurelienFT (approved):

massa-pos-exports/src/pos_final_state.rs
```diff
@@ -732,11 +718,19 @@ impl PoSFinalState {
...
+            if rest.len() < SLOT_KEY_SIZE {
+                panic!("{}", DEFERRED_CREDITS_DESER_ERROR);
+            }
+            let slot = Slot::from_bytes_key(
+                rest[..SLOT_KEY_SIZE]
+                    .try_into()
+                    .expect(DEFERRED_CREDITS_DESER_ERROR),
+            );
+            let rest: &[u8] = &rest[SLOT_KEY_SIZE..];
+
```
damip, line 733: A bit more compact and less error prone:

```rust
            // deserialize the slot
            let rest = &serialized_key[DEFERRED_CREDITS_PREFIX.len()..];
            let slot = Slot::from_bytes_key(
                rest.get(..SLOT_KEY_SIZE)
                    .and_then(|v| v.try_into().ok())
                    .expect(DEFERRED_CREDITS_DESER_ERROR),
            );
            let rest = &rest[SLOT_KEY_SIZE..];
```

damip:
All those variables on slices are stressing me out a bit.
Here is a less error-prone approach:

```
fn buf_to_array_ctr<F, V, const N: usize>(
    buffer: &[u8],
    constructor: F,
) -> Result<(&[u8], V), ModelsError>
where
    F: Fn(&[u8; N]) -> V,
{
    if buffer.len() < N {
        return Err(ModelsError::BufferError("buffer too short".into()));
    }
    let v = constructor(
        &buffer[..N].try_into().unwrap(), // checked above
    );
    Ok((&buffer[N..], v))
}
```

Then you can do:

```
let (rest_key, slot) =
    buf_to_array_ctr(rest_key, Slot::from_bytes_key).expect("error !");

let (rest_key, cycle) =
    buf_to_array_ctr(rest_key, |v| u64::from_be_bytes(v.clone()))
        .expect("error !");
```

massa-pos-exports/src/pos_final_state.rs
```diff
@@ -999,18 +991,15 @@ impl PoSFinalState {
...
-                Some(prev_cycle) => {
-                    let cycle_prefix = self.cycle_history_cycle_prefix(*prev_cycle);
-
-                    db.iterator_cf(
+                Some(prev_cycle) => db
+                    .iterator_cf(
                         STATE_CF,
                         MassaIteratorMode::From(
-                            &upper_limit_prefix!(cycle_prefix),
+                            &self.cycle_history_cycle_prefix(prev_cycle.saturating_add(1)),
```
damip, line 998: what is this ?

Leo-Besancon: This basically reverts the previous fix (that avoided the assumption that cycles are in order in the DB) made in https://github.com/massalabs/massa/pull/4104.

This is not needed, but it simplifies a bit the code.

damip: do we have tests to ensure that the change won't cause a regression ?

Leo-Besancon: test_pos_cache_recomputation() at the bottom of the file wil catch this. It was the test I added in #4104 to make sure the cycle_history is correctly recovered from the db (in the correct order)

massa-pos-exports/src/pos_final_state.rs
```diff
@@ -706,23 +706,17 @@ impl PoSFinalState {
...
+                        .get_next_slot(self.config.thread_count)
+                        .expect(DEFERRED_CREDITS_SER_ERROR)
+                        .to_bytes_key(),
+                );
             }
-            _ => {}
+            Unbounded => {}
         };
 
         for (serialized_key, serialized_value) in db.iterator_cf(
```
bilboquet, line 722: what about
```rust
        let iterator_cf = db.iterator_cf(
            STATE_CF,
            MassaIteratorMode::From(&start_key_buffer, MassaDirection::Forward),
        ).filter(|(k, _)| k.starts_with(DEFERRED_CREDITS_PREFIX.as_bytes()));

        for (serialized_key, serialized_value) in
        iterator_cf
        {
            // if !serialized_key.starts_with(DEFERRED_CREDITS_PREFIX.as_bytes()) {
            //     // println!("break !start_with");
            //     break;
            // }
```
if I'm correct, it's equivalent and based on few measures the loop is 10% faster

damip: isn't there a risk that it iterates starting from `start_key_buffer` and until **the end of the whole database** ? I'm afraid `filter()` is not short-circuiting, contrary to `https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.take_while`

bilboquet: Yes, `take_while` is the correct version of what I wanted to do.
In my test, I counted the number of entries at the end and they were the same. So I was wrong, but the conditions were favorable so the timing looks correct.
I tried again with `take_while` and it seems to be a bit faster (probably because it short-circuits)

Leo-Besancon: In another PR, we may want to go over similar behaviour in the code and see if it's worth it to change. (/!\ some of the DB iteration have weird logic, e.g. the iter over async_messages in a comment above, we have to make sure that everything is the same before changing it).

massa-execution-worker/src/execution.rs
```diff
@@ -1832,7 +1840,14 @@ impl ExecutionState {
             .final_state
             .read()
             .get_pos_state()
-            .get_address_deferred_credits(address);
+            .get_deferred_credits_range(.., Some(address))
+            .credits;
+        let res_final: BTreeMap<Slot, Amount> = res_final
```
bilboquet, line 1845: I think we could merge those 2 let bindings into one like this:
```rust
        let res_final: BTreeMap<Slot, Amount> = self
            .final_state
            .read()
            .get_pos_state()
            .get_deferred_credits_range(.., Some(address))
            .credits
            .iter()
            .filter_map(|(slot, addr_amount)| {
                addr_amount.get(address).map(|amount| (*slot, *amount))
            })
            .collect();
```

sydhds (dismissed):

AurelienFT (dismissed):

massa-execution-worker/src/execution.rs
Leo-Besancon: Done

AurelienFT (dismissed):

bilboquet (approved):

sydhds (dismissed):

damip (dismissed):

damip:
Only the genesis files update remains. Waiting for external approval

damip (dismissed):

PR #4620, opened 2024-01-09
damir independent db key audit

damip:


massa-pos-exports/src/pos_final_state.rs
```diff
@@ -732,11 +727,15 @@ impl PoSFinalState {
...
-                .deferred_credits_deserializer
-                .slot_deserializer
-                .deserialize::<DeserializeError>(&serialized_key[DEFERRED_CREDITS_PREFIX.len()..])
-                .expect(DEFERRED_CREDITS_DESER_ERROR);
+
+            // deserialize slot
+            let slot = Slot::from_bytes_key(
+                serialized_key[DEFERRED_CREDITS_PREFIX.len()..]
+                    .try_into()
+                    .expect(DEFERRED_CREDITS_DESER_ERROR),
```
Leo-Besancon, line 735: I think you need to have the correct length, or the try_into will not work, right?
```rs
serialized_key[DEFERRED_CREDITS_PREFIX.len()..DEFERRED_CREDITS_PREFIX.len() + SLOT_KEY_SIZE]
```
instead of
```rs
serialized_key[DEFERRED_CREDITS_PREFIX.len()..]
```

damip: you're right !

PR #4623, opened 2024-01-11
Fix array indexing code in versioning_ser_der.rs

sydhds:
  * [ ] make tests compile
  * [ ] make tests pass 

### DIFF
--- a/tools/extract_array_indexing_op.py
+++ b/tools/extract_array_indexing_op.py
@@ -43,7 +43,7 @@ if __name__ == "__main__":
     match_count = 0
     with open(filepath) as fp:
         file_content = fp.read()
-        for m in re.finditer(r"\[([\w:\+\.-]+)\]", file_content, flags = re.IGNORECASE | re.MULTILINE):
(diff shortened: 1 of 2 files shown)


Leo-Besancon (approved):
If I understand correctly, the deserializer behaviour is the same as before? It's mainly to cleanup error-prone array indexing? :)

tools/extract_array_indexing_op.py
```diff
@@ -42,7 +43,7 @@ def index_to_coordinates(s, index):
     match_count = 0
     with open(filepath) as fp:
         file_content = fp.read()
-        for m in re.finditer(r"\[([\w:\+\.-]+)\]", file_content, flags = re.IGNORECASE | re.MULTILINE):
+        for m in re.finditer(r"\[([\w\(\)\s:\+\.-]+)\]", file_content, flags = re.IGNORECASE | re.MULTILINE):
```
damip, line 46: @sydhds what is this ?

sydhds: Accepting () and space characters in the regex

damip (approved):

PR #4624, opened 2024-01-11
Zeroize on Wallet struct

sydhds:
  * [ ] make tests compile
  * [ ] make tests pass 

### DIFF
--- a/massa-wallet/src/lib.rs
+++ b/massa-wallet/src/lib.rs
@@ -19,17 +19,20 @@ use std::collections::hash_map::Entry;
 use std::collections::HashSet;
 use std::path::PathBuf;
 use std::str::FromStr;
+use zeroize::{Zeroize, ZeroizeOnDrop};
 
 mod error;
 
 const WALLET_VERSION: u64 = 1;
 
 /// Contains the keypairs created in the wallet.
-#[derive(Clone, Debug, Deserialize, Serialize)]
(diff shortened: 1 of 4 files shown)


massa-wallet/src/lib.rs
```diff
@@ -19,17 +19,20 @@ use std::collections::hash_map::Entry;
...
 mod error;
 
 const WALLET_VERSION: u64 = 1;
 
 /// Contains the keypairs created in the wallet.
-#[derive(Clone, Debug, Deserialize, Serialize)]
+#[derive(Clone, Debug, Deserialize, Serialize, Zeroize, ZeroizeOnDrop)]
 pub struct Wallet {
     /// Keypairs and addresses
+    #[zeroize(skip)]
```
bilboquet, line 32: I don't get why we would skip fields?
On drop, we want them to be zeroized, don't we?

sydhds: These field types are not supported by Zeroize and what we really want to set to 0 is the password. Note that the KeyPair struct has already Zeroize & ZeroizeDrop features (because the feature "zeroize" is enabled for the crate ed25519)

bilboquet: thanks for the explanation

bilboquet (approved):

damip (approved):

PR #4627, opened 2024-01-16
Remove roll_count argument in client sell_rolls

Leo-Besancon:
  * [ ] make tests compile
  * [ ] make tests pass 

### DIFF
--- a/massa-client/src/cmds.rs
+++ b/massa-client/src/cmds.rs
@@ -251,7 +251,7 @@ pub enum Command {
 
     #[strum(
         ascii_case_insensitive,
-        props(args = "Address RollCount Fee"),
+        props(args = "Address Fee"),
         message = "sell rolls with wallet address"
     )]
     sell_rolls,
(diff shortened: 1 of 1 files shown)


massa-client/src/cmds.rs
```diff
@@ -902,37 +902,35 @@ impl Command {
...
+                                    wallet,
+                                    OperationType::RollSell { roll_count },
+                                    fee,
+                                    addr,
+                                    json,
+                                )
+                                .await
                             }
-                            None => client_warning!(format!("address {} not found", addr)),
+                            None => bail!(format!("address {} not found", addr)),
```
bilboquet, line 929: Previously this was a warning, do we want to change this?

damip:
This solution is incomplete: the problem persists if there are subsequent manual roll buys after the manual roll sell

massa-client/src/cmds.rs
Leo-Besancon: I don't think it matters, because the operation previously sent would fail if the result of get_addresses was empty. But I'll let other people correct me if it's wrong.

PR #4629, opened 2024-01-17
Deferred credits fix

Leo-Besancon:
  * [x] make tests compile
  * [x] make tests pass 


AurelienFT (approved):

damip (approved):
Good job !

PR #4642, opened 2024-02-09
Feature gate test framework 

Leo-Besancon:
  * [ ] make tests compile
  * [ ] make tests pass 

The goal of this PR is to avoid building massa_models with test-exports feature when building massa's workspace.


bilboquet (dismissed):
looks good to me

damip (approved):

bilboquet (approved):

PR #4646, opened 2024-02-15
Add gathering of transfers in the node + prep work for ABIs

AurelienFT:
  * [x] make tests compile
  * [x] make tests pass 


AurelienFT (commented):
Good work ! Nothing really surprising on the code side. Just need to be sure it works has intended with tests.

sydhds (dismissed):

Leo-Besancon (approved):
The trace workflow in execution module seems good to me!

PR #4659, opened 2024-03-07
Reduce size caching transfers & add backup for infos in get_operation using execution-trace

AurelienFT:
  * [x] make tests compile
  * [x] make tests pass 

### DIFF
--- a/massa-execution-worker/src/execution.rs
+++ b/massa-execution-worker/src/execution.rs
@@ -24,6 +24,7 @@ use massa_ledger_exports::{SetOrDelete, SetUpdateOrDelete};
 use massa_metrics::MassaMetrics;
 use massa_models::address::ExecutionAddressCycleInfo;
 use massa_models::bytecode::Bytecode;
+
 use massa_models::datastore::get_prefix_bounds;
 use massa_models::denunciation::{Denunciation, DenunciationIndex};
 use massa_models::execution::EventFilter;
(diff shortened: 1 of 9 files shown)


massa-execution-worker/src/execution.rs
```diff
@@ -197,7 +197,7 @@ impl ExecutionState {
             #[cfg(feature = "execution-trace")]
             trace_history: Arc::new(RwLock::new(TraceHistory::new(
                 config.max_execution_traces_slot_limit as u32,
-                MAX_OPERATIONS_PER_BLOCK,
+                (MAX_GAS_PER_BLOCK / BASE_OPERATION_GAS_COST) as u32,
```
Leo-Besancon, line 200: You updated this because we are currently gas limited, but I propose to put the min value between MAX_OPERATIONS_PER_BLOCK and (MAX_GAS_PER_BLOCK / BASE_OPERATION_GAS_COST) as u32,

massa-node/base_config/config.toml
```diff
@@ -217,8 +217,8 @@
     broadcast_slot_execution_output_channel_capacity = 5000
     # slot execution traces channel capacity
     broadcast_slot_execution_traces_channel_capacity = 5000
-    # Max execution traces (group by block id) to keep in cache
-    execution_traces_limit = 5000
+    # Max slots execution traces to keep in cache
+    execution_traces_limit = 320
```
Leo-Besancon, line 221: There is still a value of 5000 in the default config for execution config for tests : https://github.com/massalabs/massa/blob/lower_sizes_execution_trace/massa-execution-exports/src/test_exports/config.rs#L74

massa-execution-worker/src/execution.rs
AurelienFT: done

massa-node/base_config/config.toml
AurelienFT: done

Leo-Besancon (approved):

PR #4660, opened 2024-03-11
Add effective received amount field

AurelienFT:
  * [x] make tests compile
  * [x] make tests pass 

Not ideal because it doesn't work with transfers made in ABI


Leo-Besancon (approved):
LGTM

PR #4685, opened 2024-05-06
Initial code for execution info

sydhds:
  * [ ] make tests compile
  * [ ] make tests pass 

### DIFF
--- a/massa-execution-exports/src/controller_traits.rs
+++ b/massa-execution-exports/src/controller_traits.rs
@@ -22,7 +22,11 @@ use std::collections::BTreeMap;
 use std::collections::HashMap;
 
 #[cfg(feature = "execution-trace")]
-use crate::types::{AbiTrace, SlotAbiCallStack, Transfer};
+use crate::types_trace_info::AbiTrace;
+#[cfg(feature = "execution-trace")]
+use crate::types_trace_info::SlotAbiCallStack;
+#[cfg(feature = "execution-trace")]
+use crate::types_trace_info::Transfer;
 
 #[cfg_attr(feature = "test-exports", mockall::automock)]
 /// interface that communicates with the execution worker thread
--- a/massa-execution-exports/src/controller_traits.rs
+++ b/massa-execution-exports/src/controller_traits.rs
@@ -22,11 +22,7 @@ use std::collections::BTreeMap;
 use std::collections::HashMap;
 
 #[cfg(feature = "execution-trace")]
-use crate::types_trace_info::AbiTrace;
-#[cfg(feature = "execution-trace")]
-use crate::types_trace_info::SlotAbiCallStack;
-#[cfg(feature = "execution-trace")]
-use crate::types_trace_info::Transfer;
+use crate::types_trace_info::{AbiTrace, SlotAbiCallStack, Transfer};
 
 #[cfg_attr(feature = "test-exports", mockall::automock)]
 /// interface that communicates with the execution worker thread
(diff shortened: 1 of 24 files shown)


massa-execution-exports/src/controller_traits.rs
```diff
@@ -22,7 +22,11 @@ use std::collections::BTreeMap;
 use std::collections::HashMap;
 
 #[cfg(feature = "execution-trace")]
-use crate::types::{AbiTrace, SlotAbiCallStack, Transfer};
+use crate::types_trace_info::AbiTrace;
```
AurelienFT, line 25: Can we wrap them in one import like before ?

massa-execution-exports/src/lib.rs
```diff
@@ -72,7 +72,15 @@ pub use types::{
...
 #[cfg(feature = "execution-trace")]
-pub use types::{
-    AbiTrace, SCRuntimeAbiTraceType, SCRuntimeAbiTraceValue, SlotAbiCallStack, Transfer,
-};
+pub use types_trace_info::AbiTrace;
+#[cfg(feature = "execution-trace")]
+pub use types_trace_info::{SCRuntimeAbiTraceType, SCRuntimeAbiTraceValue};
+
+#[cfg(feature = "execution-trace")]
+pub use types_trace_info::SlotAbiCallStack;
```
AurelienFT, line 84: Can we wrap them in one `use` ?

massa-execution-exports/src/types.rs
```diff
@@ -361,26 +241,12 @@ pub struct ExecutionOutput {
...
-    /// Effective received amount
-    pub effective_received_amount: Amount,
-    /// operation id
-    pub op_id: OperationId,
-    /// success or not
-    pub succeed: bool,
-    /// Fee
-    pub fee: Amount,
+    /// Deferred credits execution (empty if execution-info feature is NOT enabled)
+    pub deferred_credits_execution: Vec<(Address, Result<Amount, String>)>,
```
AurelienFT, line 245: Just a question to make sure I understand well, the fields will always exist even if the feature `execution-info` isn't activated but they will be empty ?

massa-execution-worker/src/execution_info.rs
```diff
@@ -0,0 +1,74 @@
...
+    }
+}
+
+pub enum OperationInfo {
+    RollBuy(u64),
+    RollSell(u64),
+}
+
+pub struct ExecutionInfoForSlot {
+    pub(crate) block_producer_reward: Option<(Address, Amount)>,
```
AurelienFT, line 24: I don't understand why there is less fields in the structure `ExecutionOutput` of `massa-execution-exports` than here. Moreover it's not possible to have common structure defined in `massa-execution-exports` and used in this module ?

massa-execution-exports/src/types.rs
bilboquet: If the feature is not defined, the field does not exist.
Example here: https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=31d2d33c65d9f35de77a00435365fbe1

massa-execution-exports/src/controller_traits.rs
sydhds: Done

massa-execution-exports/src/lib.rs
sydhds: Done

massa-execution-worker/src/execution_info.rs
sydhds: In fact, I forgot to add ExecutionInfoForSlot into ExecutionInfo (which is in ExecutionOutput) but I've just fix this...

sydhds:
ExecuteInfoForSlot has more fields than ExecutionOutput so it makes sense to have both structures as it is defined in this PR. Let me know if this is still an issue?

AurelienFT (approved):

PR #4686, opened 2024-05-06
Fix a few old links in README

m4ns0ur:
  * [ ] make tests compile
  * [ ] make tests pass 


AurelienFT (approved):
Hello,
Thank you for the contribution. We currently have an issue on the CI that will be fixed by the upcoming pull request and we will merge your contribution as soon as it's fixed.

PR #4689, opened 2024-05-08
Create LICENSE

Thykof:
  * [ ] make tests compile
  * [ ] make tests pass 

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,0 +1,21 @@
+MIT License
+
+Copyright (c) 2024 Massa
+
+Permission is hereby granted, free of charge, to any person obtaining a copy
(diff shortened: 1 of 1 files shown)


LICENSE
```diff
@@ -0,0 +1,21 @@
+MIT License
+
+Copyright (c) 2024 Massa
```
Thykof, line 3:
```suggestion
Copyright (c) 2024 Massa Labs
```

qdrn:
We don't want to open source the code yet. Or at least not without constraints (forks etc...).

Thykof:
Context of the PR: on my spare time I have looking at the licenses and I figured out that most of our repos don't have license.

PR #4695, opened 2024-05-27
Add feature slot-replayer

bilboquet:
if feature is enabled:
- don't do speculative execution
- don't take time_cursor() into account while initializing block_sequencer
- print ledger changes while replaying slots

PR #4697, opened 2024-05-29
rollback to mainnet config and Change two constants to config and add pruning block  storing

AurelienFT:
  * [x] make tests compile
  * [x] make tests pass 


Leo-Besancon (approved):
LGTM.

massa-node/src/settings.rs
```diff
@@ -386,6 +389,9 @@ pub struct VersioningSettings {
 pub struct BlockDumpSettings {
     /// Where to dump blocks
     pub(crate) block_dump_folder_path: PathBuf,
+    #[cfg(feature = "dump-block")]
```
bilboquet, line 392: isn't the whole struct depending upon the feature "dump-block"?

AurelienFT: Yes we could do a cfg_attr on the whole structure and do it for all `Serialize` introduced in https://github.com/massalabs/massa/pull/4695 I open an issue

PR #4705, opened 2024-06-12
Fix tests all features

Leo-Besancon:
The goal of this PR is to enable --all-feature on our CI tests configuration.

As a result, all tests will be triggered, and a few fix where needed:
- sandbox is active so chain_id is 77 and the default final_state changes (new SC addresses in tests)
- execution-trace is active and the TU didn't pass
- dump-block is active, along with both backend types, so a few fixes where needed
- heavy testing is disabled is the env var "NEXTEST" is set and at "1"

  * [ ] make tests compile
  * [ ] make tests pass 

### DIFF
--- a/massa-execution-worker/src/tests/universe.rs
+++ b/massa-execution-worker/src/tests/universe.rs
@@ -4,7 +4,7 @@ use std::{
     sync::Arc,
 };
 
-#[cfg(feature = "file_storage_backend")]
+#[cfg(all(feature = "file_storage_backend", not(feature = "db_storage_backend")))]
 use crate::storage_backend::FileStorageBackend;
 #[cfg(feature = "db_storage_backend")]
 use crate::storage_backend::RocksDBStorageBackend;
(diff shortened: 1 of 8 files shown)


massa-execution-worker/src/tests/universe.rs
```diff
@@ -4,10 +4,11 @@ use std::{
     sync::Arc,
 };
 
-#[cfg(feature = "file_storage_backend")]
+#[cfg(all(feature = "file_storage_backend", not(feature = "db_storage_backend")))]
```
bilboquet, line 7: maybe
```rust
#[cfg(all(feature = "db_storage_backend", feature = "db_storage_backend"))]
compile_error!("Features db_storage_backend and file_storage_backend are mutually exclusive");
#[cfg(feature = "file_storage_backend")]
use crate::storage_backend::FileStorageBackend;
#[cfg(feature = "db_storage_backend")]
use crate::storage_backend::RocksDBStorageBackend;
```
is a bit clearer?

bilboquet (approved):
One minor comment.
It's ok for me.

massa-execution-worker/src/tests/universe.rs
Leo-Besancon: But in that case the CI won't work because of the all features flag.
It's possible to set manually all the features to exclude one of them, but If it's possible to avoid I'd prefer :)

bilboquet: oh yes my bad
so simply forget my comment

AurelienFT (approved):

PR #4706, opened 2024-06-12
Improve async message checks

Leo-Besancon:
This improves checks on async messages.
- Avoid adding a message to the pool if the validity slot makes it not executable
- Better catch errors to avoid async message being partly executed (load module fail but transfer coin successful  and not reverted)

  * [ ] make tests compile
  * [ ] make tests pass 


bilboquet (approved):
I'm fine with the code.
It would be nice to have a TU for each change.
That would help to track the intent of the commit (but that's far from mandatory).

Leo-Besancon:
> I'm fine with the code. It would be nice to have a TU for each change. That would help to track the intent of the commit (but that's far from mandatory).

Thanks, I'll see what I can do :)

modship (approved):

bilboquet (approved):

PR #4715, opened 2024-06-20
Fix ledger change to take into account cancelled message balance change

Leo-Besancon:
The ledger changes returned when settling the slot did not take into account the async message cancellation.

I'll try and add a TU to confirm everything is working

  * [ ] make tests compile
  * [ ] make tests pass 


massa-execution-worker/src/context.rs
```diff
@@ -908,6 +908,13 @@ impl ExecutionContext {
             }
         }
 
+        // take the ledger changes again to take into account the balance change of canceled messages
+        let ledger_changes_canceled = self.speculative_ledger.take();
+
+        for (address, change) in ledger_changes_canceled.0 {
```
modship, line 914: haven't do rust for a while but this should works : 
```rust 
ledger_changes.0.extend(ledger_changes_canceled.0);
```

merge two collection

Leo-Besancon: It seems that we should use ledger_changes.apply(...) instead!

bilboquet:
You may want to add a small UT to cover this case :)

Leo-Besancon:
> You may want to add a small UT to cover this case :)

The problem here is that I didn't find a clean way to test the speculative ledger in the execution tests (the test universe seems to only have a handle on the final state). I'll dig a bit deeper on that

massa-execution-worker/src/context.rs
Leo-Besancon: This is outdated, resolving as a result

massa-execution-worker/src/context.rs
```diff
@@ -894,7 +894,7 @@ impl ExecutionContext {
         let deferred_credits_transfers = self.execute_deferred_credits(&slot);
 
         // take the ledger changes first as they are needed for async messages and cache
-        let ledger_changes = self.speculative_ledger.take();
+        let mut ledger_changes = self.speculative_ledger.get_snapshot();
```
bilboquet, line 897: is mut needed?

bilboquet:
> The problem here is that I didn't find a clean way to test the speculative ledger in the execution tests (the test universe seems to only have a handle on the final state). I'll dig a bit deeper on that

If you can't find your way in UT, maybe it's doable with functional tests? (just an idea).

massa-execution-worker/src/context.rs
```diff
@@ -894,7 +894,7 @@ impl ExecutionContext {
         let deferred_credits_transfers = self.execute_deferred_credits(&slot);
 
         // take the ledger changes first as they are needed for async messages and cache
```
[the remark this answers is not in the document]
damip: @Leo-Besancon I think it can be removed.

damip (approved):

Leo-Besancon:
I'll prepare a functional test with a scenario to test both this PR and https://github.com/massalabs/massa/pull/4718.

PR #4722, opened 2024-07-09
Consistent expiry period for async message and block operations

sydhds:
  * [ ] make tests compile
  * [ ] make tests pass 


Leo-Besancon (commented):
If we change the validity end for the async messages, I think we also need to update the rules for pruning them. Can you look into it?

Leo-Besancon (approved):
LGTM, thanks for the update!

PR #4726, opened 2024-07-15
Increase execution stack size

Leo-Besancon:
  * [ ] make tests compile
  * [ ] make tests pass 

### DIFF
--- a/massa-execution-worker/src/worker.rs
+++ b/massa-execution-worker/src/worker.rs
@@ -287,8 +287,11 @@ pub fn start_execution_worker(
 
     // launch the execution thread
     let input_data_clone = input_data.clone();
+
+    // We set the stack size to 200 Mb instead of the default 2 Mb
+    // as a temporary workaround fully fixed by https://github.com/massalabs/massa/pull/4729
     let thread_builder = thread::Builder::new()
-        .stack_size(200 * 1024 * 1024) //200 Mb stack size
+        .stack_size(200 * 1024 * 1024)
         .name("execution".into());
     let thread_handle = thread_builder
         .spawn(move || {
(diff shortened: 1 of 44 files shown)


massa-execution-worker/src/worker.rs
```diff
@@ -287,7 +287,9 @@ pub fn start_execution_worker(
 
     // launch the execution thread
     let input_data_clone = input_data.clone();
-    let thread_builder = thread::Builder::new().name("execution".into());
+    let thread_builder = thread::Builder::new()
+        .stack_size(200 * 1024 * 1024)
+        .name("execution".into()); //200 Mb stack size
```
sydhds, line 292: Comment should be on the previous line

sydhds (commented):
All good but I would have recommended to do the modification (stack size increase) then push another PR for the version updates

Leo-Besancon:
> All good but I would have recommended to do the modification (stack size increase) then push another PR for the version updates

That's true, it would probably be better like that

Leo-Besancon:
> All good but I would have recommended to do the modification (stack size increase) then push another PR for the version updates

The version update is done in https://github.com/massalabs/massa/pull/4730 instead of this one.

massa-execution-worker/src/worker.rs
```diff
@@ -288,8 +288,8 @@ pub fn start_execution_worker(
     // launch the execution thread
     let input_data_clone = input_data.clone();
     let thread_builder = thread::Builder::new()
-        .stack_size(200 * 1024 * 1024)
-        .name("execution".into()); //200 Mb stack size
+        .stack_size(200 * 1024 * 1024) //200 Mb stack size
```
sydhds, line 291: Sorry but I forgot to add that maybe a comment about why this size and that it is a temporary workaround would be nice :)

sydhds (approved):

PR #4729, opened 2024-07-16
Add initial code for recursion limit

Leo-Besancon:
  * [ ] make tests compile
  * [ ] make tests pass 


.github/workflows/ci.yml
```diff
@@ -4,7 +4,7 @@ on:
   push:
     branches: [main, staging, trying]
   pull_request:
-    branches: [main, 'testnet_*']
+    branches: [main, 'testnet_*', 'mainnet_2_3']
```
sydhds, line 7: mainnet_* ?

massa-execution-exports/src/settings.rs
```diff
@@ -102,4 +102,6 @@ pub struct ExecutionConfig {
     pub max_execution_traces_slot_limit: usize,
     /// Where to dump blocks
     pub block_dump_folder_path: PathBuf,
+    /// Max recursive calls depth
+    pub max_recursive_calls_depth: usize,
```
sydhds, line 106: Maybe add "recusrive calls depth in SC"?

sydhds (approved):

massa-execution-worker/src/execution.rs
```diff
@@ -1123,7 +1123,7 @@ impl ExecutionState {
             context.stack = vec![
                 ExecutionStackElement {
                     address: message.sender,
-                    coins: message.coins,
+                    coins: Default::default(),
```
damip, line 1126: maybe do this in a separate PR

damip, line 1126: this is from a separate issue ?

massa-execution-worker/src/execution.rs
```diff
@@ -1175,10 +1175,19 @@ impl ExecutionState {
...
             .write()
-            .load_module(&bytecode, message.max_gas)?;
+            .load_module(&bytecode, message.max_gas)
+        else {
+            let err =
+                ExecutionError::RuntimeError("could not load module for async execution".into());
+            let mut context = context_guard!(self);
+            context.reset_to_snapshot(context_snapshot, err.clone());
+            context.cancel_async_message(&message);
+            return Err(err);
```
damip, line 1188: this is from a separate issue ?

massa-execution-worker/src/interface_impl.rs
```diff
@@ -295,6 +320,8 @@ impl Interface for InterfaceImpl {
             bail!("call stack out of bounds")
         }
 
+        context.recursion_counter -= 1;
```
damip, line 323: beware of unsafe arithmetic

massa-execution-worker/src/interface_impl.rs
```diff
@@ -282,6 +305,8 @@ impl Interface for InterfaceImpl {
             operation_datastore: None,
         });
 
+        context.recursion_counter += 1;
```
damip, line 308: this creates an intrication that will increase future maintenance burden: developers will always need to keep in mind "should I also call `increment_recursion_counter` also when I call `init_call` from the runtime or not ?

Let's think of possible solutions

massa-execution-worker/src/interface_impl.rs
```diff
@@ -1200,6 +1232,15 @@ impl Interface for InterfaceImpl {
 
         let mut execution_context = context_guard!(self);
         let emission_slot = execution_context.slot;
+
+        if Slot::new(validity_end.0, validity_end.1) < Slot::new(validity_start.0, validity_start.1)
+        {
+            bail!("validity end is earlier than the validity start")
+        }
+        if Slot::new(validity_end.0, validity_end.1) < emission_slot {
```
damip, line 1240: maybe "<=" since you can't emit towards the current slot ?

Leo-Besancon:
Oops @damip , seems like the rebase did not work. Here is the compare link: https://github.com/massalabs/massa/compare/e25f021b344b78df3834d3b0e057850f4d9ea224...2cc4a1734845a22d353fc697ec38cdc635dafbfa

massa-execution-worker/src/speculative_async_pool.rs
```diff
@@ -221,9 +221,14 @@ impl SpeculativeAsyncPool {
         }
 
         // Query eliminated messages
-        let eliminated_msg =
+        let mut eliminated_msg =
             self.fetch_msgs(eliminated_infos.iter().map(|(id, _)| id).collect(), true);
 
+        eliminated_msg.extend(eliminated_new_messages.iter().filter_map(|(k, v)| match v {
```
damip, line 227: I don't understand this. Aren't the `eliminated_new_messages` already taken into account in `eliminated_msg` at line 187 ?

massa-execution-worker/src/execution.rs
Leo-Besancon: yes

massa-execution-worker/src/speculative_async_pool.rs
Leo-Besancon: The problem is that if the message is emitted and eliminated in the same slot, it will not be in the changes above and not in the eliminated_new_messages returned.

massa-models/src/config/constants.rs
```diff
@@ -261,6 +261,8 @@ pub const ASYNC_MSG_CST_GAS_COST: u64 = 750_000;
 pub const BASE_OPERATION_GAS_COST: u64 = 800_000; // approx MAX_GAS_PER_BLOCK / MAX_OPERATIONS_PER_BLOCK
 /// Maximum event size in bytes
 pub const MAX_EVENT_DATA_SIZE: usize = 50_000;
+/// Maximum number of recursion for calls
+pub const MAX_RECURSIVE_CALLS_DEPTH: usize = 50;
```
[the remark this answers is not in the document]
Leo-Besancon: I don't have any numbers, but when I tested in sandbox it was the value that made a noticeable lag during execution (and it seemed a reasonable value).
But I can certainly bench more precisely if needed.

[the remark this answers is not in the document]
Leo-Besancon: @damip With a limit of 50, executing the PoC we have leads to a ~420ms operation execution on my PC.
With 25, its ~150ms instead
Do we want to reduce it?

massa-execution-worker/src/interface_impl.rs
Leo-Besancon: @damip Now, we only increment/decrement the recursion counter in the sc-runtime, before and after calling `run_function`. This is clearer I think.

massa-models/src/config/constants.rs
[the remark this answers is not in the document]
Leo-Besancon: Settled for 25

massa-execution-exports/src/test_exports/config.rs
```diff
@@ -76,6 +76,7 @@ impl Default for ExecutionConfig {
             broadcast_slot_execution_traces_channel_capacity: 5000,
             max_execution_traces_slot_limit: 320,
             block_dump_folder_path,
+            max_recursive_calls_depth: 50,
```
sydhds, line 79: this is set to 25 in constant.rs so maybe we should be consistent here. Unless there is a good reason for not being to?

massa-execution-exports/src/settings.rs
```diff
@@ -102,6 +102,8 @@ pub struct ExecutionConfig {
     pub max_execution_traces_slot_limit: usize,
     /// Where to dump blocks
     pub block_dump_folder_path: PathBuf,
+    /// Max recursive calls depth in SC
+    pub max_recursive_calls_depth: usize,
```
sydhds, line 106: Could we use a u32 or u16 here so we known the size in advance?

massa-execution-worker/src/tests/scenarios_mandatories.rs
```diff
@@ -449,6 +449,192 @@ fn test_nested_call_gas_usage() {
     );
 }
 
+/// Test the recursion depth limit in nested calls using call SC operation
+///
+/// Create a smart contract and send it in the blockclique.
```
sydhds, line 454: Looks like the doc here is from another tests?

massa-execution-worker/src/tests/scenarios_mandatories.rs
```diff
@@ -449,6 +449,192 @@ fn test_nested_call_gas_usage() {
...
+/// This smart contract have his sources in the sources folder.
+/// It calls the test function that have a sub-call to the receive function and send it to the blockclique.
+/// We are checking that the gas is going down through the execution even in sub-calls.
+///
+/// This test can fail if the gas is going up in the execution
+#[test]
+fn test_nested_call_recursion_limit_reached() {
+    // setup the period duration
+    let exec_cfg = ExecutionConfig {
+        max_recursive_calls_depth: 0,
```
sydhds, line 464: Add a small comment here indicating that we expect the test to fail as soon as the SC did a first call/local_call?

massa-execution-worker/src/tests/scenarios_mandatories.rs
```diff
@@ -449,6 +449,192 @@ fn test_nested_call_gas_usage() {
...
+/// This smart contract have his sources in the sources folder.
+/// It calls the test function that have a sub-call to the receive function and send it to the blockclique.
+/// We are checking that the gas is going down through the execution even in sub-calls.
+///
+/// This test can fail if the gas is going up in the execution
+#[test]
+fn test_nested_call_recursion_limit_not_reached() {
+    // setup the period duration
+    let exec_cfg = ExecutionConfig {
+        max_recursive_calls_depth: 2,
```
sydhds, line 558: same as previous

massa-execution-worker/src/tests/scenarios_mandatories.rs
```diff
@@ -449,6 +449,192 @@ fn test_nested_call_gas_usage() {
...
+            ..Default::default()
+        });
+    assert!(events.len() >= 2);
+    //println!("events: {:?}", events);
+    assert!(events[1].data.contains("recursion depth limit reached"));
+}
+
+/// Test the recursion depth limit in nested calls using call SC operation
+///
+/// Create a smart contract and send it in the blockclique.
```
sydhds, line 548: same as previous comment

sydhds (changes requested):
All good but can we find a place where we explain why we need this to be implemented?

massa-execution-exports/src/test_exports/config.rs
Leo-Besancon: Good catch, changed :)

massa-execution-exports/src/settings.rs
Leo-Besancon: done

massa-execution-worker/src/tests/scenarios_mandatories.rs
Leo-Besancon: I reworked the comments in 8c5ff99dc5b0330439c64783f5bb4d8cd8ff14d4, let me know if it's better now!

sydhds (approved):

PR #4737, opened 2024-08-02
Fix wrong LedgerChanges::Delete handling

Leo-Besancon:
  * [ ] make tests compile
  * [ ] make tests pass 

### DIFF
--- a/massa-ledger-worker/src/ledger_db.rs
+++ b/massa-ledger-worker/src/ledger_db.rs
@@ -16,6 +16,7 @@ use massa_models::{
 use massa_serialization::{
     DeserializeError, Deserializer, Serializer, U64VarIntDeserializer, U64VarIntSerializer,
 };
+use parking_lot::{lock_api::RwLockReadGuard, RawRwLock};
 use std::collections::{BTreeSet, HashMap};
 use std::fmt::Debug;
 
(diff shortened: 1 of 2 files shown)


massa-ledger-worker/src/ledger_db.rs
```diff
@@ -284,6 +284,9 @@ impl LedgerDB {
     /// * `ledger_entry`: complete entry to be added
     /// * `batch`: the given operation batch to update
     fn put_entry(&self, addr: &Address, ledger_entry: LedgerEntry, batch: &mut DBBatch) {
+        // Ensures any potential previous entry is fully deleted.
+        self.delete_entry(addr, batch);
```
modship, line 288: why it is needed ? we use `put_or_update_entry_value`

massa-ledger-worker/src/ledger_db.rs
```diff
@@ -284,6 +284,9 @@ impl LedgerDB {
...
+
         let db = self.db.read();
```
bilboquet, line 290: Calling `self.delete_entry(addr, batch);` directly duplicates many operations (db.read(), various serialization).

Maybe calling `db.delete_key(batch, serialized_key);` right before each `db.put_or_update_entry_value(batch, serialized_key...` is a bit more efficient.

bilboquet (approved):
looks good to me

PR #4738, opened 2024-08-06
Re-design asc msg trigger ledger changes

Leo-Besancon:
We should be cautious about the cost of this redesign (locks, clones), and if they can be avoided.
  * [ ] make tests compile
  * [ ] make tests pass 


Leo-Besancon:
Closing this PR (that decrease performance) in favor of https://github.com/massalabs/massa/pull/4740

PR #4744, opened 2024-09-19
Fix join error message (was inconsistent with: VM controller thread)

sydhds:
  * [ ] make tests compile
  * [ ] make tests pass 


Leo-Besancon (dismissed):

Leo-Besancon:
Approved the change, I think just the fmt that isn't happy but otherwise ok!

Leo-Besancon (approved):

bilboquet (approved):

PR #4750, opened 2024-09-30
Update rust version to 1.81

sydhds:


massa-bootstrap/src/server.rs
```diff
@@ -63,11 +63,15 @@ use crate::{
     white_black_list::SharedWhiteBlackList,
     BootstrapConfig,
 };
+
+/*
```
Leo-Besancon, line 67: Why is this commented? If we need it for reference, add a comment explaining it, otherwise remove.

massa-client/src/settings.rs
```diff
@@ -10,6 +10,7 @@ lazy_static::lazy_static! {
     pub static ref SETTINGS: Settings = build_massa_settings("massa-client", "MASSA_CLIENT");
 }
 
+#[allow(dead_code)]
```
Leo-Besancon, line 13: There are several new dead codes. Do you know why this is needed? Is it behind a feature that is not detected?

Leo-Besancon (commented):
Thanks for this chore PR!
Will you also cherry-pick on main? I thought you needed this for the CI on non-breaking updates based on the main branch.

massa-bootstrap/src/server.rs
sydhds: Removed

massa-client/src/settings.rs
sydhds: I suggest merging this PR and add a meta issue tracking all of these clippy allow directives and remove/comment them. What do you think?

sydhds:
> Thanks for this chore PR! Will you also cherry-pick on main? I thought you needed this for the CI on non-breaking updates based on the main branch.

Doing this in main breaks many unit test so best to stay here for now I would say...

massa-client/src/settings.rs
Leo-Besancon: Seems good to me!

Leo-Besancon:
> > Thanks for this chore PR! Will you also cherry-pick on main? I thought you needed this for the CI on non-breaking updates based on the main branch.
> 
> Doing this in main breaks many unit test so best to stay here for now I would say...

Interesting, do you know what causes this different behaviour between main and mainnet_2_3?
Because that means everything we release on main will have a broken CI as I understand it, it may be acceptable if we release soonish, but we have to keep this in mind :).

Approving as is then!

Leo-Besancon (approved):

massa-client/src/settings.rs
sydhds: https://github.com/massalabs/massa/issues/4752

PR #4751, opened 2024-09-30
Update actions/checkout to v4

sydhds:
  * [ ] make tests compile
  * [ ] make tests pass 


Leo-Besancon (approved):
Looks good! Just a point regarding CodeCov, we have to check that the Coverage bot still passes on PRs. But IIRC, it's only for PRs based on main, so once this is merge in mainnet_2_3 with have to check the associated PR. Ping me if you have trouble about it.

PR #4766, opened 2024-10-22
Fix async message updates and add unit test

sydhds:


Leo-Besancon (approved):
Great work! The various tested scenarios are nice

massa-execution-worker/src/active_history.rs
```diff
@@ -83,7 +83,9 @@ impl ActiveHistory {
                     return HistorySearchResult::Present(SetUpdateOrDelete::Set(msg));
                 }
                 Some(SetUpdateOrDelete::Update(msg_update)) => {
-                    current_updates.apply(msg_update.clone());
+                    let mut combined_message_update = msg_update.clone();
```
damip, line 86: I think current_updates does not need to be mut anymore

massa-execution-worker/src/active_history.rs
```diff
@@ -83,7 +83,9 @@ impl ActiveHistory {
...
+                    combined_message_update.apply(current_updates);
+                    current_updates = combined_message_update;
                 }
                 Some(SetUpdateOrDelete::Delete) => return HistorySearchResult::Absent,
```
damip, line 90: how about this ? if the `current_updates` actually creates an element while it was deleted before... can it happen?

massa-execution-worker/src/speculative_async_pool.rs
```diff
@@ -287,7 +287,7 @@ impl SpeculativeAsyncPool {
                 }
                 Present(SetUpdateOrDelete::Update(msg_update)) => {
                     current_changes.entry(message_id).and_modify(|e| {
-                        e.apply(msg_update.clone());
+                        *e = msg_update.clone();
```
damip, line 290: why was this changed ?

Leo-Besancon: It did work before, but it technically applied the current_changes twice (the returned msg_update already took them into account, so we directly replace it)

massa-execution-worker/src/active_history.rs
Leo-Besancon: I think it still does because of line 88 :)

Leo-Besancon:
@sydhds unless you think otherwise, I think we can merge this one :)

massa-execution-worker/src/speculative_async_pool.rs
```diff
@@ -287,7 +287,7 @@ impl SpeculativeAsyncPool {
...
                     });
                     return true;
                 }
```
damip, line 293: question to make sure: do we properly handle the case where there is only an update in the speculative part, and the message is in the final state ?

Leo-Besancon: Seems like it to me yes, in that case:
- On line 264 we find an Update in the pool
- Then, we find nothing in the active history so we do nothing (at line 294)
- Then, we find the message in the final state and we apply the update.
@sydhds do you agree?

PR #4768, opened 2024-10-25
Limit event msg size & event number per operation

sydhds:


massa-execution-exports/src/event_store.rs
```diff
@@ -55,48 +55,61 @@ impl EventStore {
...
+
+    /// Get events iterator optionally filtered by given EventFilter
+    pub fn get_filtered_sc_output_events_iter<'b, 'a: 'b>(
+        &'a self,
+        filter: &'b EventFilter,
+    ) -> impl Iterator<Item = &'a SCOutputEvent> + 'b {
+        // Note on lifetimes:
+        // 'a -> is the lifetime for self -> because the iterator returns items from self
+        // 'b -> is the lifetime for filter -> because the returning iterator captures filter
+        // and we have lifetime 'a > 'b because filter can live less than self
```
bilboquet, line 71: do you mean "can't live"?

massa-execution-worker/src/interface_impl.rs
```diff
@@ -1150,6 +1152,24 @@ impl Interface for InterfaceImpl {
 
         let mut context = context_guard!(self);
         let event = context.event_create(data, false);
```
bilboquet, line 1154: maybe move this line just after
```rust 
 if event_per_op >= self.config.max_event_per_operation {
    bail!("Too many event for this operation");
}
```
(only create the event if it's possible)

bilboquet (approved):
2 minor comments
it's looking good to me

massa-execution-exports/src/event_store.rs
sydhds: No here lifetime 'b has the following bound: 'a: 'b. So it means that lifetime 'b can live less than lifetime 'a

massa-execution-worker/src/interface_impl.rs
sydhds: Fixed

bilboquet (approved):

bilboquet (approved):

PR #4769, opened 2024-10-29
Add massa event cache crate

sydhds:


massa-event-cache/src/controller.rs
```diff
@@ -0,0 +1,43 @@
+use crate::config::EventCacheConfig;
+use crate::event_cache::EventCache;
+use massa_models::execution::EventFilter;
+use massa_models::output_event::SCOutputEvent;
+
+#[allow(dead_code)]
```
bilboquet, line 6: useful?

massa-event-cache/src/controller.rs
```diff
@@ -0,0 +1,43 @@
...
+    /// Event stored cache.
+    /// See the `EventCache` documentation for more information.
+    event_cache: EventCache,
+}
+
+impl EventCacheController {
+    /// Creates a new `EventCacheController`
+    pub fn new(cfg: EventCacheConfig) -> Self {
+        let event_cache = EventCache::new(
+            cfg.event_cache_path.clone(),
```
bilboquet, line 21: I think the clone() can be removed
```suggestion
            &cfg.event_cache_path,
```
and in massa-event-cache/src/event_cache.rs:44:
```rust
    pub fn new(
        path: &Path,
```

massa-event-cache/src/event_cache.rs
```diff
@@ -0,0 +1,460 @@
...
+        batch.put(event_key, event_buffer);
+        self.db.write(batch).expect(CRUD_ERROR);
+        self.entry_count = self.entry_count.saturating_add(1);
+
+        debug!("(Event insert) entry_count is: {}", self.entry_count);
+    }
+
+    /// Insert new events in the cache
+    ///
+    /// For performance reason, pass events_len to avoid cloning the iterator
```
bilboquet, line 92: the iterator seems to be cloned just below in `let events_len = events_len.unwrap_or_else(|| events.clone().count());`

massa-event-cache/src/event_cache.rs
```diff
@@ -0,0 +1,460 @@
...
+            ) {
+                (Some(addr1), Some(addr2)) if addr1 != addr2 => return None,
+                (Some(_), None) => return None,
+                _ => (),
+            }
+            Some(event)
+        })
+    }
+
+    /// Try to remove as much as `self.amount_to_snip` entries from the db
```
bilboquet, line 183: much -> many ?

that being said the comment does not in phase with the function implementation, (if snip_amount > self.amount_to_snip) more than self_amount_to_snip are removed
so I guess

```suggestion
    /// Try to remove at least `self.amount_to_snip` entries from the db
```

(and from the call above
```rust
if self.entry_count + events_len >= self.max_entry_count {
    let snip_amount = max(self.snip_amount, events_len);
    self.snip(Some(snip_amount));
}
```
)

In both cases, I don't feel the function enforce any boundary limitation. Maybe the comment is just wrong ^^

massa-event-cache/src/event_cache.rs
```diff
@@ -0,0 +1,460 @@
...
+        }
+
+        assert_eq!(prev_slot, Some(slot_2));
+        assert_eq!(prev_event_index, Some(index_2_2));
+    }
+
+    #[test]
+    #[serial]
+    fn test_insert_more_than_max_entry() {
+        // Test insert (and snip) so we do no store too much event in cache
```
bilboquet, line 311:
```suggestion
        // Test insert (and snip) so we do not store too many events in the cache
```

massa-event-cache/src/event_cache.rs
```diff
@@ -0,0 +1,460 @@
...
+            cache.entry_count,
+            cache.max_entry_count - cache.snip_amount + 1
+        );
+        dbg!(cache.entry_count);
+    }
+
+    #[test]
+    #[serial]
+    fn test_insert_more_than_max_entry_2() {
+        // Test insert_multi_it (and snip) so we do no store too much event in cache
```
bilboquet, line 350:
```suggestion
        // Test insert (and snip) so we do not store too much event in cache
```

massa-event-cache/src/controller.rs
sydhds: I've moved it to the cfg field only otherwise the compiler complains it is never read

massa-event-cache/src/controller.rs
sydhds: Done

massa-event-cache/src/event_cache.rs
sydhds: Yes and cloning the iterator can be avoided if events_len is passed (and not None).

massa-event-cache/src/event_cache.rs
sydhds: Fixed the comment. Let me know if this is clearer now?

bilboquet (approved):

bilboquet (approved):

massa-event-cache/src/config.rs
```diff
@@ -0,0 +1,14 @@
+use std::path::PathBuf;
+
+pub struct EventCacheConfig {
+    /// Path to the hard drive cache storage
+    pub event_cache_path: PathBuf,
+    /// Maximum number of entries we want to keep in the event cache
+    pub event_cache_size: usize,
```
damip, line 7: maybe a more explicit name (size makes me think of a number of bytes for caching): for example max_event_cache_length

massa-event-cache/src/controller.rs
```diff
@@ -0,0 +1,43 @@
...
+            cfg.snip_amount,
+            cfg.thread_count,
+        );
+        Self { cfg, event_cache }
+    }
+
+    pub fn save_events(
+        &mut self,
+        events: impl Iterator<Item = SCOutputEvent> + Clone,
+        events_len: Option<usize>,
```
damip, line 32: instead of passing the length as an extra, simply require the `ExactSizeIterator` trait on the iterator

massa-event-cache/src/event_cache.rs
```diff
@@ -0,0 +1,454 @@
...
+    ($event_slot:expr) => {
+        [&$event_slot.to_bytes()[..], &[MODULE_IDENT]].concat()
+    };
+}
+
+pub(crate) struct EventCache {
+    /// RocksDB database
+    db: DB,
+    /// How many entries are in the db. Count is initialized at creation time by iterating
+    /// over all the entries in the db then it is maintained in memory
```
damip, line 31: The cache is not meant to be loaded from disk, we should clear it every time.
So this counter can safely start at 0

massa-event-cache/src/event_cache.rs
```diff
@@ -0,0 +1,454 @@
...
+
+const OPEN_ERROR: &str = "critical: rocksdb open operation failed";
+const CRUD_ERROR: &str = "critical: rocksdb crud operation failed";
+const EVENT_DESER_ERROR: &str = "critical: event deserialization failed";
+
+/// Module key formatting macro
+#[macro_export]
+macro_rules! event_key {
+    ($event_slot:expr) => {
+        [&$event_slot.to_bytes()[..], &[MODULE_IDENT]].concat()
```
damip, line 23: why `MODULE_IDENT` as a post-fix ?

massa-event-cache/src/event_cache.rs
```diff
@@ -0,0 +1,454 @@
...
+use rocksdb::{DBIterator, IteratorMode, WriteBatch, DB};
+use tracing::debug;
+
+const OPEN_ERROR: &str = "critical: rocksdb open operation failed";
+const CRUD_ERROR: &str = "critical: rocksdb crud operation failed";
+const EVENT_DESER_ERROR: &str = "critical: event deserialization failed";
+
+/// Module key formatting macro
+#[macro_export]
+macro_rules! event_key {
```
damip, line 21: is this macro unused ?

massa-event-cache/src/event_cache.rs
```diff
@@ -0,0 +1,454 @@
...
+    event_ser: SCOutputEventSerializer,
+    /// Event deserializer
+    event_deser: SCOutputEventDeserializer,
+}
+
+impl EventCache {
+    /// Create a new EventCache
+    pub fn new(path: &Path, max_entry_count: usize, snip_amount: usize, thread_count: u8) -> Self {
+        let db = DB::open_default(path).expect(OPEN_ERROR);
+        let entry_count = db.iterator(IteratorMode::Start).count();
```
damip, line 48: clear the DB at start, set entry_count to 0 (see above)

massa-event-cache/src/event_cache.rs
```diff
@@ -0,0 +1,454 @@
...
+            event_key.extend(event.context.index_in_slot.to_be_bytes());
+            event_key
+        };
+        let mut event_buffer = Vec::new();
+        self.event_ser.serialize(&event, &mut event_buffer).unwrap();
+
+        let mut batch = WriteBatch::default();
+        batch.put(event_key, event_buffer);
+        self.db.write(batch).expect(CRUD_ERROR);
+        self.entry_count = self.entry_count.saturating_add(1);
```
damip, line 80: note: this assumes that events are always added, never overwritten. This is a correct assumption but should be written here as a comment

massa-event-cache/src/event_cache.rs
```diff
@@ -0,0 +1,454 @@
...
+                event_key.extend(event.context.index_in_slot.to_be_bytes());
+                event_key
+            };
+            let mut event_buffer = Vec::new();
+            self.event_ser.serialize(&event, &mut event_buffer).unwrap();
+
+            batch.put(event_key, event_buffer);
+        }
+        self.db.write(batch).expect(CRUD_ERROR);
+        self.entry_count = self.entry_count.saturating_add(events_len);
```
damip, line 113: same remark as above on overwriting

massa-event-cache/src/event_cache.rs
```diff
@@ -0,0 +1,454 @@
...
+
+    fn db_iter(&self, mode: Option<IteratorMode>) -> DBIterator {
+        self.db.iterator(mode.unwrap_or(IteratorMode::Start))
+    }
+
+    pub(crate) fn get_filtered_sc_output_events<'b, 'a: 'b>(
+        &'a self,
+        filter: &'b EventFilter,
+    ) -> impl Iterator<Item = SCOutputEvent> + 'b {
+        self.db_iter(Some(IteratorMode::Start)).filter_map(|kvb| {
```
damip, line 126: With millions of events in cache, this function will always crash the node.

TODO:
* only iterate on selected slot ranges
* create indexes with respect to the available filters

massa-event-cache/src/ser_deser.rs
```diff
@@ -0,0 +1,294 @@
...
+            slot_deser: SlotDeserializer::new(
+                (Included(0), Included(u64::MAX)),
+                (Included(0), Excluded(args.thread_count)),
+            ),
+            addr_deser: Default::default(),
+            block_id_deser: OptionDeserializer::new(BlockIdDeserializer::new()),
+            op_id_deser: OptionDeserializer::new(OperationIdDeserializer::new()),
+            data_deser: StringDeserializer::new(U64VarIntDeserializer::new(
+                Included(0),
+                Included(u64::MAX),
```
damip, line 112: note: this should be safe if we are only talking about the cache. But if in the future we use this deserializer for other purposes, this will induce a security threat. It might be worth to introduce limits here as a good practice

massa-event-cache/src/ser_deser.rs
```diff
@@ -0,0 +1,294 @@
...
+    op_id_deser: OptionDeserializer<OperationId, OperationIdDeserializer>,
+    data_deser: StringDeserializer<U64VarIntDeserializer, u64>,
+}
+
+impl SCOutputEventDeserializer {
+    pub fn new(args: SCOutputEventDeserializerArgs) -> Self {
+        Self {
+            u64_deser: U64VarIntDeserializer::new(Included(0), Included(u64::MAX)),
+            // FIXME: is there some limit for call stack len?
+            u32_deser: U32VarIntDeserializer::new(Included(0), Included(u32::MAX)),
```
damip, line 102: note: this should be safe if we are only talking about the cache. But if in the future we use this deserializer for other purposes, this will induce a security threat. It might be worth to introduce limits here as a good practice

massa-event-cache/src/ser_deser.rs
```diff
@@ -0,0 +1,294 @@
...
+    addr_deser: AddressDeserializer,
+    block_id_deser: OptionDeserializer<BlockId, BlockIdDeserializer>,
+    op_id_deser: OptionDeserializer<OperationId, OperationIdDeserializer>,
+    data_deser: StringDeserializer<U64VarIntDeserializer, u64>,
+}
+
+impl SCOutputEventDeserializer {
+    pub fn new(args: SCOutputEventDeserializerArgs) -> Self {
+        Self {
+            u64_deser: U64VarIntDeserializer::new(Included(0), Included(u64::MAX)),
```
damip, line 100: I think we should avoid having "u64_deser"  but name it after the thing it is actually supposed to deserialize so we can set the right limits

massa-event-cache/src/ser_deser.rs
damip: I think we should avoid having "u32_deser" but name it after the thing it is actually supposed to deserialize so we can set the right limits

massa-event-cache/src/ser_deser.rs
damip, line 112: TODO limit

massa-execution-worker/src/execution.rs
```diff
@@ -295,8 +307,9 @@ impl ExecutionState {
 
         // append generated events to the final event store
         exec_out.events.finalize();
-        self.final_events.extend(exec_out.events);
-        self.final_events.prune(self.config.max_final_events);
+        let final_events_count = exec_out.events.0.len();
+        self.final_events_cache
```
damip, line 311: note: this is synchronous, so the speed of save_events is critical as it might slow down the whole slot execution

massa-models/src/output_event.rs
```diff
@@ -19,7 +19,7 @@ impl Display for SCOutputEvent {
 }
 
 /// Context of the event (not generated by the user)
-#[derive(Debug, Clone, Serialize, Deserialize)]
+#[derive(Debug, Clone, Serialize, Deserialize, PartialEq)]
```
damip, line 22: in which case do we need PartialEq for such a complex structure ?

massa-node/base_config/config.toml
```diff
@@ -221,6 +221,12 @@
     broadcast_slot_execution_traces_channel_capacity = 5000
     # Max slots execution traces to keep in cache
     execution_traces_limit = 320
+    # path to the event cache storage
+    event_cache_path = "storage/event_cache/rocks_db"
+    # maximum number of entries we want to keep in the Event cache
+    event_cache_size = 2000
```
damip, line 227: I think we can safely make this much bigger (eg. aim for the equivalent of 10 gigabytes)

damip (commented):
Nice !
Made a couple comments, the most important ones are related to security

massa-event-cache/src/config.rs
sydhds: renamed

massa-event-cache/src/controller.rs
sydhds: Ah yes :)

massa-event-cache/src/event_cache.rs
sydhds: Macro unused and so removed

massa-event-cache/src/event_cache.rs
```diff
@@ -0,0 +1,454 @@
...
+        [&$event_slot.to_bytes()[..], &[MODULE_IDENT]].concat()
+    };
+}
+
+pub(crate) struct EventCache {
+    /// RocksDB database
+    db: DB,
+    /// How many entries are in the db. Count is initialized at creation time by iterating
+    /// over all the entries in the db then it is maintained in memory
+    entry_count: usize,
```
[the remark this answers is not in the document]
sydhds: Clear db at startup

massa-event-cache/src/event_cache.rs
sydhds: Macro removed

massa-event-cache/src/event_cache.rs
sydhds: Done

massa-event-cache/src/event_cache.rs
sydhds: Added a comment with this assumption

massa-event-cache/src/event_cache.rs
```diff
@@ -0,0 +1,454 @@
...
+
+    /// Insert new events in the cache
+    ///
+    /// For performance reason, pass events_len to avoid cloning the iterator
+    pub fn insert_multi_it(
+        &mut self,
+        events: impl Iterator<Item = SCOutputEvent> + Clone,
+        events_len: Option<usize>,
+    ) {
+        let events_len = events_len.unwrap_or_else(|| events.clone().count());
```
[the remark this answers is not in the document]
sydhds: Done

massa-event-cache/src/event_cache.rs
sydhds: Done

massa-event-cache/src/ser_deser.rs
sydhds: Renamed the field but not sure if a limit would make sense here?

massa-event-cache/src/ser_deser.rs
sydhds: Done (see previous comment for a field in config used a limit here)

massa-models/src/output_event.rs
sydhds: It is used in unit test so PartialEq is now only derived when feature "test-exports" is enabled

massa-node/base_config/config.toml
sydhds: Done

massa-event-cache/src/config.rs
```diff
@@ -0,0 +1,16 @@
...
+    /// Maximum number of entries we want to keep in the event cache
+    pub max_event_cache_length: usize,
+    /// Amount of entries removed when `event_cache_size` is reached
+    pub snip_amount: usize,
+    /// Maximum length of an event data (aka event message)
+    pub max_event_data_length: u64,
+    /// Thread count
+    pub thread_count: u8,
+    /// Call stack max length
+    pub max_recursive_call_depth: u16,
```
damip, line 15: WARNING: sometimes the call stack can be HIGHER than the max recursive call depth. Example: at the root level (recursive depth = 1) the call stack contains 2 elements. So at MAX_DEPTH, it would contain MAX_DEPTH + 1 elements. I'm not a big fan of the coupling being created here between very distant features.

Not sure how to improve this.
What do you think ?

massa-event-cache/src/controller.rs
```diff
@@ -0,0 +1,68 @@
...
+    pub(crate) input_data: Arc<(Condvar, Mutex<EventCacheWriterInputData>)>,
+    /// Event cache
+    pub(crate) cache: Arc<RwLock<EventCache>>,
+}
+
+impl EventCacheController for EventCacheControllerImpl {
+    fn save_events(&self, events: VecDeque<SCOutputEvent>) {
+        // lock input data
+        let mut input_data = self.input_data.1.lock();
+        input_data.events = events;
```
damip, line 59: If save_events is called twice without the previous queue being fully consumed, the old non-consumed events will be lost.

Use `input_data.events.extend(events)` instead of `input_data.events = events;`

massa-event-cache/src/event_cache.rs
```diff
@@ -0,0 +1,490 @@
...
+                max_event_data_length,
+            }),
+        };
+
+        event_cache.clear();
+        event_cache
+    }
+
+    fn clear(&mut self) {
+        let iter = self.db.iterator(IteratorMode::Start);
```
damip, line 76: There is a much faster way to clear than iterating over all elements. Check how we do it for the module cache or the state

massa-event-cache/src/controller.rs
```diff
@@ -0,0 +1,68 @@
...
+        // wake up VM loop
+        self.input_data.0.notify_one();
+    }
+
+    fn get_filtered_sc_output_events(&self, filter: &EventFilter) -> Vec<SCOutputEvent> {
```
damip, line 64: note: this function should also look into `input_data`

massa-event-cache/src/event_cache.rs
```diff
@@ -0,0 +1,490 @@
...
+use massa_serialization::{DeserializeError, Deserializer, Serializer};
+use rocksdb::{DBIterator, IteratorMode, WriteBatch, DB};
+use tracing::debug;
+
+const OPEN_ERROR: &str = "critical: rocksdb open operation failed";
+const CRUD_ERROR: &str = "critical: rocksdb crud operation failed";
+const EVENT_DESER_ERROR: &str = "critical: event deserialization failed";
+
+/*
+/// Module key formatting macro
```
damip, line 20: remove commented unused code

massa-event-cache/src/event_cache.rs
```diff
@@ -0,0 +1,490 @@
...
+        // Note:
+        // This assumes that events are always added, never overwritten
+        self.entry_count = self.entry_count.saturating_add(1);
+
+        debug!("(Event insert) entry_count is: {}", self.entry_count);
+    }
+
+    /// Insert new events in the cache
+    ///
+    /// For performance reason, pass events_len to avoid cloning the iterator
```
damip, line 116: remove this comment

massa-event-cache/src/controller.rs
sydhds: Done

massa-event-cache/src/event_cache.rs
sydhds: I've used Db::destroy to avoid the iteration

massa-event-cache/src/controller.rs
sydhds: Done

massa-event-cache/src/event_cache.rs
sydhds: Removed

massa-event-cache/src/config.rs
sydhds: I've renamed the field to max_call_stack_length to avoid the confusion. It is now not limited (u16::MAX) but a note mention the assumption (in massa node). What do you think?

massa-event-cache/src/config.rs
```diff
@@ -0,0 +1,20 @@
...
+    /// Maximum length of an event data (aka event message)
+    pub max_event_data_length: u64,
+    /// Thread count
+    pub thread_count: u8,
+    /// Call stack max length
+    pub max_call_stack_length: u16,
+    /// Maximum number of events per operation
+    pub max_events_per_operation: u64,
+    /// Maximum number of operations per block
+    pub max_operations_per_block: u64,
```
damip, line 19: why is this needed

massa-event-cache/src/config.rs
```diff
@@ -0,0 +1,20 @@
...
+    /// Amount of entries removed when `event_cache_size` is reached
+    pub snip_amount: usize,
+    /// Maximum length of an event data (aka event message)
+    pub max_event_data_length: u64,
+    /// Thread count
+    pub thread_count: u8,
+    /// Call stack max length
+    pub max_call_stack_length: u16,
+    /// Maximum number of events per operation
+    pub max_events_per_operation: u64,
```
damip, line 17: why is this needed

massa-event-cache/src/controller.rs
```diff
@@ -0,0 +1,117 @@
...
+    pub fn new() -> Self {
+        Self {
+            stop: Default::default(),
+            events: Default::default(),
+        }
+    }
+
+    /// Takes the current input data into a clone that is returned,
+    /// and resets self.
+    pub fn take(&mut self) -> Self {
```
damip, line 29: if the structure implements Default, you do not need to implement `take` explicitly, it would work with just mem::take(&mut event_cache_writer_input_data)

massa-event-cache/src/controller.rs
```diff
@@ -0,0 +1,117 @@
...
+    fn save_events(&self, events: VecDeque<SCOutputEvent>) {
+        // lock input data
+        let mut input_data = self.input_data.1.lock();
+        input_data.events.extend(events);
+        // wake up VM loop
+        self.input_data.0.notify_one();
+    }
+
+    fn get_filtered_sc_output_events(&self, filter: &EventFilter) -> Vec<SCOutputEvent> {
+        let lock_0 = self.input_data.1.lock();
```
damip, line 65: this means that while the data is being queried, the execution thread is blocked because it waits for the input_data lock to release in order to be able to send the events it needs to send

massa-event-cache/src/controller.rs
```diff
@@ -0,0 +1,117 @@
...
+                event.context.origin_operation_id,
+            ) {
+                (Some(addr1), Some(addr2)) if addr1 != addr2 => return None,
+                (Some(_), None) => return None,
+                _ => (),
+            }
+            Some(event)
+        });
+
+        let lock = self.cache.read();
```
damip, line 112: Here is what we could do:
* take a lock on the queue
* get filtered events on the queue
* drop the lock on the queue (this avoids deadlock risks from taking multiple locks and also avoids the execution thread waiting for the full query in cache to finish)
* take a lock on the cache
* query the cache
* release the cache lock
* fuse the returned events from the queue with the ones returned from the cache, and deduplicate (because some events might have moved from the queue to the cache between locks so we might list them twice)

massa-event-cache/src/controller.rs
```diff
@@ -0,0 +1,117 @@
...
+        it.cloned()
+            .chain(lock.get_filtered_sc_output_events(filter))
+            .collect()
```
damip, line 115: Is there a limit on the number of returned elements ? if a user asks for all events ever, it will crash the node through out-of-memory

massa-event-cache/src/event_cache.rs
```diff
@@ -0,0 +1,1478 @@
...
+            opts.set_merge_operator_associative("counter merge operator", counter_merge);
+            opts
+        };
+        let db = DB::open(&options, path).expect(OPEN_ERROR);
+
+        let key_builder = EventCacheKeyBuilder::new();
+        // init counters
+        let mut batch = WriteBatch::default();
+        let value = 0u64.to_be_bytes();
+        let key_counter = key_builder.get_counter_key_from_indent(&KeyIndent::EmitterAddress);
```
damip, line 228: we don't need to count the number of elements in the general emitter address index. What we need is to know how many emitter address index items there are FOR A GIVEN EMITTER ADDRESS. That's how we will more accurately determine which cache to use as pivot.

For example if someone queries an is_error=true + an emitter address I would:
* get the number of events in the emitter address index for that particular address
* get the number of events in the is_error index for which the value is true
* then query firs the one that has less elements and filter over the other criterion

massa-event-cache/src/event_cache.rs
```diff
@@ -0,0 +1,1478 @@
...
+        max_event_data_length: u64,
+        max_events_per_operation: u64,
+        max_operations_per_block: u64,
+    ) -> Self {
+        // Clear the db
+        DB::destroy(&Options::default(), path).expect(DESTROY_ERROR);
+        let options = {
+            let mut opts = Options::default();
+            opts.create_if_missing(true);
+            opts.set_merge_operator_associative("counter merge operator", counter_merge);
```
damip, line 219: interesting! didn't know about that

massa-event-cache/src/config.rs
sydhds: this is to estimate the number of events between slots

massa-event-cache/src/config.rs
sydhds: same as above.

massa-event-cache/src/controller.rs
sydhds: Done

massa-event-cache/src/controller.rs
sydhds: Add a query limit

massa-event-cache/src/worker.rs
```diff
@@ -0,0 +1,157 @@
...
+        event_cache: Arc<RwLock<EventCache>>,
+    ) -> Self {
+        Self {
+            input_data,
+            cache: event_cache,
+        }
+    }
+
+    /// Waits for an event to trigger a new iteration in the ....
+    ///
```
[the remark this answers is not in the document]
sydhds: done

massa-event-cache/src/worker.rs
```diff
@@ -0,0 +1,157 @@
...
+            if !input_data.events.is_empty() {
+                return (input_data, false);
+            }
+
+            // if we need to stop, return None
+            if input_data.stop {
+                return (input_data, true);
+            }
+
+            // FIXME / TODO: should we sleep here?
```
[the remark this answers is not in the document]
sydhds: As the slot sequencer is not avail outside of massa execution, I've used a tick delay in the event cache writer thread.

massa-models/src/config/constants.rs
```diff
@@ -398,6 +398,8 @@ pub const MAX_DENUNCIATION_CHANGES_LENGTH: u64 = 1_000;
 
 /// Maximum number of events that can be returned by a query
 pub const MAX_EVENTS_PER_QUERY: usize = 10000;
+/// Delay between writes in event cache writer thread
+pub const EVENT_CACHE_TICK_DELAY: u64 = 100;
```
Leo-Besancon, line 402: Should these two constants be in config.toml instead (one node may want to increase / decrease these values based on their config / their needs)?

massa-event-cache/src/worker.rs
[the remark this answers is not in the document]
Leo-Besancon: Looks good to me! I added another comment below related to the constants, but the tick system works for me :)

massa-event-cache/src/event_cache.rs
sydhds: Done

massa-models/src/config/constants.rs
Leo-Besancon: For `EVENT_CACHE_TICK_DELAY` I agree (it should have almost no impact).
For `MAX_EVENTS_PER_QUERY` maybe it's more useful? (Depending on your performance tests).

massa-event-cache/src/config.rs
damip: why do you need to dynamically compute that estimate ?

massa-event-cache/src/controller.rs
damip: hmm not really

https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=5e171127d993700e68e42b96a13364a0

massa-event-cache/src/controller.rs
```diff
@@ -0,0 +1,127 @@
...
+                (Some(addr1), Some(addr2)) if addr1 != addr2 => return None,
+                (Some(_), None) => return None,
+                _ => (),
+            }
+            Some(event)
+        });
+
+        let mut res_0: BTreeSet<SCOutputEvent> = it.cloned().collect();
+        // Drop the lock on the queue as soon as possible to avoid deadlocks
+        drop(lock_0);
```
damip, line 114: it is better practice to use a scope

massa-event-cache/src/controller.rs
```diff
@@ -0,0 +1,127 @@
...
+
+        let lock = self.cache.read();
+
+        let (_, res_1) = lock.get_filtered_sc_output_events(filter);
+        // Drop the lock on the event cache db asap
+        drop(lock);
```
damip, line 120: use a scope

massa-event-cache/src/event_cache.rs
```diff
@@ -0,0 +1,1730 @@
...
+    /// There is no guarantees that the key will be unique
+    /// Recommended to use high level function like:
+    /// `key_from_event`, `prefix_key_from_indent`,
+    /// `prefix_key_from_filter_item` or `counter_key_from_filter_item`
+    fn key(
+        &self,
+        indent: &KeyIndent,
+        key_type: KeyBuilderType,
+        _is_prefix: bool,
+        is_counter: bool,
```
damip, line 77: `if(is_counter)` is the same as

```
if indent == KeyIndent::Counter {
  assert!(value > 40);
}
```

right ? might as well drop the is_counter boolean to reduce redundancy and error surface

massa-event-cache/src/event_cache.rs
```diff
@@ -0,0 +1,1730 @@
...
+const OPERATION_ID_DESER_ERROR: &str = "critical: deserialization failed for op id in rocksdb";
+const COUNTER_ERROR: &str = "critical: cannot get counter";
+const COUNTER_KEY_CREATION_ERROR: &str = "critical: cannot create counter key";
+
+#[allow(dead_code)]
+/// Prefix u8 used to identify rocksdb keys
+#[derive(Debug, Clone, Copy, PartialEq, Eq, PartialOrd, Ord, IntoPrimitive)]
+#[repr(u8)]
+enum KeyIndent {
+    Counter = 0,
```
damip, line 36: The fact that Counter is an indent makes the behavior very confusing to read.

I would either have 
```
enum KeyIndent {
    EventById,
    EventByIdCounter,
    EventByEmitterAddress,
    EventByEmitterAddressCounter,
    ...
}

// here the format of the EventByEmitterAddress index would be:
//   [KeyIndent::EventByEmitterAddressCounter][len(index_value)][index_value] -> (value = number of items)
//   [KeyIndent::EventByEmitterAddress][len(index_value)][index_value][event_id] -> (no value)
```

or clearly separate the two stages:
```
enum KeyIndent {
    EventById,
    EventByEmitterAddress,
    ...
}

enum KeySubtype {
    Counter,
    Item
}

// here the format of the EventByEmitterAddress index would be:
//   [KeyIndent::EventByIdCounter][len(index_value)][index_value][KeySubtype::Counter] -> (value = number of items)
//   [KeyIndent::EventByIdCounter][len(index_value)][index_value][KeySubtype::Item][event_id] -> (no value)
```

massa-event-cache/src/worker.rs
```diff
@@ -0,0 +1,172 @@
...
+            );
+
+            if stop {
+                // we need to stop
+                break;
+            }
+
+            let mut lock = self.cache.write();
+            lock.insert_multi_it(input_data.events.into_iter());
+            drop(lock);
```
damip, line 90: use a scope

massa-event-cache/src/controller.rs
sydhds: I think a drop is more explicit for a reader to check if we really drop the lock as early as possible. Anyway I've added the scope + keep the drop.

massa-event-cache/src/controller.rs
sydhds: See above

massa-event-cache/src/worker.rs
sydhds: Same as above

massa-event-cache/src/controller.rs
sydhds: For the sake of clarity, std::mem::take do require trait Default impl for T as described here: https://doc.rust-lang.org/core/mem/fn.take.html. 

So this will not work:

```
#[derive(Debug, Clone, Default)]
struct Foo2 {
    i: u32,
}

#[derive(Debug, Clone)]
struct Wrapper3<T>(T);

fn main() {
    let f2 = Foo2 { i: 5 };
    let mut w = Wrapper3(f2);
    let f2_new = core::mem::take(&mut w);
}
```

but this will work:

```
#[derive(Debug, Clone, Default)]
struct Foo2 {
    i: u32,
}

#[derive(Debug, Clone)]
struct Wrapper3<T>(T);

impl<T> Deref for Wrapper3<T> {
    type Target = T;

    fn deref(&self) -> &Self::Target {
        &self.0
    }
}

impl<T> DerefMut for Wrapper3<T> {
    fn deref_mut(&mut self) -> &mut Self::Target {
        &mut self.0
    }
}

fn main() {
    let f2 = Foo2 { i: 5 };
    let mut w = Wrapper3(f2);
    let f1: Foo2 = core::mem::take(&mut w);
}
```

And this only works because:
* Return type is annotated
* Wrapper3 implements Deref & DerefMut traits (MutexGuard also impl these traits)...

sydhds: Fixed

massa-event-cache/src/event_cache.rs
sydhds: In the code, the key format is the following (as documented in DbKeyBuilder::key_from_event method): 

```
        // Db format:
        // * Regular keys:
        //   * Event key: [Event Indent][Slot][Index] -> Event value: Event serialized
        //   * Emitter address key: [Emitter Address Indent][Addr][Addr len][Event key] -> []
        // * Prefix keys:
        //   * Emitter address prefix key: [Emitter Address Indent][Addr][Addr len]
        // * Counter keys:
        //   * Emitter address counter key: [Counter indent][Emitter Address Indent][Addr][Addr len][Event key] -> u64
```

This format has some advantages like:
* Read or Identify the kind/type of key easily (by just looking at the first byte)
* Prefix iteration only return the "regular" key and not the counter key:
  * by using a prefix key for an emitter address, only the emitter address keys are returned and not the emitter address counter key  
* Easily iterate over some keys with just an indent. It's is used in EventCache::snip method and in unit tests (test_counter_0)

 Not sure about the advantages of your solution...

massa-event-cache/src/controller.rs
```diff
@@ -0,0 +1,136 @@
...
+                    (Some(addr1), Some(addr2)) if addr1 != addr2 => return None,
+                    (Some(_), None) => return None,
+                    _ => (),
+                }
+                Some(event)
+            });
+
+            let res_0: BTreeSet<SCOutputEvent> = it.cloned().collect();
+            // Drop the lock on the queue as soon as possible to avoid deadlocks
+            drop(lock_0);
```
damip, line 118: remove this line

massa-event-cache/src/controller.rs
```diff
@@ -0,0 +1,136 @@
...
+            res_0
+        };
+
+        let res_1 = {
+            // Read from db (on disk) events
+            let lock = self.cache.read();
+            let (_, res_1) = lock.get_filtered_sc_output_events(filter);
+            // Drop the lock on the event cache db asap
+            drop(lock);
```
damip, line 127: remove the explicit drop

massa-event-cache/src/event_cache.rs
damip: The simplified builder helps. OK with the design

massa-node/src/main.rs
```diff
@@ -473,6 +477,26 @@ async fn launch(
...
+        thread_count: THREAD_COUNT,
+        // Note: SCOutputEvent call stack comes from the execution module, and we assume
+        //       this should return a limited call stack length
+        //       The value remains for future use & limitations
+        max_call_stack_length: u16::MAX,
+
+        max_events_per_operation: MAX_EVENT_PER_OPERATION as u64,
+        max_operations_per_block: MAX_OPERATIONS_PER_BLOCK as u64,
+        max_events_per_query: MAX_EVENTS_PER_QUERY,
+        tick_delay: Duration::from_millis(EVENT_CACHE_TICK_DELAY),
```
damip, line 495: the tick delay stuff should be dropped and just use a wait() on the condition variable

massa-event-cache/src/controller.rs
damip: ok

massa-node/src/main.rs
sydhds: Fixed

massa-models/src/config/constants.rs
sydhds: Fixed

damip (approved):
Good work on this complicated topic

Leo-Besancon (approved):

PR #4783, opened 2024-12-04
Use coherent PublicKeyFormat in parse_slice

sydhds:


Leo-Besancon (approved):
Note: breaking update, to be versioned

PR #4791, opened 2024-12-17
Reset cache if version becomes active

Leo-Besancon:
  * [ ] make tests compile
  * [ ] make tests pass 

### DIFF
--- a/massa-execution-worker/src/execution.rs
+++ b/massa-execution-worker/src/execution.rs
@@ -201,7 +201,7 @@ impl ExecutionState {
         let cur_execution_version;
         {
             cur_execution_version = execution_context.lock().execution_component_version;
-        } 
+        }
 
         // Instantiate the interface providing ABI access to the VM, share the execution context with it
         let execution_interface = Box::new(InterfaceImpl::new(
@@ -1465,7 +1465,7 @@ impl ExecutionState {
     /// # Returns
     /// An `ExecutionOutput` structure summarizing the output of the executed slot
     pub fn execute_slot(
-        &self,
+        &mut self,
         slot: &Slot,
         exec_target: Option<&(BlockId, ExecutionBlockMetadata)>,
         selector: Box<dyn SelectorController>,
--- a/massa-execution-worker/src/execution.rs
+++ b/massa-execution-worker/src/execution.rs
@@ -198,7 +198,7 @@ impl ExecutionState {
             execution_trail_hash,
         )));
 
-        let cur_execution_version;
+        let cur_execution_version = execution_context.lock().execution_component_version;
         {
             cur_execution_version = execution_context.lock().execution_component_version;
         }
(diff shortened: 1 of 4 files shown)


massa-execution-worker/src/execution.rs
```diff
@@ -196,6 +197,11 @@ impl ExecutionState {
             execution_trail_hash,
         )));
 
+        let cur_execution_version;
```
modship, line 200:
```suggestion
        let cur_execution_version = execution_context.lock().execution_component_version;
```

no ?

massa-module-cache/src/hd_cache.rs
```diff
@@ -36,7 +36,7 @@ macro_rules! metadata_key {
 
 pub(crate) struct HDCache {
     /// RocksDB database
-    db: DB,
+    db: Option<DB>,
```
modship, line 39: why it is now Option<T> ?

massa-module-cache/src/hd_cache.rs
```diff
@@ -72,6 +72,22 @@ impl HDCache {
         }
     }
 
+    pub fn reset(&mut self) {
```
modship, line 75: maybe this should return error and avoid unwrap ?
And in execution we can cast it in ExecError

massa-module-cache/src/hd_cache.rs
Leo-Besancon: In order to destroy the DB to reset it, I needed a way to drop the current DB, which is not possible without the Option as I can't fully drop the `HDCache`.
The only alternative I thought was to iterate through the DB to remove elements, but this may be a heavy operation.

massa-module-cache/src/hd_cache.rs
Leo-Besancon: Here, the Option should only be None during the reset function (with the `take()`), which can only be called once at a time, so handling the error does not seem needed to me.

massa-execution-worker/src/execution.rs
Leo-Besancon: Indeed, I worried that the lock did not go out of scope and caused a deadlock, but that is unnecessary, thanks!

massa-execution-worker/src/execution.rs
```diff
@@ -196,6 +197,8 @@ impl ExecutionState {
             execution_trail_hash,
         )));
 
+        let cur_execution_version = execution_context.lock().execution_component_version;
```
damip, line 200: Instead of taking a lock here:
* create the bare execution context
* then get a copy of its execution component version
* then wrap it in arc mutex

massa-module-cache/src/hd_cache.rs
damip: the only other option I see is

```
        unsafe {
            std::ptr::drop_in_place(&mut self.db);
            std::ptr::write(&mut self.db, create_new_db());
        }
```

damip (approved):

massa-module-cache/src/hd_cache.rs
sydhds: Maye we can use the unsafe snippet here but check with Miri first if the code looks ok and sound?
If you want I can have a look at it?

PR #4792, opened 2024-12-17
fetch_message now returns NoInfo if message_id cannot be found

sydhds:


Leo-Besancon (commented):
I understand the case we want to handle here (the entry `.get(message_id)` is empty for all the history elements in the active history), but it has the side effect to remove any update found in the active history if the message is not there in its entirety.

If we have:
- a message in the final_state
- updates of this message on the active history, but only `SetUpdateOrDelete::Update(msg_update)`, no `SetUpdateOrDelete::Set(msg)`, then we would return NoInfo and fail to update the message.

As a result, I think we have two choices:
- Either we keep the code as is (the code is not very readable but should be correct in its usage)
- Or we do the following:
  - remove the `current_updates: AsyncMessageUpdate` parameter of this function
  - keep a local tallyof updates in the function
  - if we have a message, return Present::Set (potentially updated with the local tally)
  - if we only have updates, return Present::Update with combined updates of the local tally
  - if we have no message and no updates, return NoInfo
  - update the caller's handling:
    - if we have a Present::Set, apply the current changes on the returned message
    - If we have a Present::Update, apply the current changes on the return update, and search in final_state and update it based on the combined updates (current slot + active history updates)
    - if we have a NoInfo, search in final_state and apply the current changes

Let me know about what you think is the best way to approach this

Leo-Besancon (approved):

PR #4793, opened 2024-12-17
Fix entry count massa module cache

Leo-Besancon:
  * [ ] make tests compile
  * [ ] make tests pass 

Note: as we currently clear the cache in main when the node starts, this does not impact the current code.
It's only fixed to avoid potential future errors if this is changed.

As a result, this does not need to be versioned.


massa-module-cache/src/hd_cache.rs
```diff
@@ -58,9 +58,22 @@ impl HDCache {
...
     /// * max_entry_count: maximum number of entries we want to keep in the db
     /// * amount_to_remove: how many entries are removed when `entry_count` reaches `max_entry_count`
+    ///
+    /// Note: entry_count refers to the number of modules stored in the db, which is half the number of (key, value) pairs in the db
     pub fn new(path: PathBuf, max_entry_count: usize, snip_amount: usize) -> Self {
-        let db = DB::open_default(path).expect(OPEN_ERROR);
-        let entry_count = db.iterator(IteratorMode::Start).count();
+        let mut db = DB::open_default(path.clone()).expect(OPEN_ERROR);
+        let mut nb_db_entries = db.iterator(IteratorMode::Start).count();
+        if nb_db_entries % 2 != 0 {
```
modship, line 66: maybe do with it with checked_div ?

damip: checked_div does not check for divisibility, only for divisions by 0

massa-module-cache/src/hd_cache.rs
```diff
@@ -58,9 +58,22 @@ impl HDCache {
     /// * path: where to store the db
     /// * max_entry_count: maximum number of entries we want to keep in the db
     /// * amount_to_remove: how many entries are removed when `entry_count` reaches `max_entry_count`
+    ///
+    /// Note: entry_count refers to the number of modules stored in the db, which is half the number of (key, value) pairs in the db
     pub fn new(path: PathBuf, max_entry_count: usize, snip_amount: usize) -> Self {
-        let db = DB::open_default(path).expect(OPEN_ERROR);
-        let entry_count = db.iterator(IteratorMode::Start).count();
+        let mut db = DB::open_default(path.clone()).expect(OPEN_ERROR);
+        let mut nb_db_entries = db.iterator(IteratorMode::Start).count();
```
damip, line 65: this count might be super heavy (?)

Leo-Besancon:
Updated to simply clear cache in new, and setting the entry_count to 0.

PR #4795, opened 2024-12-18
Features/state changes deferred calls 1

sydhds:


Leo-Besancon (approved):
Shouldn't we handle these differently? (add them together, ...)
```
        self.effective_total_gas = changes.effective_total_gas;
        self.total_calls_registered = changes.total_calls_registered;
```
Otherwise, seems fine to me

PR #4802, opened 2025-01-02
Add TU for delete_datastore_entries function

sydhds:


Leo-Besancon (approved):
Once the clippy is fixed it's good to me

PR #4805, opened 2025-01-06
Limit datastore keys query

### LINKED ISSUE #4721: Re-design datastore key queries ( JSON RPC API / GRPC)
The JsonRPC API endpoint `get_addresses` lists, in the response, all the datastore keys of the given addresses.
One issue with this is that there is no limits in the number of datastore entries for a given address. So if the address has a lot of keys, the response may be too large (e.g. larger than the default 50Mo response limit of the API) which causes issues.
	
For example, the following explorer page errors out because of it: https://explorer.massa.net/mainnet/address/AS1Ba1T2mMpHvLEhTvsNkSDqm4djno2ezVojGrtLf1wRd8ThfZD3
A request / response to the mainnet API showing the issue:
```json
{
    "jsonrpc": "2.0",
    "id": 1,
    "method": "get_addresses",
    "params": [["AS1Ba1T2mMpHvLEhTvsNkSDqm4djno2ezVojGrtLf1wRd8ThfZD3"]]
}
```

```json
{
    "jsonrpc": "2.0",
    "error": {
        "code": -32008,
        "message": "Response is too big",
        "data": "Exceeded max limit of 52428800"
    },
    "id": 1
}
```

TODO:
- [ ] List all the current ways we list datastore key (json rpc API, GRPC)
- [ ] Design a unified way to list the keys
  - Use (Bound, Bound) to define the range of keys to query. This is a way to generalize the optional prefix in some of the listing cases, with Bounds being Included, Excluded or Unbounded.
  - Limit the number of keys returned (max num would be a constant, or maybe a config parameter instead when applicable)
  - Best if non-breaking
- [ ] Implement and test all the key listing cases

modship:
  * [ ] make tests compile
  * [ ] make tests pass 


massa-api-exports/src/address.rs
```diff
@@ -177,3 +177,31 @@ impl std::fmt::Display for AddressFilter {
...
+    /// The address for which to get the datastore keys
+    pub address: Address,
+    /// The prefix to filter the keys
+    pub prefix: Vec<u8>,
+    /// final true means final
+    pub is_final: bool,
+    /// The start key to filter the keys
+    pub start_key: Option<Vec<u8>>,
+    /// inclusive start key
+    pub inclusive_start_key: Option<bool>,
```
damip, line 193: by default it should be true, let's document that

massa-api-exports/src/address.rs
```diff
@@ -177,3 +177,31 @@ impl std::fmt::Display for AddressFilter {
...
+    /// The count of keys to return
+    pub count: Option<u32>,
```
damip, line 195: defaults to the max. Let's document that

massa-node/base_config/config.toml
```diff
@@ -39,6 +39,8 @@
     enable_broadcast = false
     # deferred credits delta (in milliseconds)
     deferred_credits_delta = 7776000000 # ~ 3 months (90×24×60×60×1000) in milliseconds
+    # max datastore keys query (also used by GRPC API)
+    #max_datastore_keys_query = 500
```
damip, line 43: there should be 2 distinct config parameters:
- max datastore keys queries: limits the number of simultaneous key queries in the api
- max datastore keys limit : limits the number of keys returned by a single query

massa-api-exports/src/node.rs
```diff
@@ -46,6 +46,8 @@ pub struct NodeStatus {
     pub chain_id: u64,
     /// minimal fees to include an operation in a block
     pub minimal_fees: Amount,
+    /// current mip version
+    pub current_mip_version: u32,
```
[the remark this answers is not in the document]
modship: cherry picked the commit, the difference will disappear when mip_versionning will be merged into main_2_3

massa-api/src/public.rs
```diff
@@ -458,6 +463,12 @@ impl MassaRpcServer for API<Public> {
         let config = CompactConfig::default();
         let now = MassaTime::now();
 
+        let current_mip_version = self
```
damip, line 466: rebase artifact ?

massa-api/src/public.rs
```diff
@@ -555,6 +566,7 @@ impl MassaRpcServer for API<Public> {
             current_cycle,
             chain_id: self.0.api_settings.chain_id,
             minimal_fees: self.0.api_settings.minimal_fees,
+            current_mip_version,
```
damip, line 569: rebase artifact ?

massa-api/src/public.rs
modship: https://github.com/massalabs/massa/pull/4805#discussion_r1910010381

massa-api/src/public.rs
modship: https://github.com/massalabs/massa/pull/4805#discussion_r1910010381

massa-api/src/public.rs
```diff
@@ -945,6 +957,53 @@ impl MassaRpcServer for API<Public> {
...
     }
 
+    async fn get_address_datastore_keys(
+        &self,
+        arg: Vec<GetAddressDatastoreKeysRequest>,
+    ) -> RpcResult<Vec<GetAddressDatastoreKeysResponse>> {
+        let requests: Vec<ExecutionQueryRequestItem> = arg
+            .into_iter()
+            .map(|request| {
+                from_get_address_datastore_keys(
```
damip, line 967: I think that you should directly use `ExecutionController::query_state` with a batched request because:
- it avoids having different requests happen at different slots (if a slot is executed while this loop is running)
- it avoids having to duplicate code

massa-api/src/public.rs
```diff
@@ -945,6 +957,53 @@ impl MassaRpcServer for API<Public> {
...
+                    request,
+                    self.0.api_settings.max_datastore_keys_queries,
+                )
+            })
+            .collect();
+        // let mut result: Vec<Vec<Vec<u8>>> = Vec::with_capacity(requests.len());
```
damip, line 973: remove comment, rename result2 -> result

massa-api/src/public.rs
```diff
@@ -945,6 +957,53 @@ impl MassaRpcServer for API<Public> {
...
+        for response_item in response.responses {
+            match response_item {
+                Ok(item) => match item {
+                    ExecutionQueryResponseItem::AddressDatastoreKeys(keys, address, is_final) => {
+                        result2.push(GetAddressDatastoreKeysResponse {
+                            address,
+                            is_final,
+                            keys: keys.into_iter().collect(),
+                        });
+                        // result.push(keys.into_iter().collect());
```
damip, line 990: remove

massa-execution-exports/src/mapping_grpc.rs
```diff
@@ -49,15 +51,53 @@ pub fn to_querystate_filter(
                 ExecutionQueryRequestItem::AddressBytecodeFinal(Address::from_str(&value.address)?),
             ),
             exec::RequestItem::AddressDatastoreKeysCandidate(value) => {
+                // take the minimum of the limit and the max_datastore_query_config if it is set
+                let count = value
+                    .limit
+                    .map(|c| {
+                        max_datastore_query_config
+                            .map(|conf| c.min(conf))
```
damip, line 59: same comment here: fail if the user sent a too high limit instead of silently capping

massa-api/src/public.rs
```diff
@@ -945,6 +957,53 @@ impl MassaRpcServer for API<Public> {
...
+                    }
+                    _ => {
+                        return Err(
+                            ApiError::ExecutionError("unexpected response".to_string()).into()
```
damip, line 994: note: this is also triggered when keys from a non-existing address are queried, which is not unexpected but could be quite common

damip: same comment elsewhere and in the grpc

damip: same comment elsewhere

massa-execution-worker/src/execution.rs
```diff
@@ -2189,27 +2191,53 @@ impl ExecutionState {
         )
     }
 
+    pub fn get_final_datastore_keys(
+        &self,
+        addr: &Address,
+        prefix: &[u8],
+        offset: Option<Bound<Vec<u8>>>,
```
damip, line 2198: do not use `Option<Bound>`, use the variant `Bound::Unbounded` when we don't want a bound

massa-api/src/public.rs
modship: Here we match for the type of enum `ExecutionQueryResponseItem`, we can't have anything else than AddressDatastoreKeys type. So i return an Error "in case if" 
For NotFound error, it will be triggered at line 996

massa-api/src/public.rs
modship: done

massa-api/src/public.rs
modship: done

massa-execution-exports/src/mapping_grpc.rs
modship: done

massa-execution-exports/src/mapping_grpc.rs
```diff
@@ -49,15 +51,53 @@ pub fn to_querystate_filter(
...
+                    count,
                 })
             }
             exec::RequestItem::AddressDatastoreKeysFinal(value) => {
+                // take the minimum of the limit and the max_datastore_query_config if it is set
+                let count = value
+                    .limit
+                    .map(|c| {
+                        max_datastore_query_config
+                            .map(|conf| c.min(conf))
```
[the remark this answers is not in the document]
modship: done

PR #4809, opened 2025-01-11
Datastore keys queries (JSON RPC / GRPC)

### LINKED ISSUE #4721: Re-design datastore key queries ( JSON RPC API / GRPC)

damip:
* [x] TODO: LIMIT REQUEST COUNT IN JSONRPC API (the GRPC one is already limited)


massa-execution-worker/src/tests/tests_scan_datastore.rs
```diff
@@ -0,0 +1,429 @@
...
+    assert_eq!(candidate_k.pop_first().unwrap(), b"34".to_vec());
+    assert_eq!(candidate_k.pop_first().unwrap(), b"4".to_vec());
+    assert_eq!(candidate_k.pop_first(), None);
+}
+
+#[test]
+fn test_scan_datastore_with_random_data() {
+    let mut rng = thread_rng();
+    for _ in 0..10 {
+        let keys_count = rng.gen_range(2..50);
```
modship, line 215:
```suggestion
        let keys_count = rng.gen_range(15..50);
```

increase to 15..50 to have at least 15 keys in datastore

modship (approved):

PR #4811, opened 2025-01-14
Disable default gas_calibration on massa-xtask

### LINKED ISSUE #4812: Ensure massa-xtask does not use gas_calibration when building the workspace
On branch `mainnet_2_3`, because of https://github.com/massalabs/massa/pull/4761, a weird behaviour of Cargo (that expects features to only be additive) makes it that a `cargo build` in the root will enable the `gas_calibration` feature on massa-sc-runtime not only for massa-xtask, but also for massa-node and massa-client.

Leo-Besancon:
  * [ ] make tests compile
  * [ ] make tests pass 

Note: tested on a labnet deployment

### DIFF
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -1,2 +1,2 @@
 [alias]
-xtask = "run --package massa_xtask --"
+xtask = "run --features gas_calibration --package massa_xtask --"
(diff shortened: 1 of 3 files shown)


Leo-Besancon:
Closed in favor of https://github.com/massalabs/massa/pull/4818

PR #4819, opened 2025-01-15
Fix event cache estimation for slot filters

### LINKED ISSUE #4816: Massa event cache key ident estimation can be optimized
In case we query events with a filter with slot_start that is after the slot of the last event in the db, the `slots_since` function returns an Err() variant and we have the following WARN:

```
2025-01-10T13:56:05.906052Z  WARN massa_event_cache::event_cache: Could not estimate count for key indent: Event - filter_item: SlotStart(Slot { period: 2014222, thread: 15 }): period overflow error
```

In that case, we also return `max_entry_count` as the estimate.

In that specific case, we want the estimation to return 0 instead (because we have no events in the given period) to avoid unnecessary work.

Leo-Besancon:
  * [ ] make tests compile
  * [ ] make tests pass 

### DIFF
--- a/massa-event-cache/src/event_cache.rs
+++ b/massa-event-cache/src/event_cache.rs
@@ -677,7 +677,10 @@ impl EventCache {
     ) -> Result<u64, ModelsError> {
         match filter_item {
             FilterItem::SlotStart(start) => {
-                let diff = self.last_slot.slots_since(start, self.thread_count)?;
+                let diff = self
+                    .last_slot
+                    .slots_since(start, self.thread_count)
+                    .unwrap_or(0);
                 // Note: Pessimistic estimation - should we keep an average count of events per slot
                 //       and use that instead?
                 Ok(diff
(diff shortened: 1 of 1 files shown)


Leo-Besancon:
Todo: unit test these changes

massa-event-cache/src/event_cache.rs
```diff
@@ -677,21 +677,26 @@ impl EventCache {
     ) -> Result<u64, ModelsError> {
         match filter_item {
             FilterItem::SlotStart(start) => {
-                let diff = self.last_slot.slots_since(start, self.thread_count)?;
+                let diff = self
+                    .last_slot
+                    .slots_since(start, self.thread_count)
+                    .unwrap_or(0);
```
sydhds, line 683: Add a small comment here explaining why we set to 0?

sydhds (changes requested):

massa-event-cache/src/event_cache.rs
Leo-Besancon: Updated with comment and unit test!

sydhds (approved):
All good

PR #4820, opened 2025-01-15
Fix db sanity check 

### LINKED ISSUE #4817: _is_db_valid sanity check does not handle deferred_calls
After completing a bootstrap on another node, we check that our state DB is consistant with the _is_db_valid function. It will try and deserialize all the key/values in our DB and panic when encountering an unknown key.

On buildnet, with the new version with deferred_calls, this new key type in the DB is not checked, and so the node panics:
```
2025-01-15T10:23:44.010632Z  WARN massa_final_state::final_state: Key/value does not correspond to any prefix: serialized_key: [100, 101, 102, 101, 114, 114, 101, 100, 95, 99, 97, 108, 108, 115, 47, 0, 0, 0, 0, 0, 31, 144, 39, 27, 2], serialized_value: [10]
thread 'main' panicked at massa-node/src/main.rs:427:9:
critical: db is not valid after bootstrap
```

We should handle the new case and checks the keys are consistant

modship:
  * [ ] make tests compile
  * [ ] make tests pass 


Leo-Besancon (approved):
Thanks a lot!

PR #4824, opened 2025-01-15
Fix versioning + Started + network restart

sydhds:


massa-versioning/src/versioning.rs
```diff
@@ -1190,7 +1190,9 @@ impl MipStoreRaw {
 
         for (mip_info, mip_state) in &self.store {
             match mip_state.state {
-                ComponentState::Defined(..) => {
+                ComponentState::Defined(..)
+                | ComponentState::Started(..)
+                | ComponentState::LockedIn(..) => {
```
Leo-Besancon, line 1195: Good catch, the fix looks good to me!

One question, if we go in LockedIn before the downtime, but the timeout is during the downtime, si we want to return an error?
It seems to me the vote succeded and so there is no issue?

Leo-Besancon: Answer during sync: To be safe, we return an error even if it's LockedIn, to avoid issues if the activation was supposed to happen during the downtime

Leo-Besancon (approved):

Issue #4839, opened 2025-02-03
Rework event limitation per op


Leo-Besancon:
Currently, in https://github.com/massalabs/massa/pull/4768/files, the limitation of the number of events in a given op is done by filtering on the event db with a given filter.
This has two main drawbacks:
- The limit does not apply to ASCs, DeferredCalls, readonly contexts
- For each generated event, we query the event db and iterate on disk

Instead, we should rely on a counter in the context, that is reset to 0 at each new execution (of an operation, a readonly operation, a async message, a deferred call), and increments only on generate_event() (system events should not be counted towards the limit).

PR #4840, opened 2025-02-03
Use event counter for event limitation per op / asc / deferred_call

Leo-Besancon:
  * [ ] make tests compile
  * [ ] make tests pass 

Cf: https://github.com/massalabs/massa/issues/4839

### DIFF
--- a/massa-execution-worker/src/execution.rs
+++ b/massa-execution-worker/src/execution.rs
@@ -529,6 +529,7 @@ impl ExecutionState {
         if execution_component_version == 0 {
             context.origin_operation_id = Some(operation_id);
         }
+        context.event_count_in_current_exec = 0;
 
         Ok(context_snapshot)
     }
@@ -1199,6 +1200,7 @@ impl ExecutionState {
                     operation_datastore: None,
                 },
             ];
+            context.event_count_in_current_exec = 0;
 
             // check the target address
             if let Err(err) = context.check_target_sc_address(message.destination) {
@@ -1357,6 +1359,7 @@ impl ExecutionState {
                             operation_datastore: None,
                         },
                     ];
+                    context.event_count_in_current_exec = 0;
 
                     // Ensure that the target address is an SC address
                     // Ensure that the target address exists
--- a/massa-execution-worker/src/execution.rs
+++ b/massa-execution-worker/src/execution.rs
@@ -496,9 +496,7 @@ impl ExecutionState {
         // set the context origin operation ID
         // Note: set operation ID early as if context.transfer_coins fails, event_create will use
         // operation ID in the event message
-        if execution_component_version >= 1 {
-            context.origin_operation_id = Some(operation_id);
-        }
+        context.origin_operation_id = Some(operation_id);
 
         // debit the fee from the operation sender
         if let Err(err) =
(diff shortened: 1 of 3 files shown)


massa-execution-worker/src/execution.rs
```diff
@@ -529,6 +529,7 @@ impl ExecutionState {
         if execution_component_version == 0 {
```
damip, line 529: not the same topic, but why is this IF there ? does it mean that we always keep  `origin_operation_id` to None in the new version ??

massa-execution-worker/src/execution.rs
```diff
@@ -1199,6 +1200,7 @@ impl ExecutionState {
                     operation_datastore: None,
                 },
```
damip, line 1201: unrelated again, but where do we set `origin_operation_id` ?

massa-execution-worker/src/execution.rs
Leo-Besancon: We do set the  origin_operation_id , but depending on the version, we set it after or before the transfer_coins (the fix, for newer version, is to set it before : https://github.com/massalabs/massa/pull/4840/files/c6a6910b894cf1d03e87598655a6f070fecb46a0#diff-3c20f3c5be10cedee411058391c72ef3b17e6102f3325e0b9c98b5ca3a3edbf1L500 )

massa-execution-worker/src/execution.rs
Leo-Besancon: https://github.com/massalabs/massa/blob/rework_event_limits/massa-execution-worker/src/execution.rs#L1957

damip:
* rename context.event_count_in_current_exec to context.user_event_count_in_current_exec because system events are not counted (update the comment as well to explain that)
* change the comment description of created_event_index, created_addr_index, created_message_index to explain that those absolute per-slot counters, not per execution 
* make sure the following are explicitly set in the context:
  * gas_remaining_before_subexecution
    * [ ] OPERATION
    * [ ] ASYNC_MESSAGE
    * [ ] DEFERRED_CALL
  * recursion_counter
    * [ ] OPERATION
    * [ ] ASYNC_MESSAGE
    * [ ] DEFERRED_CALL
  * event_count_in_current_exec
    * [ ] OPERATION
    * [ ] ASYNC_MESSAGE
    * [ ] DEFERRED_CALL
  * stack
    * [ ] OPERATION
    * [ ] ASYNC_MESSAGE
    * [ ] DEFERRED_CALL

Leo-Besancon:
> * [x]  rename context.event_count_in_current_exec to context.user_event_count_in_current_exec because system events are not counted (update the comment as well to explain that)
> 
>     * [x]  change the comment description of created_event_index to explain that those absolute per-slot counters, not per execution
> 
>     * [x]  change the comment description of created_addr_index to explain that those absolute per-slot counters, not per execution
> 
>     * [x]  change the comment description of created_message_index to explain that those absolute per-slot counters, not per execution
> 
>     * [x]  make sure the following are explicitly set in the context:
>       
>       * [x]  gas_remaining_before_subexecution
>         
>         * [x]  before executing an OPERATION
>         * [x]  before executing an ASYNC_MESSAGE
>         * [x]  before executing a DEFERRED_CALL
>       * [x]  recursion_counter
>         
>         * [x]  before executing an OPERATION
>         * [x]  before executing an ASYNC_MESSAGE
>         * [x]  before executing a DEFERRED_CALL
>       * [x]  event_count_in_current_exec
>         
>         * [x]  before executing an OPERATION
>         * [x]  before executing an ASYNC_MESSAGE
>         * [x]  before executing a DEFERRED_CALL
>       * [x]  stack
>         
>         * [x]  before executing an OPERATION
>         * [x]  before executing an ASYNC_MESSAGE
>         * [x]  before executing a DEFERRED_CALL

Done, I also explicitely set the origin_operation_id to None before executing an Async message and deferred call (beforehand, we only did it once after exec the operations, but it's clearer that way I think)

damip:
All good ! Gg

damip (approved):

PR #4844, opened 2025-02-04
Allow bypass event limits

Leo-Besancon:
  * [ ] make tests compile
  * [ ] make tests pass 

### DIFF
--- a/massa-execution-worker/src/interface_impl.rs
+++ b/massa-execution-worker/src/interface_impl.rs
@@ -180,17 +180,13 @@ impl InterfaceImpl {
         // NOTE: the router addresses are smart contract SC_1 that will call a smart contract SC_2,
         // and SC_2 will generate the event.
         // This is why we do not check against the last address in the call stack, but the one before (call_stack.len() - 2)
+        // These addresses are respectively the the Mainnet and Buildnet routers for Dusa 
         let allowed_router_addresses = [
             Address::from_str("AS12UMSUxgpRBB6ArZDJ19arHoxNkkpdfofQGekAiAJqsuE6PEFJy").unwrap(),
             Address::from_str("AS1XqtvX3rz2RWbnqLfaYVKEjM3VS5pny9yKDdXcmJ5C1vrcLEFd").unwrap(),
         ];
         let call_stack = context_guard!(self).get_call_stack();
-        if call_stack.len() > 1
-            && allowed_router_addresses.contains(&call_stack[call_stack.len() - 2])
-        {
-            return true;
-        }
-        false
+        call_stack.len() > 1 && allowed_router_addresses.contains(&call_stack[call_stack.len() - 2])
     }
 }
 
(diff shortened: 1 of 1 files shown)


massa-execution-worker/src/interface_impl.rs
```diff
@@ -172,6 +172,26 @@ impl InterfaceImpl {
...
+    fn bypass_event_limitation(&self) -> bool {
+        // NOTE: the router addresses are smart contract SC_1 that will call a smart contract SC_2,
+        // and SC_2 will generate the event.
+        // This is why we do not check against the last address in the call stack, but the one before (call_stack.len() - 2)
+        let allowed_router_addresses = [
+            Address::from_str("AS12UMSUxgpRBB6ArZDJ19arHoxNkkpdfofQGekAiAJqsuE6PEFJy").unwrap(),
+            Address::from_str("AS1XqtvX3rz2RWbnqLfaYVKEjM3VS5pny9yKDdXcmJ5C1vrcLEFd").unwrap(),
+        ];
+        let call_stack = context_guard!(self).get_call_stack();
+        if call_stack.len() > 1
```
bilboquet, line 188: I think you can simply return:
```suggestion
        call_stack.len() > 1 && allowed_router_addresses.contains(&call_stack[call_stack.len() - 2])
```
(no need for `if`)

massa-execution-worker/src/interface_impl.rs
```diff
@@ -172,6 +172,26 @@ impl InterfaceImpl {
...
     }
+
+    // Allow certain addresses to bypass:
+    // - the event size limit
+    // - the user event count per operation / asc / deferred_call
+    fn bypass_event_limitation(&self) -> bool {
+        // NOTE: the router addresses are smart contract SC_1 that will call a smart contract SC_2,
+        // and SC_2 will generate the event.
+        // This is why we do not check against the last address in the call stack, but the one before (call_stack.len() - 2)
+        let allowed_router_addresses = [
```
modship, line 183: small comment why these addresses

modship (approved):

massa-execution-worker/src/interface_impl.rs
```diff
@@ -172,6 +172,22 @@ impl InterfaceImpl {
...
+    fn bypass_event_limitation(&self) -> bool {
+        // NOTE: the router addresses are smart contract SC_1 that will call a smart contract SC_2,
+        // and SC_2 will generate the event.
+        // This is why we do not check against the last address in the call stack, but the one before (call_stack.len() - 2)
+        // These addresses are respectively the the Mainnet and Buildnet routers for Dusa
+        let allowed_router_addresses = [
+            Address::from_str("AS12UMSUxgpRBB6ArZDJ19arHoxNkkpdfofQGekAiAJqsuE6PEFJy").unwrap(),
+            Address::from_str("AS1XqtvX3rz2RWbnqLfaYVKEjM3VS5pny9yKDdXcmJ5C1vrcLEFd").unwrap(),
+        ];
+        let call_stack = context_guard!(self).get_call_stack();
```
damip, line 188: we need to make sure there is no double lock

Leo-Besancon: It is only called in generate_events, where the context is locked after the call to this function.
Otherwise, we could pass the callstack as a parameter of this helper.

damip: yeah let it take just the callstack for safety

Leo-Besancon: done

PR #4851, opened 2025-02-06
Deferred_call : cancelled gas

### LINKED ISSUE #4850: Deferred call : Slot gas on cancel
When a call was cancelled, the call gas shouldn"t be substract from the gas of the slot

modship:
  * [x] make tests compile
  * [x] make tests pass 

### DIFF
--- a/massa-deferred-calls/src/lib.rs
+++ b/massa-deferred-calls/src/lib.rs
@@ -548,6 +548,13 @@ impl DeferredCallRegistry {
                 .deserialize::<DeserializeError>(serialized_value)
                 .is_ok();
         } else if serialized_key.eq(DEFERRED_CALL_TOTAL_REGISTERED.as_bytes()) {
+            let db = self.db.read();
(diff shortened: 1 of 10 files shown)


Leo-Besancon (approved):
LGTM

massa-deferred-calls/src/lib.rs
```diff
@@ -554,11 +548,7 @@ impl DeferredCallRegistry {
                 .deserialize::<DeserializeError>(serialized_value)
                 .is_ok();
         } else if serialized_key.eq(DEFERRED_CALL_TOTAL_REGISTERED.as_bytes()) {
-            return self
-                .registry_changes_deserializer
-                .total_calls_registered_deserializer
-                .deserialize::<DeserializeError>(serialized_value)
-                .is_ok();
+            return true;
```
damip, line 551: maybe directly delete the entry here if it is present and create a followup issue to remove this check and the DEFERRED_CALL_TOTAL_REGISTERED constant completely from the code

PR #4853, opened 2025-02-06
unidirectional stream

### LINKED ISSUE #4661: Unidirectional GRPC stream
We should change all the bidirectional gRPC streams that are not truly necessary into unidirectional streams (server -> client).

This would enable gRPC streaming to work from the web (such as a React application).

Simple example : https://github.com/modship/react-grpc

todo : 

- [x] transactions_thoughput_stream
- [x] new_operations
- [x] new_blocks
- [x] new_filled_blocks
- [x] new_slot_execution_outputs
- [x] new_endorsements

modship:
  * [ ] make tests compile
  * [ ] make tests pass 


Leo-Besancon (approved):
The implementation seems clean, good job!

PR #4866, opened 2025-02-27
Several bootstrap and rocksdb improvements

### LINKED ISSUE #4865: Several bootstrap improvements
- when a client receives the Bootstrap error "all our changes are strictly after last_change_id, we can't be sure we did not miss any", we do not reset the state, and so the same slot is asked to the server (and the bootstrap fails in a loop). We should propagate the SlotTooOld error in the server instead
- Investigate and improve memory usage behaviour for bootstrap servers
- We recently increased the bootstrap limits, but we should also increase the default cache_history len
```
[ledger]
    # length of the changes history. Higher values allow bootstrapping nodes with slower connections
    final_history_length = 100
```
It is the max number of slots we allow between two bootstrap message.
With a default of 300s, it should be 600 slots instead of 100.

Leo-Besancon:
  * [ ] make tests compile
  * [ ] make tests pass 

Solves:
- Bad error handling (SlotTooOld error was not propagated correctly to client)
- Increase the default final_history_length to be consistent with the bootstrap default timeouts
- Avoid filling RocksDB read cache when iterating for bootstrap and other DB options
- Feature gate the use of jemalloc allocator (to be used to avoid memory fragmentation on linux)

### DIFF
--- a/massa-db-worker/src/massa_db.rs
+++ b/massa-db-worker/src/massa_db.rs
@@ -14,7 +14,7 @@ use massa_serialization::{DeserializeError, Deserializer, Serializer, U64VarIntS
 use parking_lot::Mutex;
 use rocksdb::{
     checkpoint::Checkpoint, BlockBasedOptions, Cache, ColumnFamilyDescriptor, Direction,
-    IteratorMode, Options, ReadOptions, WriteBatch, WriteBufferManager, DB,
+    IteratorMode, Options, ReadOptions, WriteBatch, DB,
 };
 use std::path::PathBuf;
 use std::{
@@ -79,6 +79,13 @@ where
     ChangeIDSerializer: Serializer<ChangeID>,
     ChangeIDDeserializer: Deserializer<ChangeID>,
 {
+    /// If we read the db with these options, cache will be disabled (useful for scanning the whole db)
+    pub fn default_read_opts() -> ReadOptions {
+        let mut read_opts = ReadOptions::default();
+        read_opts.fill_cache(false);
+        read_opts
+    }
+
     /// Used for bootstrap servers (get a new batch of data from STATE_CF to stream to the client)
     ///
     /// Returns a StreamBatch<ChangeID>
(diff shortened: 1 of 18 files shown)


massa-db-worker/src/massa_db.rs
```diff
@@ -81,6 +81,13 @@ where
     ChangeIDSerializer: Serializer<ChangeID>,
     ChangeIDDeserializer: Deserializer<ChangeID>,
 {
+    /// If we read the db with these options, cache will be disabled (useful for scanning the whole db)
+    pub fn default_read_opts() -> ReadOptions {
```
Leo-Besancon, line 85: rename to read_opts_for_full_db_traversal

massa-db-worker/src/massa_db.rs
```diff
@@ -621,27 +638,63 @@ impl RawMassaDB<Slot, SlotSerializer, SlotDeserializer> {
...
+
+        // Set hybrid bloom and ribbon filter, to reduce both memory and cpu usage, and add to cache
+        block_opts.set_hybrid_ribbon_filter(5.0, 2);
+        block_opts.set_cache_index_and_filter_blocks(true);
+        block_opts.set_pin_l0_filter_and_index_blocks_in_cache(true);
+        block_opts.set_pin_top_level_index_and_filter(true);
+        // Note: optimized it for memory depends on rocksdb feature "malloc-usable-size"
+        //block_opts.set_optimize_filters_for_memory(true);
+
+        db_opts.set_block_based_table_factory(&block_opts);
```
Leo-Besancon, line 667: Restrain to only the most useful ones (to solve the RAM issue). Otherwise, remove all of them.

massa-db-worker/src/massa_db.rs
```diff
@@ -815,9 +868,9 @@ impl MassaDBController for RawMassaDB<Slot, SlotSerializer, SlotDeserializer> {
                 IteratorMode::From(key, Direction::Reverse)
             }
         };
-
+        let read_opts = Self::default_read_opts();
         Box::new(
-            db.iterator_cf(handle, rocksdb_mode)
+            db.iterator_cf_opt(handle, read_opts, rocksdb_mode)
```
Leo-Besancon, line 873: Check if this function is used for small db iteration, if we want to use the cache in some cases

massa-node/src/jemalloc_init.rs
```diff
@@ -0,0 +1,4 @@
+extern crate tikv_jemalloc_sys;
+
+#[used]
+static INIT: unsafe extern "C" fn(usize) -> *mut std::ffi::c_void = tikv_jemalloc_sys::malloc;
```
Leo-Besancon, line 4: We may want to test this out to better understand the impact, e.g.
- does this mean all allocations are done with Jemalloc?
- are there any edge-cases to consider?

Leo-Besancon: @damip By using the "stats" feature of tikv_jemalloc_sys, and printing the allocation stats, I can confirm jemalloc is used for all allocations of the binary, not just RocksDB allocations.
In my case, I printed stats in the main before and after allocating a huge Vec<u8>.

damip: this is quite surprising to me but ok x)

modship (approved):

PR #4872, opened 2025-03-20
Refactor & cleanup asc message fetch and cache recompute

Leo-Besancon:
  * [ ] make tests compile
  * [ ] make tests pass 

### DIFF
--- a/massa-async-pool/src/pool.rs
+++ b/massa-async-pool/src/pool.rs
@@ -10,8 +10,8 @@ use crate::{
     AsyncMessageSerializer,
 };
 use massa_db_exports::{
-    DBBatch, MassaDirection, MassaIteratorMode, ShareableMassaDBController, ASYNC_POOL_PREFIX,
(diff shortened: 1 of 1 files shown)


massa-async-pool/src/pool.rs
```diff
@@ -320,21 +317,44 @@ impl AsyncPool {
...
+            (STATE_CF, fee_key!(serialized_message_id)),
+            (STATE_CF, coins_key!(serialized_message_id)),
+            (STATE_CF, validity_start_key!(serialized_message_id)),
+            (STATE_CF, validity_end_key!(serialized_message_id)),
+            (STATE_CF, function_params_key!(serialized_message_id)),
+            (STATE_CF, trigger_key!(serialized_message_id)),
+            (STATE_CF, can_be_executed_key!(serialized_message_id)),
+        ];
+
+        let results = db.multi_get_cf(query);
```
Leo-Besancon, line 337: Cleanup: instead, deserialize each field, and then create an AsyncMessage::new() based on these fields.

PR #4873, opened 2025-03-21
Ensure ASC msg ordering is deterministic

Leo-Besancon:
  * [ ] make tests compile
  * [ ] make tests pass 

### DIFF
--- a/massa-async-pool/src/pool.rs
+++ b/massa-async-pool/src/pool.rs
@@ -1124,13 +1124,12 @@ mod tests {
             .map(|(k, v)| {
                 (
                     *k,
-                    v.clone().expect(
-                        format!(
+                    v.clone().unwrap_or_else(|| {
+                        panic!(
                             "message_id {:?} should have been inserted in the pool above",
                             k
                         )
-                        .as_str(),
-                    ),
+                    }),
                 )
             })
             .collect();
@@ -1195,13 +1194,12 @@ mod tests {
             .map(|(k, v)| {
                 (
                     *k,
-                    v.clone().expect(
-                        format!(
+                    v.clone().unwrap_or_else(|| {
+                        panic!(
                             "message_id {:?} should have been inserted in the pool above",
                             k
                         )
-                        .as_str(),
-                    ),
+                    }),
                 )
             })
             .collect();
@@ -1267,13 +1265,12 @@ mod tests {
             .map(|(k, v)| {
                 (
                     *k,
-                    v.clone().expect(
-                        format!(
+                    v.clone().unwrap_or_else(|| {
+                        panic!(
                             "message_id {:?} should have been inserted in the pool above",
                             k
                         )
-                        .as_str(),
-                    ),
+                    }),
                 )
             })
             .collect();
--- a/massa-async-pool/src/pool.rs
+++ b/massa-async-pool/src/pool.rs
@@ -347,7 +347,7 @@ impl AsyncPool {
         &self,
         message_ids: &[AsyncMessageId],
     ) -> Vec<(AsyncMessageId, Option<AsyncMessage>)> {
-        let mut fetched_messages = Vec::new();
+        let mut fetched_messages = Vec::with_capacity(message_ids.len());
 
         for message_id in message_ids {
             let message = self.fetch_message(message_id);
@@ -1121,7 +1121,18 @@ mod tests {
         let to_ser_ = pool.fetch_messages(&message_ids);
         let to_ser = to_ser_
             .iter()
-            .map(|(k, v)| (*k, v.clone().unwrap()))
+            .map(|(k, v)| {
+                (
+                    *k,
+                    v.clone().expect(
+                        format!(
+                            "message_id {:?} should have been inserted in the pool above",
+                            k
+                        )
+                        .as_str(),
+                    ),
+                )
+            })
             .collect();
         serializer.serialize(&to_ser, &mut serialized).unwrap();
 
@@ -1181,7 +1192,18 @@ mod tests {
         let to_ser_ = pool.fetch_messages(&message_ids);
         let to_ser = to_ser_
             .iter()
-            .map(|(k, v)| (*k, v.clone().unwrap()))
+            .map(|(k, v)| {
+                (
+                    *k,
+                    v.clone().expect(
+                        format!(
+                            "message_id {:?} should have been inserted in the pool above",
+                            k
+                        )
+                        .as_str(),
+                    ),
+                )
+            })
             .collect();
         serializer.serialize(&to_ser, &mut serialized).unwrap();
         assert_eq!(to_ser.len(), 2);
@@ -1242,7 +1264,18 @@ mod tests {
         let to_ser_ = pool.fetch_messages(&message_ids);
         let to_ser = to_ser_
             .iter()
-            .map(|(k, v)| (*k, v.clone().unwrap()))
+            .map(|(k, v)| {
+                (
+                    *k,
+                    v.clone().expect(
+                        format!(
+                            "message_id {:?} should have been inserted in the pool above",
+                            k
+                        )
+                        .as_str(),
+                    ),
+                )
+            })
             .collect();
         serializer.serialize(&to_ser, &mut serialized).unwrap();
         assert_eq!(to_ser.len(), 2);
--- a/massa-async-pool/src/pool.rs
+++ b/massa-async-pool/src/pool.rs
@@ -343,13 +343,13 @@ impl AsyncPool {
     ///
     /// This should only be called when we know we want to execute the messages.
     /// Otherwise, we should use the `message_info_cache`.
-    pub fn fetch_messages<'a>(
+    pub fn fetch_messages(
         &self,
-        message_ids: Vec<&'a AsyncMessageId>,
-    ) -> Vec<(&'a AsyncMessageId, Option<AsyncMessage>)> {
+        message_ids: &[AsyncMessageId],
+    ) -> Vec<(AsyncMessageId, Option<AsyncMessage>)> {
         let mut fetched_messages = Vec::new();
 
-        for message_id in message_ids.iter() {
+        for message_id in message_ids {
             let message = self.fetch_message(message_id);
             fetched_messages.push((*message_id, message));
         }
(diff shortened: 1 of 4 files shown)


massa-execution-worker/src/speculative_async_pool.rs
```diff
@@ -351,6 +342,114 @@ impl SpeculativeAsyncPool {
...
+                    msgs_opt.push((**message_id, Some(msg.clone())));
+                }
+                Some(SetUpdateOrDelete::Update(msg_update)) => {
+                    current_changes.entry(message_id).and_modify(|e| {
+                        e.apply(msg_update.clone());
+                    });
+                    msgs_opt.push((**message_id, None));
+                    still_wanted_ids_index.push(index);
+                }
+                Some(SetUpdateOrDelete::Delete) => {
```
modship, line 379: Delete and None has same code 
```suggestion
              Some(SetUpdateOrDelete::Delete) | None => {
```

massa-execution-worker/src/speculative_async_pool.rs
```diff
@@ -351,6 +342,114 @@ impl SpeculativeAsyncPool {
...
+
+        // Then, fetch all the remaining messages from the final state
+        let fetched_msgs = self.final_state.read().get_async_pool().fetch_messages(
+            still_wanted_ids_index
+                .into_iter()
+                .map(|index| wanted_ids[index])
+                .collect(),
+        );
+
+        for (index, (message_id, message)) in fetched_msgs.iter().enumerate() {
```
modship, line 423: into_iter() ?

massa-execution-worker/src/speculative_async_pool.rs
```diff
@@ -351,6 +342,110 @@ impl SpeculativeAsyncPool {
...
+                msg.apply(current_changes.get(message_id).cloned().unwrap_or_default());
+                if delete_existing {
+                    self.pool_changes.push_delete(*message_id);
+                }
+                *msgs_opt.get_mut(index).unwrap() = (*message_id, Some(msg.clone()));
+            }
+        }
+
+        msgs_opt
+    }
```
damip, line 431: A couple issues:
* consecutive changes are applied in the wrong order (active changes then active history not in the right order).
* it is quite complicated and confusing
* delete_existing might happen even if the entry was later deleted in active history for example

Here is a counter-proposal:

```
   fn fetch_msgs_v1(
        &mut self,
        wanted_ids: Vec<AsyncMessageId>,
        delete_existing: bool,
    ) -> Vec<(AsyncMessageId, Option<AsyncMessage>)> {
        // fetch final state
        let mut retrieved = self
            .final_state
            .read()
            .get_async_pool()
            .fetch_messages(&wanted_ids);

        // function to accumulate changes to the retrieved list
        let mut apply_changes = |changes: &AsyncPoolChanges| {
            for (id, msg_opt) in retrieved.iter_mut() {
                match changes.0.get(id) {
                    Some(SetUpdateOrDelete::Set(msg)) => {
                        *msg_opt = Some(msg.clone());
                    }
                    Some(SetUpdateOrDelete::Update(msg_update)) => {
                        if msg_opt.is_none() {
                            *msg_opt = Some(AsyncMessage::default());
                        }
                        msg_opt.as_mut().unwrap().apply(msg_update.clone()); // unwrap checked above
                    }
                    Some(SetUpdateOrDelete::Delete) => {
                        *msg_opt = None;
                    }
                    None => {}
                }
            }
        };

        // fetch active history
        for hist_item in self.active_history.read().0.iter() {
            apply_changes(&hist_item.state_changes.async_pool_changes);
        }

        // fetch current changes
        apply_changes(&self.pool_changes);

        // apply deletions
        if delete_existing {
            for (id, msg_opt) in retrieved.iter() {
                if msg_opt.is_some() {
                    self.pool_changes.push_delete(*id);
                }
            }
        }

        retrieved
    }
```

damip: Note: the async pool fetching in the ActiveHistory is also not great so I didn't use it in this proposal. Ideally we might even want to drop it completely

damip: Note: AsyncPool::fetch_messages is also uselessly complicated. Can be simplified:

```rust
    pub fn fetch_messages(
        &self,
        message_ids: &[AsyncMessageId],
    ) -> Vec<(AsyncMessageId, Option<AsyncMessage>)> {
        let mut fetched_messages = Vec::new();

        for message_id in message_ids {
            let message = self.fetch_message(message_id);
            fetched_messages.push((*message_id, message));
        }

        fetched_messages
    }
```

modship: we have two iter on wanted_ids, can we do it with one ?

Leo-Besancon: > * consecutive changes are applied in the wrong order (active changes then active history not in the right order).
I think it was an issue before but now we ensure to apply the "current updates" to the older updates, so it is in the correct order, but the code is indeed quite confusing (see point two!)

> * it is quite complicated and confusing
very true, we previously had other issues regarding this code (and we even have an ongoing issue to refactor it. We previously wanted to avoid refactoring too much to avoid regressions, but we can certainly do better yes

> * delete_existing might happen even if the entry was later deleted in active history for example
true, I think for the current code we rely on some assumptions that may not be true (now or in the future)

damip: @Leo-Besancon  In most current use cases we would end up querying the final state anyways except if we are executing messages that "just got emitted"

damip: @modship yeah I completely removed that part in my proposal

massa-async-pool/src/pool.rs
```diff
@@ -343,13 +343,13 @@ impl AsyncPool {
...
     /// This should only be called when we know we want to execute the messages.
     /// Otherwise, we should use the `message_info_cache`.
-    pub fn fetch_messages<'a>(
+    pub fn fetch_messages(
         &self,
-        message_ids: Vec<&'a AsyncMessageId>,
-    ) -> Vec<(&'a AsyncMessageId, Option<AsyncMessage>)> {
+        message_ids: &[AsyncMessageId],
+    ) -> Vec<(AsyncMessageId, Option<AsyncMessage>)> {
         let mut fetched_messages = Vec::new();
```
damip, line 350: wish_capacity(message_ids.len())

massa-async-pool/src/pool.rs
```diff
@@ -1117,11 +1117,11 @@ mod tests {
...
         );
 
-        let message_ids: Vec<&AsyncMessageId> = vec![];
-        let to_ser_ = pool.fetch_messages(message_ids);
+        let message_ids = vec![];
+        let to_ser_ = pool.fetch_messages(&message_ids);
         let to_ser = to_ser_
             .iter()
-            .map(|(k, v)| (*(*k), v.clone().unwrap()))
+            .map(|(k, v)| (*k, v.clone().unwrap()))
```
damip, line 1124: Is this unwrap()  safe ?

massa-async-pool/src/pool.rs
```diff
@@ -1177,11 +1177,11 @@ mod tests {
...
             .write_batch(batch, versioning_batch, Some(slot_1));
 
-        let message_ids: Vec<&AsyncMessageId> = vec![&message_id, &message2_id];
-        let to_ser_ = pool.fetch_messages(message_ids);
+        let message_ids = vec![message_id, message2_id];
+        let to_ser_ = pool.fetch_messages(&message_ids);
         let to_ser = to_ser_
             .iter()
-            .map(|(k, v)| (*(*k), v.clone().unwrap()))
+            .map(|(k, v)| (*k, v.clone().unwrap()))
```
damip, line 1184: is the unwrap() safe ?

massa-async-pool/src/pool.rs
```diff
@@ -1238,11 +1238,11 @@ mod tests {
...
```
damip, line 1245: is the unwrap safe

massa-async-pool/src/pool.rs
Leo-Besancon: (this change is in unit tests) here and in the 2 other cases, we only fetch messages that have been added previously into the pool, so it is safe yes.

massa-async-pool/src/pool.rs
```diff
@@ -1117,11 +1117,21 @@ mod tests {
...
-        let to_ser_ = pool.fetch_messages(message_ids);
+        let message_ids = vec![];
+        let to_ser_ = pool.fetch_messages(&message_ids);
         let to_ser = to_ser_
             .iter()
-            .map(|(k, v)| (*(*k), v.clone().unwrap()))
+            .map(|(k, v)| {
+                (
+                    *k,
+                    v.clone().unwrap_or_else(|| {
```
damip, line 1127: can this be factored as `v.clone().expect("blablabla")` ?

massa-async-pool/src/pool.rs
```diff
@@ -1177,11 +1187,21 @@ mod tests {
...
-        let to_ser_ = pool.fetch_messages(message_ids);
+        let message_ids = vec![message_id, message2_id];
+        let to_ser_ = pool.fetch_messages(&message_ids);
         let to_ser = to_ser_
             .iter()
-            .map(|(k, v)| (*(*k), v.clone().unwrap()))
+            .map(|(k, v)| {
+                (
+                    *k,
+                    v.clone().unwrap_or_else(|| {
```
damip, line 1197: can this be factored as `v.clone().expect("blablabla")` ?

massa-async-pool/src/pool.rs
```diff
@@ -1238,11 +1258,21 @@ mod tests {
...
```
damip, line 1268: can this be factored as `v.clone().expect("blablabla")` ?

massa-async-pool/src/pool.rs
Leo-Besancon: at first I had `expect(format!("message_id {:?} blablabla", k)`, but clippy does not like it.

damip (commented):
looks good to me, worth testing a bit

Leo-Besancon:
> looks good to me, worth testing a bit

I've tested that there does not seem to have any regression with nodes connected to Buildnet and comparing the executed ASC messages.

modship (approved):

PR #4875, opened 2025-03-25
Add massa-node and massa-client disclaimer

Leo-Besancon:
TODO:
- [x] Update disclaimer when known
- [ ] Test behaviour in scripts (without the -a flag to automatically accept the terms)

Note: @modship suggested that we may load the disclaimer content dynamically instead of using a constant, but we did not find a good way to avoid duplicating it (in `massa-client/base_config` and `massa-node/base_config`.
We could use build.rs scripts, but it seems overkill. Don't hesitate to propose something else if it makes sence!


Soulthym:
Looks good overall!

I only found one potential issue:

When following instructions in the docs, it recommends appending `&| tee logs.txt`

This causes the binary to not detect that a terminal is running, and panics if the charter was not already approved.

Possible solutions:
1. Improve the error displayed in case it hasnt been accepted to suggest to re-run with the -a flag. Potentially print the charter at this point too (shouldnt be too hard).
Main caveat is that all users will run the command twice the first time.

2. Add the disclaimer and the -a on the commands in the docs. (Fairly easy)
Main caveat is that we dont guarantee tutorials will always do that. We also need the disclaimer there, or by default most users will never see it.

3. Add a cli flag to specify a logging path and use that instead of piping to tee in the docs. (A bit more involved but would solve the problem in future similar cases).

4. Find a better bash formulation that doesnt "hide" the tty from the binary (easy if possible)
Main caveats are that it may not be possible, and might be too involved for some users to manually copy

modship (approved):

PR #4897, opened 2025-05-05
Update MIP activation threshold

Leo-Besancon:
  * [ ] make tests compile
  * [ ] make tests pass 

Note: this will break current functional tests, as their scenario rely on the previous activation constant.

### DIFF
--- a/massa-models/src/config/constants.rs
+++ b/massa-models/src/config/constants.rs
@@ -388,7 +388,7 @@ pub const MAX_LISTENERS_PER_PEER: u64 = 100;
 // Constants used in versioning
 //
 /// Threshold to accept a new versioning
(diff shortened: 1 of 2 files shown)


massa-models/src/config/constants.rs
```diff
@@ -388,7 +388,7 @@ pub const MAX_LISTENERS_PER_PEER: u64 = 100;
 // Constants used in versioning
 //
 /// Threshold to accept a new versioning
-pub const VERSIONING_THRESHOLD_TRANSITION_ACCEPTED: Ratio<u64> = Ratio::new_raw(75, 100);
+pub const VERSIONING_THRESHOLD_TRANSITION_ACCEPTED: Ratio<u64> = Ratio::new_raw(50, 100);
```
bilboquet, line 391: 51% would sound more logical to me.
That being said, it's an arbitrary choice that is totally acceptable.

bilboquet (approved):

PR #4898, opened 2025-05-07
Mip massa client

### LINKED ISSUE #4896: Voting from massa client
Multiple stakers are scared to copy their staking private key into a browser enabled (non-server) personal computer as it presents a true security risk. But this is currently required to use mip.massa.net.

The idea would be to enable users to interact with the voting system from the massa CLI client (eg. List proposals, create a proposal, see details of a proposal, vote, see their supply of masog).

This would allow users to use the same command line server software they already use for managing their nodes in order to vote. And that way they won't be exposing their staking secret keys.

modship:
  * [ ] make tests compile
  * [ ] make tests pass 

### DIFF
--- a/massa-client/src/cmds.rs
+++ b/massa-client/src/cmds.rs
@@ -297,8 +297,8 @@ pub enum Command {
 
     #[strum(
         ascii_case_insensitive,
-        props(args = "Address ProposalId yes/no/blank", pwd_not_needed = "true"),
-        message = "mip_vote"
+        props(args = "Address ProposalId yes/no/blank"),
+        message = "vote for a proposal"
     )]
     mip_vote,
 
@@ -1299,6 +1299,9 @@ impl Command {
                 std::process::exit(0);
             }
             Command::mip_vote => {
+                if parameters.is_empty() || parameters.len() != 3 {
+                    bail!("wrong number of parameters");
+                }
                 let wallet = wallet_opt.as_mut().unwrap();
 
                 let voting_config = VotingConfig::new(client.chain_id);
--- a/massa-client/src/cmds.rs
+++ b/massa-client/src/cmds.rs
@@ -297,7 +297,7 @@ pub enum Command {
 
     #[strum(
         ascii_case_insensitive,
-        props(args = "Address ProposalId yes/no/blank"),
+        props(args = "Address ProposalId Fee yes/no/blank"),
         message = "vote for a proposal"
     )]
     mip_vote,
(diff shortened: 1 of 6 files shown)


massa-client/src/cmds.rs
```diff
@@ -1307,10 +1298,138 @@ impl Command {
             Command::exit => {
                 std::process::exit(0);
             }
+            Command::mip_vote => {
+                if parameters.is_empty() || parameters.len() != 3 {
```
bilboquet, line 1302: you can simplify the condition to `parameters.len() != 3`

modship: done

massa-client/src/cmds.rs
```diff
@@ -1307,10 +1298,138 @@ impl Command {
...
+                    bail!("wrong number of parameters");
+                }
+                let wallet = wallet_opt.as_mut().unwrap();
+
+                let voting_config = VotingConfig::new(client.chain_id);
+                let addr = parameters[0].parse::<Address>()?;
+                let proposal_id = parameters[1].parse::<u64>()?;
+                let vote_str = parameters[2].parse::<String>()?;
+                let fee = Amount::from_str("0.01")?;
```
bilboquet, line 1311: IIRC, fees are explicit for other commands.

It's very few, but as it touches the balance, I think it's better to not remove them "silently".

massa-client/src/cmds.rs
```diff
@@ -1307,10 +1298,138 @@ impl Command {
...
+                    .public
+                    .get_datastore_entries(vec![balance_requests])
+                    .await?;
+
+                let balance_bytes = balance_results
+                    .first()
+                    .and_then(|result| result.final_value.as_ref())
+                    .ok_or_else(|| anyhow!("No balance found"))?;
+                let balance_masog = convert_args_u256_to_u128(balance_bytes).unwrap();
+                if balance_masog < 1 {
```
bilboquet, line 1328: Shouldn't this `1` be a constant in the config file?
(MINIMUM_MASOG_TO_VOTE or something)

massa-client/src/cmds.rs
```diff
@@ -1307,10 +1298,138 @@ impl Command {
...
+                let mut buffer = vec![];
+                buffer.extend_from_slice(&proposal_id.to_le_bytes());
+                buffer.extend_from_slice(&vote.to_le_bytes());
+
+                let op = OperationType::CallSC {
+                    target_addr: voting_config.governance_sc_addr,
+                    target_func: "vote".to_string(),
+                    param: buffer,
+                    max_gas: 4894131,
+                    coins: Amount::from_str("1")?,
```
bilboquet, line 1348: is it required to pass 1 MAS?

massa-client/src/cmds.rs
modship: i can add it to cmd parameters

bilboquet (approved):

PR #4899, opened 2025-05-07
Fix 4894

bilboquet:
  * [ ] make tests compile
  * [X] make tests pass 

### DIFF
--- a/massa-ledger-worker/src/ledger_db.rs
+++ b/massa-ledger-worker/src/ledger_db.rs
@@ -217,12 +217,9 @@ impl LedgerDB {
         };
         let end_bound = match end_bound {
             std::ops::Bound::Unbounded => {
-                let end_bound = get_prefix_bounds(&datastore_prefix_from_address(addr, &[])).1;
-                if end_bound == std::ops::Bound::Unbounded {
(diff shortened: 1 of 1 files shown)


massa-ledger-worker/src/ledger_db.rs
```diff
@@ -216,7 +216,13 @@ impl LedgerDB {
             }
         };
         let end_bound = match end_bound {
-            std::ops::Bound::Unbounded => std::ops::Bound::Unbounded,
+            std::ops::Bound::Unbounded => {
+                let end_bound = get_prefix_bounds(&datastore_prefix_from_address(addr, &[])).1;
+                if end_bound == std::ops::Bound::Unbounded {
+                    panic!("End prefix can't be Unbounded for existing address");
```
damip, line 222: no need to perform this check and panic. If `get_prefix_bounds` returns Unbounded, it means it's fine to go till the end

Leo-Besancon (approved):
Aside from @damip's comment, LGTM, and the added test case make sence!

massa-ledger-worker/src/ledger_db.rs
bilboquet: Done, thanks

PR #4908, opened 2025-05-20
Remove versioning of MIP-0001

Leo-Besancon:
  * [ ] make tests compile
  * [ ] make tests pass 

Corresponding PR on sc-runtime: https://github.com/massalabs/massa-sc-runtime/pull/351

Left TODO:
- [ ] Update Version to MAIN.3.0: https://github.com/massalabs/massa/pull/4909
- [ ] Ensure if MIP needs to be removed or not, I don't think so because it is part of the state so it is needed to stay synced?
- [ ] Merge related PRs to main and reference commit ids


massa-execution-exports/src/settings.rs
```diff
@@ -46,7 +46,7 @@ pub struct ExecutionConfig {
     /// block creation reward (v0)
     pub block_reward_v0: Amount,
     /// block creation reward (v1)
```
bilboquet, line 48: remove `(v1)`

massa-execution-exports/src/settings.rs
```diff
@@ -96,7 +96,7 @@ pub struct ExecutionConfig {
     /// max size of event data, in bytes
     pub max_event_size_v0: usize,
```
bilboquet, line 97: Still needed?

Leo-Besancon: This one yes because we still need to cap the events that bypass the v1 limitation

massa-execution-worker/src/execution.rs
```diff
@@ -1568,93 +1548,44 @@ impl ExecutionState {
...
+                            exec_info.deferred_calls_messages.push(Ok(_exec));
                         }
                     }
                 }
+                Err(err) => {
+                    deferred_calls_stats.1 += 1;
+                    let msg = format!("failed executing deferred call: {}", err);
+                    #[cfg(feature = "execution-info")]
+                    exec_info.deferred_calls_messages.push(Err(msg.clone()));
+                    dbg!(msg);
```
bilboquet, line 1587: not sure we want to keep this `dbg!()` we'd better use `tracing::debug!()`

massa-execution-worker/src/interface_impl.rs
```diff
@@ -257,20 +257,15 @@ impl Interface for InterfaceImpl {
     }
 
     fn get_interface_version(&self) -> Result<u32> {
```
bilboquet, line 259: isn't this function now useless?

Leo-Besancon: It won't be used in the curent code, but I'm not sure if it's worth removing from the interface (e.g. if we just have to add it back on the next versioning). What do you think?

Leo-Besancon: Thanks! Updated in latest commit, and I made sure we currently do not call it in the codebase (massa or massa-sc-runtime)

bilboquet (approved):

modship (approved):

massa-execution-worker/src/context.rs
```diff
@@ -1701,7 +1464,7 @@ fn init_prng(execution_trail_hash: &massa_hash::Hash) -> Xoshiro256PlusPlus {
 }
 
 /// Truncates a string to a maximum byte length, ensuring it remains valid UTF-8 in the process.
-fn truncate_string(string: String, max_byte_len: usize) -> String {
+pub fn truncate_string(string: String, max_byte_len: usize) -> String {
```
damip, line 1467: why did this become public ?

Leo-Besancon: To use it in execution on the following error:
`let mut error = format!("could not spend fees: {}", err);`

I forgot to do it before, but:
- It should not matter as the error is for transfer_coins and should never get truncated anyway
- I can make it pub(crate) instead

damip: ok all good

PR #4918, opened 2025-06-17
Devn 28 16

Leo-Besancon:
  * [ ] make tests compile
  * [ ] make tests pass 


Leo-Besancon:
Note: now buildnet should be up to date with Mainnet, so going from one to the other should only require changing:
- The node version (MAIN > DEVN)
- The initial / config files (bootstrap list, initial peers, initial ledger, initial rolls..)
- Potentially, the tests to have a passing CI

PR #4921, opened 2025-06-25
Win sign

peterjah:


Leo-Besancon (approved):
Looks good to me if the CD succeeds and binaries are signed.
Thank you!

PR #4931, opened 2025-09-15
fix: only consider .yaml files as wallets

Leo-Besancon:
  * [ ] make tests compile
  * [ ] make tests pass 


Leo-Besancon:
> OK if tested and works !

Tested and works on Windows at least. Do you think it's enough or should we also test on macos/ubuntu?

damip:
I think it should be ok ! bien joué

PR #4933, opened 2025-09-15
feat: Hide secret_keys from client history

Leo-Besancon:
  * [ ] make tests compile
  * [ ] make tests pass 

### DIFF
--- a/massa-client/src/cmds.rs
+++ b/massa-client/src/cmds.rs
@@ -41,6 +41,7 @@ use strum_macros::{Display, EnumIter, EnumString};
 /// the order they are defined is the order they are displayed in so be careful
 /// Maybe it would be worth renaming some of them for consistency
 /// Use props(pwd_not_needed = "true") if the command does not need an access to the wallet, to avoid unnecessary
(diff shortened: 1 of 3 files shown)


massa-client/src/cmds.rs
```diff
@@ -424,8 +425,15 @@ impl Command {
 
     /// Returns true if the command needs wallet access
     pub(crate) fn is_pwd_needed(&self) -> bool {
-        !(self.get_str("pwd_not_needed").is_some()
-            && self.get_str("pwd_not_needed").unwrap() == "true")
+        let pwd_not_needed = self.get_str("pwd_not_needed");
+        !pwd_not_needed.map_or(false, |f| f == "true")
```
bilboquet, line 429: Thank you for the double negation x)

Maybe a match is more readable? (but it's only a suggestion)
```suggestion
    match self.get_str("pwd_not_needed") {
        Some("true") => false, // pwd_not_needed si "true", so return false
        _ => true,            // Else, pwd is needed
    }
```

bilboquet (approved):

PR #4942, opened 2025-10-13
decouple block production from blocking on pool slowness

bilboquet:
[ ] document all added functions
  * [ ] make tests compile
  * [ ] make tests pass 

### DIFF
--- a/massa-api/src/public.rs
+++ b/massa-api/src/public.rs
@@ -521,14 +521,22 @@ impl MassaRpcServer for API<Public> {
         };
 
         let pool_stats = (
-            self.0
+            match self
+                .0
                 .pool_command_sender
-                .get_operation_count(None)
-                .unwrap_or(0),
-            self.0
+                .get_operation_count(Some(self.0.api_settings.pool_api_timeout))
+            {
+                Ok(count) => count,
+                Err(e) => return Err(ApiError::PoolError(e).into()),
+            },
+            match self
+                .0
                 .pool_command_sender
-                .get_endorsement_count(None)
-                .unwrap_or(0),
+                .get_endorsement_count(Some(self.0.api_settings.pool_api_timeout))
+            {
+                Ok(count) => count,
+                Err(e) => return Err(ApiError::PoolError(e).into()),
+            },
         );
 
         let next_slot_result = last_slot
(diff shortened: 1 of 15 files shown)


massa-api/src/public.rs
```diff
@@ -521,8 +521,14 @@ impl MassaRpcServer for API<Public> {
         };
 
         let pool_stats = (
-            self.0.pool_command_sender.get_operation_count(),
-            self.0.pool_command_sender.get_endorsement_count(),
+            self.0
+                .pool_command_sender
+                .get_operation_count(Some(self.0.api_settings.pool_api_timeout))
+                .unwrap_or(0),
```
Leo-Besancon: The timeout looks good to me, but I'm unsure if we want to unwrap_or() or return an ApiError instead (saying that the Pool timed-out). What do you think?

massa-pool-worker/src/tests/endorsement_pool_tests.rs
```diff
@@ -7,42 +7,72 @@ use massa_signature::KeyPair;
 
 use super::tools::{create_endorsement, default_mock_execution_controller, pool_test};
 
-fn default_mock_selector(address: Address) -> Box<MockSelectorController> {
```
Leo-Besancon: as a side-note: why did we need to change this?

massa-api/src/public.rs
bilboquet: agreed, I'm going to change this. (see next commit)

massa-pool-worker/src/tests/endorsement_pool_tests.rs
bilboquet: Thinking about it twice, I think it's a simple code factorization

massa-node/.gitignore
```diff
@@ -0,0 +1 @@
+.qodo
```
Leo-Besancon: massa-node/.gitignore to uncommit?

Leo-Besancon (commented):
just a few nitpicks, otherwise all good!

massa-node/massa.code-workspace
```diff
@@ -0,0 +1,12 @@
+{
+	"folders": [
```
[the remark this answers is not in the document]
bilboquet: Sorry for the inconvenience.

massa-node/.gitignore
bilboquet: Sorry for the inconvenience.

PR #4949, opened 2025-11-13
fix: use THREAD_COUNT instead of hardcoded 32 in operation_injector

radik878:
Replace hardcoded 32 with THREAD_COUNT in operation_injector to prevent index panics, infinite loops, and incorrect TPS distribution when the protocol thread count changes, keeping the injector aligned with the global THREAD_COUNT constant used across the node


Leo-Besancon:
Closing as stale

PR #4955, opened 2026-01-14
add macOs entitlements

peterjah:
Add macOs entitlements to allow running local api


Leo-Besancon (approved):
LGTM (but I'm not on MacOS to test)

PR #4958, opened 2026-01-16
add jemalloc by default on unix targets

### LINKED ISSUE #4957: Add jemalloc feature by default
The cargo feature `jemalloc` has been added to help RAM usage with RocksDB on the default allocator (at least for linux targets, potentially with macos target too).
The feature has not been added as a default feature, and it seems the recent MAIN.4.1 release has worse performances without it.

We should add the feature as a default (and ensure the pre-built binaries use them), and test on all targets with a MAIN.4.2 release.

Leo-Besancon:
  * [ ] make tests compile
  * [ ] make tests pass 

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -16,8 +16,6 @@ members = [
   "massa-factory-exports",
   "massa-factory-worker",
   "massa-channel",
-  "massa-consensus-exports",
-  "massa-consensus-worker",
   "massa-hash",
   "massa-logging",
   "massa-metrics",
--- a/massa-node/Cargo.toml
+++ b/massa-node/Cargo.toml
@@ -5,6 +5,7 @@ authors = ["Massa Labs <info@massa.net>"]
 edition = "2021"
 
 [features]
+default = []
 beta = []
 resync_check = []
 deadlock_detection = []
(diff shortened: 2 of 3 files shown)


Leo-Besancon:
This PR has been tested on buildnet, it seem to indeed automatically enable jemalloc on unix systems without the need to provide the `--features jemalloc` flag. However, disabling the feature means editing the Cargo.toml file, there is no equivalent `-no-default-features` here.

It seems that either:
- we add in the documentation the --features jemalloc flag when building on unix systems, and generate the prebuilt binaries with the feature enabled (apart from windows).
- or we consider the drawback of this PR ok and we merge this

bilboquet:
> This PR has been tested on buildnet, it seem to indeed automatically enable jemalloc on unix systems without the need to provide the `--features jemalloc` flag. However, disabling the feature means editing the Cargo.toml file, there is no equivalent `-no-default-features` here.
> 
> It seems that either:
> 
>     * we add in the documentation the --features jemalloc flag when building on unix systems, and generate the prebuilt binaries with the feature enabled (apart from windows).
> 
>     * or we consider the drawback of this PR ok and we merge this

If I understand correctly, the default allocator does not work correctly in our case, and we need jemalloc as default (on Unixes). So I'd go with this PR (including its drawback), and it's totally fine to me that someone who wants to use something known to be broken has to edit the code.

bilboquet (approved):

modship (approved):

PR #4960, opened 2026-01-23
Fix too many open files

### LINKED ISSUE #4959: Investigate `Too many open files` IO error
A node has crashed with the following:

```txt
thread 'execution' panicked at massa-db-worker/src/massa_db.rs:769:14:
Failed to create checkpoint: Error { message: "IO error: While opening a file for sequentially reading: storage/ledger/rocks_db/MANIFEST-000005: Too many open files" }
```

RocksDB has its own config for this here:
```rust
    // 2. Limit the max number of open files (which decreases memory requirements)
    db_opts.set_max_open_files(820);
```

But the OS also seems to have a default limit of 1024 (that could be increased, but for each machine).

We should why we would hit the 1024 limit with rocksdb only being allowed 820 files.

Leo-Besancon:
  * [ ] make tests compile
  * [ ] make tests pass 

### DIFF
--- a/massa-event-cache/src/event_cache.rs
+++ b/massa-event-cache/src/event_cache.rs
@@ -318,6 +318,7 @@ impl EventCache {
             let mut opts = Options::default();
             opts.create_if_missing(true);
             opts.set_merge_operator_associative("counter merge operator", counter_merge);
+            opts.set_max_open_files(128);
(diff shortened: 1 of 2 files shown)


bilboquet (approved):
that should do it

PR #4972, opened 2026-04-24
fix execution stream

damip:
When emitting a "final" execution stream event, info was taken from the current context, which relates to the latest speculative state, instead of the actual finalized data


Leo-Besancon (approved):
Nice find! The fix looks good to me, as long as it was tested in practice (no need to UT it I think).

PR #4978, opened 2026-07-21
fix(serialization): reject zero-denominator ratio on deserialization

damip:
## Summary

`RatioDeserializer::deserialize` unconditionally called `Ratio::new(numer, denom)`, which **panics when `denom == 0`**. The `u64` denominator deserializer used by `ComponentStateDeserializer::new` and `AdvanceLWDeserializer::new` is configured with `U64VarIntDeserializer::new(Included(0), Included(u64::MAX))`, so `0` is accepted as a denominator.

As a result, a malformed versioning value encoding a ratio with denominator `0` could **panic during `MipStore::is_key_value_valid`** — a path that is supposed to return `false` for invalid data. This is reachable during final-state checks and bootstrap-fed DB validation, turning a validation failure into a process abort (DoS).

Addresses AI-report **Finding 145** (severity: medium).

## Fix

Validate the denominator inside the deserializer using `map_opt`: a zero denominator now produces a normal parse error instead of reaching `Ratio::new`.

```rust
|(numer, denom)| {
    if num::Zero::is_zero(&denom) {
        None // -> parse error
    } else {
        Some(Ratio::new(numer, denom))
    }
}
```

## Why this is non-breaking (no consensus / wire / deployment impact)

- `Ratio::new` can **never** construct a zero-denominator value (it panics), so no valid in-memory ratio — and therefore no valid *serialized* ratio — has a zero denominator.
- The only inputs whose behavior changes are malformed byte sequences that previously **panicked**; they now return a clean deserialization error.
- Valid ratios round-trip identically (same `Ratio::new` normalization for non-zero denominators).
- Pure local robustness fix in `massa-serialization`; no change to serialization output, consensus rules, or network messages.

## Testing

- Added unit test `test_ratio_deserializer_rejects_zero_denominator` (asserts a `[1, 0]` buffer yields `Err`, not a panic).
- `cargo test -p massa_serialization`: 23 passed, 0 failed (+ 2 doc-tests).
- `cargo clippy -p massa_serialization --all-targets`: clean.
- `cargo fmt`: clean.

## Checklist
* [x] document all added functions (n/a — no new public API; behavior documented inline)
  * [x] make tests compile
  * [x] make tests pass
* [x] add logs allowing easy debugging (n/a — returns a descriptive parse error)
* [ ] if the API has changed, update the API specification (n/a)

### DIFF
--- a/massa-serialization/src/lib.rs
+++ b/massa-serialization/src/lib.rs
@@ -465,7 +465,7 @@ where
         &self,
         buffer: &'a [u8],
     ) -> IResult<&'a [u8], Ratio<T>, E> {
-        let mut parser = context(
+        context(
             "Ratio<_> deserializer failed",
             map_opt(
                 tuple((
@@ -487,8 +487,8 @@ where
                     }
                 },
             ),
-        );
-        parser.parse(buffer)
+        )
+        .parse(buffer)
     }
 }
 
--- a/massa-serialization/src/lib.rs
+++ b/massa-serialization/src/lib.rs
@@ -7,7 +7,7 @@ use displaydoc::Display;
 use nom::{
     branch::alt,
     bytes::complete::tag,
-    combinator::value,
+    combinator::{map_opt, value},
     error::{ContextError, ParseError},
     sequence::preceded,
     sequence::tuple,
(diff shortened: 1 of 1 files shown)


massa-serialization/src/lib.rs
```diff
@@ -465,19 +465,30 @@ where
         &self,
         buffer: &'a [u8],
     ) -> IResult<&'a [u8], Ratio<T>, E> {
-        context(
+        let mut parser = context(
```
Leo-Besancon: We don't need the let binding for the parser, we can revert this change for consistency with other parsers.
LGTM otherwise.

Leo-Besancon: Solved in f312158471ad80c23117224501128bf9

Leo-Besancon (approved):

PR #5078, opened 2026-07-21
fix(models): don't reflect rejected Amount string in deserialization error

damip:
## Summary
`AmountVisitor::visit_str` mapped parse failures to `E::invalid_value(Unexpected::Str(value), &self)`, embedding the full, attacker-controlled input string into the serde/JSON-RPC error message. On public JSON-RPC methods that deserialize `Amount` (e.g. `execute_read_only_call`), this makes error-path memory allocation scale with the invalid field length (bounded by `max_request_body_size`, but still avoidable amplification).

Addresses AI-report **Finding 85** (severity: minor).

## Fix
Use `Unexpected::Other("string")` instead of `Unexpected::Str(value)`, so the error no longer reflects the rejected input.

## Why this is non-breaking
- Only the **text of the error message** for *invalid* input changes; valid `Amount` values deserialize identically.
- No change to serialization, wire format, consensus, or the success path.

## Testing
- `test_valid_amount_still_deserializes` — valid input still parses.
- `test_invalid_amount_error_does_not_reflect_input` — a 4096-char bogus string is absent from the error message.
- `cargo test -p massa_models --features test-exports amount`: 2/2 pass.
- `cargo clippy -p massa_models --all-targets --features test-exports`: clean. `cargo fmt`: clean.

## Checklist
* [x] document all added functions (n/a)
  * [x] make tests compile
  * [x] make tests pass
* [x] if the API has changed, update the API specification (n/a — only error text)

### DIFF
--- a/massa-models/src/amount.rs
+++ b/massa-models/src/amount.rs
@@ -6,7 +6,6 @@ use massa_serialization::{U64VarIntDeserializer, U64VarIntSerializer};
 use nom::error::{context, ContextError, ParseError};
 use nom::{IResult, Parser};
 use rust_decimal::prelude::*;
-use serde::de::Unexpected;
 use std::fmt;
 use std::ops::Bound;
 use std::str::FromStr;
@@ -414,10 +413,14 @@ impl<'de> serde::de::Visitor<'de> for AmountVisitor {
     where
         E: serde::de::Error,
     {
-        // Do not reflect the (attacker-controlled, possibly large) input back into the
-        // error message: use a generic descriptor instead of `Unexpected::Str(value)` so
-        // error-path allocation does not scale with the rejected field length.
-        Amount::from_str(value).map_err(|_| E::invalid_value(Unexpected::Other("string"), &self))
+        // The parse error is propagated instead of reflecting the (attacker-controlled,
+        // possibly large) input via `Unexpected::Str(value)`: every message reachable from
+        // `Amount::from_str` is a fixed static string (rust_decimal's parse errors are
+        // `&'static str` literals, and the sign/precision/range checks use constant
+        // messages), so error-path allocation does not scale with the rejected field length.
+        // Matches the `map_err(E::custom)` pattern used by the other string-based
+        // deserializers (Address, Hash, PublicKey, Signature).
+        Amount::from_str(value).map_err(E::custom)
     }
 
     fn expecting(&self, formatter: &mut fmt::Formatter) -> fmt::Result {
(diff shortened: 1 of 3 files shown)


massa-models/src/amount.rs
```diff
@@ -414,7 +414,10 @@ impl<'de> serde::de::Visitor<'de> for AmountVisitor {
     where
         E: serde::de::Error,
     {
-        Amount::from_str(value).map_err(|_| E::invalid_value(Unexpected::Str(value), &self))
+        // Do not reflect the (attacker-controlled, possibly large) input back into the
+        // error message: use a generic descriptor instead of `Unexpected::Str(value)` so
+        // error-path allocation does not scale with the rejected field length.
+        Amount::from_str(value).map_err(|_| E::invalid_value(Unexpected::Other("string"), &self))
```
modship:
```suggestion
        Amount::from_str(value).map_err(|_| E::invalid_value(Unexpected::Other("AmountVisitor visit_str error"), &self))
```

modship (changes requested):

massa-models/src/amount.rs
Leo-Besancon: Alternatively, we could directly return the error message, it's coming from the Decimal crate and may be more explicit (https://docs.rs/rust_decimal/1.37.0/rust_decimal/struct.Decimal.html#method.from_str_exact)

Leo-Besancon: In the latest commit (6326459d40922df6f181c26519777d81e279f495), I've chosen to propagate the error with E::custom.

Do you find it better @modship @damip? (otherwise don't hesitate to drop the commit or do something else)

Leo-Besancon (dismissed):
LGTM

damip (commented):
@Leo-Besancon yes, keep `E::custom`, it is the best of the three: bounded and it says what is wrong. I checked that every message reachable from `Amount::from_str` is a constant: rust_decimal 1.37 `parse_str_radix_10_exact` only produces `Error::from("Invalid decimal: ...")` literals (no `format!` of the input anywhere in `str.rs`), the other `Error` variants have fixed `Display` text, and `from_decimal` uses four constant messages. `AmountParseError` is built nowhere else, so the new `{0}` in its `Display` cannot leak anything.

@modship your suggestion is superseded by Leo's commit (same idea, plus the actual reason), can you re-review?

One inline fix on the comment, then merge. No consensus or wire impact: serde `Amount` is JSON only (API, config, genesis files), the binary format uses the nom deserializer.

For the record: the same `Unexpected::Str(value)` line exists in `version.rs`, but `Version` is only ever in responses (`NodeStatus.version`), never deserialized from untrusted input. Not worth touching.

massa-models/src/amount.rs
```diff
@@ -414,7 +413,14 @@ impl<'de> serde::de::Visitor<'de> for AmountVisitor {
...
         E: serde::de::Error,
     {
-        Amount::from_str(value).map_err(|_| E::invalid_value(Unexpected::Str(value), &self))
+        // The parse error is propagated instead of reflecting the (attacker-controlled,
+        // possibly large) input via `Unexpected::Str(value)`: every message reachable from
+        // `Amount::from_str` is a fixed static string (rust_decimal's parse errors are
+        // `&'static str` literals, and the sign/precision/range checks use constant
+        // messages), so error-path allocation does not scale with the rejected field length.
+        // Matches the `map_err(E::custom)` pattern used by the other string-based
+        // deserializers (Address, Hash, PublicKey, Signature).
```
damip, line 422: Not quite: the pattern matches, the property does not. `Address::from_str` starts with `let err = Err(ModelsError::AddressParseError(s.to_string()))` and returns it on every prefix failure, so `Address` reflects the full input, on a lot more API methods than `Amount`. Same non-issue (bounded by `max_request_body_size`, constant factor), same one-line fix. Either make that message static in this PR or drop the last sentence of this comment so it does not claim `Address` is clean.

Leo-Besancon: Thanks for the check, I've updated the address parsing errors to the following in the latest commit:

```rust
        let err = Err(ModelsError::AddressParseError(
            "Invalid address: Address prefix 'AU' or 'AS' not found".to_string(),
        ));
```

I've also checked that bs58::decode do return static strings (except for checksums, but it's only 4 chars).

modship (approved):

PR #5080, opened 2026-07-21
fix(grpc): avoid panic on zero throughput interval

damip:
## Summary
In `transactions_throughput` / `transactions_throughput_server`, a client-supplied `interval` of `0` reached `tokio::time::interval(Duration::from_secs(0))`, which **panics** (`interval period must be greater than zero`) and terminates the streaming task.

Addresses AI-report **Finding 80** (severity: major).

## Fix
Resolve the interval via `resolve_interval_secs(Option<u64>)`, which falls back to `DEFAULT_THROUGHPUT_INTERVAL` when the requested value is absent **or zero**.

## Not an interface/proto change
- `tx_throughput.rs` is the hand-written stream handler; the proto types (`massa_proto_rs`) are **not** touched.
- No request/response schema change. The only behavioral difference is that `interval = 0` — which previously crashed the task — now uses the default interval. No valid tooling depended on crashing the stream.

## Testing
- Unit test `test_resolve_interval_secs` covering `None`, `Some(0)`, and positive values.
- `cargo test -p massa_grpc tx_throughput`: pass. `cargo clippy -p massa_grpc --all-targets`: clean. `cargo fmt`: clean.

## Checklist
  * [x] make tests compile
  * [x] make tests pass
* [x] if the API has changed, update the API specification (n/a — no interface change)

### DIFF
--- a/massa-grpc/src/stream/tx_throughput.rs
+++ b/massa-grpc/src/stream/tx_throughput.rs
@@ -15,6 +15,18 @@ use tracing::error;
 /// end user can override this value by sending a request with a custom interval
 const DEFAULT_THROUGHPUT_INTERVAL: u64 = 10;
 
+/// Resolve the throughput interval (in seconds) from an optional client-supplied
+/// value, falling back to the default when it is absent or zero.
+///
+/// A zero interval must never reach `tokio::time::interval`, which panics when the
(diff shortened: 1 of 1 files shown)


Leo-Besancon (approved):
LGTM

PR #5089, opened 2026-07-21
fix(channel): only unregister channel metrics when the last receiver clone drops

damip:
## Summary
`MassaReceiver<T>` is `Clone`. When the channel disconnects, **every** clone observes the disconnection, but the `recv` / `recv_deadline` / `recv_timeout` / `try_recv` disconnect branches called `unregister_metrics()` unconditionally. The first clone to hit the disconnect therefore removed the `*_channel_actual_size` gauge and `*_channel_total_receive` counter from the prometheus registry while sibling clones were still alive and calling `update_metrics()` on them.

Addresses AI-report **Finding 18** (severity: minor).

## Fix
As shows the discussion below, the finding does not need any fix. We explained why in a doc comment, and that is all that PR updates.

### DIFF
--- a/massa-channel/src/receiver.rs
+++ b/massa-channel/src/receiver.rs
@@ -25,11 +25,8 @@ pub struct MassaReceiver<T> {
 
 impl<T> Drop for MassaReceiver<T> {
     fn drop(&mut self) {
-        let ref_count = Arc::strong_count(&self.ref_counter);
-        if ref_count == 1 {
-            // this is the last ref so we can unregister metrics
-            self.unregister_metrics();
-        }
+        // this only unregisters if this is the last live clone
+        self.unregister_metrics_if_last();
     }
 }
 
@@ -43,6 +40,18 @@ impl<T> MassaReceiver<T> {
         self.actual_len.set(self.receiver.len() as f64);
 
         self.received.inc();
+    }
+
+    /// Unregister metrics only if this is the last live receiver clone.
+    ///
+    /// Channel disconnection is observed by *every* clone, so unregistering
+    /// unconditionally from a `recv*`/`try_recv` disconnect branch would remove
+    /// the metrics from the registry while sibling clones are still alive and
+    /// updating them. Only the last remaining clone should unregister.
+    fn unregister_metrics_if_last(&self) {
+        if Arc::strong_count(&self.ref_counter) == 1 {
+            self.unregister_metrics();
+        }
     }
 
     /// unregister metrics
@@ -73,7 +82,7 @@ impl<T> MassaReceiver<T> {
             }
             Err(crossbeam::channel::TryRecvError::Empty) => Err(TryRecvError::Empty),
             Err(crossbeam::channel::TryRecvError::Disconnected) => {
-                self.unregister_metrics();
+                self.unregister_metrics_if_last();
                 Err(TryRecvError::Disconnected)
             }
         }
@@ -87,7 +96,7 @@ impl<T> MassaReceiver<T> {
             }
             Err(RecvTimeoutError::Timeout) => Err(RecvTimeoutError::Timeout),
             Err(RecvTimeoutError::Disconnected) => {
-                self.unregister_metrics();
+                self.unregister_metrics_if_last();
                 Err(RecvTimeoutError::Disconnected)
             }
         }
@@ -101,7 +110,7 @@ impl<T> MassaReceiver<T> {
             }
             Err(RecvTimeoutError::Timeout) => Err(RecvTimeoutError::Timeout),
             Err(RecvTimeoutError::Disconnected) => {
-                self.unregister_metrics();
+                self.unregister_metrics_if_last();
                 Err(RecvTimeoutError::Disconnected)
             }
         }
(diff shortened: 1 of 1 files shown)


Leo-Besancon:
I understand the fix and it looks ok to me, but I'm not sure I understand the issue.
Given the scenario in the unit test (sender disconnect -> receiver errors out and disconnects), the second alive receiver would either never touch the metrics, or would try to receive something and error out immediately, no? So I don't see any useful info we gain here.

Could you check @damip?

damip:
@Leo-Besancon you are right, there is no problem here. I dug into it.

WHY

1. A crossbeam `Disconnected` is a property of the channel, not of the clone. It means all senders are dropped and the buffer is empty. So when one clone sees it, every other clone is dead too: nothing will ever arrive again, the gauge is frozen at 0, the counter is frozen.
2. Un-exporting the frozen metrics of a dead channel right away (current code) or when the last clone drops (this PR) is not observable by anyone. A sibling clone calling `update_metrics()` on an unregistered gauge just writes to a local object: no panic, no error. The second unregister from `Drop` fails with "not registered" and is swallowed at `trace!`.
3. Disconnects only happen at shutdown and at the `NeedSync` relaunch. The relaunch is the one place where the current behavior is actually the safer one: `launch()` re-creates every channel with the same name, `prometheus::register` fails on a still-registered name, and that failure is only a `debug!` in `MassaChannel::new`, so the new channel's metrics silently never exist. Unregistering as soon as the channel is dead frees the name even if a clone leaks past `stop()`. This PR removes that.
4. Bonus: the new test never runs in CI. `ci.yml` runs `cargo nextest run` at workspace level, `massa-protocol-worker` dev-deps enable `massa_channel/test-exports`, cargo builds one `massa_channel` with the union of features, so `cfg(not(feature = "test-exports"))` compiles the module out. It only runs with `cargo test -p massa_channel` alone.

WHAT TO DO

Do not merge the code change. Keep only a comment so the next scanner does not re-report it. I will reduce this PR to that comment (no code change, no test). Proposed doc comment on `unregister_metrics` in `massa-channel/src/receiver.rs`:

```rust
/// Unregister the channel metrics.
///
/// Called from every `recv*`/`try_recv` disconnect branch on purpose, even
/// though `MassaReceiver` is `Clone`: a crossbeam disconnect means all senders
/// are dropped and the buffer is empty, so the channel is dead for every clone
/// at once and the metrics are frozen. Unregistering right away also frees the
/// metric names for the relaunch (`NeedSync`) that re-creates the channels
/// with the same names; `MassaChannel::new` only logs registration failures
/// at debug level. Later calls (from other clones or from `Drop`) fail with
/// "not registered" and are swallowed at trace level.
```

Leo-Besancon:
> @Leo-Besancon you are right, there is no problem here. I dug into it.
> 
> WHY
> 
>     1. A crossbeam `Disconnected` is a property of the channel, not of the clone. It means all senders are dropped and the buffer is empty. So when one clone sees it, every other clone is dead too: nothing will ever arrive again, the gauge is frozen at 0, the counter is frozen.
> 
>     2. Un-exporting the frozen metrics of a dead channel right away (current code) or when the last clone drops (this PR) is not observable by anyone. A sibling clone calling `update_metrics()` on an unregistered gauge just writes to a local object: no panic, no error. The second unregister from `Drop` fails with "not registered" and is swallowed at `trace!`.
> 
>     3. Disconnects only happen at shutdown and at the `NeedSync` relaunch. The relaunch is the one place where the current behavior is actually the safer one: `launch()` re-creates every channel with the same name, `prometheus::register` fails on a still-registered name, and that failure is only a `debug!` in `MassaChannel::new`, so the new channel's metrics silently never exist. Unregistering as soon as the channel is dead frees the name even if a clone leaks past `stop()`. This PR removes that.
> 
>     4. Bonus: the new test never runs in CI. `ci.yml` runs `cargo nextest run` at workspace level, `massa-protocol-worker` dev-deps enable `massa_channel/test-exports`, cargo builds one `massa_channel` with the union of features, so `cfg(not(feature = "test-exports"))` compiles the module out. It only runs with `cargo test -p massa_channel` alone.
> 
> 
> WHAT TO DO
> 
> Do not merge the code change. Keep only a comment so the next scanner does not re-report it. I will reduce this PR to that comment (no code change, no test). Proposed doc comment on `unregister_metrics` in `massa-channel/src/receiver.rs`:
> 
> ```rust
> /// Unregister the channel metrics.
> ///
> /// Called from every `recv*`/`try_recv` disconnect branch on purpose, even
> /// though `MassaReceiver` is `Clone`: a crossbeam disconnect means all senders
> /// are dropped and the buffer is empty, so the channel is dead for every clone
> /// at once and the metrics are frozen. Unregistering right away also frees the
> /// metric names for the relaunch (`NeedSync`) that re-creates the channels
> /// with the same names; `MassaChannel::new` only logs registration failures
> /// at debug level. Later calls (from other clones or from `Drop`) fail with
> /// "not registered" and are swallowed at trace level.
> ```

Thanks, reverted your commit and added this comment in the latest commit!

bilboquet (approved):

PR #5090, opened 2026-07-21
fix(ledger): validate key version and bounds-check prefix skip in KeyDeserializer

damip:
## Summary
`KeyDeserializer::deserialize` (`massa-ledger-exports/src/key.rs`) had two robustness gaps:
1. It skipped the ledger prefix via a raw slice `&buffer[LEDGER_PREFIX.as_bytes().len()..]`, which **panics** on a buffer shorter than the prefix.
2. It decoded the version byte into `_version` and discarded it, never checking it against `KEY_VERSION`. Since `version_byte_deserializer` accepts any `u64`, non-canonical encodings (version != 0) deserialized into the same logical `Key` — a **version-malleability** flaw.

Addresses AI-report **Finding 81** (severity: info).

## Fix
- Replace the raw slice with a checked `buffer.get(prefix_len..)`, returning a parse error (`Eof`) instead of panicking on undersized input.
- Enforce `version == KEY_VERSION`, returning a parse error (`Verify`) otherwise.

## Why this is non-breaking
- `KeySerializer` **always** writes `KEY_VERSION` (0), so every key produced by any honest node — and every key already on disk — decodes exactly as before.
- The version check only rejects non-canonical/malformed encodings that the serializer never emits.
- Scope is node-local DB-key parsing: **no** wire/consensus format, JSON-RPC, or gRPC surface is changed.

## Testing
- `canonical_key_still_round_trips`: normal serialize/deserialize unchanged.
- `undersized_buffer_errors_instead_of_panicking`: short buffer → error, no panic.
- `non_canonical_version_is_rejected`: flipping the version byte → parse error.
- `cargo test -p massa_ledger_exports`: pass. `cargo clippy --all-targets`: clean. `cargo fmt`: clean.

## Checklist
  * [x] make tests compile
  * [x] make tests pass

### DIFF
--- a/massa-ledger-exports/src/key.rs
+++ b/massa-ledger-exports/src/key.rs
@@ -254,12 +254,83 @@ impl Deserializer<Key> for KeyDeserializer {
         &self,
         buffer: &'a [u8],
     ) -> nom::IResult<&'a [u8], Key, E> {
-        let (rest, _version) = self
-            .version_byte_deserializer
-            .deserialize(&buffer[LEDGER_PREFIX.as_bytes().len()..])?;
+        // Skip the ledger prefix without slicing past the end of the buffer:
+        // an undersized input must yield a clean parse error rather than an
+        // out-of-bounds panic.
+        let prefix_len = LEDGER_PREFIX.as_bytes().len();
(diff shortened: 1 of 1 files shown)


Leo-Besancon (approved):
LGTM

PR #5095, opened 2026-07-21
fix(bootstrap): don't spin the listener on persistent accept errors

damip:
## Summary
`BootstrapTcpListener::poll` drained `self.server.accept()` in a loop that only broke on `WouldBlock`; every other error just logged a warning and `continue`d. A persistent listener-level failure (e.g. `EMFILE` / `ENFILE`) could spin this loop indefinitely, consuming CPU and — because control never returned to `mio::Poll` — preventing the `STOP_LISTENER` token from being serviced.

Addresses AI-report **Finding 7** (severity: minor).

## Fix
Extract the accept drain into a small `drain_accept` helper that ends on the first error, whether `WouldBlock` (no more pending connections) or any other error (sticky failure). Connection post-processing (deregister, convert to std, collect) now happens after the drain, which also keeps the `self.server` / `self.poll` borrows disjoint.

## Scope / safety
Node-local listener liveness only. No consensus, wire-format, JSON-RPC, or gRPC surface is touched. Normal accept behavior (drain until `WouldBlock`) is unchanged.

## Testing
- `drain_stops_on_persistent_error_instead_of_spinning`: a closure that always returns a non-`WouldBlock` error is called exactly once (no infinite loop).
- `drain_collects_ready_connections_until_would_block`: items are collected up to the first `WouldBlock`.
- `cargo test -p massa_bootstrap --lib`: pass (the unrelated `test_binders_multiple_send` is a pre-existing timing-sensitive test that passes in isolation). `cargo clippy --all-targets`: clean. `cargo fmt`: clean.

## Checklist
  * [x] make tests compile
  * [x] make tests pass

### DIFF
--- a/massa-bootstrap/src/listener.rs
+++ b/massa-bootstrap/src/listener.rs
@@ -19,6 +19,26 @@ pub struct BootstrapTcpListener {
 
 pub struct BootstrapListenerStopHandle(pub(crate) Waker);
 
+/// Drain all currently-ready connections from `accept` into `out`.
+///
+/// The drain ends on the first error, whether it is `WouldBlock` (no more
+/// pending connections) or any other error such as a persistent `EMFILE` /
+/// `ENFILE`. Ending the drain on a non-`WouldBlock` error (instead of
+/// `continue`-ing) is what prevents a sticky listener-level failure from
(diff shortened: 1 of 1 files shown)


Leo-Besancon (approved):
LGTM

PR #5096, opened 2026-07-21
fix(bootstrap): keep ACL refresher alive when a list file fails to reload

damip:
## Summary
`run_updater` called `list.update()?` on every tick. A single failed refresh — e.g. malformed JSON in a whitelist/blacklist file (`WhiteBlackListInner::load_list` → `BootstrapError::InitListError`) — propagated out and **terminated the updater thread permanently**. The bootstrap event loop then kept enforcing the last cached ACL snapshot, with no further periodic reloads until process restart.

Addresses AI-report **Finding 16** (severity: minor).

## Fix
Catch the refresh error inside the tick arm, log it, and keep the previously loaded lists — retrying on the next tick. The stop path is unchanged.

## Scope / safety
Node-local ACL-refresh robustness only. No consensus, wire-format, JSON-RPC, or gRPC surface is touched. Enforcement already relied on the cached lists between ticks; this simply keeps refreshes going after a transient/bad file.

## Testing
- New regression test `updater_survives_a_malformed_acl_reload`: runs `run_updater` on a thread, corrupts the blacklist file, and asserts the thread is still alive after several ticks and still stops cleanly on request. (Under the old code the thread would have exited on the first bad reload.)
- `cargo test -p massa_bootstrap --lib`: pass. `cargo clippy --all-targets`: clean. `cargo fmt`: clean.

## Checklist
  * [x] make tests compile
  * [x] make tests pass

### DIFF
--- a/massa-bootstrap/src/server.rs
+++ b/massa-bootstrap/src/server.rs
@@ -230,7 +230,16 @@ impl BootstrapServer<'_> {
                         Err(e) => return Err(BootstrapError::GeneralError(format!("update stopper error : {}", e))),
                     }
                 },
-                recv(ticker) -> _ => {list.update()?;},
+                recv(ticker) -> _ => {
+                    // A failed refresh (e.g. malformed JSON in an ACL file)
+                    // must not kill the updater: log it and keep the previously
+                    // loaded lists, retrying on the next tick. Propagating the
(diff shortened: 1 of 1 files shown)


Leo-Besancon (approved):
LGTM

PR #5098, opened 2026-07-21
fix(protocol): make BlockHandler::stop tolerant of a panicked thread

damip:
## Summary
`BlockHandler::stop` signaled and joined the retrieval thread with `thread.join().unwrap()` **before** handling the propagation thread. `JoinHandle::join()` returns `Err` when the target thread has already panicked, so if `block_retrieval_thread` had panicked, the `unwrap()` unwound `stop()` and the propagation thread was never sent `Stop` nor joined.

Addresses AI-report **Finding 96** (severity: minor).

## Fix
- Send `Stop` to **both** threads first, so a failure while joining one cannot prevent the other from stopping.
- Join both via `join_logging_panic`, which logs an already-panicked thread instead of propagating, so shutdown of the remaining thread always completes.

## Scope / safety
Node-local shutdown robustness only. No consensus, wire-format, JSON-RPC, or gRPC surface is touched.

## Testing
- `join_logging_panic_tolerates_a_panicked_thread`: joining a thread that panicked does not itself panic (and a normal thread joins cleanly).

## Checklist
  * [x] make tests compile
  * [x] make tests pass

### DIFF
--- a/massa-protocol-worker/src/handlers/block_handler/mod.rs
+++ b/massa-protocol-worker/src/handlers/block_handler/mod.rs
@@ -8,6 +8,7 @@ use massa_pos_exports::SelectorController;
 use massa_protocol_exports::ProtocolConfig;
 use massa_storage::Storage;
 use massa_versioning::versioning::MipStore;
+use tracing::warn;
 
 use crate::wrap_network::ActiveConnectionsTrait;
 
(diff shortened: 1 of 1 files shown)


Leo-Besancon (commented):
LGTM. Before merging, it would be great to test if the warn!() is effectively displayed (during normal shut down) or not. I feel like if it's "expected" to trigger, then we shouldn't even log it if we can detect it's due to shutdown.

Leo-Besancon (approved):
Tested and confirmed that the warn is not displayed normally, so I'm approving!

PR #5099, opened 2026-07-21
fix(protocol): don't ban peers for local selector errors on endorsements

damip:
## Summary
In `massa-protocol-worker/src/handlers/endorsement_handler/retrieval.rs`, `process_message` banned the sending peer whenever `note_endorsements_from_peer` returned an error. That function propagates `selector_controller.get_selection(..)?` failures directly, so a **local** selector-state failure (e.g. a PoS draw not available locally) could cause an honest peer sending a valid endorsement to be banned.

Addresses AI-report **Finding 116** (severity: minor).

## Fix
Introduce `is_local_selector_error`, which matches `ProtocolError::PosError(_)` (the variant a `get_selection` failure maps to). Local selector failures are now only logged (no ban); genuine peer-caused validation errors (bad signature, wrong producer selection, etc.) still ban exactly as before.

## Scope / safety
Node-local ban-decision correctness only. No consensus, wire-format, JSON-RPC, or gRPC surface is touched.

## Testing
- `selector_errors_are_not_bannable_but_validation_errors_are`: `ProtocolError::PosError(CycleUnavailable)` is treated as non-bannable, while `WrongSignature` / `GeneralProtocolError` remain bannable.

## Checklist
  * [x] make tests compile
  * [x] make tests pass

### DIFF
--- a/massa-protocol-worker/src/handlers/endorsement_handler/retrieval.rs
+++ b/massa-protocol-worker/src/handlers/endorsement_handler/retrieval.rs
@@ -132,14 +132,27 @@ impl RetrievalThread {
                     &self.internal_sender,
                     self.pool_controller.as_mut(),
                 ) {
-                    warn!(
-                        "peer {} sent us critically incorrect endorsements, \
-                        which may be an attack attempt by the remote node or a \
-                        loss of sync between us and the remote node. Err = {}",
-                        peer_id, err
-                    );
-                    if let Err(err) = self.ban_peer(&peer_id) {
-                        warn!("Error while banning peer {} err: {:?}", peer_id, err);
+                    if is_local_selector_error(&err) {
+                        // A selector failure (e.g. a locally missing draw for
+                        // the endorsement's slot) is a *local* state issue, not
+                        // peer misbehavior. Banning an honest peer for our own
+                        // inability to compute the draw is incorrect, so only
+                        // log it here.
+                        warn!(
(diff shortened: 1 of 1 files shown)


damip (commented):
Warning: was this was also used to eliminate endorsements for which slots were too early or late, which SHOULD result in a ban?

damip:
Closing in favor of #5112.

The reviewer's concern is correct: `note_endorsements_from_peer` reaches `selector_controller.get_selection(endorsement.content.slot)?` with an attacker-controlled slot, and `get_selection` returns the same `PosError::CycleUnavailable(cycle)` for two very different situations:

- an **in-window** slot whose cycle draws are just not available locally yet (honest peer, local lag) — should **not** ban; and
- an **out-of-range / structurally invalid** slot (far future/past, beyond thread bounds) — a peer offense that **should** ban.

So this PR's approach (never ban on `PosError`) fixes the over-ban (F116) but introduces an under-ban / ban-evasion gap: an attacker can send endorsements with bogus out-of-range slots without penalty, and it also short-circuits the producer-selection ban checks for in-window slots.

There is no simple/clean fix here: distinguishing the two cases requires an explicit endorsement slot-acceptance window (too-early/too-late ⇒ ban) evaluated independently of local selector availability. That is a protocol-validation design decision that overlaps with the already-filed F113 (endorsement freshness) and F117 (validation order / replay). Tracking the full problem — original over-ban + this new under-ban dimension + suggested direction — in #5112.

PR #5100, opened 2026-07-21
fix(protocol): re-check operation size before propagation

damip:
## Summary
The operation propagation loop accepted every `PropagateOperations(Storage)` command and marked all contained operations checked, stored them, and announced them **without** re-checking the protocol rule `serialized_size() <= max_serialized_operations_size_per_block`. Locally submitted operations (JSON-RPC / gRPC) are not validated against this limit on submission, so an oversized-but-otherwise-valid operation could be advertised to peers, causing honest peers to ban/disconnect from this node.

Addresses AI-report **Finding 125** (severity: medium).

## Fix
Filter operations through the new `operations_within_size_limit` helper before marking them checked / storing for propagation / announcing. This mirrors the size check already enforced on the peer-received (retrieval) path (`note_operations_from_peer`). Oversized operations are dropped (and logged), not propagated.

## Scope / safety
No wire/consensus format change and no JSON-RPC/gRPC interface change — only which **locally supplied** operations get gossiped. Operations that honest peers would reject as oversized are simply not advertised.

## Testing
- `oversized_operations_are_filtered_out_of_propagation`: with a generous limit an operation is kept; with a limit below its serialized size it is filtered out (so never marked checked/stored/announced).

## Checklist
  * [x] make tests compile
  * [x] make tests pass


massa-protocol-worker/src/handlers/operation_handler/propagation.rs
```diff
@@ -51,16 +85,28 @@ impl PropagationThread {
                 Ok(internal_message) => {
                     match internal_message {
                         OperationHandlerPropagationCommand::PropagateOperations(operations) => {
-                            // Note operations as checked.
+                            // Re-validate the protocol size limit before propagating.
```
Leo-Besancon, line 114: Is preventing propagation enough? Would it be possible to directly check this limit against submitted ops (JSON-RPC / gRPC) and return an error there?

Here I think we basically silently drop the operation, no?

Leo-Besancon: Added the checks in 0afc9a6ae84766afd9a1fcf9ec98c9ea73ab4e72

bilboquet (approved):

PR #5103, opened 2026-07-21
fix(protocol): record last_success on successful connection for peer prioritization

damip:
## Summary

`ConnectionMetadata.last_success` feeds the peer-priority comparator
(`ConnectionMetadata::cmp`), but it was **never written anywhere** in the codebase.

The successful-handshake path in `peer_handler::mod` and the dial path in
`connectivity.rs` both route through `set_try_connect_success_or_insert`, whose name
suggests it records success but which only updates `last_try_connect` (the attempt
timestamp) via `ConnectionMetadata::try_connect()`. `ConnectionMetadata::success()`
existed but had no callers, so `last_success` stayed `None` forever and the
`last_success`-based prioritization branch was effectively dead.

## Change

- Add a dedicated `set_success_or_insert` method to `PeerDBTrait` / `PeerDB` that
  records an actual successful connection via `ConnectionMetadata::success()`.
- Call it on the successful-handshake path (`Ok((peer_id, Some(announcement)))`), in
  addition to the existing attempt bookkeeping.

This keeps the attempt vs. success semantics distinct: `last_try_connect` still tracks
every dial attempt, while `last_success` now reflects real successful connections and
can drive peer prioritization as intended.

## Testing

- New unit test `set_success_records_last_success_for_priority_comparator` on the real
  `PeerDB`: verifies that a mere connect attempt leaves `last_success` unset, and that
  `set_success_or_insert` populates it.
- `cargo test -p massa_protocol_worker --features test-exports` passes; `clippy` clean.

## Risk

Low — additive metadata bookkeeping. It only starts populating a field that was
previously always empty; it does not change connection acceptance or banning logic.

Tracking: findings F103 and F137.

### DIFF
--- a/massa-protocol-worker/src/handlers/peer_handler/models.rs
+++ b/massa-protocol-worker/src/handlers/peer_handler/models.rs
@@ -315,7 +315,7 @@ impl PeerDBTrait for PeerDB {
         self.try_connect_history.entry(*addr).or_default().success();
     }
 
-    fn set_try_connect_success_or_insert(&mut self, addr: &SocketAddr) {
+    fn set_try_connect_or_insert(&mut self, addr: &SocketAddr) {
         self.try_connect_history
             .entry(*addr)
             .or_default()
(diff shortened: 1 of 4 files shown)


massa-protocol-worker/src/handlers/peer_handler/models.rs
```diff
@@ -308,6 +308,10 @@ impl PeerDBTrait for PeerDB {
             .unwrap_or(ConnectionMetadata::default())
     }
 
+    fn set_success_or_insert(&mut self, addr: &SocketAddr) {
+        self.try_connect_history.entry(*addr).or_default().success();
+    }
+
     fn set_try_connect_success_or_insert(&mut self, addr: &SocketAddr) {
```
Leo-Besancon: I think this should be renamed to `set_try_connect_or_insert`, as it doesn't record the success of the attempt, no?

Leo-Besancon: Updated in be0fcec929a0bb7d3d3ae6cfff5f2d4b4e724d9b

Leo-Besancon (approved):
LGTM

PR #5111, opened 2026-07-21
fix(execution): enforce block-dump retention cap across restarts

damip:
## Summary

In `storage_backend.rs`, both `StorageBackend::write` implementations enforce the
`max_blocks` block-dump retention cap only against the in-memory `slots_saved` queue.
But `FileStorageBackend::new` / `RocksDBStorageBackend::new` recreate `slots_saved` as
**empty** and never rebuild it from the block dumps already present on disk / in
RocksDB. Since the dump folder / DB is reopened in place (not cleaned) on startup,
pre-restart entries survive outside the retention accounting, so after every restart
eviction only ever removes entries written since the most recent start — the configured
cap is bypassed and dumps accumulate unbounded across restarts.

## Change

Rebuild `slots_saved` at construction time (oldest-first) from the persisted state:

- `FileStorageBackend::new` scans the folder for `block_slot_{thread}_{period}.bin`
  files and reconstructs the retention queue (missing/unreadable folder → empty).
- `RocksDBStorageBackend::new` iterates existing keys and reconstructs slots via
  `Slot::from_bytes_key`.

Both sort by `Slot` order (period, thread) so eviction removes the genuinely oldest
dump. Retention now spans restarts.

## Testing

- New unit test `file_backend_retention_survives_restart`: writes two dumps under
  `max_blocks = 2`, drops the backend, recreates it (simulated restart), writes a third
  dump, and asserts the oldest dump was evicted while the two newest remain.
- Existing `test_file_storage_backend` / `test_rocksdb_storage_backend` still pass.
- `cargo clippy` clean for `massa_execution_worker`.

## Risk

Low — the reconstruction is read-only at startup and only affects which dump is evicted
next; on a fresh start the queue is empty as before. Severity: minor.

Tracking: finding F59.

### DIFF
--- a/massa-execution-worker/src/storage_backend.rs
+++ b/massa-execution-worker/src/storage_backend.rs
@@ -2,11 +2,51 @@ use std::{
     collections::VecDeque,
     fs::File,
     io::{Read, Write},
-    path::PathBuf,
+    path::{Path, PathBuf},
 };
 
-use massa_models::slot::Slot;
+use massa_models::slot::{Slot, SLOT_KEY_SIZE};
 use rocksdb::{DBCompressionType, Options};
+
+/// Parse a block-dump file name of the form `block_slot_{thread}_{period}.bin`
+/// back into its [`Slot`]. Returns `None` for names that don't match.
+fn parse_block_dump_file_name(name: &str) -> Option<Slot> {
+    let rest = name.strip_prefix("block_slot_")?.strip_suffix(".bin")?;
+    let (thread_str, period_str) = rest.split_once('_')?;
+    let thread: u8 = thread_str.parse().ok()?;
+    let period: u64 = period_str.parse().ok()?;
+    Some(Slot::new(period, thread))
+}
+
+/// Rebuild the retention queue (oldest first) from block dumps already present
+/// on disk, so that the `max_blocks` cap keeps accounting for pre-restart entries.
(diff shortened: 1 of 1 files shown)


massa-execution-worker/src/storage_backend.rs
```diff
@@ -2,12 +2,52 @@ use std::{
...
+
+/// Rebuild the retention queue (oldest first) from block dumps already present
+/// in the RocksDB store, so that the `max_blocks` cap survives restarts.
+fn load_saved_slots_from_db(db: &rocksdb::DB) -> VecDeque<Slot> {
+    let mut slots: Vec<Slot> = db
+        .iterator(rocksdb::IteratorMode::Start)
+        .filter_map(|res| res.ok())
+        .filter_map(|(key, _)| {
+            let key: [u8; SLOT_KEY_SIZE] = key.as_ref().try_into().ok()?;
+            Some(Slot::from_bytes_key(&key))
```
Leo-Besancon: Note: IIRC from_bytes_key does not check the thread is in bound, as it's "external" data (even if it should be created from the node), we may want to add a check.

LGTM besides that.

Leo-Besancon: Added checks in both `parse_block_dump_file_name` and `load_saved_slots_from_db` in 06dd5b3357.

Leo-Besancon (approved):

PR #5123, opened 2026-07-22
[f72]  gRPC get_blocks Enforces Incorrect Request Limit

### LINKED ISSUE #5086: [F72] gRPC `get_blocks` Enforces Incorrect Request Limit, Bypassing `max_block_ids_per_request`

---

## Source
AI security-scan **Finding 72** (severity: **minor**). Tracked in `SUBMISSIONS.md` as F72.

## Summary
The public GRPC `get_blocks` handler is meant to process block ID queries and is advertised as having its own limit, `max_block_ids_per_request`. However, it checks the request size against `max_operation_ids_per_request` and also truncates using that same operation-related limit. If `max_block_ids_per_request` is lower than `max_operation_ids_per_request`, callers can send more block IDs than the block query configuration is supposed to allow, causing extra storage lookups and status resolution.

## Recommendation
Update the `get_blocks` handler to enforce the block-specific limit consistently by using `max_block_ids_per_request` for both request validation and any truncation logic. Review the surrounding configuration and documentation to ensure the advertised limit matches the value actually enforced at runtime. Add targeted tests to confirm `get_blocks` rejects or truncates requests according to `max_block_ids_per_request`, especially when it differs from `max_operation_ids_per_request`.

## Why P0 / breaking
The fix changes **API / gRPC behavior** (gRPC `get_blocks` request limits). Per maintainer guidance, API/gRPC interface or observable-behavior changes are treated as **breaking** (they can break external tooling), so this is filed as an issue for design/versioning discussion rather than a direct PR. Note: the gRPC proto types are auto-generated and must not be hand-edited. See `SUBMISSIONS.md` Part D.

modship:
  * [ ] make tests compile
  * [ ] make tests pass 

### DIFF
--- a/massa-grpc/src/public.rs
+++ b/massa-grpc/src/public.rs
@@ -196,7 +196,6 @@ pub(crate) fn get_blocks(
 
     let mut block_ids: Vec<BlockId> = ids
         .into_iter()
-        .take(grpc.grpc_config.max_block_ids_per_request as usize + 1)
         .map(|id| {
             BlockId::from_str(id.as_str())
                 .map_err(|_| GrpcError::InvalidArgument(format!("invalid block id: {}", id)))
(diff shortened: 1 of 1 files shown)


massa-grpc/src/public.rs
```diff
@@ -196,7 +196,7 @@ pub(crate) fn get_blocks(
 
     let mut block_ids: Vec<BlockId> = ids
         .into_iter()
-        .take(grpc.grpc_config.max_operation_ids_per_request as usize + 1)
+        .take(grpc.grpc_config.max_block_ids_per_request as usize + 1)
```
Leo-Besancon: We should proably remove this .take(), the error return is enough.

Leo-Besancon (approved):
LGTM

PR #5124, opened 2026-07-22
[F69] gRPC datastore entry queries silently drop malformed filters

### LINKED ISSUE #5084: [F69] gRPC datastore entry queries silently drop malformed filters

---

## Source
AI security-scan **Finding 69** (severity: **minor**). Tracked in `SUBMISSIONS.md` as F69.

## Summary
Several gRPC handlers accept requests whose top-level structure is valid, then silently discard malformed nested filter or IP entries instead of rejecting the request. In `massa-main/massa-grpc/src/public.rs`, `get_datastore_entries` drops invalid inner datastore filters and may continue with a reduced filter set or an empty vector. In `massa-main/massa-grpc/src/private.rs`, bootstrap whitelist and blacklist mutation handlers similarly skip malformed IPs and still return success even if persistence through `SharedWhiteBlackList` fails. The original claim that these code paths broaden datastore queries or create immediate external data exposure is not supported.

## Recommendation
Reject malformed nested datastore filters and IP entries with `InvalidArgument` instead of silently skipping them. Validate every requested element before applying any state change or executing the query, and fail the entire request if any nested filter is missing, any address or IP cannot be parsed, or the resulting validated set is empty where at least one valid element is required.

For datastore queries, enforce validation after decoding the nested filters and before calling the execution layer so the request cannot proceed with a reduced subset or an empty validated filter vector. Return an explicit error that identifies the request as malformed rather than producing a partial or empty success response.

For bootstrap whitelist and blacklist updates, use all-or-nothing handling for batched IP changes. Parse and validate the full batch first, apply updates only if all entries are valid, and propagate persistence failures back to the caller instead of logging them and still returning success. Ensure the RPC response reflects whether the requested change was fully applied and durably stored.

## Why P0 / breaking
The fix changes **API / gRPC behavior** (gRPC datastore-entry query behavior). Per maintainer guidance, API/gRPC interface or observable-behavior changes are treated as **breaking** (they can break external tooling), so this is filed as an issue for design/versioning discussion rather than a direct PR. Note: the gRPC proto types are auto-generated and must not be hand-edited. See `SUBMISSIONS.md` Part D.

modship:
  * [ ] make tests compile
  * [ ] make tests pass 

[gRPC datastore entry queries silently drop malformed filters
](https://github.com/massalabs/massa/issues/5084)

### DIFF
--- a/massa-grpc/src/private.rs
+++ b/massa-grpc/src/private.rs
@@ -16,7 +16,6 @@ use massa_proto_rs::massa::model::v1 as grpc_model;
 use massa_protocol_exports::{PeerConnectionType, PeerId};
 use massa_signature::KeyPair;
 use massa_time::MassaTime;
-use tracing::warn;
 // use massa_proto_rs::massa::model::v1 "add_to_bootstrap_blacklist"as grpc_model;
 
 /// Add IP addresses to node bootstrap blacklist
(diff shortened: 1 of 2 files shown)


bilboquet (approved):

Leo-Besancon (approved):
LGTM! Have you checked if other filters where handled in the same way for other endpoints?

PR #5128, opened 2026-07-23
[F71] Block slot-range filter caps

### LINKED ISSUE #5085: [F71] Block slot-range filter caps are enforced one range too late

---

## Source
AI security-scan **Finding 71** (severity: **info**). Tracked in `SUBMISSIONS.md` as F71.

## Summary
`search_blocks` in `massa-main/massa-grpc/src/public.rs`, along with `FilterNewBlocks::build_from_request` and `FilterNewFilledBlocks::build_from_request`, enforces `max_slot_ranges_per_request` with an off-by-one check. Because the code compares `slot_ranges.len() > max_slot_ranges_per_request` before inserting the newly parsed `SlotRange`, one extra distinct slot range can be accepted beyond the configured cap. The original claim that this necessarily enlarges scan work or meaningfully expands the scanned block window is not supported.

## Recommendation
Enforce `max_slot_ranges_per_request` after accounting for the newly parsed distinct `SlotRange`, or change the condition so a request is rejected as soon as it would exceed the configured cap. Apply the same correction consistently in `search_blocks`, `FilterNewBlocks::build_from_request`, and `FilterNewFilledBlocks::build_from_request` to ensure uniform limit enforcement across all request-building paths. Also add boundary-focused tests covering exactly the maximum allowed number of distinct slot ranges and one above the limit to prevent regressions.

## Why P0 / breaking
The fix changes **API / gRPC behavior** (gRPC block slot-range filtering). Per maintainer guidance, API/gRPC interface or observable-behavior changes are treated as **breaking** (they can break external tooling), so this is filed as an issue for design/versioning discussion rather than a direct PR. Note: the gRPC proto types are auto-generated and must not be hand-edited. See `SUBMISSIONS.md` Part D.

modship:
  * [ ] make tests compile
  * [ ] make tests pass 

### DIFF
--- a/massa-grpc/src/public.rs
+++ b/massa-grpc/src/public.rs
@@ -848,7 +848,7 @@ pub(crate) fn get_selector_draws(
                 }
                 grpc_api::selector_draws_filter::Filter::SlotRange(s_range) => {
                     let slot_ranges = slot_ranges_filter.get_or_insert_with(HashSet::new);
-                    if slot_ranges.len() as u32 > grpc.grpc_config.max_slot_ranges_per_request {
+                    if slot_ranges.len() as u32 >= grpc.grpc_config.max_slot_ranges_per_request {
                         return Err(GrpcError::InvalidArgument(format!(
                             "too many slot ranges received. Only a maximum of {} slot ranges are accepted per request",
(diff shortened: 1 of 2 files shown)


Leo-Besancon (approved):
LGTM

PR #5130, opened 2026-07-23
[F93] (cache) reorder save_module to check lru_cache first

### LINKED ISSUE #5116: [F93] Inverted cache-check order in save_module causes redundant work (resource amplification)

---

**Severity:** medium
**Component:** `massa-module-cache/src/controller.rs` (`save_module`)

## Summary

`save_module` checks `hd_cache` before `lru_cache`, even when the same module may already be
present in both. This causes unnecessary RocksDB reads and `RuntimeModule::deserialize` work,
followed by redundant reinsertion into `lru_cache`, instead of reusing the existing in-memory
entry — a resource-amplification problem.

The specific low-cost exploit narrative and the "deserialization = proven full recompilation"
characterization from the original report are overstated.

Reordering the cache-check logic is small, but `save_module` is on a hot path shared by
execution and module loading; changing lookup/insertion order must preserve correctness of the
LRU/HD cache coherence and eviction. Needs review to ensure no cache-consistency regression.

## Suggested direction

Check the in-memory `lru_cache` first and short-circuit if the module is already present,
before touching `hd_cache`/RocksDB; add a test asserting no redundant HD read/deserialize when
the module is already in the LRU cache.

Tracking: finding F93.

peterjah:
  * [ ] make tests compile
  * [ ] make tests pass 

### DIFF
--- a/massa-module-cache/src/controller.rs
+++ b/massa-module-cache/src/controller.rs
@@ -72,15 +72,15 @@ impl ModuleCache {
     /// Save a new or an already existing module in the cache
     pub fn save_module(&mut self, bytecode: &[u8], condom_limits: CondomLimits) {
         let hash = Hash::compute_from(bytecode);
-        if let Some(hd_module_info) =
+        if let Some(lru_module_info) = self.lru_cache.get(hash) {
+            debug!("save_module: {} present in lru", hash);
+            self.hd_cache.insert(hash, lru_module_info);
+        } else if let Some(hd_module_info) =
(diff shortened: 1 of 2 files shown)


Leo-Besancon (approved):

Leo-Besancon (approved):
LGTM, IMO the added test read counter on the hd cache could be removed to avoid bloating, but it's not too many lines so we can also keep it.

bilboquet (approved):

PR #5131, opened 2026-07-23
[F108] add locally produced blocks to checked_headers

### LINKED ISSUE #5118: [F108] Missing checked-header cache update on propagated blocks enables reflected header CPU amplification

---

**Severity:** medium
**Component:** `massa-protocol-worker` block handler (`perform_propagations` / `note_header_from_peer`, `cache.checked_headers`)

## Summary

`perform_propagations` sends integrated block headers to all peers but does **not** store those
headers in `cache.checked_headers`. `note_header_from_peer` decides whether a header is new by
checking only that cache. Because of the mismatch, a peer can send back the same
already-announced header and make the node process it again as if new, repeating validation and
consensus-registration work (reflected header CPU amplification).

The fix (record propagated/integrated headers in `checked_headers`) is localized, but
`checked_headers` is a bounded cache with its own eviction and is consulted across the header
validation path. Needs review to ensure inserting propagated headers doesn't evict entries that
are relied upon elsewhere or change dedup/propagation semantics.

## Suggested direction

When propagating an integrated block header, insert it into `cache.checked_headers` (as already
done for headers received/validated from peers), so re-sent headers are recognized as known.
Add a test that a re-sent propagated header is not re-validated.

Tracking: finding F108.

peterjah:
  * [ ] make tests compile
  * [ ] make tests pass 

### DIFF
--- a/massa-protocol-worker/src/handlers/block_handler/propagation.rs
+++ b/massa-protocol-worker/src/handlers/block_handler/propagation.rs
@@ -93,6 +93,16 @@ impl PropagationThread {
                                 }
                             };
 
+                            // Mark the integrated header as already checked so that a peer
+                            // bouncing it back to us (after we announce it) is recognized as
+                            // known and not re-validated / re-registered with consensus.
+                            // Locally-produced blocks never go through `note_header_from_peer`,
+                            // so without this they would be absent from `checked_headers`.
+                            self.cache
+                                .write()
+                                .checked_headers
+                                .insert(block_id, header.clone());
+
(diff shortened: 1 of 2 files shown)


Leo-Besancon (approved):
Aside from the fact that we insert twice non-local blocks into the cache (both during note_header_from_peer and in the added path here), the fix looks good.
Alternatives would probably require more changes to the codebase to cover all behaviors  so that would not be worth it.

So the current PR LGTM!

bilboquet (approved):

PR #5132, opened 2026-07-23
[F104] Deduplicates requested operation IDs before cloning/assembling responses

### LINKED ISSUE #5117: [F104] Duplicate OperationId in DataRequest::Operations enables response-amplification DoS

---

**Severity:** medium
**Component:** `massa-protocol-worker` block handler (`BlockMessage::DataRequest` → `Operations` branch)

## Summary

`BlockMessage::DataRequest`'s `Operations` branch accepts a `Vec<OperationId>` and only checks
that its length is below `max_operations_per_block`; it does **not** require the requested
`OperationId`s to be unique. The handler filters the list against the block's operation set but
**keeps duplicates**, then returns one full `SecureShareOperation` per remaining entry. An
unauthenticated peer can send a small request with repeated valid IDs and make an honest node
build and send a much larger response (response-amplification DoS).

The fix (deduplicate requested IDs before responding) is conceptually simple, but it lives on
the block-info request/response path and interacts with peer-knowledge accounting and the
existing size checks. Needs review to confirm dedup doesn't break legitimate repeated-request
semantics or block-sync behaviour, and to decide whether duplicates should also be penalised.

## Suggested direction

Deduplicate the requested `OperationId` set (and cap it) before assembling the response;
optionally treat excessive duplicates as a peer offense. Add a test that a request with
repeated IDs yields at most one response entry per distinct operation.

Tracking: finding F104.

peterjah:
  * [ ] make tests compile
  * [ ] make tests pass 

### DIFF
--- a/massa-protocol-worker/src/handlers/block_handler/retrieval.rs
+++ b/massa-protocol-worker/src/handlers/block_handler/retrieval.rs
@@ -304,11 +304,14 @@ impl RetrievalThread {
             (Some((_, block_op_ids)), AskForBlockInfo::Operations(mut asked_ops)) => {
                 // the peer asked for a list of full operations from the block
 
-                // retain only ops that belong to the block
+                // Retain only unique operations that belong to the block. Request vectors may
+                // contain duplicate IDs, but returning a full operation for every occurrence
+                // would allow a peer to amplify a small request into a much larger response.
                 {
                     let block_op_ids_set: PreHashSet<OperationId> =
                         block_op_ids.iter().copied().collect();
-                    asked_ops.retain(|id| block_op_ids_set.contains(id));
+                    let mut seen = PreHashSet::default();
+                    asked_ops.retain(|id| block_op_ids_set.contains(id) && seen.insert(*id));
                 }
 
                 // Send the operations that are available in storage
--- a/massa-protocol-worker/src/handlers/block_handler/retrieval.rs
+++ b/massa-protocol-worker/src/handlers/block_handler/retrieval.rs
@@ -319,11 +319,14 @@ impl RetrievalThread {
             (Some((_, block_op_ids)), AskForBlockInfo::Operations(mut asked_ops)) => {
                 // the peer asked for a list of full operations from the block
 
-                // retain only ops that belong to the block
+                // Retain only unique operations that belong to the block. Request vectors may
+                // contain duplicate IDs, but returning a full operation for every occurrence
+                // would allow a peer to amplify a small request into a much larger response.
                 {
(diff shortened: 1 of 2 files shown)


massa-protocol-worker/src/handlers/block_handler/retrieval.rs
```diff
@@ -304,11 +304,14 @@ impl RetrievalThread {
...
 
-                // retain only ops that belong to the block
+                // Retain only unique operations that belong to the block. Request vectors may
+                // contain duplicate IDs, but returning a full operation for every occurrence
+                // would allow a peer to amplify a small request into a much larger response.
                 {
                     let block_op_ids_set: PreHashSet<OperationId> =
                         block_op_ids.iter().copied().collect();
-                    asked_ops.retain(|id| block_op_ids_set.contains(id));
+                    let mut seen = PreHashSet::default();
```
bilboquet, line 328: That looks good.
But I think we could catch this issue earlier, in `massa-protocol-worker/src/handlers/block_handler/messages.rs:287` in 
```rust
impl Deserializer<BlockMessage> for BlockMessageDeserializer {
    fn deserialize<'a, E: ParseError<&'a [u8]> + ContextError<&'a [u8]>>()
```

applying the same principle, Operations deserializer would become something like:
```rust
BlockInfoType::Operations => self
    .operation_ids_deserializer
    .deserialize(rest)
    .map(|(rest, operation_ids)| {
        let deduplicated: Vec<OperationId> = operation_ids
            .into_iter()
            .collect::<PreHashSet<_>>()
            .into_iter()
            .collect();
        (rest, AskForBlockInfo::Operations(deduplicated))
    }),
```

what do you think?

Leo-Besancon: Second opinion: I think we could go with @bilboquet's approach if we don't intend to punish the peer for duplicated ops.
(in the issue recomandations: `optionally treat excessive duplicates as a peer offense`). If we want to punish our peer, I think the current location of this dedup makes more sence.

I could go either way personally.

bilboquet: Reading my comment a second time, I realize I was not clear.
It was not a counter proposal: I was suggesting to do it in both places.

EDIT: never mind, that's dumb. My "proposal for an early filter" would prevent any action against the guilty peer. 
So it's probably better to keep the PR version and decide what to do with peer.
(thanks leo for the remark)

peterjah:
@Leo-Besancon I think we can merge this

Leo-Besancon (approved):
LGTM

PR #5133, opened 2026-07-23
[F39] filter out readonly execute from cache-import block

### LINKED ISSUE #5113: [F39] Read-only executions can poison the shared module cache with arbitrary bytecode

---

**Severity:** info
**Component:** `massa-execution-worker` (`execute_readonly_request` → `settle_slot`) + `massa-module-cache`

## Summary

`execute_readonly_request` always finalizes through `settle_slot(None)`, and `settle_slot`
unconditionally imports speculative bytecode updates into the shared module cache. Read-only
execution can still reach bytecode-writing paths (`create_new_sc_address`, `set_bytecode`), so
uncommitted simulated bytecode can be compiled and persisted into the global RAM + RocksDB
module cache from a read-only request.

Validated as a shared-cache side effect in read-only mode; the broader "fully unmetered
post-execution DoS" claim is not supported.

The fix touches the read-only settle path and the module-cache import policy. Preventing
read-only executions from mutating the shared cache must not regress the intended
cache-warming behaviour for real execution, and needs care around speculative-vs-final
bytecode handling. Needs design review.

## Suggested direction

Do not import speculative bytecode into the persistent/shared module cache from read-only
settle (`settle_slot(None)` on the read-only path), or scope such compiled modules to a
throwaway/ephemeral cache for the duration of the read-only call.

Tracking: finding F39.

peterjah:
  * [ ] make tests compile
  * [ ] make tests pass 

### DIFF
--- a/massa-execution-worker/src/context.rs
+++ b/massa-execution-worker/src/context.rs
@@ -1054,8 +1054,12 @@ impl ExecutionContext {
         }
 
         // update module cache
-        let bc_updates = self.speculative_ledger.added_changes.get_bytecode_updates();
-        {
+        // Skip this for read-only executions: their speculative bytecode is never
+        // finalized, so importing it would only pollute the shared RAM + on-disk module
+        // cache (potentially evicting legitimately warmed modules) with no cache-warming
+        // benefit. Real executions (read_only == false) still warm the cache as intended.
+        if !self.read_only {
+            let bc_updates = self.speculative_ledger.added_changes.get_bytecode_updates();
             let mut cache_write_lock = self.module_cache.write();
             for bytecode in bc_updates {
                 cache_write_lock.save_module(&bytecode.0, self.config.condom_limits.clone());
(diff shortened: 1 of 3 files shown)


bilboquet (approved):

Leo-Besancon:
I think this fix is still needed and the PR looks good, but I'm wondering if there are not cases where it is favorable that readonly executions fill up the cache?
If everybody does the same readonly request that relies on deploying the same bytecode, wouldn't we want the module to be cached?

What do you think @bilboquet @peterjah?

bilboquet:
> I think this fix is still needed and the PR looks good, but I'm wondering if there are not cases where it is favorable that readonly executions fill up the cache? If everybody does the same readonly request that relies on deploying the same bytecode, wouldn't we want the module to be cached?
> 
The case you're mentioning is legit.
The only way I see to take care of it without reenabling the poisoning problem is by having a separate cache for each kind of execution.
And maybe add some fee to readonly.
Could be a follow-up.

peterjah:
agree with you. but imho we should priorize security over performance here ! and eventually plan a followup if we need perf improvement for that special case

bilboquet:
> agree with you. but imho we should priorize security over performance here ! and eventually plan a followup if we need perf improvement for that special case

agreed, security before perf, that's what I was trying to say^^

Leo-Besancon (approved):
Ok then this seems good!

PR #5134, opened 2026-07-23
[F94] enforce max_module_length when loading temp module

### LINKED ISSUE #5088: [F94] Missing Bytecode Length Validation in `load_tmp_module`

---

## Source
AI security-scan **Finding 94** (severity: **medium**). Tracked in `SUBMISSIONS.md` as F94.

Corresponds to external bug-bounty PDF #2 (missing bytecode size check).

**Affected file:** `massa-module-cache/src/controller.rs`

## Summary
`load_tmp_module` in `massa-main/massa-module-cache/src/controller.rs` compiles temporary modules without enforcing `self.cfg.max_module_length`. The public read-only path can forward arbitrary `bytecode` into this function, so oversized modules can reach `RuntimeModule::new` despite the configured size cap. This creates a real local denial-of-service risk on exposed nodes. The original claim about the `save_module` and `compile_cached` path bypassing the same protection is not supported because on-chain bytecode updates are size-checked before reaching that code.

## Recommendation
Enforce the same bytecode size validation in `load_tmp_module` as in `load_module_info` before invoking `RuntimeModule::new`. Reject empty inputs and any `bytecode` whose length exceeds `self.cfg.max_module_length` so temporary, read-only compilation cannot bypass the configured module-size cap. Apply this validation as early as possible in the read-only execution flow, ideally at both the API boundary and within `load_tmp_module`, to ensure oversized payloads are denied before any compilation work or significant resource allocation occurs.

## Why P0 / breaking
The fix changes **API / gRPC behavior** (read-only bytecode execution acceptance (`load_tmp_module`)). Per maintainer guidance, API/gRPC interface or observable-behavior changes are treated as **breaking** (they can break external tooling), so this is filed as an issue for design/versioning discussion rather than a direct PR. Note: the gRPC proto types are auto-generated and must not be hand-edited. See `SUBMISSIONS.md` Part D.

peterjah:
  * [ ] make tests compile
  * [ ] make tests pass 

### DIFF
--- a/massa-module-cache/src/controller.rs
+++ b/massa-module-cache/src/controller.rs
@@ -196,6 +196,20 @@ impl ModuleCache {
         condom_limits: CondomLimits,
     ) -> Result<RuntimeModule, CacheError> {
         debug!("load_tmp_module");
+        if bytecode.is_empty() {
+            let error_msg = "load_tmp_module: bytecode is absent".to_string();
+            debug!(error_msg);
+            return Err(CacheError::LoadError(error_msg));
+        }
+        if bytecode.len() > self.cfg.max_module_length as usize {
+            let error_msg = format!(
+                "load_tmp_module: bytecode length {} exceeds max module length {}",
+                bytecode.len(),
+                self.cfg.max_module_length
+            );
+            debug!(error_msg);
+            return Err(CacheError::LoadError(error_msg));
+        }
         // Do not actually debit the instance creation cost from the provided gas
         // This is only supposed to be a check
         limit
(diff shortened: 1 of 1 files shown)


bilboquet (dismissed):

massa-module-cache/src/controller.rs
```diff
@@ -196,6 +196,20 @@ impl ModuleCache {
         condom_limits: CondomLimits,
     ) -> Result<RuntimeModule, CacheError> {
         debug!("load_tmp_module");
+        if bytecode.is_empty() {
```
Leo-Besancon, line 199: Note that this is redundent as the runtime checks:
```rust
if bytecode.len() <= 2 {
    return Err(anyhow!("Too small bytecode"));
}
```

Leo-Besancon (dismissed):
Left a comment, but I saw that load_module does the same check, you can keep the empty check for consistency if you prefer.

Otherwise LGTM

Leo-Besancon (approved):

PR #5135, opened 2026-07-23
[F61] Add ExecutionController::get_slot_abi_call_stack_and_transfers

### LINKED ISSUE #5061: [F61] Speculative Rollback Does Not Revert Trace and Transfer Indexes, Exposing Stale Execution Data

---

## Source
AI security-scan **Finding 61** (severity: **medium**). Tracked in `SUBMISSIONS.md` as F61.

**Affected file:** `massa-execution-worker/src/trace_history.rs`

## Summary
`ExecutionState::execute_candidate_slot` in `massa-execution-worker/src/execution.rs` rolls back speculative execution by truncating `active_history`, but it does not roll back the related data stored in `trace_history`. During execution, traces and transfers are saved through `save_traces_for_slot` and `save_transfers_for_slot`, and `save_traces_for_slot` also updates the reverse index `op_per_slot`. Because these trace and transfer indexes are not cleaned when a slot is rolled back, overwritten, or evicted, queries can return stale data from an abandoned branch. Over time, stale `op_per_slot` entries can also crowd out valid entries and break operation-based lookups.

## Root cause
The main rollback bug is in `massa-execution-worker/src/execution.rs` inside `ExecutionState::execute_candidate_slot`. When execution restarts from a candidate `slot` or earlier, the code only calls `self.active_history.write().truncate_from(slot, self.config.thread_count);` and rewinds `self.active_cursor`. It does *not* truncate or reconcile `trace_history`, even though traces and transfers for those rolled-back slots were already stored there.

The trace and transfer data are written earlier in `massa-execution-worker/src/execution.rs` during slot execution through:

- `self.trace_history.write().save_traces_for_slot(*slot, slot_trace.clone());`
- `self.trace_history.write().save_transfers_for_slot(*slot, transfers.clone());`

Because `trace_history` is not updated during rollback, old entries in `trace_per_slot`, `transfer_per_slot`, and `op_per_slot` can remain after the corresponding candidate branch has been abandoned.

There is also an index maintenance bug in `massa-execution-worker/src/trace_history.rs` inside `save_traces_for_slot(&mut self, slot: Slot, traces: SlotAbiCallStack)`. The function adds new reverse-index entries with:

- `for (op_id, _) in traces.operation_call_stacks.iter() { self.op_per_slot.insert(*op_id, slot); }`
- `self.trace_per_slot.insert(slot, traces);`

This insertion path only adds fresh `op_id -> slot` mappings. It does not remove mappings tied to the previous contents of the same `slot` when that slot is re-executed, and it does not remove mappings for slots later removed from `trace_per_slot` or `transfer_per_slot` by rollback or normal eviction.

As a result, `op_per_slot` can point to data from an overwritten slot, a rolled-back slot, or a slot-level cache entry that is no longer aligned with the reverse index. The input also stat …

## Affected code
```rust
pub(crate) fn save_traces_for_slot(&mut self, slot: Slot, traces: SlotAbiCallStack) {
    for (op_id, _) in traces.operation_call_stacks.iter() {
        self.op_per_slot.insert(*op_id, slot);
    }
    self.trace_per_slot.insert(slot, traces);
}
```

## Recommendation
Ensure rollback and cache lifecycle handling keep `trace_history` fully consistent with `active_history`. When `ExecutionState::execute_candidate_slot` truncates speculative state, it should also remove all trace and transfer entries for rolled-back slots and clean any associated `op_per_slot` mappings. Apply the same cleanup when a slot is re-executed or when slot-level entries are evicted, so reverse indexes never outlive the underlying `trace_per_slot` or `transfer_per_slot` data.

Treat `save_traces_for_slot` and `save_transfers_for_slot` as replacement operations for an existing slot: before inserting new data, remove any prior slot contents and their reverse-index references. More generally, enforce an invariant that every `op_per_slot` entry must point to a live slot entry and that no abandoned branch data remains queryable after rollback.

Add targeted tests covering rollback, slot overwrite, cache eviction, and operation-based lookups to confirm reverted or evicted data is no longer returned and valid recent entries remain accessible.

## Why P0 / breaking

peterjah:
  * [ ] make tests compile
  * [ ] make tests pass 

### DIFF
--- a/massa-api/src/public.rs
+++ b/massa-api/src/public.rs
@@ -139,10 +139,13 @@ impl MassaRpcServer for API<Public> {
                 continue;
             };
             let mut transfers = Vec::new();
-            let abi_calls = self
+            // Read the ABI call stack and the direct transfers for this slot from a single
+            // consistent snapshot, so the combined response cannot mix data from two
+            // different executions of the same slot (a re-execution between two reads).
+            let (abi_calls, slot_transfers) = self
                 .0
                 .execution_controller
-                .get_slot_abi_call_stack(slot.clone().into());
+                .get_slot_abi_call_stack_and_transfers(slot);
             if let Some(abi_calls) = abi_calls {
                 // flatten & filter transfer trace in asc_call_stacks
 
@@ -221,10 +224,7 @@ impl MassaRpcServer for API<Public> {
                     }
                 }
             }
-            let transfers_op: Vec<Transfer> = self
-                .0
-                .execution_controller
-                .get_transfers_for_slot(slot)
+            let transfers_op: Vec<Transfer> = slot_transfers
                 .unwrap_or_default()
                 .iter()
                 .map(|t| Transfer {
--- a/massa-execution-exports/src/controller_traits.rs
+++ b/massa-execution-exports/src/controller_traits.rs
@@ -134,6 +134,19 @@ pub trait ExecutionController: Send + Sync {
     fn get_transfers_for_slot(&self, slot: Slot) -> Option<Vec<Transfer>>;
 
     #[cfg(feature = "execution-trace")]
+    /// Get both the ABI call stack and the direct transfers for a given slot, read from a
+    /// single consistent snapshot of the execution trace history.
+    ///
+    /// Callers that need both datasets for a slot must use this instead of combining
+    /// `get_slot_abi_call_stack` and `get_transfers_for_slot`: those are two separate reads,
+    /// and a slot re-execution occurring between them can yield a response mixing data from
+    /// two different executions of the same slot.
+    fn get_slot_abi_call_stack_and_transfers(
+        &self,
+        slot: Slot,
+    ) -> (Option<SlotAbiCallStack>, Option<Vec<Transfer>>);
+
+    #[cfg(feature = "execution-trace")]
     /// Get the transfer of MAS for a given operation id
     fn get_transfer_for_op(&self, op_id: &OperationId) -> Option<Transfer>;
 
--- a/massa-execution-worker/src/controller.rs
+++ b/massa-execution-worker/src/controller.rs
@@ -581,6 +581,22 @@ impl ExecutionController for ExecutionControllerImpl {
     }
 
     #[cfg(feature = "execution-trace")]
+    fn get_slot_abi_call_stack_and_transfers(
+        &self,
+        slot: Slot,
+    ) -> (Option<SlotAbiCallStack>, Option<Vec<Transfer>>) {
+        // Hold a single `execution_state` (and thus `trace_history`) read lock for both
+        // lookups. A slot re-execution takes the `execution_state` write lock, so it cannot
+        // run between the two reads; the returned ABI call stack and transfers are therefore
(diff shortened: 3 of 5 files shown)


bilboquet (dismissed):

Leo-Besancon (commented):
Not doing a full review, but for me this PR mostly answers finding 60 (https://github.com/massalabs/massa/pull/5135), not 61 related to rollback mechanism, no?

peterjah:
i would say they are both related... and the fix is to handle atomically slots and slots transfers.

Leo-Besancon:
@peterjah I think I understand how handling both slots and slot transfers atomically prevents having a read from inconsistant data, but I'm not sure how it prevents stale data after a rollback.

In the linked issue, one recommandation:
> When ExecutionState::execute_candidate_slot truncates speculative state, it should also remove all trace and transfer entries for rolled-back slots and clean any associated op_per_slot mappings

Here, when there is a rollback, we could still read data from a stale execution, no?

peterjah:
@Leo-Besancon you were right, and thanks for insisting. Atomic writes only prevent a torn read; they do nothing about data left behind by an abandoned branch.

The concrete window: execute_candidate_slot truncated active_history from the rollback slot but re-executed only that slot, and the worker takes the execution-state write lock once per slot. So every slot after the rollback point kept the traces and transfers of the abandoned branch until a later loop iteration re-executed it, and the sequencer can interleave final slots, new input and read-only requests in between. A query landing in that window got the abandoned branch's data — atomically, which made it indistinguishable from valid data. Operation lookups were affected too: op_per_slot still pointed at those slots, and since their stale content still contained the operation, the consistency check in fetch_traces_for_op passed.

That part is fixed in #5136 (merged):

TraceHistory::truncate_from(slot) drops traces, transfers and the matching op_per_slot entries for the rollback slot and every later one. It is called next to active_history.truncate_from in execute_candidate_slot, and where execute_final_slot clears the speculative queue on cache mismatch.
save_for_slot is now a replacement operation: a re-executed slot removes its previous contents first, so the reverse index can no longer hold operations that are not part of the slot anymore.
save_for_slot also evicts the oldest slot explicitly instead of letting the two LRUs drift apart, so an op_per_slot entry cannot outlive the slot entry it points to.
Unit tests cover rollback, slot replacement and eviction.
What remains in this PR is the read side, which #5136 does not address: get_slots_transfers (JSON-RPC) and get_slot_transfers (gRPC) query the trace history twice, acquiring the execution-state lock separately for the call stack and for the transfers. Atomic writes don't help there — a re-execution landing between the two reads still returns traces from one execution and transfers from the next. get_slot_abi_call_stack_and_transfers returns both under a single lock and both handlers now use it. I rebased on main and added a test pinning the guard in remove_slot (replacing a slot must not unindex an operation that another slot has since recorded).

So F61's rollback and index-lifecycle recommendations are covered by #5136, and this PR closes the remaining two-reads window. Happy to drop the [F61] tag from the title if you'd rather it not read as the fix for the finding.

Leo-Besancon (approved):
LGTM, thanks for the explanation!

PR #5138, opened 2026-07-23
[f92] fix balance underflow error msg

### LINKED ISSUE #5027: [F92] Incorrect Error Handling Logic Emits Overflow Error on Balance Underflow

---

## Source
AI security-scan **Finding 92** (severity: **info**). Tracked in `SUBMISSIONS.md` as F92.

**Affected file:** `massa-models/src/ledger.rs`

## Summary
`LedgerData::apply_change` in `massa-main/massa-models/src/ledger.rs` mislabels a subtraction failure as an overflow. In the `checked_sub` branch, a balance underflow condition returns `ModelsError::InvalidLedgerChange` with the message `"balance overflow in LedgerData::apply_change"`, which is diagnostically incorrect. The confirmed issue is therefore an *error-message bug* in the failure path, not a failure of arithmetic safety. The broader claim that this bypasses distinct underflow handling or enables attacker-driven operational abuse is not supported.

## Root cause
The root cause is a copy-paste mistake in `LedgerData::apply_change`: both the `checked_add` and `checked_sub` failure paths construct the same error message, `"balance overflow in LedgerData::apply_change"`, even though the `checked_sub` path corresponds to a subtraction underflow or insufficient-balance condition.

In the subtraction branch, the arithmetic itself remains correctly protected because `checked_sub` returns `None` on failure and the function exits with `ModelsError::InvalidLedgerChange`. The ruled-out portion of the original theory is that any separate underflow-specific handling is skipped; both branches already return the same `ModelsError::InvalidLedgerChange` variant, so no distinct handling path is being bypassed by this mislabeled string.

## Affected code
```rust
pub fn apply_change(&mut self, change: &LedgerChange) -> Result<()> {
        if change.balance_increment {
            self.balance = self
                .balance
                .checked_add(change.balance_delta)
                .ok_or_else(|| {
                    ModelsError::InvalidLedgerChange(
                        "balance overflow in LedgerData::apply_change".into(),
                    )
                })?;
        } else {
            self.balance = self
                .balance
                .checked_sub(change.balance_delta)
                .ok_or_else(|| {
                    ModelsError::InvalidLedgerChange(
                        "balance overflow in LedgerData::apply_change".into(),
                    )
                })?;
        }
        Ok(())
    }
```

## Recommendation
Update the error message returned in the `checked_sub` failure path of `LedgerData::apply_change` so it accurately describes a subtraction underflow or insufficient-balance condition rather than an overflow. Review similar arithmetic error paths in the module to ensure each message matches the actual failure condition and remains consistent with the surrounding error taxonomy. If this message is surfaced in logs, tests, or operator tooling, update those expectations accordingly to avoid diagnostic confusion.

## Why P0 / breaking

peterjah:
  * [ ] make tests compile
  * [ ] make tests pass 

### DIFF
--- a/massa-models/src/ledger.rs
+++ b/massa-models/src/ledger.rs
@@ -126,7 +126,7 @@ impl LedgerData {
     }
 
     /// apply a `LedgerChange` for an entry
-    /// Can fail in overflow or underflow occur
+    /// Can fail if an overflow or underflow occurs
     pub fn apply_change(&mut self, change: &LedgerChange) -> Result<()> {
         if change.balance_increment {
             self.balance = self
@@ -143,7 +143,7 @@ impl LedgerData {
                 .checked_sub(change.balance_delta)
                 .ok_or_else(|| {
                     ModelsError::InvalidLedgerChange(
-                        "balance overflow in LedgerData::apply_change".into(),
+                        "balance underflow in LedgerData::apply_change".into(),
                     )
                 })?;
         }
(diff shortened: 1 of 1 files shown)


bilboquet (approved):

Leo-Besancon (approved):
LGTM

modship (approved):

PR #5139, opened 2026-07-23
[F75] gRPC operation stream subscribes before handshake

### LINKED ISSUE #5064: [F75] gRPC operation stream subscribes before handshake and gains hidden backlog

---

## Source
AI security-scan **Finding 75** (severity: **info**). Tracked in `SUBMISSIONS.md` as F75.

**Affected file:** `massa-grpc/src/stream/new_operations.rs`

## Summary
The gRPC stream handlers in `massa-main/massa-grpc/src/stream/new_operations.rs` and `new_slot_abi_call_stacks` begin observing broadcast traffic before the client has provided its initial subscription selectors. In `new_operations`, events broadcast after `operation_sender.subscribe()` but before the first `NewOperationsRequest` can remain buffered and later be emitted if they match the first filter. In `new_slot_abi_call_stacks`, the handler starts receiving immediately with `FinalityLevel::Unspecified`, so traces can be sent before the client sets its finality preference. The unsupported portion of the original claim is the characterization of this behavior as a handshake, authorization, or access-control bypass.

## Root cause
In `new_operations`, `grpc.pool_broadcasts.operation_sender.subscribe()` is called before the first request is read from `in_stream`, while `FilterNewOperations::build_from_request(request.filters, &config)` is deferred until after that first message arrives. This creates a gap in which broadcast operations can accumulate in the subscriber buffer and later be processed under the first client-provided filter.

In `new_slot_abi_call_stacks`, the stream enters its receive loop immediately after subscribing and initializes `finality` to `FinalityLevel::Unspecified`. Because messages may be consumed before any `NewSlotAbiCallStacksRequest` is received, pre-request traces can be forwarded without the client's intended finality selection having been applied.

The ruled-out portion of the original theory is that these selectors act as security gates. The validator determined that the gRPC stream is already established and the affected streams are public, so `filters` and `finality` are subscription-shaping inputs, not authorization checks.

## Affected code
```rust
pub(crate) async fn new_operations(
    grpc: &MassaPublicGrpc,
    request: Request<Streaming<grpc_api::NewOperationsRequest>>,
) -> Result<NewOperationsStreamType, GrpcError> {
    let (tx, rx) = tokio::sync::mpsc::channel(grpc.grpc_config.max_channel_size);
    let mut in_stream = request.into_inner();
    let mut subscriber = grpc.pool_broadcasts.operation_sender.subscribe();
    let config = grpc.grpc_config.clone();

    tokio::spawn(async move {
        if let Some(Ok(request)) = in_stream.next().await {
            let mut filters =
                match FilterNewOperations::build_from_request(request.filters, &config) { ... };

            loop {
                select! {
                    event = subscriber.recv() => {
                        match event {
                            Ok(massa_operation) => {
                                if let Some(data) = filters.filter_output(massa_operation, &config) {
                                    if let Err(e) = tx.send(Ok(grpc_api::NewOperationsResponse {signed_operation: Some(data.into())})).await {
                                        error!("failed to send operation : {}", e);
                                        break;
                                    }
                                }
                            },
                            Err(e) => error!("{}", e)
                        }
                    },
                    res = in_stream.next() => { ... }
                }
            }
        }
    });
}

async fn broadcast_via_ws<T: Serialize + Send + Clone + 'static>(
    sender: tokio::sync::broadcast::Sender<T>,
    pending: PendingSubscriptionSink,
) -> SubscriptionResult {
    let sink = pending.accept().await?;
    let closed = sink.closed();
    let stream = BroadcastStream::new(sender.subscribe());
    ...
}
```

## Recommendation
Delay subscription and event consumption until the server has received and validated the first client request, so the stream begins only after the initial selectors have been applied. For `new_operations`, create the broadcast subscriber only after `FilterNewOperations::build_from_request(...)` succeeds. For `new_slot_abi_call_stacks`, avoid receiving or forwarding messages while `finality` is still `FinalityLevel::Unspecified`, and require the first request to establish the intended finality before processing any trace events.

To preserve consistent behavior across streaming endpoints, standardize initialization semantics so all streams either wait for the first request before subscribing or explicitly document and enforce any alternative model. If early subscription is unavoidable, discard any events observed before the initial request rather than buffering and replaying them under later client-provided parameters.

## Why P0 / breaking

modship:
  * [ ] make tests compile
  * [ ] make tests pass 

### DIFF
--- a/massa-grpc/src/stream/new_operations.rs
+++ b/massa-grpc/src/stream/new_operations.rs
@@ -39,8 +39,8 @@ pub(crate) async fn new_operations(
     let (tx, rx) = tokio::sync::mpsc::channel(grpc.grpc_config.max_channel_size);
     // Get the inner stream from the request
     let mut in_stream = request.into_inner();
-    // Subscribe to the new operations channel
-    let mut subscriber = grpc.pool_broadcasts.operation_sender.subscribe();
+    // Clone the new operations channel sender to subscribe from the spawned task
+    let operation_sender = grpc.pool_broadcasts.operation_sender.clone();
     // Clone grpc to be able to use it in the spawned task
     // let grpc = grpc.clone();
 
@@ -61,6 +61,11 @@ pub(crate) async fn new_operations(
                         return;
                     }
                 };
+
+            // Subscribe to the new operations channel only once the initial
+            // filters are known, so that operations broadcast before the
+            // handshake are not buffered and replayed
+            let mut subscriber = operation_sender.subscribe();
 
             loop {
                 select! {
--- a/massa-grpc/src/stream/new_slot_abi_call_stacks.rs
+++ b/massa-grpc/src/stream/new_slot_abi_call_stacks.rs
@@ -42,21 +42,32 @@ pub(crate) async fn new_slot_abi_call_stacks(
     // Extract the incoming stream of abi call stacks messages
     let mut in_stream = request.into_inner();
 
-    // Subscribe to the new slot execution events channel
+    // Clone the slot execution traces channel sender to subscribe from the spawned task
     #[cfg(feature = "execution-trace")]
-    let mut subscriber = grpc
-        .execution_channels
-        .slot_execution_traces_sender
-        .subscribe();
-    #[cfg(not(feature = "execution-trace"))]
-    let (mut subscriber, _receiver) = {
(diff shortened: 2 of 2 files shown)


bilboquet (approved):

Leo-Besancon (approved):
LGTM!

PR #5142, opened 2026-07-24
[F4] JSON-RPC reports send failure after local pool admission

### LINKED ISSUE #5048: [F4] JSON-RPC reports send failure after local pool admission already happened

---

## Source
AI security-scan **Finding 4** (severity: **minor**). Tracked in `SUBMISSIONS.md` as F4.

**Affected file:** `massa-api/src/public.rs`

## Summary
`API<Public>::send_operations` in `massa-main/massa-api/src/public.rs` submits operations to the local pool path through `cmd_sender.add_operations(to_send.clone())` before it waits for `protocol_sender.propagate_operations(to_send)` to succeed. This creates *non-atomic submission semantics*: the JSON-RPC call can return an error even though the operation was already accepted into the local submission pipeline and may later appear on local subscriptions. The original finding is correct on that core behavior, but it overstated the determinism and security framing. In particular, local pool admission is asynchronous after `try_send`, and the WebSocket visibility path is an intended local broadcast mechanism rather than a standalone confidentiality break.

## Root cause
The confirmed issue is the ordering in `send_operations`: after `check_input_operation(...)` succeeds, the function stores the operations, derives `ids`, and calls `cmd_sender.add_operations(to_send.clone())` *before* awaiting `tokio::task::spawn_blocking(move || protocol_sender.propagate_operations(to_send))`.

`add_operations` does not propagate queueing failures back to the RPC layer. Its `operations_input_sender.try_send(Command::AddItems(ops))` path only logs `TrySendError::Disconnected(_)` and `TrySendError::Full(_)`, while `send_operations` can still later fail on the protocol propagation step and return `ApiError::InternalServerError(...)`.

The ruled-out portion of the original theory is the claim that local admission is already complete by the time the RPC error is returned. That does not hold as stated because pool admission after `try_send` is *asynchronous*, so the code guarantees only that the operation was handed to the local submission path before propagation completed, not that full pool insertion had synchronously finished at the exact error boundary.

## Affected code
```rust
async fn send_operations(&self, ops: Vec<OperationInput>) -> RpcResult<Vec<OperationId>> {
    ...
    let verified_ops = ops
        .into_iter()
        .map(|op_input| check_input_operation(op_input, api_cfg, last_slot))
        ...
        .collect::<RpcResult<Vec<SecureShareOperation>>>()?;

    to_send.store_operations(verified_ops.clone());
    let ids: Vec<OperationId> = verified_ops.iter().map(|op| op.id).collect();
    cmd_sender.add_operations(to_send.clone());

    tokio::task::spawn_blocking(move || protocol_sender.propagate_operations(to_send))
        .await
        .map_err(|err| ApiError::InternalServerError(err.to_string()))?
        .map_err(|err| {
            ApiError::InternalServerError(format!("Failed to propagate operations: {}", err))
        })?;
    Ok(ids)
}

fn add_operations(&mut self, ops: Storage) {
    match self.operations_input_sender.try_send(Command::AddItems(ops)) {
        Err(TrySendError::Disconnected(_)) => warn!("Could not add operations to pool: worker is unreachable."),
        Err(TrySendError::Full(_)) => warn!("Could not add operations to pool: worker channel is full."),
        Ok(_) => {}
    }
}
```

## Recommendation
Reorder `API<Public>::send_operations` so that it returns success only after both local submission and protocol propagation have succeeded, or clearly separates these outcomes in the API contract. Avoid enqueuing operations into the local pool before the propagation result is known unless the method explicitly documents and exposes partial-success semantics.

Ensure failures from the local submission path are surfaced to the caller instead of being only logged. If atomic behavior is intended, introduce coordinated error handling so that a propagation failure cannot leave the operation pending in the local processing pipeline while the RPC reports an error. If atomicity is not practical, document that the method may return a propagation-related failure after local queuing has already occurred, so clients can implement correct retry and reconciliation logic.

## Why P0 / breaking

modship:
  * [ ] make tests compile
  * [ ] make tests pass 

### DIFF
--- a/massa-grpc/src/stream/send_operations.rs
+++ b/massa-grpc/src/stream/send_operations.rs
@@ -141,24 +141,10 @@ pub(crate) async fn send_operations(
                                     let mut operation_storage = storage.clone_without_refs();
                                     operation_storage
                                         .store_operations(verified_ops.values().cloned().collect());
-                                    // Add the received operations to the operations pool
-                                    if let Err(e) =
-                                        pool_controller.add_operations(operation_storage.clone())
-                                    {
-                                        // If pool admission failed, send an error message back to the client
-                                        let error =
-                                            format!("failed to add operations to pool: {}", e);
-                                        report_error(
-                                            tx.clone(),
-                                            tonic::Code::Internal,
-                                            error.to_owned(),
-                                        )
-                                        .await;
-                                    };
-
-                                    // Propagate the operations to the network
-                                    if let Err(e) =
-                                        protocol_controller.propagate_operations(operation_storage)
+                                    // Propagate the operations to the network before admitting
+                                    // them into the local pool (same order as the JSON-RPC API)
+                                    if let Err(e) = protocol_controller
+                                        .propagate_operations(operation_storage.clone())
                                     {
                                         // If propagation failed, send an error message back to the client
                                         let error =
@@ -169,6 +155,23 @@ pub(crate) async fn send_operations(
                                             error.to_owned(),
                                         )
                                         .await;
+                                        continue;
+                                    };
+
+                                    // Add the received operations to the operations pool
+                                    if let Err(e) =
+                                        pool_controller.add_operations(operation_storage)
+                                    {
+                                        // If pool admission failed, send an error message back to the client
+                                        let error =
+                                            format!("failed to add operations to pool: {}", e);
+                                        report_error(
+                                            tx.clone(),
+                                            tonic::Code::Internal,
+                                            error.to_owned(),
+                                        )
+                                        .await;
+                                        continue;
                                     };
 
                                     // Build the response message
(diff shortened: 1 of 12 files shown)


massa-grpc/src/stream/send_operations.rs
```diff
@@ -142,7 +142,19 @@ pub(crate) async fn send_operations(
                                     operation_storage
                                         .store_operations(verified_ops.values().cloned().collect());
                                     // Add the received operations to the operations pool
-                                    pool_controller.add_operations(operation_storage.clone());
+                                    if let Err(e) =
+                                        pool_controller.add_operations(operation_storage.clone())
```
bilboquet: In `public.rs` you:
1. try to propagate the op
2. if propagation is ok, try to add locally.
Which seems to be the correct order.

So, shouldn't the order be the same here (only add locally after a successful propagation)?

modship: good catch !

bilboquet (dismissed):
LGTM ^^

bilboquet (approved):

PR #5143, opened 2026-07-24
[F36] Flushed event-cache queries ignore finality-only filters

### LINKED ISSUE #5066: [F77] `new_slot_transfers` uses a different default finality window than `new_slot_execution_outputs`

---

## Source
AI security-scan **Finding 77** (severity: **info**). Tracked in `SUBMISSIONS.md` as F77.

**Affected file:** `massa-grpc/src/stream/new_slot_transfers.rs`

## Summary
`new_slot_transfers` in `massa-main/massa-grpc/src/stream/new_slot_transfers.rs` and `new_slot_execution_outputs` apply different effective defaults before any client-specific filter is established. The confirmed issue is that `new_slot_transfers` begins streaming immediately and suppresses non-final events when `FinalityLevel::Unspecified` is used, while `new_slot_execution_outputs` waits for the first client message and then uses `FilterNewSlotExec::default()`, whose empty `status_filter` allows both candidate and final outputs. As a result, the same slot can appear earlier on the execution-output stream than on the transfer stream, and final transfer data can be emitted before the client sends an initial filter update. The original claim is narrowed to an API-consistency issue; a standalone security exploit based on mandatory slot-by-slot synchronization is not supported.

## Root cause
The root cause is inconsistent default stream behavior across two gRPC subscriptions.

In `new_slot_transfers`, the send loop starts immediately with `FinalityLevel::Unspecified`, and the implementation explicitly drops non-final traces, making the effective behavior `final-only` until a client changes the filter. In `new_slot_execution_outputs`, the stream instead waits for the first client request and then applies `FilterNewSlotExec::default()`, where an empty `status_filter` does not restrict delivery to final events and therefore permits both candidate and final outputs.

The ruled-out portion of the original theory is the implication that this difference violates an enforced invariant or a guaranteed cross-stream synchronization contract. The validated record does not show any explicit requirement that these streams remain aligned slot-by-slot, nor any direct attacker-controlled path beyond the inconsistent API behavior itself.

## Affected code
```rust
let mut subscriber = grpc
    .execution_channels
    .slot_execution_traces_sender
    .subscribe();

...

Ok((massa_slot_execution_trace, is_final)) => {
    ...
    for op_call_stack in massa_slot_execution_trace.operation_call_stacks {
        let op_id = op_call_stack.0;
        let op_call_stack = op_call_stack.1;
        for abi_trace in op_call_stack {
            let only_transfer = abi_trace.flatten_filter(&transfer_abi_names);
            for transfer in only_transfer {
                let (t_from, t_to, t_amount) = transfer.parse_transfer();
                ret_transfers.push(TransferInfo {
                    from: t_from.clone(),
                    to: t_to.clone(),
                    amount: t_amount,
                    operation_id_or_asc_index: Some(
                        grpc_api::transfer_info::OperationIdOrAscIndex::OperationId(op_id.to_string()),
                    ),
                });
            }
        }
    }
    let transfers =
        execution_controller
        .get_transfers_for_slot(massa_slot_execution_trace.slot);
    if let Some(transfers) = transfers {
        for transfer in transfers {
            ret_transfers.push(TransferInfo {
                from: transfer.from.to_string(),
                to: transfer.to.to_string(),
                amount: transfer.amount.to_raw(),
                operation_id_or_asc_index: Some(
                    grpc_api::transfer_info::OperationIdOrAscIndex::OperationId(
                        transfer.op_id.to_string(),
                    ),
                ),
            });
        }
    }
    let ret = grpc_api::NewSlotTransfersResponse {
// … (truncated)
```

## Recommendation
Align the default initialization and finality semantics of `new_slot_transfers` and `new_slot_execution_outputs` so both streams apply a consistent policy before any client-specific filter is received. Choose a single documented behavior for uninitialized or unspecified filters, and ensure it is enforced uniformly across both subscriptions.

If the intended default is final-only delivery, make both streams suppress candidate events until the client explicitly requests them. If the intended default is to allow both candidate and final events, apply that behavior consistently to both endpoints. Also ensure that neither stream begins emitting data under implicit defaults unless that startup behavior is intentional, documented, and consistent across the API.

Update the API documentation to clearly describe the default behavior for each stream, especially how `FinalityLevel::Unspecified` and empty status filters are interpreted before the first client message. Add regression tests to verify that both subscriptions exhibit the same default finality window and startup behavior under equivalent client input.

## Why P0 / breaking

### LINKED ISSUE #5082: [F36] Flushed event-cache queries ignore finality-only filters

---

## Source
AI security-scan **Finding 36** (severity: **minor**). Tracked in `SUBMISSIONS.md` as F36.

**Affected file:** `massa-event-cache/src/event_cache.rs`

## Summary
`massa-event-cache/src/event_cache.rs` handles `EventFilter.is_final` inconsistently between in-memory queued events and the persistent final-events cache. The confirmed issue is that queries using `is_final = Some(true)` can return matching events before the async flush, then return no persisted matches after flush because the RocksDB-backed path drops that filter. The original claim about `is_final = Some(false)` returning finalized events on the user-facing path is not supported.

## Recommendation
Handle `EventFilter.is_final` consistently across both the in-memory queue and the persisted `final_events_cache` path. Ensure that queries with `is_final = Some(true)` remain valid after events are flushed, rather than being dropped because `from_event_filter()` does not translate that field and `EventCache::get_filtered_sc_output_events()` rejects empty filter sets. At a minimum, finalized-event queries should either map `is_final` into the persistent filtering logic or treat `is_final = Some(true)` as an explicit request to search finalized entries even when no other filter fields are provided. Add regression tests covering the full lifecycle of an event before and after flush to confirm identical results for equivalent final-only queries across both storage layers.

## Why P0 / breaking
The fix changes **API / gRPC behavior** (event-query results (get_filtered_sc_output_event)). Per maintainer guidance, API/gRPC interface or observable-behavior changes are treated as **breaking** (they can break external tooling), so this is filed as an issue for design/versioning discussion rather than a direct PR. Note: the gRPC proto types are auto-generated and must not be hand-edited. See `SUBMISSIONS.md` Part D.

modship:
  * [ ] make tests compile
  * [ ] make tests pass 


bilboquet (dismissed):

bilboquet (dismissed):

massa-event-cache/src/event_cache.rs
```diff
@@ -468,18 +468,31 @@ impl EventCache {
...
-            // Note: will return too many event - user should restrict the filter
-            warn!("Filter item only on is final field, please add more filter parameters");
-            return (vec![], vec![]);
+            if filter.is_final == Some(true) {
+                // Explicit request for all finalized events with no other
+                // criteria: scan the whole event column (bounded below by
+                // max_events_per_query) instead of matching nothing
+                filter_items.push((KeyIndent::Event, FilterItem::SlotStart(Slot::new(0, 0))));
+            } else {
+                // Note: will return too many event - user should restrict the filter
```
Leo-Besancon, line 492: This comment does not make sence anymore, as the number of events in both branches are the same (all events).

Leo-Besancon (commented):
A small comment on my part, the rest looks good!

massa-event-cache/src/event_cache.rs
modship: removed else path

Leo-Besancon (approved):
LGTM

bilboquet (approved):

PR #5144, opened 2026-07-28
ci: drop macOS from full test matrix

damip:
## Summary
- Remove `macOS-latest` from the CI `full` job matrix (Linux-only tests on PRs/merges).
- Leave CD binary builds (including macOS sign/notarize) unchanged.

## Why
Org Actions analysis: macOS was ~$640 YTD gross, mostly from running the full Rust test suite on every PR. macOS runners cost ~10× Linux. `p-layer` already made this change in May.

## Test plan
- [ ] CI passes on this PR (Linux `full` only)
- [ ] Confirm `cd.yml` still lists macOS matrix entries for releases

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -134,7 +134,7 @@ jobs:
   #         save-if: ${{ github.ref_name == 'main' }}
   #     - run: cargo xtask check_gas_cost_definitions
 
-  # Full cross-platform tests required to merge on main branch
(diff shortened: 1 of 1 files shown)


bilboquet:
We have made some other improvements on p-layer recently, like dropping "check" which is a subset of "clippy" and can be enforced by a pre-commit hook locally. You may want to take a look?

damip:
ok let's merge this and make the proposed changes as followup

PR #5145, opened 2026-08-03
[F50] Failed Read-Only Execution Can Leak Stale Execution Context

### LINKED ISSUE #5083: [F50] Failed Read-Only Execution Can Leak Stale Execution Context Into Subsequent RPC Queries

---

## Source
AI security-scan **Finding 50** (severity: **medium**). Tracked in `SUBMISSIONS.md` as F50.

**Affected file:** `massa-execution-worker/src/worker.rs`

## Summary
`execute_one_readonly_request()` calls `ExecutionState::execute_readonly_request()` while holding the shared execution-state write lock. During this read-only flow, a mutable `ExecutionContext::readonly` is written into shared state, but it is only cleared on success. If execution fails, the temporary context can remain in shared memory and later RPC queries can read this stale, attacker-controlled data. This can corrupt off-chain query results even though no real on-chain state transition was committed.

## Recommendation
Ensure the read-only execution context is always cleared, regardless of whether `ExecutionState::execute_readonly_request()` succeeds or fails. Add unconditional cleanup on every exit path so that temporary `ExecutionContext::readonly` state cannot persist in shared memory after an aborted or reverted read-only call.

Prefer a scoped or guard-based lifecycle for the temporary context so cleanup is guaranteed even when execution returns early or propagates an error. More generally, avoid exposing speculative read-only mutations through shared execution state; isolate them from subsequent RPC handlers unless execution completes successfully and the state is explicitly finalized.

## Why P0 / breaking
The fix changes **API / gRPC behavior** (read-only execution results returned via RPC). Per maintainer guidance, API/gRPC interface or observable-behavior changes are treated as **breaking** (they can break external tooling), so this is filed as an issue for design/versioning discussion rather than a direct PR. Note: the gRPC proto types are auto-generated and must not be hand-edited. See `SUBMISSIONS.md` Part D.

modship:
  * [ ] make tests compile
  * [ ] make tests pass 

### DIFF
--- a/massa-execution-worker/src/execution.rs
+++ b/massa-execution-worker/src/execution.rs
@@ -2140,9 +2140,38 @@ impl ExecutionState {
         &self,
         req: ReadOnlyExecutionRequest,
     ) -> Result<ReadOnlyExecutionOutput, ExecutionError> {
-        // TODO ensure that speculative things are reset after every execution ends (incl. on error and readonly)
-        // otherwise, on prod stats accumulation etc... from the API we might be counting the remainder of this speculative execution
+        // Run the read-only execution request.
+        let result = self.execute_readonly_request_inner(req);
 
+        // Always reset the shared execution context to a clean read-only context,
+        // regardless of whether the execution succeeded or failed.
+        // Otherwise, a failed read-only execution could leave stale, caller-controlled
+        // speculative changes in the shared context, which would then be exposed
+        // to subsequent RPC queries.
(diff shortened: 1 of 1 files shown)


Leo-Besancon (approved):
I initially had some trouble with the finding's description as it is not clear that the previous "success path cleanup" was the call to `settle_slot` on the context.
The way you handle the cleanup seems ok to me!

PR #5149, opened 2026-08-12
[F95] drain command before shutdown

### LINKED ISSUE #5121: [F95] Silent data loss of pool items during shutdown sequence

---

**Severity:** minor
**Component:** `massa-pool-worker/src/controller_impl.rs` (`PoolManagerImpl::stop()`)

## Summary

`PoolManagerImpl::stop()` shuts pool workers down by sending `Command::Stop` to the worker
channels and then joining the threads. This creates a shutdown race: if any sender still
enqueues a command behind `Command::Stop`, the worker exits as soon as it reads the stop command
and does not process later-queued items, silently losing them.

The original exploit narrative (attacker-driven submission via RPC/network during shutdown) is
not supported: those components are stopped earlier in the sequence. The validated issue is
best-effort shutdown-drain semantics.

The exploit path is unreachable (submission components are stopped before `pool_manager.stop()`),
and the lost items are transient mempool entries that are re-gossiped after restart. A correct
fix would change shutdown-drain semantics across all three pool worker loops, interacting with
buffer-swap and pool locks. Given the "be very careful of unintended consequences" directive,
this needs per-item design review before any change — hence tracked as an issue rather than a PR.

## Suggested direction (if pursued)

Drain remaining queued commands before honoring `Stop` (or use a two-phase stop that processes
already-enqueued items), consistently across the operation/endorsement/denunciation pool loops,
with care around the buffer-swap and pool locks.

Tracking: finding F95.

modship:
  * [ ] make tests compile
  * [ ] make tests pass 

### DIFF
--- a/massa-pool-worker/src/worker.rs
+++ b/massa-pool-worker/src/worker.rs
@@ -16,7 +16,76 @@ use massa_wallet::Wallet;
 use parking_lot::RwLock;
 use std::time::Instant;
 use std::{sync::Arc, thread, thread::JoinHandle};
-use tracing::warn;
+
+/// Process commands already queued behind a `Stop` before shutting down.
+fn drain_commands_after_stop<F>(receiver: &MassaReceiver<Command>, process: &mut F)
+where
+    F: FnMut(Command),
+{
+    while let Ok(cmd) = receiver.try_recv() {
+        if !matches!(cmd, Command::Stop) {
+            process(cmd);
+        }
+    }
+}
+
+fn process_endorsement_pool_command(
+    cmd: Command,
+    buffer: &mut EndorsementPool,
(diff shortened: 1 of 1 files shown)


Leo-Besancon (commented):
The fix looks good, but I'm not sure I understand the finding's issue. The pool persists nothing on disk, every command after stop is just putting data in ram that will get cleaned out just after, no?
Is there a good reason to drain the commands I'm not seeing?

modship:
> The fix looks good, but I'm not sure I understand the finding's issue. The pool persists nothing on disk, every command after stop is just putting data in ram that will get cleaned out just after, no? Is there a good reason to drain the commands I'm not seeing?

You're right, and I'm closing this PR.

PR #5150, opened 2026-08-13
[F138] Update signature verification chunk_size formula

### LINKED ISSUE #5077: [F138] Flawed Chunk Size Calculation Disables Batch Verification and Enables CPU Exhaustion DoS

---

## Source
AI security-scan **Finding 138** (severity: **info**). Tracked in `SUBMISSIONS.md` as F138.

**Affected file:** `massa-protocol-worker/src/sig_verifier.rs`

## Summary
In `verify_sigs_batch` at `massa-main/massa-protocol-worker/src/sig_verifier.rs`, the chunking heuristic can reduce `chunk_size` to `1` when `ops.len()` is greater than `2` but still below `2 * rayon::current_num_threads()`. In that range, `ops.par_chunks(1)` sends single-element slices to `verify_signature_batch`, which falls back to per-signature verification via `public_key.verify_signature` instead of using `ed25519_dalek::verify_batch`. This is a confirmed performance inefficiency in the batching logic. The original claim that this behavior by itself enables a realistic CPU-exhaustion DoS is not supported.

## Root cause
The confirmed root cause is the `chunk_size` computation in `verify_sigs_batch`:

- `let chunk_size = std::cmp::max(1, ops.len() / rayon::current_num_threads());`
- when `ops.len() > 2` but `ops.len() < 2 * rayon::current_num_threads()`, integer division makes `ops.len() / rayon::current_num_threads()` evaluate to `0` or `1`, and the `max(1, ...)` result becomes `1`
- `ops.par_chunks(1)` then invokes `verify_signature_batch` on single-element chunks, preventing those chunks from using `ed25519_dalek::verify_batch`

The ruled-out portion of the original theory is that this logic disables parallel verification entirely. It does not: verification still runs through Rayon in parallel, but with inefficient one-item chunks.

## Affected code
```rust
pub fn verify_sigs_batch(ops: &[(Hash, Signature, PublicKey)]) -> Result<(), ProtocolError> {
    // if it's a small batch, use single-core verification
    if ops.len() <= SMALL_BATCH_LIMIT {
        return verify_signature_batch(ops).map_err(|_err| ProtocolError::WrongSignature);
    }

    // otherwise, use parallel batch verif

    // compute chunk size for parallelization
    let chunk_size = std::cmp::max(1, ops.len() / rayon::current_num_threads());
    // process chunks in parallel
    ops.par_chunks(chunk_size)
        .try_for_each(verify_signature_batch)
        .map_err(|_err| ProtocolError::WrongSignature)
}
```

## Recommendation
Adjust the chunking logic in `verify_sigs_batch` so that batches above `SMALL_BATCH_LIMIT` are split into chunks large enough to preserve the `ed25519_dalek::verify_batch` path whenever practical. Avoid producing single-element chunks for medium-sized inputs unless the entire batch is intentionally handled as individual verifications. The chunk-size calculation should balance parallelism with batching efficiency, ensuring each Rayon task processes more than one signature when possible.

Additionally, define and document the intended batching thresholds and validate them with benchmarks across different batch sizes and thread counts. Include tests that cover the range where `ops.len()` is greater than `2` but below `2 * rayon::current_num_threads()` to ensure the implementation does not regress into one-item chunking and continues to use the expected batch-verification path.

## Why P0 / breaking

Leo-Besancon:
  * [ ] make tests compile
  * [ ] make tests pass 

### DIFF
--- a/massa-protocol-worker/src/sig_verifier.rs
+++ b/massa-protocol-worker/src/sig_verifier.rs
@@ -23,7 +23,10 @@ pub fn verify_sigs_batch(ops: &[(Hash, Signature, PublicKey)]) -> Result<(), Pro
 
     // compute chunk size for parallelization with ceiling division to ensure we batch properly.
     // note that current_num_threads() will not return 0.
-    let chunk_size = ops.len().div_ceil(rayon::current_num_threads());
+    let chunk_size = std::cmp::max(
+        SMALL_BATCH_LIMIT,
+        ops.len().div_ceil(rayon::current_num_threads()),
+    );
     // process chunks in parallel
     ops.par_chunks(chunk_size)
         .try_for_each(verify_signature_batch)
--- a/massa-protocol-worker/src/sig_verifier.rs
+++ b/massa-protocol-worker/src/sig_verifier.rs
@@ -21,8 +21,9 @@ pub fn verify_sigs_batch(ops: &[(Hash, Signature, PublicKey)]) -> Result<(), Pro
 
     // otherwise, use parallel batch verif
 
-    // compute chunk size for parallelization
-    let chunk_size = std::cmp::max(1, ops.len() / rayon::current_num_threads());
+    // compute chunk size for parallelization with ceiling division to ensure we batch properly.
+    // note that current_num_threads() will not return 0.
+    let chunk_size = ops.len().div_ceil(rayon::current_num_threads());
     // process chunks in parallel
     ops.par_chunks(chunk_size)
         .try_for_each(verify_signature_batch)


peterjah (dismissed):

massa-protocol-worker/src/sig_verifier.rs
```diff
@@ -21,8 +21,9 @@ pub fn verify_sigs_batch(ops: &[(Hash, Signature, PublicKey)]) -> Result<(), Pro
 
     // otherwise, use parallel batch verif
 
-    // compute chunk size for parallelization
-    let chunk_size = std::cmp::max(1, ops.len() / rayon::current_num_threads());
+    // compute chunk size for parallelization with ceiling division to ensure we batch properly.
+    // note that current_num_threads() will not return 0.
+    let chunk_size = ops.len().div_ceil(rayon::current_num_threads());
```
bilboquet: as 5.div_ceil(8) -> 1
I'm not sur this fix it fully correct.
I'd propose.
```rs
let chunk_size = cmp::max(SMALL_BATCH_LIMIT, ops.len().div_ceil(rayon::current_num_threads()));
```

bilboquet (changes requested):

massa-protocol-worker/src/sig_verifier.rs
Leo-Besancon: The two formulas differ only when `ops.len()` is between `SMALL_BATCH_LIMIT` and `rayon::current_num_threads()`.
My take on it was that we are ok with rayon's overhead when `ops.len() > SMALL_BATCH_LIMIT`.
And if we're ok with the pool thread overhead, it would be better to use all the threads available in order to keep the batch small (at 1 in fact).

But your suggestion made me think a bit more about how batch verification works, and it seems a bit more coherent, so I'll apply it!

bilboquet (approved):

PR #5157, opened 2026-08-13
[f132] make Announcement::new listeners map match its signed payload

### LINKED ISSUE #5074: [F132] State Divergence in Announcement Listeners via Unmodified IP Map

---

## Source
AI security-scan **Finding 132** (severity: **info**). Tracked in `SUBMISSIONS.md` as F132.

**Affected file:** `massa-protocol-worker/src/handlers/peer_handler/announcement.rs`

## Summary
`Announcement::new` in `massa-main/massa-protocol-worker/src/handlers/peer_handler/announcement.rs` serializes each listener using `routable_ip` when one is provided, but returns `Self { listeners, ... }` with the original `HashMap` keys unchanged. As a result, a freshly constructed `Announcement` can contain `listeners` entries that do not match its own `serialized` payload, `hash`, and `signature`. The original claim that this creates a practically exploitable, critical listener-state divergence in downstream peer-management logic is not supported in this codebase.

## Root cause
The confirmed issue is a correctness mismatch inside `Announcement::new`: during serialization, the function replaces each listener IP with `routable_ip` via `let ip = routable_ip.unwrap_or_else(|| listener.0.ip());`, but when constructing the return value it uses the unmodified `listeners` map in `Ok(Self { listeners, ... })`.

Because the deserializer rebuilds listener keys from the serialized IP and port bytes, a serialize-deserialize round trip produces an `Announcement` whose `listeners` map reflects the routable addresses rather than the original local keys. The ruled-out portion of the original theory is the claim that this mismatch propagates into normal peer-management or connection logic: according to the validated analysis, `Announcement::new` is only used to build an outbound handshake message, and that transient local struct is immediately serialized and discarded.

## Affected code
```rust
pub fn new(
        mut listeners: HashMap<SocketAddr, TransportType>,
        routable_ip: Option<IpAddr>,
        keypair: &KeyPair,
    ) -> PeerNetResult<Self> {
        let mut buf: Vec<u8> = vec![];
        let length_serializer = U64VarIntSerializer::new();
        //TODO: Hacky to fix and adapt to support multiple ip/listeners
        if routable_ip.is_none() {
            listeners = HashMap::default()
        }
        length_serializer
            .serialize(&(listeners.len() as u64), &mut buf)
            .map_err(|err| {
                PeerNetError::HandlerError
                    .error("Announcement serialization", Some(err.to_string()))
            })?;
        for listener in &listeners {
            let ip = routable_ip.unwrap_or_else(|| listener.0.ip());
            let ip_bytes = match ip {
                IpAddr::V4(ip) => {
                    buf.push(4);
                    ip.octets().to_vec()
                }
                IpAddr::V6(ip) => {
                    buf.push(6);
                    ip.octets().to_vec()
                }
            };
            buf.extend_from_slice(&ip_bytes);
            let port_bytes = listener.0.port().to_be_bytes();
            buf.extend_from_slice(&port_bytes);
            buf.push(*listener.1 as u8);
        }
        let timestamp = MassaTime::now().as_millis();
        buf.extend(timestamp.to_be_bytes());
        let hash = Hash::compute_from(&buf);
        Ok(Self {
            listeners,
            timestamp,
            hash,
            signature: keypair.sign(&hash).map_err(|err| {
                PeerNetError::SignError.error("Announcement serialization", Some(err.to_string()))
            })?,
            serialized: buf,
// … (truncated)
```

## Recommendation
Ensure `Announcement::new` constructs a canonical in-memory object that matches the data being signed and serialized. When `routable_ip` is provided, apply the same IP substitution to the `listeners` map stored in the returned `Announcement`, not only to the serialized payload. This removes the mismatch between `listeners`, `serialized`, `hash`, and `signature`, and ensures that a serialize-deserialize round trip preserves equivalent state.

Additionally, document and enforce an invariant that all fields of `Announcement` must represent the same listener set in the same form. Consider validating this invariant during construction or testing to prevent similar internal consistency issues from being reintroduced.

## Why P0 / breaking

peterjah:
  * [ ] make tests compile
  * [ ] make tests pass 

### DIFF
--- a/massa-protocol-worker/src/handlers/peer_handler/announcement.rs
+++ b/massa-protocol-worker/src/handlers/peer_handler/announcement.rs
@@ -171,16 +171,25 @@ impl Deserializer<Announcement> for AnnouncementDeserializer {
 
 impl Announcement {
     pub fn new(
-        mut listeners: HashMap<SocketAddr, TransportType>,
+        listeners: HashMap<SocketAddr, TransportType>,
         routable_ip: Option<IpAddr>,
         keypair: &KeyPair,
     ) -> PeerNetResult<Self> {
         let mut buf: Vec<u8> = vec![];
         let length_serializer = U64VarIntSerializer::new();
         //TODO: Hacky to fix and adapt to support multiple ip/listeners
-        if routable_ip.is_none() {
-            listeners = HashMap::default()
-        }
+        // Build the canonical listener set that gets serialized, hashed and signed:
+        // when a routable IP is provided, advertise every listener under that IP;
+        // otherwise advertise no listeners. Storing this same map in the returned
+        // struct keeps `listeners`, `serialized`, `hash` and `signature` describing
+        // the exact same listener set (so a serialize/deserialize round trip is stable).
+        let listeners: HashMap<SocketAddr, TransportType> = match routable_ip {
+            Some(ip) => listeners
+                .into_iter()
+                .map(|(addr, transport)| (SocketAddr::new(ip, addr.port()), transport))
+                .collect(),
+            None => HashMap::default(),
+        };
         length_serializer
             .serialize(&(listeners.len() as u64), &mut buf)
             .map_err(|err| {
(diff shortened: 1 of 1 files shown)


Leo-Besancon (approved):
Good fix!

PR #5158, opened 2026-08-13
[F101] make BootstrapPeers serialization canonical by sorting listeners

### LINKED ISSUE #5068: [F101] Non-Deterministic Serialization of Bootstrap Peers Due to HashMap Iteration

---

## Source
AI security-scan **Finding 101** (severity: **info**). Tracked in `SUBMISSIONS.md` as F101.

**Affected file:** `massa-protocol-exports/src/bootstrap_peers.rs`

## Summary
`BootstrapPeersSerializer::serialize` in `massa-main/massa-protocol-exports/src/bootstrap_peers.rs` serializes listener entries by iterating a `HashMap<SocketAddr, TransportType>`, and `BootstrapPeersDeserializer::deserialize` reconstructs those listeners back into a new `HashMap`. As a result, logically identical `BootstrapPeers` values can produce different byte encodings for the listener list, so the format is *non-canonical* at the byte level. The original claim is only valid to that extent: the asserted signature-break, relay, duplicate-filter bypass, and network desynchronization impacts are not supported by the implemented bootstrap flow.

## Root cause
The confirmed issue is the use of `HashMap` iteration as part of the serialized output in `BootstrapPeersSerializer::serialize`, specifically the `for (addr, transport_type) in listeners.iter()` loop over listener endpoints. Because Rust `HashMap` iteration order is not stable, equal `SocketAddr -> TransportType` sets may be emitted in different orders.

The deserialization path in `BootstrapPeersDeserializer::deserialize` reinforces this property by rebuilding listeners with `listeners.into_iter().collect::<HashMap<SocketAddr, TransportType>>()`, which preserves the logical mapping but not any canonical ordering for later serialization.

The ruled-out portion of the original theory is that this non-determinism breaks signature verification in the implemented codebase. In the actual bootstrap flow, the server signs the exact wire bytes before parsing in `massa-bootstrap/src/bindings/server.rs`, and the client verifies those same bytes before deserialization in `massa-bootstrap/src/bindings/client.rs`, so a deserialize-to-reserialize cycle is not part of signature verification.

## Affected code
```rust
        &self,
        buffer: &'a [u8],
    ) -> IResult<&'a [u8], BootstrapPeers, E> {
        length_count(
                self.length_deserializer.deserialize(input)
            }),
            context("Failed Peer deserialization", |input| {
                tuple((
                    context("Failed PeerId deserialization", |input: &'a [u8]| {
                        self.peer_id_deserializer.deserialize(input)
                    }),
                    length_count(
                            self.length_listeners_deserializer.deserialize(input)
                        }),
                        context("Failed listener deserialization", |buffer: &'a [u8]| {
                            tuple((
                                tuple((
                                    context("Failed ip deserialization", |buffer| {
                                        self.ip_addr_deserializer.deserialize(buffer)
                                    }),
                                    context("Failed port deserialization", |buffer| {
                                        self.port_deserializer.deserialize(buffer)
                                    }),
                                ))
                                .map(|(addr, ip)| SocketAddr::new(addr, ip)),
                                context("Failed transport deserialization", |buffer| {
                                    let (rest, id) = nom::number::complete::be_u8(buffer)?;
                                    match id {
                                        0 => Ok((rest, TransportType::Tcp)),
                                        1 => Ok((rest, TransportType::Quic)),
                                        _ => Err(nom::Err::Error(ParseError::from_error_kind(
                                            buffer,
                                            nom::error::ErrorKind::MapRes,
                                        ))),
                                    }
                                }),
                            ))
                            .parse(buffer)
                        }),
                    )
                    .map(|listeners| {
                        listeners
// … (truncated)
```

## Recommendation
Ensure `BootstrapPeers` uses a canonical serialization format by avoiding direct iteration over `HashMap` entries when encoding listener data. Before serialization, impose a deterministic order on listener endpoints, such as by sorting entries on their address and transport type or by storing them in an ordered map structure. Apply the same canonical ordering assumptions throughout serialization and deserialization so that logically identical `BootstrapPeers` values always produce the same byte representation. If byte-level stability is relied upon elsewhere, document this requirement explicitly and add tests that verify repeated serialize–deserialize–serialize cycles remain deterministic for equivalent inputs.

## Why P0 / breaking

peterjah:
  * [ ] make tests compile
  * [ ] make tests pass 

### DIFF
--- a/massa-protocol-exports/src/bootstrap_peers.rs
+++ b/massa-protocol-exports/src/bootstrap_peers.rs
@@ -91,7 +91,13 @@ impl Serializer<BootstrapPeers> for BootstrapPeersSerializer {
             self.peer_id_serializer.serialize(peer_id, buffer)?;
             self.u32_serializer
                 .serialize(&(listeners.len() as u32), buffer)?;
-            for (addr, transport_type) in listeners.iter() {
+            // Emit listeners in a deterministic order: `HashMap` iteration order is not
+            // stable, so without sorting, logically identical peer sets could produce
+            // different byte encodings (a non-canonical format). Sorting by socket
+            // address (unique per map) yields a canonical, reproducible encoding.
+            let mut listeners: Vec<(&SocketAddr, &TransportType)> = listeners.iter().collect();
+            listeners.sort_unstable_by(|(a, _), (b, _)| a.cmp(b));
+            for (addr, transport_type) in listeners {
                 self.ip_addr_serializer.serialize(&addr.ip(), buffer)?;
                 self.port_serializer.serialize(&addr.port(), buffer)?;
                 buffer.push(*transport_type as u8);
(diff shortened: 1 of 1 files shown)


Leo-Besancon (approved):
LGTM

PR #5159, opened 2026-08-14
[F58] fix: only auto-sell rolls that were active at cycle with…

### LINKED ISSUE #5010: [F58] Auto-sell liquidates current rolls instead of the rolls that were active for the penalized cycle

---

## Source
AI security-scan **Finding 58** (severity: **minor**). Tracked in `SUBMISSIONS.md` as F58.

**Affected file:** `massa-execution-worker/src/speculative_roll_state.rs`

## Summary
In `massa-execution-worker/src/speculative_roll_state.rs`, `settle_production_stats` settles penalties for cycle `C` using `production_stats` from that cycle, but liquidates `self.get_rolls(&addr)`, which is the address's *current* roll balance. Elsewhere, production eligibility is based on the delayed active-roll snapshot for `cycle - 3`. As a result, rolls purchased later on the same address, after the obligations for cycle `C` were already determined, can still be auto-sold and reimbursed only at `cycle + 3`. The original claim is overstated to the extent it presents this as a general exploit or loss-of-funds issue rather than an address-level deactivation behavior with delayed liquidity.

## Root cause
The confirmed issue is a mismatch between the rolls used to evaluate missed production and the rolls liquidated when the penalty is applied. In `settle_production_stats`, once `stats.is_satisfying(&max_miss_ratio)` fails, the function reads `owned_count` from `self.get_rolls(&addr)` and then sets `self.added_changes.roll_changes.insert(addr, 0)`, which removes the address's *entire current* roll balance.

This liquidation basis is broader than the rolls that were actually active for the penalized cycle, because active production eligibility elsewhere relies on the delayed snapshot for `cycle - 3`, not the current balance at settlement time. The ruled-out portion of the original theory is that this necessarily constitutes an external attack path: the validator reason confirms that only the penalized staker can cause newly purchased rolls on that same address to be caught by voluntarily adding rolls before settlement.

## Affected code
```rust
/// Settle the production statistics at `slot`.
    ///
    /// IMPORTANT: This function should only be used at the end of a cycle.
    ///
    /// # Arguments:
    /// `slot`: the final slot of the cycle to compute
    pub fn settle_production_stats(
        &mut self,
        slot: &Slot,
        periods_per_cycle: u64,
        thread_count: u8,
        roll_price: Amount,
        max_miss_ratio: Ratio<u64>,
    ) -> Vec<(Address, Amount)> {
        #[allow(unused_mut)]
        let mut result = vec![];
        let cycle = slot.get_cycle(periods_per_cycle);

        let (production_stats, full) =
            self.get_production_stats_at_cycle(cycle, periods_per_cycle, thread_count, slot);
        if !full {
            panic!(
                "production stats were not fully ready when settle_production_stats was executed"
            )
        }

        let target_slot = Slot::new_last_of_cycle(
            cycle
                .checked_add(3)
                .expect("unexpected cycle overflow in settle_production_stats"),
            periods_per_cycle,
            thread_count,
        )
        .expect("unexpected slot overflow in settle_production_stats");

        let mut target_credits = PreHashMap::default();
        for (addr, stats) in production_stats {
            if !stats.is_satisfying(&max_miss_ratio) {
                let owned_count = self.get_rolls(&addr);
                if owned_count != 0 {
                    if let Some(amount) = roll_price.checked_mul_u64(owned_count) {
                        let new_deferred_credits = self
                            .get_address_deferred_credit_for_slot(&addr, &target_slot)
                            .unwrap_or_default()
                            .saturating_add(amount);
// … (truncated)
```

## Recommendation
Limit liquidation during `settle_production_stats` to the roll balance that was active for the penalized cycle, rather than the address’s current roll balance at settlement time. Align the liquidation basis with the same delayed active-roll snapshot used to determine production eligibility and missed-production obligations, so that rolls acquired after those obligations were fixed are not auto-sold. If full-address deactivation is intended instead, explicitly document this behavior and enforce consistent accounting and user-facing semantics to avoid unexpected liquidity delays for newly purchased rolls.

## Why P0 / breaking

Leo-Besancon:
  * [ ] make tests compile
  * [ ] make tests pass 

Note: targetting `dev_breaking` branch as this is a versioned breaking change.

### DIFF
--- a/massa-execution-worker/src/context.rs
+++ b/massa-execution-worker/src/context.rs
@@ -15,7 +15,10 @@ use crate::speculative_deferred_calls::{
 use crate::speculative_executed_denunciations::SpeculativeExecutedDenunciations;
 use crate::speculative_executed_ops::SpeculativeExecutedOps;
 use crate::speculative_ledger::SpeculativeLedger;
-use crate::{active_history::ActiveHistory, speculative_roll_state::SpeculativeRollState};
+use crate::{
+    active_history::ActiveHistory,
+    speculative_roll_state::{SpeculativeRollState, SETTLE_ACTIVE_ROLLS_EXEC_VERSION},
+};
 use massa_async_pool::AsyncPoolChanges;
 
 use massa_deferred_calls::registry_changes::DeferredCallRegistryChanges;
@@ -1101,6 +1104,7 @@ impl ExecutionContext {
                 self.config.thread_count,
                 self.config.roll_price,
                 self.config.max_miss_ratio,
+                self.is_execution_component_version_at_least(SETTLE_ACTIVE_ROLLS_EXEC_VERSION),
             )
         } else {
             vec![]
--- a/massa-execution-worker/src/speculative_roll_state.rs
+++ b/massa-execution-worker/src/speculative_roll_state.rs
@@ -21,6 +21,13 @@ use std::cmp::min;
 use std::collections::{BTreeMap, HashMap};
 use std::sync::Arc;
 
+/// Execution component version from which missed-production auto-sell
+/// liquidates only the rolls that were active for the penalized cycle
+/// (the delayed cycle − 3 snapshot), not the address's entire current balance.
+///
+/// Rides on `MIP-0002-BugFix`, which bumps `MipComponent::Execution` to 2.
+pub(crate) const SETTLE_ACTIVE_ROLLS_EXEC_VERSION: u32 = 2;
+
 /// Speculative state of the rolls
 #[allow(dead_code)]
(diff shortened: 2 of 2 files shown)


peterjah (approved):

damip:
@Leo-Besancon are we confident on the MIP transition for this one?

Leo-Besancon:
> @Leo-Besancon are we confident on the MIP transition for this one?

I think so yes.

As there are not many versioned changes, I think we could make a deployment on buildnet first and test out all the transitions (especially with the tricky transition for F90).

bilboquet (approved):

PR #5161, opened 2026-08-14
[F143] Clone peernet send channels so send_to_peer can serialize with…

### LINKED ISSUE #5109: [F143] Extended RwLock hold during synchronous message serialization in send_to_peer

---

**Severity:** minor (lock-scope / performance)
**Component:** `massa-protocol-worker/src/wrap_network.rs` (`send_to_peer`) +
upstream `peernet::peer::SendChannels`

## Summary

In `send_to_peer`, `self.read().connections.get(peer_id)` holds the global
active-connections `RwLock` **read guard** for the whole `if let` body. The call to
`connection.send_channels.try_send(message_serializer, message, high_priority)` — which
performs **synchronous message serialization** before enqueueing — therefore runs while
the read lock is held, extending the lock scope beyond the map lookup.

(Note: the original "attacker forces unboundedly large responses → node-wide DoS"
framing is not supported by the code; the concrete issue is the lock lifetime.)

## Root cause

`connections.get(peer_id)` returns a reference tied to the read guard, so the guard
cannot be dropped before serialization/enqueue runs inside `SendChannels::try_send`.

The clean fix is to look up the connection, obtain a cheap handle to its send channels,
drop the read guard, then serialize/enqueue. That is **not currently possible from the
massa side**: `peernet::peer::SendChannels` does not derive `Clone` and its
`Sender<Vec<u8>>` fields are private, so the guarded reference cannot be detached.
Fixing this requires an **upstream peernet change** (e.g. `#[derive(Clone)]` on
`SendChannels`, whose fields are already `crossbeam` senders and cheap to clone, or an
accessor returning cloned senders). Coordinating a dependency change places this beyond
a local one-file PR.

- Upstream: make `SendChannels` cloneable (or expose a cloneable send handle).
- massa: in `send_to_peer`, clone the send handle under the read lock, drop the guard,
  then serialize + `try_send` outside the lock.

## Refs

Findings F143.

peterjah:
…out holding the connections lock.

  * [ ] make tests compile
  * [ ] make tests pass 

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -110,7 +110,7 @@ massa_event_cache = { path = "./massa-event-cache" }
 massa-proto-rs = { git = "https://github.com/massalabs/massa-proto-rs.git", "rev" = "9bfbbe1b5dc62815aa2b9f9c5b827b0198da65e3" }
 massa-sc-runtime = { git = "https://github.com/massalabs/massa-sc-runtime", "rev" = "3218b591dcbbc2b88b22a91817434208eb909d12" }
 
-peernet = { git = "https://github.com/massalabs/PeerNet", "rev" = "4f0f69571021de8d7f4c0a70327274c2b0976c71" }
+peernet = { git = "https://github.com/massalabs/PeerNet", "rev" = "c26659997602386d3c9883361d8867139d8de246" }
 # Dev only - use local dependencies
 # massa-proto-rs = { path = "../massa-proto-rs" }
 # massa-sc-runtime = { path = "../massa-sc-runtime" }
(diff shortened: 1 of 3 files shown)


Leo-Besancon (approved):
LGTM, don't forget to make a PR on the [PeerNet ](https://github.com/massalabs/PeerNet) repo and merging it to main before merging this one!

Cargo.toml
```diff
@@ -110,7 +110,7 @@ massa_event_cache = { path = "./massa-event-cache" }
 massa-proto-rs = { git = "https://github.com/massalabs/massa-proto-rs.git", "rev" = "9bfbbe1b5dc62815aa2b9f9c5b827b0198da65e3" }
 massa-sc-runtime = { git = "https://github.com/massalabs/massa-sc-runtime", "rev" = "3218b591dcbbc2b88b22a91817434208eb909d12" }
 
-peernet = { git = "https://github.com/massalabs/PeerNet", "rev" = "04b05ddd320fbe76cc858115af7b5fc28bdb8310" }
+peernet = { git = "https://github.com/massalabs/PeerNet", "rev" = "4f0f69571021de8d7f4c0a70327274c2b0976c71" }
```
Leo-Besancon: This should target a commit_id on the main branch of the [PeerNet](https://github.com/massalabs/PeerNet) repo. Please create a PR there, merge it to main, and update the commit ref here before merging this one!

Leo-Besancon (changes requested):
Just one comment related to Peernet dep, otherwise the implementation LGTM!

Leo-Besancon (approved):

PR #5162, opened 2026-08-14
[F78] bind slot transfers to the trace broadcast event to avoid hybri…

### LINKED ISSUE #5067: [F78] `new_slot_transfers` can combine candidate trace data with a later final transfer cache for the same slot

---

## Source
AI security-scan **Finding 78** (severity: **info**). Tracked in `SUBMISSIONS.md` as F78.

**Affected file:** `massa-grpc/src/stream/new_slot_transfers.rs`

## Summary
`new_slot_transfers` in `massa-grpc/src/stream/new_slot_transfers.rs` constructs each response from two independently read sources: the dequeued `massa_slot_execution_trace` event and `execution_controller.get_transfers_for_slot(slot)`. Because `TraceHistory` stores transfer data only by `slot` and overwrites that entry when the same slot is re-executed, a delayed subscriber can receive an older trace event but append newer cached transfers for that slot. This produces a *hybrid* transfer list that does not correspond to any single execution instance. The original claim is narrower than stated, because this is *not* triggered by the normal candidate-to-final path where finalization reuses speculative results instead of re-executing the slot.

## Root cause
The confirmed issue is caused by `new_slot_transfers` flattening transfers from the broadcasted `massa_slot_execution_trace` and then separately fetching additional transfers through `execution_controller.get_transfers_for_slot(massa_slot_execution_trace.slot)`. These two reads are not tied to the same execution instance.

`TraceHistory` caches transfers by `slot` only, so when a slot is re-executed the cached transfers for that slot replace the earlier entry. If a subscriber processes an older queued event after such a re-execution, the response combines stale trace data with newer cached transfer data.

The ruled-out portion of the original theory is the claim that this occurs in the *common* candidate-to-final transition. In the normal finalization flow, `execute_final_slot` typically reuses speculative `active_history` output via `apply_final_execution_output` rather than calling `execute_slot` again, so that path does not perform the overwrite required for this mismatch.

## Affected code
```rust
let mut subscriber = grpc
    .execution_channels
    .slot_execution_traces_sender
    .subscribe();

...

Ok((massa_slot_execution_trace, is_final)) => {
    ...
    for op_call_stack in massa_slot_execution_trace.operation_call_stacks {
        let op_id = op_call_stack.0;
        let op_call_stack = op_call_stack.1;
        for abi_trace in op_call_stack {
            let only_transfer = abi_trace.flatten_filter(&transfer_abi_names);
            for transfer in only_transfer {
                let (t_from, t_to, t_amount) = transfer.parse_transfer();
                ret_transfers.push(TransferInfo {
                    from: t_from.clone(),
                    to: t_to.clone(),
                    amount: t_amount,
                    operation_id_or_asc_index: Some(
                        grpc_api::transfer_info::OperationIdOrAscIndex::OperationId(op_id.to_string()),
                    ),
                });
            }
        }
    }
    let transfers =
        execution_controller
        .get_transfers_for_slot(massa_slot_execution_trace.slot);
    if let Some(transfers) = transfers {
        for transfer in transfers {
            ret_transfers.push(TransferInfo {
                from: transfer.from.to_string(),
                to: transfer.to.to_string(),
                amount: transfer.amount.to_raw(),
                operation_id_or_asc_index: Some(
                    grpc_api::transfer_info::OperationIdOrAscIndex::OperationId(
                        transfer.op_id.to_string(),
                    ),
                ),
            });
        }
    }
    let ret = grpc_api::NewSlotTransfersResponse {
// … (truncated)
```

## Recommendation
Bind streamed transfer data to a single execution instance instead of combining an event payload with a separate slot-based cache lookup. `new_slot_transfers` should build each response from one authoritative source captured at emission time, or query transfer data using an identifier that uniquely distinguishes re-executions of the same slot rather than `slot` alone.

If a separate cache must be retained, ensure it is versioned per execution instance and cannot overwrite earlier data for the same slot while older events are still consumable. As an additional safeguard, detect and reject mismatched trace and transfer data rather than emitting a mixed response. Finally, document the consistency guarantees of the stream so downstream consumers can rely on stable semantics.

## Why P0 / breaking

peterjah:
…d NewSlotTransfers responses

  * [ ] make tests compile
  * [ ] make tests pass 

### DIFF
--- a/massa-execution-exports/src/channels.rs
+++ b/massa-execution-exports/src/channels.rs
@@ -3,7 +3,7 @@
 use crate::types::SlotExecutionOutput;
 
 #[cfg(feature = "execution-trace")]
-use crate::types_trace_info::SlotAbiCallStack;
+use crate::types_trace_info::{SlotAbiCallStack, Transfer};
 
 #[cfg(feature = "execution-info")]
 use crate::execution_info::ExecutionInfoForSlot;
@@ -13,9 +13,11 @@ use crate::execution_info::ExecutionInfoForSlot;
 pub struct ExecutionChannels {
     /// Broadcast channel for new slot execution outputs
     pub slot_execution_output_sender: tokio::sync::broadcast::Sender<SlotExecutionOutput>,
-    /// Broadcast channel for execution traces (abi call stacks, boolean true if the slot is finalized, false otherwise)
+    /// Broadcast channel for execution traces (abi call stacks, the slot's transfers, boolean true if the slot is finalized, false otherwise).
+    /// The transfers are bound to the same execution instance as the abi call stacks so subscribers never need a separate slot-keyed lookup (which could be overwritten by a re-execution of the same slot).
     #[cfg(feature = "execution-trace")]
-    pub slot_execution_traces_sender: tokio::sync::broadcast::Sender<(SlotAbiCallStack, bool)>,
+    pub slot_execution_traces_sender:
+        tokio::sync::broadcast::Sender<(SlotAbiCallStack, Vec<Transfer>, bool)>,
     /// Broadcast channel for execution info
     #[cfg(feature = "execution-info")]
     pub slot_execution_info_sender: tokio::sync::broadcast::Sender<ExecutionInfoForSlot>,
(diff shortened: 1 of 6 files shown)


Leo-Besancon (dismissed):
Once you fix the test build failure shown in the CI, it looks good to me!

Leo-Besancon (approved):
LGTM

PR #5163, opened 2026-08-14
[F1] Apply the default 50-staker page whenever get_largest_stakers om…

### LINKED ISSUE #5046: [F1] Explicit null pagination bypasses the API v2 staker default cap

---

## Source
AI security-scan **Finding 1** (severity: **info**). Tracked in `SUBMISSIONS.md` as F1.

**Affected file:** `massa-api/src/api.rs`

## Summary
`API<ApiV2>::get_largest_stakers` in `massa-main/massa-api/src/api.rs` applies a `50`-item fallback only when the outer `Option<ApiRequest>` is `None`. When the request is present but `page_request` is absent, such as `Some(ApiRequest { page_request: None })`, the call flows into `PagedVec::new(..., None)`, which returns the full staker list. This creates an inconsistency between omitting `api_request` entirely and providing an explicit or implicit null `page_request`. The original characterization as a security bypass is not supported, because optional pagination is part of the documented API shape and the same `None => full list` behavior is used elsewhere.

## Root cause
The confirmed issue is a branching inconsistency in `massa-main/massa-api/src/api.rs`: when `api_request` is `None`, `get_largest_stakers` constructs `Some(PageRequest { offset: 0, limit: 50 })`, but when `api_request` is `Some(...)`, it forwards `api_request.page_request` directly to `PagedVec::new`.

In `massa-api-exports/src/page.rs`, `PagedVec::new` treats `page_request: None` as `limit = total_count` and `offset = 0`, which returns the complete vector rather than applying the `50`-item fallback. As a result, `None` at the outer `ApiRequest` layer and `None` at the inner `page_request` layer are not equivalent.

The ruled-out portion is the claim that this behavior bypasses an enforced cap. `ApiRequest.page_request` is explicitly optional in the Rust types and OpenRPC schema, and the same full-list behavior for `None` pagination is used elsewhere, so the evidence supports an API design inconsistency rather than circumvention of a mandatory restriction.

## Affected code
```rust
let paged_vec = if let Some(api_request) = api_request {
    PagedVec::new(staker_vec, api_request.page_request)
} else {
    PagedVec::new(
        staker_vec,
        Some(PageRequest {
            offset: 0,
            limit: 50,
        }),
    )
};

impl<T: Serialize> PagedVec<T> {
    pub fn new(elements: Vec<T>, page_request: Option<PageRequest>) -> Self {
        let total_count = elements.len();

        let (limit, offset) = match page_request {
            Some(PageRequest { limit, offset }) => (limit, offset),
            None => (total_count, 0),
        };
```

## Recommendation
Normalize pagination handling in `API<ApiV2>::get_largest_stakers` so that omitted `api_request` and absent or null `page_request` are treated consistently. If the intended behavior is to enforce a default `50`-item response when pagination is not specified, apply the same fallback whenever `page_request` is `None`, regardless of whether `api_request` is present. Otherwise, if returning the full list without pagination is intentional, remove the special `50`-item default from the `api_request == None` branch and align the endpoint with the broader API behavior.

In either case, document the chosen semantics clearly in the OpenRPC schema and endpoint documentation so clients can rely on consistent behavior. Add regression tests covering all equivalent input shapes, including omitted `api_request`, empty request objects, and explicit `null` `page_request`, to prevent future divergence.

## Why P0 / breaking

peterjah:
…its pagination

  * [ ] make tests compile
  * [ ] make tests pass 

### DIFF
--- a/massa-api/src/api.rs
+++ b/massa-api/src/api.rs
@@ -94,17 +94,16 @@ impl MassaApiServer for API<ApiV2> {
         staker_vec
             .sort_by(|&(_, roll_counts_a), &(_, roll_counts_b)| roll_counts_b.cmp(&roll_counts_a));
 
-        let paged_vec = if let Some(api_request) = api_request {
-            PagedVec::new(staker_vec, api_request.page_request)
-        } else {
-            PagedVec::new(
-                staker_vec,
-                Some(PageRequest {
-                    offset: 0,
-                    limit: 50,
-                }),
-            )
-        };
+        // Omitted api_request and a present request with no page_request must
+        // behave the same: default to the first 50 largest stakers.
+        const DEFAULT_PAGE_LIMIT: usize = 50;
+        let page_request = api_request
+            .and_then(|request| request.page_request)
+            .or(Some(PageRequest {
+                offset: 0,
+                limit: DEFAULT_PAGE_LIMIT,
+            }));
+        let paged_vec = PagedVec::new(staker_vec, page_request);
 
         Ok(paged_vec.into())
     }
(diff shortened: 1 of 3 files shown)


Leo-Besancon (approved):
LGTM!

PR #5164, opened 2026-08-14
[F14] bound bootstrap update-section length and fold updates directly…

### LINKED ISSUE #5055: [F14] Client-controlled cursors can trigger unbounded delta serialization and memory-heavy replies

---

## Source
AI security-scan **Finding 14** (severity: **medium**). Tracked in `SUBMISSIONS.md` as F14.

**Affected file:** `massa-bootstrap/src/server.rs`

## Summary
`BootstrapServerMessageDeserializer` in `massa-bootstrap/src/messages.rs` accepts very large update-section lengths and parses them into an intermediate `Vec` before collecting into a `BTreeMap`. This allows a malicious bootstrap server to send a single large `BootstrapPart` frame with many tiny update entries and force substantial client-side CPU and memory consumption. On the server side, `stream_bootstrap_information` in `massa-bootstrap/src/server.rs` also trusts the initial client-provided resume cursors, so an attacker can select an old in-history `last_slot` together with advanced state or versioning steps and trigger a delta-heavy reply that is fully materialized before write throttling applies. The original claim is overstated to the extent it describes these replies as *unbounded* or repeatedly rewindable within a single stream.

## Root cause
The confirmed client-side issue is caused by `BootstrapServerMessageDeserializer::new` configuring `stream_batch_updates_length_deserializer` with `U64VarIntDeserializer::new(Included(0), Included(u64::MAX))`, and by `BootstrapServerMessageDeserializer::deserialize` parsing update sections with `length_value(..., many0(...))`. This causes large update payloads to be accumulated in an intermediate `Vec` and then allocated again during `into_iter().collect()` into `StreamBatch::<Slot>::updates_on_previous_elements`, making deserialization of a large but valid bootstrap frame memory- and CPU-intensive.

The confirmed server-side issue is narrower. In `stream_bootstrap_information`, the initial `last_slot`, `last_state_step`, and `last_versioning_step` supplied by the client are used to call `get_batch_to_stream(&last_state_step, last_slot)` and `get_versioning_batch_to_stream(&last_versioning_step, last_slot)` before the server replaces them with updated cursors for the next loop iteration. This allows a client to choose an old in-history `last_slot` with advanced streaming steps and induce a delta-heavy response that is serialized in memory before backpressure from `send_msg` applies.

The ruled-out portion is the claim that the server permits *unbounded* replay or rewind of progress throughout a single bootstrap session. That does not hold because each loop iteration in `stream_bootstrap_information` updates `last_state_step`, `last_versioning_step`, `last_consensus_step`, and `last_slot` internally, so the client does not keep resubmitting arbitrary cursors mid-stream as described. The consensus-specific claim is also materially overstated, because consensus output remains bounded by `bootstrap_part_size` and cursor size limits.

## Affected code
```rust
#[allow(clippy::too_many_arguments)]
pub fn stream_bootstrap_information(
    server: &mut BootstrapServerBinder,
    final_state: Arc<RwLock<dyn FinalStateController>>,
    consensus_controller: Box<dyn ConsensusController>,
    mut last_slot: Option<Slot>,
    mut last_state_step: StreamingStep<Vec<u8>>,
    mut last_versioning_step: StreamingStep<Vec<u8>>,
    mut last_consensus_step: StreamingStep<PreHashSet<BlockId>>,
    mut send_last_start_period: bool,
    bs_deadline: &Instant,
    write_timeout: Duration,
) -> Result<(), BootstrapError> {
    loop {
        let current_slot;
        let state_part;
        let versioning_part;
        let last_start_period;
        let last_slot_before_downtime;

        // Scope of the final state read
        {
            let final_state_read = final_state.read();

            last_start_period = if send_last_start_period {
                Some(final_state_read.get_last_start_period())
            } else {
                None
            };
            last_slot_before_downtime = if send_last_start_period {
                Some(*final_state_read.get_last_slot_before_downtime())
            } else {
                None
            };

            let state_part_res = final_state_read
                .get_database()
                .read()
                .get_batch_to_stream(&last_state_step, last_slot);

            match state_part_res {
                    Ok(part) => state_part = part,
                    Err(MassaDBError::CacheMissError(s)) if s.as_str() == "all our changes are strictly after last_change_id, we can't be sure we did not miss any" => {
                        return server.send_msg(write_timeout, BootstrapServerMessage::SlotTooOld);
                    }
// … (truncated)
```

## Recommendation
Bound the accepted sizes of all bootstrap update sections to protocol-level limits that are consistent with `MAX_BOOTSTRAP_MESSAGE_SIZE`, and reject frames whose declared lengths would require excessive allocation. Avoid deserializing update entries into an intermediate `Vec` before building the final map; deserialize directly into the target structure or otherwise enforce strict per-section and per-entry caps to prevent duplicate allocation and excessive CPU usage.

On the server side, validate the initial client-provided resume cursors for consistency before generating the first reply. In particular, reject or clamp combinations such as an old `last_slot` paired with advanced state or versioning steps if they would force an unnecessarily large delta response. Ensure the cost of any single bootstrap part remains bounded before serialization, and apply size-aware chunking so large deltas are split across multiple parts rather than fully materialized in one response.

Additionally, add defensive accounting for both parsing and reply construction, including explicit limits on the number of update entries, cumulative serialized size, and memory budget per bootstrap part. If these limits are exceeded, terminate the bootstrap exchange early and log the event for monitoring.

## Why P0 / breaking

peterjah:
… into the map

  * [ ] make tests compile
  * [ ] make tests pass 

### DIFF
--- a/massa-bootstrap/src/messages.rs
+++ b/massa-bootstrap/src/messages.rs
@@ -31,7 +31,7 @@ use std::collections::BTreeMap;
 
 use massa_time::{MassaTime, MassaTimeDeserializer, MassaTimeSerializer};
 use nom::error::context;
-use nom::multi::{fold_many0, length_data, length_value, many0};
+use nom::multi::{fold_many0, length_data, length_value};
 use nom::sequence::tuple;
 use nom::Parser;
 use nom::{
@@ -506,10 +506,25 @@ impl Deserializer<BootstrapServerMessage> for BootstrapServerMessageDeserializer
                                         self.state_new_elements_length_deserializer
                                             .deserialize(input)
                                     }),
-                                    many0(tuple((
-                                        |input| self.datastore_key_deserializer.deserialize(input),
-                                        |input| self.datastore_val_deserializer.deserialize(input),
-                                    ))),
+                                    // Fold directly into the target `BTreeMap` instead of
+                                    // collecting into an intermediate `Vec` first: the new_elements
+                                    // section can be large, so avoiding the extra full-size
+                                    // allocation roughly halves the peak memory of parsing it.
+                                    fold_many0(
+                                        tuple((
+                                            |input| {
+                                                self.datastore_key_deserializer.deserialize(input)
+                                            },
+                                            |input| {
+                                                self.datastore_val_deserializer.deserialize(input)
+                                            },
+                                        )),
+                                        BTreeMap::new,
+                                        |mut acc, (key, value)| {
+                                            acc.insert(key, value);
+                                            acc
+                                        },
+                                    ),
                                 ),
                             ),
                             context(
@@ -553,10 +568,25 @@ impl Deserializer<BootstrapServerMessage> for BootstrapServerMessageDeserializer
                                         self.versioning_part_new_elements_length_deserializer
                                             .deserialize(input)
                                     }),
-                                    many0(tuple((
-                                        |input| self.datastore_key_deserializer.deserialize(input),
-                                        |input| self.datastore_val_deserializer.deserialize(input),
-                                    ))),
+                                    // Fold directly into the target `BTreeMap` instead of
+                                    // collecting into an intermediate `Vec` first: the new_elements
+                                    // section can be large, so avoiding the extra full-size
+                                    // allocation roughly halves the peak memory of parsing it.
+                                    fold_many0(
+                                        tuple((
+                                            |input| {
+                                                self.datastore_key_deserializer.deserialize(input)
+                                            },
+                                            |input| {
+                                                self.datastore_val_deserializer.deserialize(input)
+                                            },
+                                        )),
+                                        BTreeMap::new,
+                                        |mut acc, (key, value)| {
+                                            acc.insert(key, value);
+                                            acc
+                                        },
+                                    ),
                                 ),
                             ),
                             context(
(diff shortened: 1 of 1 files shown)


massa-bootstrap/src/messages.rs
```diff
@@ -513,10 +519,23 @@ impl Deserializer<BootstrapServerMessage> for BootstrapServerMessageDeserializer
...
                                     }),
-                                    many0(tuple((
-                                        |input| self.datastore_key_deserializer.deserialize(input),
-                                        |input| self.opt_vec_u8_deserializer.deserialize(input),
-                                    ))),
+                                    // Fold directly into the target `BTreeMap` instead of
+                                    // collecting into an intermediate `Vec` first: the updates
+                                    // section can be large, so avoiding the extra full-size
+                                    // allocation roughly halves the peak memory of parsing it.
+                                    fold_many0(
```
Leo-Besancon, line 541: I think we can parse the `new_elements` parts in the same way, directly into a BTreeMap, for consistency (and even more RAM optim).

Leo-Besancon (changes requested):
Left one comment, but the fix looks good!

Note: the issue desribes a server fix to validate client inputs, but it does not make sense, every cursors asked by the client can be valid. So once my comment below is resolved, we can merge this PR!

massa-bootstrap/src/messages.rs
peterjah: done

Leo-Besancon (approved):
Thanks, LGTM!

PR #5165, opened 2026-08-18
[F91] feat: make chain() atomic in case of error

### LINKED ISSUE #5026: [F91] Lack of Atomicity in LedgerChanges::chain Leads to Partial State Mutation

---

## Source
AI security-scan **Finding 91** (severity: **info**). Tracked in `SUBMISSIONS.md` as F91.

**Affected file:** `massa-models/src/ledger.rs`

## Summary
`LedgerChanges::chain` in `massa-models/src/ledger.rs` applies entries from another `LedgerChanges` instance directly into `self` in a loop. If a later `self.apply(addr, change)` returns `Err`, any earlier successful mutations remain in `self`, so the operation is not atomic at the type level. The original finding is only valid to that extent: the claim that this behavior lets an external attacker submit a crafted batch and mint funds is not supported.

## Root cause
The confirmed issue is that `LedgerChanges::chain` performs incremental mutation without rollback:

- `chain` iterates over `other.0` and calls `self.apply(addr, change)?` for each entry.
- If one call fails, the `?` operator exits immediately, but prior successful `apply` calls have already modified `self`.
- As a result, `self` can be left in a partially updated state after an error.

The ruled-out portion of the original root-cause theory does not hold:

- A single `Amount::MAX` change on a fresh address is not sufficient to trigger the described failure path, because the vacant-entry case starts from `LedgerChange::default()` and that first merge does not overflow.
- The broader protocol-exploit theory is unsupported because this `LedgerChanges` type in `massa-models/src/ledger.rs` does not appear to be used by the runtime execution or final-state paths, which instead use `massa_ledger_exports::LedgerChanges`.

## Affected code
```rust
pub fn chain(&mut self, other: &LedgerChanges) -> Result<()> {
        for (addr, change) in other.0.iter() {
            self.apply(addr, change)?;
        }
        Ok(())
    }
```

## Recommendation
Make `LedgerChanges::chain` transactional so that it either applies all changes or leaves `self` unchanged on failure. Avoid mutating `self` incrementally while an error can still occur. Instead, validate that all merges can succeed before committing, or apply changes to a temporary copy and replace `self` only after the full operation completes successfully. If atomic behavior is not intended, document the partial-application semantics explicitly and rename or redesign the API to prevent callers from assuming all-or-nothing behavior.

## Why P0 / breaking

Leo-Besancon:
Note: there is no need to make it a breaking change (and add versioning to it) as this method is not part of the execution path, it's currently unused.

### DIFF
--- a/massa-models/src/ledger.rs
+++ b/massa-models/src/ledger.rs
@@ -461,9 +461,12 @@ impl LedgerChanges {
 
     /// chain with another `LedgerChange`
     pub fn chain(&mut self, other: &LedgerChanges) -> Result<()> {
+        // We avoid mutating self directly to ensure atomicity in error cases.
+        let mut updated = self.clone();
         for (addr, change) in other.0.iter() {
-            self.apply(addr, change)?;
+            updated.apply(addr, change)?;
         }
+        *self = updated;
         Ok(())
     }
 


peterjah (approved):

bilboquet:
As the function is not used I would have just dropped it.
That being said, I'm ok with the fix ^^

bilboquet (approved):

Leo-Besancon:
> As the function is not used I would have just dropped it. That being said, I'm ok with the fix ^^

True, I should have, but as I made the fix let's merge it, we can avoid future hassle if we need it at some point!

PR #5166, opened 2026-08-18
[F67] fix: reset whole db before retrying bootstrap

### LINKED ISSUE #4994: [F31] Database reset leaves stale on-disk state across bootstrap retries
> ⚠️ **Warning — This finding was routed to an issue rather than a PR because it affects persisted final-state / ledger / DB recovery and bootstrap-restart handling.** A fix changes on-disk or state-recovery behavior that must stay consistent across nodes and restarts; a wrong change can corrupt or diverge state, so it needs design review and careful (possibly coordinated) rollout.

---

## Source
AI security-scan **Finding 31** (severity: **medium**). Tracked in `SUBMISSIONS.md` as F31.

**Affected file:** `massa-db-worker/src/massa_db.rs`

## Summary
`RawMassaDB::reset` in `massa-db-worker/src/massa_db.rs` only resets in-memory history and the initial change id; it does not wipe on-disk RocksDB entries. In the bootstrap retry path, this becomes a real issue for deferred-call state because `FinalState::reset` does not clear `DeferredCallRegistry`, and the subsequent bootstrap replay overlays streamed keys without sending deletions for keys absent from the server snapshot. As a result, stale entries under `DEFERRED_CALLS_PREFIX` and `DEFERRED_CALL_TOTAL_GAS` can persist across retries. The broader claim that this causes a general full-database corruption across ledger, PoS, executed ops, MIP, or most `STATE_CF`/`VERSIONING_CF` content is not supported because those practical final-state prefixes are explicitly cleared before retry.

## Root cause
The confirmed root cause is the interaction between three concrete behaviors:

- `RawMassaDB::reset` only calls `set_initial_change_id(slot)` and clears `change_history` and `change_history_versioning`; it does not delete existing on-disk keys from the database.
- `FinalState::reset` does not reset `DeferredCallRegistry`, so deferred-call data stored under `DEFERRED_CALLS_PREFIX` and `DEFERRED_CALL_TOTAL_GAS` is left behind during a bootstrap retry.
- Bootstrap replay through `write_batch_bootstrap_client` and `write_changes(..., reset_history = true)` applies incoming state as an overlay, and an initial stream from `StreamingStep::Started` does not include deletion tombstones for keys that are absent from the remote snapshot.

Because of that combination, deferred-call entries that already exist locally but are missing from the new snapshot are not removed during the next bootstrap. The original theory that this mechanism leaves most other final-state prefixes stale is ruled out here, because the retry flow explicitly clears nearly all practical final-state prefixes before bootstrap resumes.

## Affected code
```rust
/// Reset the database, and attach it to the given slot.
    ///
    /// This function is used in the FinalStateController::reset method which is used in the Bootstrap
    /// process when the bootstrap fails (Bootstrap slot too old). A bootstrap to another node will likely occur
    /// after this reset.
    fn reset(&mut self, slot: Slot) {
        // For dev: please take care of correctly reset the db to avoid any issue when the bootstrap
        //          process is restarted
        self.set_initial_change_id(slot); // Note: this also reset the field: current_batch
        self.change_history.clear();
        self.change_history_versioning.clear();
    }
```

## Recommendation
Ensure the bootstrap retry path fully clears deferred-call state before replaying a new snapshot. In particular, `FinalState::reset` should also reset `DeferredCallRegistry`, including all keys under `DEFERRED_CALLS_PREFIX` and `DEFERRED_CALL_TOTAL_GAS`, so that no stale deferred-call data survives across retries.

More generally, treat bootstrap from `StreamingStep::Started` as a full-state replacement rather than an overlay for components that are not explicitly deleted by the streamed snapshot. If full clearing at the `FinalState` layer is not feasible, add an explicit cleanup step for deferred-call on-disk entries before applying bootstrap batches. This ensures absent deferred-call records are removed and prevents stale scheduling data and gas accounting from persisting after a retry.

## Why P0 / breaking

### LINKED ISSUE #5016: [F67] Incomplete Bootstrap Reset Can Persist Stale Deferred-Call State Across Retries
> ⚠️ **Warning — This finding was routed to an issue rather than a PR because it affects persisted final-state / ledger / DB recovery and bootstrap-restart handling.** A fix changes on-disk or state-recovery behavior that must stay consistent across nodes and restarts; a wrong change can corrupt or diverge state, so it needs design review and careful (possibly coordinated) rollout.

---

## Source
AI security-scan **Finding 67** (severity: **major**). Tracked in `SUBMISSIONS.md` as F67.

**Affected file:** `massa-final-state/src/final_state.rs`

## Summary
`FinalState::reset()` is used to restart final state reconstruction during bootstrap. It clears many parts of the state, but it does *not* remove deferred-call data that was already written to RocksDB. If bootstrap later restarts after a `SlotTooOld` response, those old deferred-call entries can stay on disk and be loaded again, which can make the rebuilt final state contain stale data.

## Root cause
The main issue is that `FinalState::reset()` rewinds the database change id and clears ledger, async pool, PoS, executed ops, executed denunciations, MIP data, and the execution trail hash, but it does *not* clear deferred-call rows.

`RawMassaDB::reset()` also contributes to the problem. It only rewinds metadata and does not delete stored value keys, so previously written deferred-call keys remain in RocksDB.

During bootstrap, streamed state batches are flushed directly to RocksDB before bootstrap completes. This includes deferred-call entries and the related deferred total-gas/base-fee accounting. If the server later returns `SlotTooOld`, the retry path calls `FinalState::reset()` and restarts from genesis, but the deferred-call keys written by the failed attempt are still present on disk.

A fresh bootstrap snapshot does not delete keys that are absent from the new image. Because of this, stale deferred-call keys from an earlier bootstrap attempt can survive across retries and then be reloaded into the final state.

## Affected code
```rust
fn reset(&mut self) {
    let slot = Slot::new(0, self.config.thread_count.saturating_sub(1));
    self.db.write().reset(slot);
    self.ledger.reset();
    self.async_pool.reset();
    self.pos_state.reset();
    self.executed_ops.reset();
    self.executed_denunciations.reset();
    self.mip_store.reset_db(self.db.clone());
    self.db.write().delete_prefix(EXECUTION_TRAIL_HASH_PREFIX, STATE_CF, None);
}

let (updated, added) = guard
    .get_mip_store_mut()
    .extend_from_db(db)
    .map_err(|e| BootstrapError::from(FinalStateError::from(e)))?;

fn extend_from_db(
    &mut self,
    db: ShareableMassaDBController,
) -> Result<(Vec<MipInfo>, BTreeMap<MipInfo, MipState>), ExtendFromDbError> {
    ...
    if !update_data.is_empty() {
        let store_raw_ = MipStoreRaw { ... };
        let (updated_2, added_2) = self.update_with(&store_raw_)?;
        updated.extend(updated_2);
        added.extend(added_2);
    }
    Ok((updated, added))
}
```

## Recommendation
Ensure `FinalState::reset()` fully clears all deferred-call state persisted during bootstrap, including deferred-call entries and their related total-gas/base-fee accounting, so no stale data survives a retry. The reset path should also guarantee that any keys written by an incomplete bootstrap attempt are removed from RocksDB rather than only rewinding metadata. More generally, bootstrap reconstruction should make the persisted final state exactly match the received snapshot, deleting any keys that are absent from the new bootstrap image. Consider treating bootstrap writes as isolated or temporary until bootstrap completes successfully, and only committing them once the snapshot has been fully validated.

## Why P0 / breaking

Leo-Besancon:
  * [ ] make tests compile
  * [ ] make tests pass 

### DIFF
--- a/massa-db-exports/src/controller.rs
+++ b/massa-db-exports/src/controller.rs
@@ -35,7 +35,8 @@ pub trait MassaDBController: Send + Sync + Debug {
     fn delete_prefix(&mut self, prefix: &str, handle_str: &str, change_id: Option<Slot>);
 
     /// Reset the database, and attach it to the given slot.
-    fn reset(&mut self, slot: Slot);
+    /// Note that this does not clear the db's content, only the change_id and the history.
+    fn reset_slot_and_history(&mut self, slot: Slot);
 
     /// Exposes RocksDB's "get_cf" function
     fn get_cf(&self, handle_cf: &str, key: Key) -> Result<Option<Value>, MassaDBError>;
--- a/massa-db-worker/src/massa_db.rs
+++ b/massa-db-worker/src/massa_db.rs
@@ -816,8 +816,8 @@ impl MassaDBController for RawMassaDB<Slot, SlotSerializer, SlotDeserializer> {
     ///
     /// This function is used in the FinalStateController::reset method which is used in the Bootstrap
     /// process when the bootstrap fails (Bootstrap slot too old). A bootstrap to another node will likely occur
-    /// after this reset.
-    fn reset(&mut self, slot: Slot) {
+    /// after this reset. Note that this does not clear the db's content, only the change_id and the history.
+    fn reset_slot_and_history(&mut self, slot: Slot) {
         // For dev: please take care of correctly reset the db to avoid any issue when the bootstrap
         //          process is restarted
         self.set_initial_change_id(slot); // Note: this also reset the field: current_batch
--- a/massa-final-state/src/final_state.rs
+++ b/massa-final-state/src/final_state.rs
@@ -862,10 +862,9 @@ impl FinalStateController for FinalState {
         self.executed_ops.reset();
         self.executed_denunciations.reset();
         self.mip_store.reset_db(self.db.clone());
-        // delete the execution trail hash
-        self.db
-            .write()
-            .delete_prefix(EXECUTION_TRAIL_HASH_PREFIX, STATE_CF, None);
+        // delete all remaining data from the database (execution trail hash, deferred calls, etc.)
+        self.db.write().delete_prefix("", STATE_CF, None);
+        self.db.write().delete_prefix("", VERSIONING_CF, None);
     }
 
     fn get_ledger(&self) -> &Box<dyn LedgerController> {
--- a/massa-final-state/src/final_state.rs
+++ b/massa-final-state/src/final_state.rs
@@ -855,7 +855,7 @@ impl FinalStateController for FinalState {
 
     fn reset(&mut self) {
         let slot = Slot::new(0, self.config.thread_count.saturating_sub(1));
-        self.db.write().reset(slot);
+        self.db.write().reset_slot_and_history(slot);
         self.ledger.reset();
         self.async_pool.reset();
         self.pos_state.reset();


peterjah (dismissed):

Leo-Besancon:
Note that https://github.com/massalabs/massa/issues/4994 was a duplicate of finding 67 solved by this PR. I've just added a small clarifying comment and rename to make it clear that RawMassaDB::reset only reset the slot and history, not the DB's content.

bilboquet (approved):

PR #5168, opened 2026-08-18
 fix(grpc): gate get_slot_transfers on blockclique block to match JSO…

peterjah:
…N-RPC query semantics

This is a followup from F76

### DIFF
--- a/massa-grpc/src/public.rs
+++ b/massa-grpc/src/public.rs
@@ -534,7 +534,11 @@ pub fn into_element(abi_trace: &AbiTrace) -> AbiCallStackElementParent {
 }
 
 #[cfg(feature = "execution-trace")]
-/// Get slot transfers
+/// Get slot transfers.
+///
+/// A slot is only reported if it currently has a block in the blockclique; otherwise its
+/// transfers are returned empty. This matches the JSON-RPC `get_slots_transfers` availability
+/// semantics so both query APIs agree for a given slot.
 pub(crate) fn get_slot_transfers(
     grpc: &MassaPublicGrpc,
     request: tonic::Request<grpc_api::GetSlotTransfersRequest>,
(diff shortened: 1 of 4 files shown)


massa-grpc/src/public.rs
```diff
@@ -549,6 +553,18 @@ pub(crate) fn get_slot_transfers(
...
         };
+        // Only report transfers for a slot that currently has a block in the blockclique,
+        // mirroring the JSON-RPC `get_slots_transfers` behavior so both query APIs return the
+        // same batch for a given slot. Without this gate the two queries could disagree for a
+        // slot whose block is not (or no longer) in the blockclique.
+        if grpc
+            .consensus_controller
+            .get_blockclique_block_at_slot(slot.clone().into())
+            .is_none()
+        {
```
Leo-Besancon: I agree that we should unify the behaviour for both APIs, but I'm wondering about potential consequences of this change (it could potentially impact the transfers-indexer, and as a result the centralized exchanges that rely on it).

@damip do you agree?

damip: I think the JSON one has an issue already: why do we force it to have a block in the blockclique ? why not count final blocks as well ? This should not break compat, just return more stuff when possible (worth checking the transfers indexer just in case)
Basically what I would do is fix the original JSON one and leave the GRPC one as it was because I think it is correct.

peterjah: done

peterjah: "why force a block in the blockclique, why not count final blocks as well?" — get_blockclique_block_at_slot already does both. massa-consensus-worker/src/state/mod.rs:128:

/// get the blockclique (or final) block ID at a given slot, if any
// block not found in the blockclique: search in the final blocks
blocks_at_slot.into_iter().find(|b_id| matches!(... a_block.is_final))

Leo-Besancon (approved):
If I understand correctly this PR aligns with how the JSON rpc behaves on slots that are neither in blockclique nor final.
@peterjah If that's the case I approve!

PR #5169, opened 2026-08-18
[F5] [F6] bound the bootstrap resume cursor with max_consensus_block_ids i…

### LINKED ISSUE #5049: [F5] Bootstrap Resume Can Fail Due to Incorrect Per-Part Size Limit on Cumulative Consensus Cursor

---

## Source
AI security-scan **Finding 5** (severity: **medium**). Tracked in `SUBMISSIONS.md` as F5.

**Affected file:** `massa-bootstrap/src/bindings/server.rs`

## Summary
The bootstrap server lets a client resume bootstrap by sending `AskBootstrapPart.last_consensus_step`. This field contains the set of consensus block ids already received by the client. The server deserializes that field using `consensus_bootstrap_part_size` as the size limit, even though the resume cursor is cumulative across all previously streamed consensus blocks. If the client reconnects after receiving enough parts, the resume message can be rejected, which can prevent the node from finishing bootstrap.

## Root cause
The issue is in the bootstrap server handling of `AskBootstrapPart.last_consensus_step`. The server uses `consensus_bootstrap_part_size` as the deserialization limit for this field.

This is incorrect because `AskBootstrapPart.last_consensus_step` is a *resume cursor*, not a single-part payload. The client stores previously received consensus block ids cumulatively across all consensus blocks already streamed. After an interruption, the client retries by sending all of those ids in `last_consensus_step`.

Because the server still applies the per-part limit `consensus_bootstrap_part_size` during deserialization, a valid resume request is rejected once `last_consensus_step` grows larger than one part.

## Affected code
```rust
pub fn new(
    duplex: TcpStream,
    local_keypair: KeyPair,
    cfg: BootstrapSrvBindCfg,
    rw_limit: Option<u64>,
) -> Self {
    let BootstrapSrvBindCfg {
        rate_limit: _limit,
        thread_count,
        max_datastore_key_length,
        randomness_size_bytes,
        consensus_bootstrap_part_size,
        write_error_timeout,
    } = cfg;

    ...
    BootstrapServerBinder {
        max_consensus_block_ids: consensus_bootstrap_part_size,
        ...
    }
}

let (_, msg) = BootstrapClientMessageDeserializer::new(
    self.thread_count,
    self.max_datastore_key_length,
    self.max_consensus_block_ids,
)
.deserialize::<DeserializeError>(&msg_bytes)
.map_err(|err| BootstrapError::GeneralError(format!("{}", err)))?;

let last_consensus_step = StreamingStep::Ongoing(
    global_bootstrap_state
        .graph
        .as_ref()
        .unwrap()
        .final_blocks
        .iter()
        .map(|b_export| b_export.block.id)
        .collect(),
);
*next_bootstrap_message = BootstrapClientMessage::AskBootstrapPart {
    last_slot: Some(slot),
    last_state_step,
    last_versioning_step,
    last_consensus_step,
// … (truncated)
```

## Recommendation
Use a deserialization limit for `AskBootstrapPart.last_consensus_step` that matches its role as a cumulative resume cursor rather than a single streamed part. The server should accept a cursor large enough to represent all previously received consensus block IDs, or validate it with dedicated bounds derived from the maximum resumable bootstrap state. Ensure the same sizing assumptions are applied consistently on both serialization and deserialization paths so that valid resume requests cannot be rejected after multiple consensus parts have already been received.

## Why P0 / breaking

### LINKED ISSUE #5050: [F6] Bootstrap retry can mix state across sessions and accept inconsistent bootstrap snapshots

---

## Source
AI security-scan **Finding 6** (severity: **medium**). Tracked in `SUBMISSIONS.md` as F6.

**Affected file:** `massa-bootstrap/src/client.rs`

## Summary
The bootstrap retry logic in `massa-main/massa-bootstrap/src/client.rs` keeps `next_bootstrap_message` and `global_bootstrap_state` alive across retry attempts. If one bootstrap session fails after partial progress, the next session can continue with leftover state from the previous one instead of starting cleanly. This can make the client accept a final bootstrap result built from mixed data taken from different servers or different failed sessions. The result is an inconsistent local view that may not match any single valid bootstrap snapshot.

## Root cause
The main issue is in `massa-main/massa-bootstrap/src/client.rs`, where the outer retry loop preserves two mutable objects across calls to `bootstrap_from_server(...)`:

- `next_bootstrap_message`
- `global_bootstrap_state`

They are initialized once as:

- `BootstrapClientMessage::AskBootstrapPart { last_slot: None, last_state_step: StreamingStep::Started, last_versioning_step: StreamingStep::Started, last_consensus_step: StreamingStep::Started, send_last_start_period: true }`
- `GlobalBootstrapState::new(final_state)`

They are then passed by mutable reference into `bootstrap_from_server(...)` and reused on later retries. Because they are not rebuilt per server or per bootstrap session, partially consumed state can survive failures and be combined with data from a later session.

Restart metadata is stored early and not always requested again. After a first `BootstrapPart`, the client writes:

- `write_final_state.set_last_start_period(last_start_period);`
- `write_final_state.set_last_slot_before_downtime(last_slot_before_downtime);`

It then rewrites the cursor to:

- `BootstrapClientMessage::AskBootstrapPart { ..., send_last_start_period: false }`

If the connection fails after that point, a retry keeps both the stored metadata in `global_bootstrap_state.final_state` and the updated `next_bootstrap_message`. The next server is therefore asked to continue without sending restart metadata again, so its `state_part`, `versioning_part`, and `consensus_part` can be combined with `last_start_period` or `last_slot_before_downtime` from the previous server, without a consistency check.

The reset path for `BootstrapServerMessage::SlotTooOld` is incomplete. On that path, the client logs `Slot is too old retry bootstrap from scratch`, resets `next_bootstrap_message` to a fres …

## Affected code
```rust
let mut next_bootstrap_message: BootstrapClientMessage =
    BootstrapClientMessage::AskBootstrapPart {
        last_slot: None,
        last_state_step: StreamingStep::Started,
        last_versioning_step: StreamingStep::Started,
        last_consensus_step: StreamingStep::Started,
        send_last_start_period: true,
    };
let mut global_bootstrap_state = GlobalBootstrapState::new(final_state);

...

let bs = bootstrap_from_server(
    bootstrap_config,
    &mut client,
    &mut next_bootstrap_message,
    &mut global_bootstrap_state,
    version,
);

...

if let Some(last_start_period) = last_start_period {
    write_final_state.set_last_start_period(last_start_period);
}
if let Some(last_slot_before_downtime) = last_slot_before_downtime {
    write_final_state.set_last_slot_before_downtime(last_slot_before_downtime);
}

...

*next_bootstrap_message = BootstrapClientMessage::AskBootstrapPart {
    last_slot: Some(slot),
    last_state_step,
    last_versioning_step,
    last_consensus_step,
    send_last_start_period: false,
};
```

## Recommendation
Reinitialize all bootstrap session state on every retry and treat each retry as a fully independent bootstrap attempt. In particular, recreate both `next_bootstrap_message` and `global_bootstrap_state` before each call to `bootstrap_from_server(...)` so that no partially received final-state, versioning, consensus, or restart metadata can be reused across sessions.

Ensure that any restart triggered by failure, including `BootstrapServerMessage::SlotTooOld`, performs a complete reset of all accumulated bootstrap state rather than only resetting `final_state` or the request cursor. This reset should also clear retained consensus data such as `global_bootstrap_state.graph` and any cached optional restart fields, so a new server cannot inherit stale data from a previous attempt.

Require restart metadata to be bound to the same bootstrap session as the streamed snapshot data. If fields such as `last_start_period` or `last_slot_before_downtime` are optional, explicitly clear previously stored values when a new session begins and reject completion if the final accepted snapshot depends on metadata carried over from an earlier session.

Delay advancing the bootstrap phase to `AskBootstrapPeers` until all final-state and versioning post-processing has completed successfully, including `MipStore::extend_from_db`. If that step fails, discard any partially mutated in-memory state and restart bootstrap from a fresh `AskBootstrapPart` session instead of resuming from the peer phase.

More generally, enforce an atomic bootstrap outcome: either one complete snapshot from one session is fully accepted, or the entire attempt is discarded and replayed from scratch.

## Why P0 / breaking

peterjah:
…nstead of the per-part size

  * [ ] make tests compile
  * [ ] make tests pass 

### DIFF
--- a/massa-bootstrap/src/bindings/server.rs
+++ b/massa-bootstrap/src/bindings/server.rs
@@ -42,6 +42,7 @@ struct ClientMessageLeader {
 
 /// Bootstrap server binder
 pub struct BootstrapServerBinder {
+    /// max number of block ids accepted in the client's cumulative bootstrap cursor
     max_consensus_block_ids: u64,
     thread_count: u8,
     max_datastore_key_length: u8,
@@ -73,7 +74,8 @@ impl BootstrapServerBinder {
             thread_count,
             max_datastore_key_length,
             randomness_size_bytes,
-            consensus_bootstrap_part_size,
+            consensus_bootstrap_part_size: _part_size,
+            max_consensus_block_ids,
             write_error_timeout,
         } = cfg;
 
@@ -82,7 +84,7 @@ impl BootstrapServerBinder {
         });
         let duplex = Limiter::new(duplex, limit_opts.clone(), limit_opts);
         BootstrapServerBinder {
-            max_consensus_block_ids: consensus_bootstrap_part_size,
+            max_consensus_block_ids,
             local_keypair,
             duplex,
             prev_message: None,
--- a/massa-bootstrap/src/client.rs
+++ b/massa-bootstrap/src/client.rs
@@ -65,7 +65,7 @@ impl BSConnector for DefaultConnector {
 /// This function will send the starting point to receive a stream of the ledger and will receive and process each part until receive a `BootstrapServerMessage::FinalStateFinished` message from the server.
 /// `next_bootstrap_message` passed as parameter must be `BootstrapClientMessage::AskFinalStatePart` enum variant.
 /// `next_bootstrap_message` will be updated after receiving each part so that in case of connection lost we can restart from the last message we processed.
-fn stream_final_state_and_consensus(
+pub(crate) fn stream_final_state_and_consensus(
     cfg: &BootstrapConfig,
     client: &mut BootstrapClientBinder,
     next_bootstrap_message: &mut BootstrapClientMessage,
@@ -150,8 +150,6 @@ fn stream_final_state_and_consensus(
                 }
                 BootstrapServerMessage::BootstrapFinished => {
                     info!("State bootstrap complete");
-                    // Set next bootstrap message
-                    *next_bootstrap_message = BootstrapClientMessage::AskBootstrapPeers;
 
                     // Update MIP store by reading from the disk
                     let mut guard = global_bootstrap_state.final_state.write();
@@ -162,6 +160,11 @@ fn stream_final_state_and_consensus(
                         .map_err(|e| BootstrapError::from(FinalStateError::from(e)))?;
 
                     warn_user_about_versioning_updates(updated, added);
+
+                    // Only advance to the next phase once the streamed state has been fully
+                    // post-processed: on failure the retry must replay the state streaming
+                    // instead of resuming from the peers phase with a half-updated store.
+                    *next_bootstrap_message = BootstrapClientMessage::AskBootstrapPeers;
 
                     return Ok(());
                 }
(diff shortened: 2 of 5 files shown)


Leo-Besancon (approved):
These fixes looks good to me!

PR #5170, opened 2026-08-18
5018 f79 grpc send operations underestimates smart contract operation gas accepting operations exceeding the block gas limit

### LINKED ISSUE #5018: [F79] gRPC send_operations Underestimates Smart-Contract Operation Gas, Accepting Operations Exceeding the Block Gas Limit

---

## Source
AI security-scan **Finding 79** (severity: **medium**). Tracked in `SUBMISSIONS.md` as F79.

**Affected file:** `massa-grpc/src/stream/send_operations.rs`

## Summary
The public gRPC submission path in `massa-main/massa-grpc/src/stream/send_operations.rs` validates smart-contract gas limits too loosely. For `OperationType::CallSC` and `OperationType::ExecuteSC`, it checks only the raw `max_gas` against `config.max_gas_per_block`, but actual block feasibility depends on the full gas computed by `get_gas_usage(base_operation_gas_cost, sp_compilation_cost)`. As a result, gRPC can accept signed operations that can never fit in a block. This creates inconsistent behavior with the JSON-RPC path and can lead to pool, storage, and network pollution.

## Root cause
The main issue is in `massa-main/massa-grpc/src/stream/send_operations.rs` inside the public gRPC `send_operations` path. For `OperationType::CallSC { max_gas, .. }` and `OperationType::ExecuteSC { max_gas, .. }`, the code rejects only when `max_gas > config.max_gas_per_block`, using `GrpcError::InvalidArgument("Gas limit of the operation is higher than the block gas limit. Your operation will never be included in a block.")`. This validation ignores mandatory gas overhead.

The real gas usage is computed elsewhere through `get_gas_usage(base_operation_gas_cost, sp_compilation_cost)`. According to that logic:
- `CallSC` uses `base_operation_gas_cost + max_gas`
- `ExecuteSC` uses `base_operation_gas_cost + sp_compilation_cost + max_gas`

Because `send_operations` compares only `max_gas` and does not use the full value from `get_gas_usage(base_operation_gas_cost, sp_compilation_cost)`, it underestimates true gas usage for both `CallSC` and `ExecuteSC`.

This creates an admission-policy mismatch between interfaces. The input states that the JSON-RPC path evaluates full gas usage and rejects these impossible-to-include operations, while the gRPC path accepts them.

## Affected code
```rust
match res_operation.content.op {
    OperationType::CallSC { max_gas, .. } | OperationType::ExecuteSC { max_gas, .. } => {
        if max_gas > config.max_gas_per_block {
            return Err(GrpcError::InvalidArgument("Gas limit of the operation is higher than the block gas limit. Your operation will never be included in a block.".into()));
        }
    },
    _ => {}
};
...
operation_storage.store_operations(verified_ops.values().cloned().collect());
pool_controller.add_operations(operation_storage.clone());
protocol_controller.propagate_operations(operation_storage)

pub fn get_gas_usage(&self, base_operation_gas_cost: u64, sp_compilation_cost: u64) -> u64 {
    match &self.content.op {
        OperationType::ExecuteSC { max_gas, .. } => max_gas.saturating_add(sp_compilation_cost),
        OperationType::RollBuy { .. } => 0,
        OperationType::RollSell { .. } => 0,
        OperationType::Transaction { .. } => 0,
    }
    .saturating_add(base_operation_gas_cost)
}
```

## Recommendation
Update the gRPC `send_operations` validation for `OperationType::CallSC` and `OperationType::ExecuteSC` to enforce the full operation gas usage rather than checking `max_gas` alone. The validation should use the same gas computation as block inclusion logic, including `base_operation_gas_cost` and, for `ExecuteSC`, `sp_compilation_cost`, and reject any operation whose total gas exceeds `config.max_gas_per_block`.

Align the gRPC admission rules with the JSON-RPC path so both interfaces apply the same includability criteria. If similar gas-limit checks exist elsewhere, centralize this validation in a shared helper to avoid future inconsistencies. Also ensure the rejection message clearly states that the operation’s total gas usage exceeds the block gas limit and therefore it can never be included in a block.

## Why P0 / breaking

peterjah:
  * [ ] make tests compile
  * [ ] make tests pass 

### DIFF
--- a/massa-api/src/public.rs
+++ b/massa-api/src/public.rs
@@ -46,7 +46,7 @@ use massa_models::{
     error::ModelsError,
     execution::EventFilter,
     node::NodeId,
-    operation::{OperationDeserializer, OperationId, OperationType, SecureShareOperation},
+    operation::{OperationDeserializer, OperationId, SecureShareOperation},
     output_event::SCOutputEvent,
     prehash::{PreHashMap, PreHashSet},
     secure_share::SecureShareDeserializer,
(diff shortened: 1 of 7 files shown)


Leo-Besancon (approved):
Looks great! Nice code reuse to avoid duplicating the code in both APIs

PR #5171, opened 2026-08-18
docs(execution): record that async gas intentionally absorbs unused b…

### LINKED ISSUE #5001: [F44] Async execution can exceed the configured per-slot gas cap

---

## Source
AI security-scan **Finding 44** (severity: **info**). Tracked in `SUBMISSIONS.md` as F44.

**Affected file:** `massa-execution-worker/src/execution.rs`

## Summary
In `massa-execution-worker/src/execution.rs`, `execute_slot` derives async execution capacity from both the async budget and unused block gas. Specifically, `async_msg_gas_available` is computed as `max_async_gas - deferred_calls_slot_gas + remaining_block_gas`, so on empty or underfilled blocks `take_async_batch` can execute async work well above the configured `MAX_ASYNC_GAS`. The confirmed issue is therefore a mismatch between the configured async cap and the actual async capacity available at runtime. The original claim that this creates an unbounded throttling bypass or increases total slot execution beyond the system budget is not supported.

## Root cause
The root cause is the formula used in `execute_slot` to compute `async_msg_gas_available` after block execution:

- `remaining_block_gas` is initialized to `self.config.max_gas_per_block` and decreases only as block operations consume gas.
- Async execution capacity is then set to `self.config.max_async_gas.saturating_sub(deferred_calls_slot_gas).saturating_add(remaining_block_gas)`.

As a result, any unused portion of `max_gas_per_block` is reassigned to async execution, allowing async messages to consume more than `MAX_ASYNC_GAS` in a given slot. The ruled-out portion of the original theory is that deferred-call quoting and booking against `MAX_ASYNC_GAS` creates an exploitable enforcement failure; per the validated analysis, that logic is merely conservative relative to the runtime's elastic capacity.

## Affected code
```rust
/// Executes a full slot (with or without a block inside) without causing any changes to the state,
    /// just yielding the execution output.
    ///
    /// # Arguments
    /// * `slot`: slot to execute
    /// * `exec_target`: metadata of the block to execute, if not miss
    /// * `selector`: Reference to the selector
    ///
    /// # Returns
    /// An `ExecutionOutput` structure summarizing the output of the executed slot
    pub fn execute_slot(
        &mut self,
        slot: &Slot,
        exec_target: Option<&(BlockId, ExecutionBlockMetadata)>,
        selector: Box<dyn SelectorController>,
    ) -> ExecutionOutput {
        // Start slot execution timer
        let slot_start_time = Instant::now();
        let mut slot_trace = SlotAbiCallStack {
            slot: *slot,
            operation_call_stacks: PreHashMap::default(),
            asc_call_stacks: vec![],
            deferred_call_stacks: HashMap::default(),
        };
        let mut transfers = vec![];

        // Create a new execution context for the whole active slot
        let mut execution_context = ExecutionContext::active_slot(
            self.config.clone(),
            *slot,
            exec_target.as_ref().map(|(b_id, _)| *b_id),
            self.final_state.clone(),
            self.active_history.clone(),
            self.module_cache.clone(),
            self.mip_store.clone(),
        );

        #[cfg(feature = "execution-info")]
        let mut exec_info = ExecutionInfoForSlot::new(
            slot.clone(),
            execution_context.execution_trail_hash.clone(),
            exec_target.as_ref().map(|(b_id, _)| *b_id),
        );
// … (truncated)
```

## Recommendation
Enforce `MAX_ASYNC_GAS` as a hard per-slot limit for async execution and avoid increasing async capacity with unused block gas unless that behavior is explicitly intended and documented. In particular, revise the `async_msg_gas_available` calculation so that async execution remains bounded by the configured async budget after accounting for deferred calls, rather than inheriting `remaining_block_gas`. If elastic reallocation between block and async execution is a desired design choice, document it clearly and update the related configuration, naming, and operator-facing guidance so `MAX_ASYNC_GAS` is not presented as a strict cap.

## Why P0 / breaking

peterjah:
…lock gas since MIP-0001

  * [ ] make tests compile
  * [ ] make tests pass 

### DIFF
--- a/massa-execution-worker/src/execution.rs
+++ b/massa-execution-worker/src/execution.rs
@@ -1599,7 +1599,7 @@ impl ExecutionState {
         // Block execution
 
         let mut block_info: Option<ExecutedBlockInfo> = None;
-        // Set block gas (max_gas_per_block - gas used by deferred calls)
+        // Set block gas (deferred calls have their own budget, see async gas below)
         let mut remaining_block_gas = self.config.max_gas_per_block;
 
         // Check if there is a block at this slot
@@ -1887,6 +1887,9 @@ impl ExecutionState {
 
         // Get asynchronous messages to execute
         // The gas available for async messages is the remaining block gas + async remaining gas (max_async - gas used by deferred calls)
+        // Reusing the gas the block left unused is intentional (activated by MIP-0001, MipComponent::Execution v1):
+        // it keeps the whole-slot budget bounded by max_gas_per_block + max_async_gas while letting
+        // underfilled slots do more async work. Changing this formula is a consensus change.
         let async_msg_gas_available = self
             .config
             .max_async_gas
--- a/massa-models/src/config/constants.rs
+++ b/massa-models/src/config/constants.rs
@@ -315,7 +315,11 @@ pub const PROTOCOL_EVENT_CHANNEL_SIZE: usize = 1024;
 
 /// Maximum of GAS allowed for a block
 pub const MAX_GAS_PER_BLOCK: u64 = u32::MAX as u64;
-/// Maximum of GAS allowed for asynchronous messages execution on one slot
+/// Base GAS budget for asynchronous messages execution on one slot.
+/// This is not a strict per-slot cap: since MIP-0001, `execute_slot` tops this budget
+/// up with the gas left unused by the block at that slot (and subtracts the gas booked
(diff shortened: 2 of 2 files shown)


Leo-Besancon (approved):
Good explanation, thanks! it was indeed deliberate as otherwise we could have queued async messages that won't be executed even though the "global" block time allotation has not been filled.

Issue #5127, opened 2026-07-23
Cleanup outdated code after breaking update release


peterjah:
TODO:
- Cleanup versioning related code introduced in MAIN.5.1 version in PR https://github.com/massalabs/massa/pull/5125
  - remove wmas issue mitigation code introduced by https://github.com/massalabs/massa/pull/5126
  - cleanup legacy autosell behaviour: https://github.com/massalabs/massa/pull/5159
  - cleanup legacy signatures for structs without chain_id: https://github.com/massalabs/massa/pull/5173
  - If the PR is merged and depending on the implementation chosen, clean up legacy ASC priority Ord or legacy compute_id behaviour (as well as the migration code from old ids to new ones): https://github.com/massalabs/massa/pull/5230
  - cleanup legacy deferred call slot index computation: https://github.com/massalabs/massa/pull/5245
- Update to clean MAIN.6.0 version

PR #5173, opened 2026-08-18
[F90] feat: include chain_id in signed block_header, endorsement and denunciation

### LINKED ISSUE #4979: [F90] Missing chain_id domain separation in Endorsement/BlockHeader signatures enables cross-chain replay & false slashing
> ⚠️ **Warning — This finding was routed to an issue rather than a PR because it changes what bytes are signed/verified.** Existing nodes would reject signatures produced under the new scheme (and vice-versa), so this is **consensus-breaking**. It can only ship as a coordinated, version-gated (MIP / hard-fork) rollout, never as a drop-in patch.

---

## Severity
**P0 / Critical.** Enables slashing of honest validators' rolls (loss of staked funds) with no misbehavior required. AI-report **Finding 90** (major); external bug-bounty submission **PDF #11** (Critical). Consensus-breaking fix (hard fork).

## Problem

`Operation` includes `chain_id` in its signed hash (EIP-155 style, to prevent cross-chain replay), but **`Endorsement` and `BlockHeader` do not**, and the denunciation verifier reconstructs the same `chain_id`-free hash. Because validators commonly run multiple Massa chains (e.g. MAINNET + a testnet like LABNET) with the same keypair, an attacker can take a validator's two *legitimate* same-`(slot, index)` endorsements (or block headers) from two different chains and submit them as a **denunciation** on either chain. Signature verification passes (the hashes are chain-agnostic), and the validator is slashed for entirely normal cross-chain operation.

The attacker needs no cooperation from the victim; normal endorsement/header production on both chains is sufficient.

## Affected code

Signing (no `chain_id`):
- `massa-models/src/endorsement.rs` — `Endorsement::compute_signed_hash` hashes only `public_key + EndorsementDenunciationData(slot, index) + content_hash`.
- `massa-models/src/block_header.rs` — `BlockHeader::compute_signed_hash` hashes only `public_key + BlockHeaderDenunciationData(slot) + content_hash`.

Denunciation verification (also no `chain_id`), which must mirror the signing layout:
- `massa-models/src/denunciation.rs` — `EndorsementDenunciation::compute_hash_for_sig_verif` (keyed on `slot`, `index`) and `BlockHeaderDenunciation::compute_hash_for_sig_verif` (keyed on `slot`).

Slashing path:
- `massa-execution-worker/src/execution.rs` — `execute_denunciation` (no `chain_id` check).

For reference, `Operation::compute_hash` in `massa-models/src/operation.rs` already folds in `chain_id`; `chain_id` constants live in `massa-models/src/config/constants.rs` (MAINNET=77658377, LABNET=77658376, ...).

## Fix (consensus rule change)

Include `chain_id` in the signed hash for `Endorsement` and `BlockHeader` (e.g. fold it into `EndorsementDenunciationData` / `BlockHeaderDenunciationData`), and mirror it in `compute_hash_for_sig_verif` so denunciations verify against the chain-scoped hash. Signing, verification, and denunciation reconstruction must all change together.

## ⚠️ Deployment — must NOT be a flag-day change

`compute_signed_hash` verifies **every** endorsement and block header on the network, not just denunciations. A naive change makes upgraded nodes reject un-upgraded nodes' endorsements/headers (and any block carrying a denunciation of them) as invalid signatures, and vice-versa — **partitioning consensus**, not just fixing the bug.

Deploy via the existing MIP network-versioning mechanism (`massa-versioning`, block header `current_version`/`announced_version`, `Defined → Started → LockedIn → Active`):
1. Define a new consensus-signature MIP (Block/consensus component).
2. Gate the hashing rule on the **network-agreed active version for the message's slot** (not the self-claimed header field): legacy `chain_id`-free layout before activation, `chain_id`-inclusive at/after the activation slot.
3. Flip signing + verification + denunciation reconstruction together at the deterministic activation slot so all nodes switch simultaneously.
4. Since denunciations have a bounded validity horizon, the legacy branch can be retired after `activation_slot + denunciation_expiry`.

## Acceptance criteria
- [ ] `chain_id` included in `Endorsement` and `BlockHeader` signed hashes, gated by an MIP-activated network version.
- [ ] Denunciation verification reconstructs the chain-scoped hash consistently.
- [ ] A denunciation combining endorsements/headers from two different `chain_id`s no longer verifies.
- [ ] Same-chain denunciations still work.
- [ ] Migration/versioning test proving no partition across the activation boundary.

_Tracked in SUBMISSIONS.md as F90 / PDF #11._

Leo-Besancon:
Note: I'm not sure if it's the simplest way to implement this.
Here, I pass in argument an `Option<u64>` representing an optional `CHAIN_ID` to put in the signed hash.
This means that we can pass `None` or `Some(chain_id)` depending on the active version, using the same API.

We could also duplicate the API (have two versions of the functions) and dispatch to one or the other depending on the version. This would add code duplication, but may be easier to cleanup in the future. What do you think @peterjah @damip?


peterjah:
Seems good to me. I don't see a better way

massa-api/src/public.rs
```diff
@@ -1431,7 +1431,7 @@ impl MassaRpcServer for API<Public> {
                         .into());
                     }
 
-                    let _verify_signature = match operation.verify_signature() {
+                    let _verify_signature = match operation.verify_signature(None) {
```
damip, line 1441: this one is hardcoded to none?

Leo-Besancon: Yes, Operation overrides compute_hash and folds chain_id into the content hash / OperationId.
This means that before this PR, the chain_id finds itself in the OperationID, and the OperationID is signed.
With this PR, we don't do that for Endorsement / BlockHeader / Denunciation, the IDs are still valid on multiple network. We just add chain_id to the hash that is signed in SecureShare.

massa-versioning/src/consensus_signature.rs
```diff
@@ -0,0 +1,220 @@
...
+///
+/// Every signing / verification / denunciation-reconstruction site must
+/// call this with the message's own slot timestamp so all nodes switch
+/// layouts simultaneously at the activation slot (no partition).
+pub fn sig_chain_id_for_slot(
+    mip_store: &MipStore,
+    chain_id: u64,
+    slot_ts: MassaTime,
+) -> Option<u64> {
+    if mip_store.get_latest_component_version_at(&MipComponent::Execution, slot_ts)
```
damip, line 56: **Activation-boundary determinism / spurious-ban risk.** `get_latest_component_version_at` first filters MIPs on the store's *current* in-memory state being `Active(_)`, and only then evaluates `state_at(ts)`:

```rust
// versioning.rs
.filter(|(mi, ms)| {
    mi.components.contains_key(component)
        && matches!(ms.state, ComponentState::Active(_))
})
.find_map(|(mi, ms)| {
    let res = ms.state_at(ts, mi.start, mi.timeout, mi.activation_delay);
    ...
```

`state_at(ts)` is deterministic given the same recorded history, but the outer filter is not a pure function of `ts`: a node whose in-memory MIP state hasn't advanced to `Active` yet (e.g. lagging in finalization across the activation instant, or catching up after being offline) will return the **legacy** layout for a slot that up-to-date peers treat as **scoped**. Since this helper now sits on the hot verification path and `note_endorsements_from_peer` **bans** peers on signature failure, a lagging node crossing the boundary could transiently ban honest peers (or be banned by them).

For synced nodes the window is small, but consider either (a) making the decision rely on `state_at(ts)` alone (drop the current-state pre-filter for this use), or (b) softening ban-on-invalid-signature in a window around the activation slot. At minimum this deserves a comment documenting the assumption that the local store state has advanced past the activation instant before messages for post-activation slots are verified.

massa-protocol-worker/src/handlers/endorsement_handler/retrieval.rs
```diff
@@ -196,13 +201,22 @@ pub(crate) fn note_endorsements_from_peer(
...
         &new_endorsements
             .values()
             .map(|endorsement| {
+                let slot_ts = get_block_slot_timestamp(
+                    config.thread_count,
+                    config.t0,
+                    config.genesis_timestamp,
+                    endorsement.content.slot,
+                )
+                .unwrap_or_else(|_| MassaTime::from_millis(0));
```
damip: **Inconsistent `get_block_slot_timestamp` error handling across sites.** Here (and in `block_factory.rs`, `endorsement_factory.rs`, `send_blocks.rs`, `send_endorsements.rs`, `denunciation_pool.rs`) a timestamp-computation failure silently falls back to `MassaTime::from_millis(0)` — i.e. the legacy layout — while `block_handler/retrieval.rs` and `execute_denunciation` propagate the error and reject the message.

The error is practically unreachable (arithmetic overflow on absurd slot values only), but the sites should agree on the policy, and rejecting is the safer default: a silent legacy fallback means a message with an overflow-inducing slot gets its signature checked against the wrong layout instead of being refused outright. Suggest propagating the error everywhere (or at least in all verification paths).

damip (changes requested):
## Review (F90 / #4979)

Reviewed the full diff against the acceptance criteria of #4979, verified every signing/verification/denunciation call site, and — since **CI has not run on this branch** — compiled the seven affected crates and ran the test suites locally (`massa_versioning` consensus_signature tests, `massa_models` denunciation/endorsement/block tests with `--features test-exports`): all green, including `test_endorsement_denunciation_cross_chain_rejected` and the migration/no-partition tests.

The implementation is correct on all acceptance criteria:
- `chain_id` folded into `Endorsement`/`BlockHeader` signed hashes, gated on the network-agreed `Execution` component version **at the message's own slot timestamp** (not the self-claimed header field) — matches the MIP-0002 definition already on `dev_breaking`.
- Legacy layout is byte-for-byte identical to pre-fix when `sig_chain_id = None`, so pre-activation signatures keep verifying.
- Denunciation pool and `execute_denunciation` both gate on the **denounced** slot — the correct slot.
- `Operation` and the block wrapper correctly keep their unchanged layouts; genesis (`None`) is consistent since genesis predates the MIP start.
- All production verification sites covered (factories, block handler, endorsement handler batch verify, gRPC `send_blocks`/`send_endorsements`, API); `is_also_for_*` has no production callers.

Requesting changes for the two inline points below (activation-boundary ban risk, and inconsistent timestamp-error handling). Also:

1. **Run CI before merge** — no checks have executed on this branch.
2. **Deployment coupling (conscious decision needed):** the signature-layout flip rides MIP-0002-BugFix together with the WMAS execution patch (`Execution` 2 + `FinalState` 2). Both activate at the same instant and cannot be rolled out separately. Fine if intentional — just flagging it.
3. **On the API question in the PR description** (single `Option<u64>` parameter vs duplicated functions): the `Option` approach is the right call. One code path, the compiler forces every call site to choose explicitly, and post-retirement cleanup (after `activation + denunciation_expiry`, as #4979 notes) is trivial: inline `Some(chain_id)` and delete the `None` branches — the `new_legacy` constructors are already staged for that.

massa-versioning/src/consensus_signature.rs
Leo-Besancon: Nice catch! Indeed, the prefilter could be relaxed to Active or LockedIn, as then the activation status can already be known at a later date.
As we plan on having an activation delay of a few days to a week, it solves the raised issue.
Note that I needed to update a few tests that relied on the fact that at least one finalized slot needed to be produced before the MIPs were taken into account.

massa-protocol-worker/src/handlers/endorsement_handler/retrieval.rs
Leo-Besancon: I've updated the error handling at all call sites of this PR, matching the behaviour of similar errors depending on call site.

bilboquet:
That one was hard to read ^^
It looks good to me and I'm about to approve it.

I'd like to propose an idea that could be treated as a follow up (it supposes that I have understood correctly):
Basically the `Option<u64>` that appears in many places is a binary choice (legacy vs chain-scope), so `Some(chain_id)` is redundant. Hence, the layout could live in the type so a type-state pattern would apply.

Today, every call must restate the layout:
```rs
Endorsement::new_verifiable(..., CHAINID, Some(CHAINID));   // redundant
header.verify_signature(sig_chain_id_for_slot(&mip_store, chain_id, slot_ts));
Operation::verify_signature(None);                           // what the None means depends on the type
let de = Denunciation::try_from((&e1, &e2, Some(CHAINID)))?;   // cross-chain check
assert!(de.is_valid_with_chain(Some(CHAINID)));                 // same layout, restated
```
With the type-state pattern, no argument should be needed
```rs
pub struct SecureShare<T, ID, Scope = Legacy>;
pub type SecureShareEndorsement = SecureShare<Endorsement, EndorsementId, ChainScoped>;
pub type SecureShareOperation   = SecureShare<Operation, OperationId, Legacy>;

let e = Endorsement::new_verifiable::<SecureShareEndorsement>(content, ser, &keypair, CHAINID);
e.verify_signature();                    // scope known from the type, no arg
// denunciation: scope is carried by the two message types, so TryFrom can't mix them
let de = Denunciation::try_from((&e1, &e2))?;   // scopes must match, compiler-checked
assert!(de.is_valid());                         // layout inferred from the type
```

That's a first thought, there is a possible caveat: for received messages the layout is known at runtime. But overall I think it could simplify the code. I'll be glad to ear your thoughts about this.

bilboquet (approved):

Leo-Besancon:
> That one was hard to read ^^ It looks good to me and I'm about to approve it.
> 
> I'd like to propose an idea that could be treated as a follow up (it supposes that I have understood correctly): Basically the `Option<u64>` that appears in many places is a binary choice (legacy vs chain-scope), so `Some(chain_id)` is redundant. Hence, the layout could live in the type so a type-state pattern would apply.
> 
> Today, every call must restate the layout:
> 
> ```rust
> Endorsement::new_verifiable(..., CHAINID, Some(CHAINID));   // redundant
> header.verify_signature(sig_chain_id_for_slot(&mip_store, chain_id, slot_ts));
> Operation::verify_signature(None);                           // what the None means depends on the type
> let de = Denunciation::try_from((&e1, &e2, Some(CHAINID)))?;   // cross-chain check
> assert!(de.is_valid_with_chain(Some(CHAINID)));                 // same layout, restated
> ```
> 
> With the type-state pattern, no argument should be needed
> 
> ```rust
> pub struct SecureShare<T, ID, Scope = Legacy>;
> pub type SecureShareEndorsement = SecureShare<Endorsement, EndorsementId, ChainScoped>;
> pub type SecureShareOperation   = SecureShare<Operation, OperationId, Legacy>;
> 
> let e = Endorsement::new_verifiable::<SecureShareEndorsement>(content, ser, &keypair, CHAINID);
> e.verify_signature();                    // scope known from the type, no arg
> // denunciation: scope is carried by the two message types, so TryFrom can't mix them
> let de = Denunciation::try_from((&e1, &e2))?;   // scopes must match, compiler-checked
> assert!(de.is_valid());                         // layout inferred from the type
> ```
> 
> That's a first thought, there is a possible caveat: for received messages the layout is known at runtime. But overall I think it could simplify the code. I'll be glad to ear your thoughts about this.

Thank you! I haven't thought too much about this, but seems like a good idea, I'll create a followup issue based on this (edit: https://github.com/massalabs/massa/issues/5127)
The only insight I have is that this should be probably studied / implemented after the versioning cleanup to make it easier to see the best design.

PR #5174, opened 2026-08-19
[F126] mark operations known by a peer only after the announcement wa…

### LINKED ISSUE #5072: [F126] Premature State Updates Before Confirmed Delivery Cause Silent Message and Command Loss

---

## Source
AI security-scan **Finding 126** (severity: **major**). Tracked in `SUBMISSIONS.md` as F126.

**Affected file:** `massa-protocol-worker/src/handlers/operation_handler/propagation.rs`

## Summary
Several paths in this codebase treat *non-blocking sends* and *best-effort propagation* as if delivery were guaranteed. In multiple places, the code updates local state, marks data as processed, or drops owned payloads *before* confirming that the message or action was actually delivered or executed. When bounded queues are full, peers disconnect, or network sends fail, this can cause *permanent message loss*, *peer state desynchronization*, *missed security actions*, *conflicting acknowledgments*, and even *loss of deferred credits*.

## Root cause
The common root cause is a *non-atomic state transition*: code removes, marks, or advances state first, then performs a non-blocking send or later execution step, and on failure it only logs or converts the error without restoring the original item or guaranteeing retry.

- In `massa-main/massa-protocol-worker/src/handlers/operation_handler/propagation.rs`, `PropagationThread::announce_ops` takes the full pending batch with `mem::take(&mut self.next_batch)` before delivery is guaranteed. It then computes `new_ops` with ``.filter(|id| ops.peek(&id.prefix()).is_none())`` and inserts each prefix into `cache_write.ops_known_by_peer[peer_id]` with `ops.insert(id.prefix(), ())` *before* calling `self.active_connections.send_to_peer(...)`. If `send_to_peer` fails, the code only logs `warn!("Failed to send OperationsAnnouncement message to peer: {}", err);`. It does not roll back `ops_known_by_peer`, and it does not reinsert unsent operation IDs into `self.next_batch`.

- The same function, `PropagationThread::announce_ops`, also marks *all* prefixes in `new_ops` as known before chunked delivery starts with `new_ops.chunks(self.config.max_operations_per_message as usize)`. This means a failure after only some `OperationsAnnouncement` chunks have been sent can still suppress the remaining chunks in later rounds, because future filtering uses `ops.peek(&id.prefix()).is_none()`.

- In `massa-main/massa-protocol-worker/src/handlers/endorsement_handler/propagation.rs`, `propagate_endorsements(&mut self, endorsements: Storage)` updates `cache_write.checked_endorsements` with `cache_write.checked_endorsements.insert(endorsement.id, ())` before reliable propagation is ensured. It sends endorsements only to peers connected at that moment, in chunks through `self.active_connections.send_ …

## Affected code
```rust
fn run(&mut self) {
        let mut batch_deadline = std::time::Instant::now()
            .checked_add(self.config.operation_announcement_interval.to_duration())
            .expect("Can't init interval op propagation");
        loop {
            match self.internal_receiver.recv_deadline(batch_deadline) {
                Ok(internal_message) => {
                    match internal_message {
                        OperationHandlerPropagationCommand::PropagateOperations(operations) => {
                            // Note operations as checked.
                            {
                                let mut cache_write = self.cache.write();
                                for op_id in operations.get_op_refs().iter().copied() {
                                    cache_write.insert_checked_operation(op_id);
                                }
                            }

                            // add to propagation storage
                            let new_ops = operations.get_op_refs().clone();
                            self.stored_for_propagation
                                .push_back((std::time::Instant::now(), new_ops.clone()));
                            self.op_storage.extend(operations);
                            self.prune_propagation_storage();

                            for op_id in new_ops {
                                self.next_batch.insert(op_id);
                                if self.next_batch.len()
                                    >= self.config.operation_announcement_buffer_capacity
                                {
                                    self.announce_ops();
                                    batch_deadline = std::time::Instant::now()
                                        .checked_add(
                                            self.config
                                                .operation_announcement_interval
                                                .to_duration(),
                                        )
                                        .expect("Can't init interval op propagation");
                                }
                            }
                        }
                        OperationHandlerPropagationCommand::Stop => {
                            info!("Stop operation propagation thread");
                            return;
                        }
                    }
// … (truncated)
```

## Recommendation
Refactor these flows so state changes occur only after delivery or execution is confirmed, or make the transition atomic by preserving the original item until success is guaranteed. Do not mark operations or endorsements as known, processed, acknowledged, removed, or finalized before the corresponding send, enqueue, or payout step has succeeded.

For peer propagation, update per-peer caches only for chunks that were actually sent successfully, and requeue any unsent items for later retry. Avoid taking ownership of the full pending batch unless failed sends can restore the remaining items. If partial chunk delivery is possible, track delivery at chunk granularity so later rounds do not suppress unsent data.

For bounded-channel paths using `try_send`, treat `Full` and disconnection as non-delivery rather than soft failures. Preserve the unsent payload, return it to the caller when possible, or place it into a retry buffer with explicit backpressure handling. Do not advance controller state, drop security-relevant commands, or continue with dependent processing when one of the required enqueue steps failed.

For multi-step pipelines, ensure all required downstream actions succeed before reporting success or committing visible state. In particular, avoid sending success responses after propagation failures, and avoid zeroing or removing deferred credits until the transfer has completed successfully. If rollback is not feasible, introduce durable retry semantics and surface the failure explicitly so the system cannot silently lose messages, commands, or funds.

## Why P0 / breaking

peterjah:
…s sent, and requeue unsent ones

  * [ ] make tests compile
  * [ ] make tests pass 

### DIFF
--- a/massa-protocol-worker/src/handlers/operation_handler/propagation.rs
+++ b/massa-protocol-worker/src/handlers/operation_handler/propagation.rs
@@ -10,7 +10,6 @@ use massa_models::prehash::CapacityAllocator;
 use massa_models::prehash::PreHashSet;
 use massa_protocol_exports::PeerId;
 use massa_protocol_exports::ProtocolConfig;
-use massa_protocol_exports::ProtocolError;
 use massa_storage::Storage;
 use tracing::{debug, info, log::warn};
 
@@ -200,6 +199,9 @@ impl PropagationThread {
         massa_trace!("protocol.protocol_worker.announce_ops.begin", {
             "operation_ids": operation_ids
         });
+        // Operations whose announcement could not be delivered to at least one peer.
+        // They are put back into the next batch so that a later round retries them.
+        let mut to_retry: PreHashSet<OperationId> = PreHashSet::default();
         {
             let mut cache_write = self.cache.write();
             let peers_connected = self.active_connections.get_peer_ids_connected();
@@ -207,7 +209,8 @@ impl PropagationThread {
 
             // Propagate to peers
             let all_keys: Vec<PeerId> = cache_write.ops_known_by_peer.keys().cloned().collect();
-            for peer_id in all_keys {
+            let chunk_size = self.config.max_operations_per_message as usize;
+            'peer_loop: for peer_id in all_keys {
                 let ops = cache_write.ops_known_by_peer.get_mut(&peer_id).unwrap();
                 let new_ops: Vec<OperationId> = operation_ids
                     .iter()
@@ -215,16 +218,12 @@ impl PropagationThread {
                     .copied()
                     .collect();
                 if !new_ops.is_empty() {
-                    for id in &new_ops {
-                        ops.insert(id.prefix(), ());
-                    }
                     debug!(
                         "Send operations announcement of len {} to {}",
                         new_ops.len(),
                         peer_id
                     );
-                    for sub_list in new_ops.chunks(self.config.max_operations_per_message as usize)
-                    {
+                    for (chunk_index, sub_list) in new_ops.chunks(chunk_size).enumerate() {
                         if let Err(err) = self.active_connections.send_to_peer(
                             &peer_id,
                             &self.operation_message_serializer,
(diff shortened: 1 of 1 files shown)


massa-protocol-worker/src/handlers/operation_handler/propagation.rs
```diff
@@ -174,31 +173,31 @@ impl PropagationThread {
         massa_trace!("protocol.protocol_worker.announce_ops.begin", {
             "operation_ids": operation_ids
         });
+        // Operations whose announcement could not be delivered to at least one peer.
```
Leo-Besancon, line 202: It adds a bit of complexity and overhead, but would it make sense to track which peers failed to receive them too?

If we have 20 peers, I assume maybe one or two would fail to receive the announcement most of the time, so retrying to all the peers may be a bit spammy.

Leo-Besancon (commented):
Left a comment, but otherwise, the linked issue also talks about endorsement propagation, is there not the same issue?

massa-protocol-worker/src/handlers/operation_handler/propagation.rs
peterjah: "Track which peers failed, retrying all peers may be spammy"

Already handled — the retry is not spammy. The requeue only puts op ids back into next_batch; the per-peer fan-out still runs through the existing filter at propagation.rs:184-188:
```
let new_ops: Vec<OperationId> = operation_ids
    .iter()
    .filter(|id| ops.peek(&id.prefix()).is_none())
```

peterjah: "Does endorsement propagation have the same issue?"

No — it was already correct. endorsement_handler/propagation.rs:118-136 sends the chunk first and only then does peer_knowledge.insert(endorsement.id, ()), with continue 'peer_loop on error. git blame dates that to d935fc8a71 (Jul 2023), so it predates this finding; the operation-handler fix brings ops in line with the endorsement handler, not the reverse. Block propagation (block_handler/propagation.rs:197-212) is also correct — it inserts into blocks_known_by_peer in the Ok(()) arm only.

Leo-Besancon: great, thanks for the explanations

Leo-Besancon (dismissed):
LGTM

Leo-Besancon (approved):

PR #5175, opened 2026-08-19
[F130] gate operation dedup cache on successful local retention

### LINKED ISSUE #5073: [F130] Premature Operation Deduplication Can Blackhole Valid Operations Before Local Retention

---

## Source
AI security-scan **Finding 130** (severity: **medium**). Tracked in `SUBMISSIONS.md` as F130.

**Affected file:** `massa-protocol-worker/src/handlers/operation_handler/retrieval.rs`

## Summary
In `massa-main/massa-protocol-worker/src/handlers/operation_handler/retrieval.rs`, `note_operations_from_peer(...)` marks valid operations as already processed *before* it knows they were actually kept locally. It inserts operation IDs into `checked_operations` and records prefixes with `insert_peer_known_ops(...)` right after `verify_sigs_batch(...)`, but local retention happens only later through a temporary `Storage` created with `base_storage.clone_without_refs()`. If propagation or pool handling does not retain that `Storage`, the operation can be lost while the caches still say it was handled. This can make the node ignore later announcements or re-deliveries of a valid operation until the cache entry is evicted.

## Root cause
The main root cause is in `massa-main/massa-protocol-worker/src/handlers/operation_handler/retrieval.rs` inside `note_operations_from_peer(...)`. After the function filters operations and successfully runs `verify_sigs_batch(...)`, it updates shared caches too early:
- it inserts each operation ID into `checked_operations` with `insert_checked_operation(...)`,
- and it marks the peer as knowing the operation prefixes with `insert_peer_known_ops(...)`.

This cache update happens *before* the function confirms that the operation was actually retained locally. The only local holding created afterward is a temporary `Storage`:
- `let mut ops = base_storage.clone_without_refs();`
- `ops.store_operations(...)`
- `ops_propagation_sender.try_send(OperationHandlerPropagationCommand::PropagateOperations(ops.clone()))`
- `pool_controller.add_operations(ops)`

Because `ops` is temporary, the operation survives only if some downstream component keeps a reference or admits it. If that does not happen, the `Storage` can be dropped when `note_operations_from_peer(...)` exits, even though the caches already mark the operation as known.

A related root cause is that downstream handoff failures do not prevent this inconsistent state. In particular, if `ops_propagation_sender.try_send(...)` fails, the function only logs `warn!("Error sending operations to propagation channel")` and still returns `Ok(())`. The input also states that pool-side handling may be `Full`, `Disconnected`, or may later drop the operation due to capacity or excess filtering. In these cases, the cache state can still indicate successful processing although no surviving local reference exists.

This becomes visible across module boundaries because later logic consults these caches:
- `checked_operations` is used to sk …

## Affected code
```rust
pub(crate) fn note_operations_from_peer(
    base_storage: &Storage,
    operations_cache: &mut SharedOperationCache,
    config: &ProtocolConfig,
    source_peer_id: &PeerId,
    ops_propagation_sender: &mut MassaSender<OperationHandlerPropagationCommand>,
    pool_controller: &mut Box<dyn PoolController>,
) -> Result<(), ProtocolError> {
    massa_trace!("protocol.protocol_worker.note_operations_from_peer", { "peer": source_peer_id, "operations": operations });
    let now = MassaTime::now();

    let mut new_operations = PreHashMap::with_capacity(operations.len());
    for operation in operations {
        // ignore if op is too old
        let expire_period_timestamp = get_block_slot_timestamp(
            config.thread_count,
            config.t0,
            config.genesis_timestamp,
            Slot::new(
                operation.content.expire_period,
                operation
                    .content_creator_address
                    .get_thread(config.thread_count),
            ),
        );
        match expire_period_timestamp {
            Ok(slot_timestamp) => {
                if slot_timestamp.saturating_add(config.max_operations_propagation_time) < now {
                    continue;
                }
            }
            Err(_) => continue,
        }

        // quit if op is too big
        if operation.serialized_size() > config.max_serialized_operations_size_per_block {
            return Err(ProtocolError::InvalidOperationError(format!(
                "Operation {} exceeds max block size,  maximum authorized {} bytes but found {} bytes",
                operation.id,
                operation.serialized_size(),
                config.max_serialized_operations_size_per_block
            )));
        };

// … (truncated)
```

## Recommendation
Update `note_operations_from_peer(...)` so that `checked_operations` and `insert_peer_known_ops(...)` are updated only after the operation is durably retained by at least one local component. Treat signature verification as necessary but not sufficient for deduplication.

Reorder the flow to first confirm successful admission or ownership transfer to a persistent local holder, such as storage, propagation, or pool handling, and only then mark the operation as processed or known. If retention cannot be guaranteed, do not populate deduplication or peer-known caches.

Also ensure downstream handoff failures are reflected in the function outcome instead of only being logged. If `ops_propagation_sender.try_send(...)` fails, or if pool admission does not guarantee continued retention, keep the operation eligible for future announcements or re-delivery rather than recording it as already handled.

More generally, define a clear invariant: an operation may be inserted into `checked_operations` and peer-known prefix tracking only once the node can prove it still has a live local reference or another durable guarantee of availability. This prevents valid operations from being blackholed during temporary backpressure, channel saturation, or pool-side dropping.

## Why P0 / breaking

peterjah:
  * [ ] make tests compile
  * [ ] make tests pass 

### DIFF
--- a/massa-protocol-worker/src/handlers/operation_handler/retrieval.rs
+++ b/massa-protocol-worker/src/handlers/operation_handler/retrieval.rs
@@ -439,15 +439,9 @@ pub(crate) fn note_operations_from_peer(
     )?;
 
     {
-        // add to checked operations
+        // mark the sender as knowing the ops it sent us:
+        // this holds regardless of whether we end up retaining them locally
         let mut cache_write = operations_cache.write();
-
-        // add checked operations
-        for op_id in new_operations.keys().copied() {
-            cache_write.insert_checked_operation(op_id);
-        }
-
-        // add to known ops
         cache_write.insert_peer_known_ops(
             source_peer_id,
             &all_received_ids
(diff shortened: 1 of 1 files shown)


Leo-Besancon (approved):
LGTM

PR #5176, opened 2026-08-19
[F106] ban only peers that actually sent us a block, not everyone the…

### LINKED ISSUE #5069: [F106] Improper Peer Banning Based on Mutable Propagation Cache Enables Ban Evasion and False Positives

---

## Source
AI security-scan **Finding 106** (severity: **major**). Tracked in `SUBMISSIONS.md` as F106.

**Affected file:** `massa-protocol-worker/src/handlers/block_handler/propagation.rs`

## Summary
In `massa-main/massa-protocol-worker/src/handlers/block_handler/propagation.rs`, the ban flow in `PropagationThread::run()` uses `blocks_known_by_peer` to decide which peers to ban when `BlockHandlerPropagationCommand::AttackBlockDetected(block_id)` arrives. This is unsafe because `blocks_known_by_peer` is a mutable propagation cache, not a reliable record of who created or sent the malicious block. As a result, the real attacker may avoid banning if the cache entry disappears before the check, while honest peers may be banned only because they relayed, received, or requested the block. This can enable repeated malicious submissions and also cause collateral peer isolation.

## Root cause
The main issue is that `PropagationThread::run()` in `massa-main/massa-protocol-worker/src/handlers/block_handler/propagation.rs` treats `self.cache.read().blocks_known_by_peer` as the source of truth for a permanent security action. When handling `BlockHandlerPropagationCommand::AttackBlockDetected(block_id)`, it scans `blocks_known_by_peer.iter()`, checks `knowledge.peek(&block_id)`, builds `peers_to_ban`, and passes that list to `self.ban_peers(&peers_to_ban)`. The cache only records whether a peer is currently marked as knowing `block_id`. It does not reliably identify the peer that originated the malicious block.

A separate but related problem is a *time-of-check/time-of-use* gap. The peer-to-block association is checked only after asynchronous validation has finished, but `blocks_known_by_peer` can change before that happens.

- In `PropagationThread::perform_propagations()`, the code calls `let peers_connected = self.active_connections.get_peer_ids_connected();`, then `let mut cache_lock = self.cache.write();`, then `cache_lock.update_cache(&peers_connected);`. This can remove disconnected peers from the cache entirely.
- The per-peer knowledge store is bounded and described as LRU-style. Later header or block knowledge insertions can evict the earlier malicious `block_id`, so `knowledge.peek(&block_id)` may return `None` by the time `AttackBlockDetected(block_id)` is processed.
- `perform_propagations()` runs during normal operation, including on `RecvTimeoutError::Timeout` driven by `config.block_propagation_tick`, and also after `BlockHandlerPropagationCommand::IntegratedBlock { block_id, storage }`. This means pruning and cache mutation can happen before the later attack decision.

The same cache also causes false attribution because its meaning is too broad …

## Affected code
```rust
pub fn start_propagation_thread(
    active_connections: Box<dyn ActiveConnectionsTrait>,
    receiver: MassaReceiver<BlockHandlerPropagationCommand>,
    peer_cmd_sender: MassaSender<PeerManagementCmd>,
    config: ProtocolConfig,
    cache: SharedBlockCache,
) -> JoinHandle<()> {
    std::thread::Builder::new()
        .name(THREAD_NAME.to_string())
        .spawn(move || {
            let block_serializer = MessagesSerializer::new()
                .with_block_message_serializer(BlockMessageSerializer::new());
            let mut propagation_thread = PropagationThread {
                stored_for_propagation: LruMap::new(ByLength::new(
                    config
                        .max_blocks_kept_for_propagation
                        .try_into()
                        .expect("max_blocks_kept_for_propagation does not fit in u32"),
                )),
                receiver,
                config,
                cache,
                peer_cmd_sender,
                active_connections,
                block_serializer,
            };
            propagation_thread.run();
        })
        .expect("OS failed to start block propagation thread")
}
```

## Recommendation
Do not derive banning decisions from `blocks_known_by_peer`. Treat this cache only as propagation state, not as evidence of authorship or malicious origin. Instead, maintain a dedicated and durable attribution record at the moment a block or header is first received, preserving the peer identity that directly supplied the data and, if needed, the full forwarding path for later investigation. Use that immutable reception metadata as the sole basis for punitive actions when `BlockHandlerPropagationCommand::AttackBlockDetected(block_id)` is emitted.

Ensure the attribution record is not affected by propagation-cache updates, peer disconnections, LRU eviction, or other routine pruning, so the later ban decision cannot be bypassed through cache mutation. Also distinguish clearly between peers that originated a malicious item and peers that merely received, requested, or relayed it. Only ban peers for whom there is direct evidence of malicious submission, and avoid penalizing peers based solely on having the block marked as known.

If immediate attribution is not available with sufficient confidence, prefer discarding the block, recording the event, and increasing monitoring or scoring rather than issuing a permanent ban from ambiguous propagation data.

## Why P0 / breaking

peterjah:
… propagation cache marks as knowing it

  * [ ] make tests compile
  * [ ] make tests pass 

### DIFF
--- a/massa-protocol-worker/src/handlers/block_handler/cache.rs
+++ b/massa-protocol-worker/src/handlers/block_handler/cache.rs
@@ -9,6 +9,10 @@ use massa_protocol_exports::PeerId;
 use parking_lot::RwLock;
 use schnellru::{ByLength, LruMap};
 
+/// Maximum number of peers recorded as direct senders of a given block.
+/// Bounds the memory used by the attribution record in case many peers send us the same block.
+const MAX_SENDERS_PER_BLOCK: usize = 32;
+
 /// Cache on block knowledge by our node and its peers
 pub struct BlockCache {
     /// cache of previously checked headers
@@ -17,6 +21,13 @@ pub struct BlockCache {
     pub blocks_known_by_peer: HashMap<PeerId, LruMap<BlockId, (bool, Instant)>>,
     /// max number of blocks known in peer knowledge cache
     pub max_known_blocks_by_peer: u32,
+    /// For each recently seen block, the peers that directly sent us data for it.
+    ///
+    /// This is attribution data, not propagation state: unlike `blocks_known_by_peer` it only
+    /// records peers from which we actually received the block's header or contents, it never
+    /// records peers that merely learned about the block from us, and it is not pruned when a
+    /// peer disconnects. It is therefore the only sound basis for punitive actions.
+    pub block_senders: LruMap<BlockId, HashSet<PeerId>>,
 }
 
 impl BlockCache {
@@ -43,6 +54,31 @@ impl BlockCache {
             known_blocks.insert(*block_id, (known, now));
         }
     }
+
+    /// Record that a peer directly sent us data (header or contents) for a given block.
+    ///
+    /// # Arguments
+    ///
+    /// * `block_id` - The block the data belongs to
+    /// * `from_peer_id` - The peer that sent us the data
+    pub fn insert_block_sender(&mut self, block_id: &BlockId, from_peer_id: &PeerId) {
+        let senders = self
+            .block_senders
+            .get_or_insert(*block_id, HashSet::new)
+            .expect("failed to insert block senders entry");
+        // bound the number of recorded senders per block
+        if senders.len() < MAX_SENDERS_PER_BLOCK {
+            senders.insert(*from_peer_id);
+        }
+    }
+
+    /// Get the peers that directly sent us data for a given block.
+    pub fn get_block_senders(&self, block_id: &BlockId) -> Vec<PeerId> {
+        self.block_senders
+            .peek(block_id)
+            .map(|senders| senders.iter().copied().collect())
+            .unwrap_or_default()
+    }
 }
 
 impl BlockCache {
@@ -51,6 +87,7 @@ impl BlockCache {
             checked_headers: LruMap::new(ByLength::new(max_known_blocks)),
             blocks_known_by_peer: HashMap::new(),
             max_known_blocks_by_peer,
+            block_senders: LruMap::new(ByLength::new(max_known_blocks)),
         }
     }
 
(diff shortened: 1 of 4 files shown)


Leo-Besancon (commented):
I've always thought that the goal of this was not just to ban the bad block's creator or the peer that sent it to us, but to punish all the peers that propagated a bad block (as a genuine node would have checked the block before propagating it).

If we want to keep this behaviour, we could do an alternative fix to your PR, we could split blocks_known_by_peer and blocks_headers_known_by_peer.

Sometimes we wrongly infer that:

- If a peer knows a block header it knows the full block
- If a peer knows a block header it knows all the parents too (as the full block)

If we check that we don't send bad blocks (likely), I think it means we would only need to make sure that we don't ban a peer that knows the header of a bad full block. (However, we can keep banning them if the header itself is bad).

Does this make sense?

I'm not saying that my alternative implem would be better, just that it may preserve more of the current (wanted?) behaviour.

peterjah:
> I've always thought that the goal of this was not just to ban the bad block's creator or the peer that sent it to us, but to punish all the peers that propagated a bad block (as a genuine node would have checked the block before propagating it).
> 
> If we want to keep this behaviour, we could do an alternative fix to your PR, we could split blocks_known_by_peer and blocks_headers_known_by_peer.
> 
> Sometimes we wrongly infer that:
> 
> * If a peer knows a block header it knows the full block
> * If a peer knows a block header it knows all the parents too (as the full block)
> 
> If we check that we don't send bad blocks (likely), I think it means we would only need to make sure that we don't ban a peer that knows the header of a bad full block. (However, we can keep banning them if the header itself is bad).
> 
> Does this make sense?
> 
> I'm not saying that my alternative implem would be better, just that it may preserve more of the current (wanted?) behaviour.

your criterion — don't ban a peer that only knows the *header* of a
  bad *full block* — is a real gap in my PR, and I'd like to fold it in:
  `mark_block_as_invalid` fires on a contents-level defect (oversized op set,
  `retrieval.rs:1202`), which a header-only sender couldn't have detected. I'll
  tag each `block_senders` entry with what the peer actually sent (header vs
  contents) and:

  - `AttackBlockDetected` (header-level: all `DiscardReason::Invalid` in
    `verifications.rs` come from `check_header`) → ban header *and* contents
    senders;
  - `mark_block_as_invalid` (contents-level) → ban contents senders only.

Leo-Besancon (approved):
Looks better now, thank you!

PR #5177, opened 2026-08-19
[F114] bound the per-round endorsement propagation drain with a confi…

### LINKED ISSUE #5107: [F114] Unbounded endorsement queue draining enables memory/CPU exhaustion

---

**Severity:** medium (resource exhaustion / DoS)
**Component:** `massa-protocol-worker/src/handlers/endorsement_handler/propagation.rs`

## Summary

The endorsement propagation `run` loop drains the propagation channel by repeatedly
calling `try_recv()` and merges **all** pending `PropagateEndorsements` commands into a
single batch before sending anything. There is no bound on how many commands /
endorsements accumulate per round. `propagate_endorsements` then clones the full batch
into a `Vec` and builds a per-peer `to_send` vector before chunking. A sustained
source (e.g. the public `send_endorsements` stream) can force large allocations and
long CPU-heavy propagation rounds, delaying propagation and potentially exhausting
memory.

## Root cause

`run` merges every queued `PropagateEndorsements` into one batch with no cap on the
number of endorsements collected, and `propagate_endorsements` cost scales with both
batch size and peer count.

The fix requires **choosing a bound** (max endorsements or max drain iterations per
round) and deciding how to handle the overflow (defer remaining to the next loop
iteration vs. drop). There is no existing config field that obviously represents this
per-round propagation budget (`max_endorsements_per_message` only sizes wire chunks),
so this needs a design decision and likely a new `ProtocolConfig` field — a judgment
call best made by maintainers rather than an arbitrary constant.

- Cap the per-round drain (e.g. stop merging once a configurable endorsement budget is
  reached) and re-loop to process the remainder, keeping propagation responsive and
  memory bounded.
- Add a `ProtocolConfig` field for the budget with a sensible default.
- Unit-test the bounded-drain helper.

## Refs

Findings F114.

peterjah:
…gurable budget

  * [ ] make tests compile
  * [ ] make tests pass 

### DIFF
--- a/massa-node/base_config/config.toml
+++ b/massa-node/base_config/config.toml
@@ -347,6 +347,8 @@
     timeout_connection = 1000
     # max number of operations kept for propagation
     max_ops_kept_for_propagation = 320000
+    # max number of endorsements merged into a single propagation round
+    max_endorsements_per_propagation_round = 10000
     # time threshold after which operation are not propagated
     max_operations_propagation_time = 32000
     # time threshold after which endorsement are not propagated
--- a/massa-node/base_config/config.toml
+++ b/massa-node/base_config/config.toml
@@ -348,7 +348,7 @@
     # max number of operations kept for propagation
     max_ops_kept_for_propagation = 320000
     # max number of endorsements merged into a single propagation round
-    max_endorsements_per_propagation_round = 10000
+    max_endorsements_per_propagation_round = 1024
     # time threshold after which operation are not propagated
     max_operations_propagation_time = 32000
     # time threshold after which endorsement are not propagated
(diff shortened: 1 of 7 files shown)


massa-node/base_config/config.toml
```diff
@@ -347,6 +347,8 @@
     timeout_connection = 1000
     # max number of operations kept for propagation
     max_ops_kept_for_propagation = 320000
+    # max number of endorsements merged into a single propagation round
+    max_endorsements_per_propagation_round = 10000
```
Leo-Besancon: Would having a lower default config make more sense? 1 period should produce at most 512 endorsement so I thought that a config of e.g. 1024 would be enough.

peterjah: done

Leo-Besancon (approved):
LGTM

PR #5178, opened 2026-08-19
[PDF8] fix endorsement pool index desync and bound conflicting endors…

### LINKED ISSUE #5079: [PDF#8] Endorsement-pool index/storage desync leaks memory and can panic a producer node

---

## Source
External bug-bounty submission **PDF #8** (salmanmeizar@gmail.com, severity Medium). **Not covered by the AI scan** — `massa-pool-worker/src/endorsement_pool.rs` was not scanned. Tracked in `SUBMISSIONS.md` as PDF #8 / B4.

## Summary
The endorsement pool keeps two structures keyed by `(slot, index, endorsed_block)`: `endorsements_indexed: HashMap<...>` and `endorsements_sorted: Vec<BTreeMap<...>>`. The per-thread size-cap prune pops from `endorsements_sorted` only and drops the key, orphaning the matching `endorsements_indexed` entry permanently. Finalization cleanup only removes indexed keys still present in sorted, so orphans are never reclaimed.

Endorsement ingress validates only signature and the PoS draw on `(slot, index)` and never constrains `endorsed_block` (`massa-protocol-worker/src/handlers/endorsement_handler/retrieval.rs`), so a drawn endorser can mint unlimited valid endorsements differing only in `endorsed_block`, exceed the per-thread cap, and grow `endorsements_indexed` without bound on a victim producer. A later `get_block_endorsements` lookup of an orphaned id whose storage was dropped can panic, halting the pool worker.

## Affected code (`./massa`, verified)
- `massa-pool-worker/src/endorsement_pool.rs` — `endorsements_indexed` / `endorsements_sorted`; prune uses `pop_first()` + `.expect(...)`; finalization cleanup keyed on `endorsements_sorted`.
- Ingress: `massa-protocol-worker/src/handlers/endorsement_handler/retrieval.rs`.

## Precondition
Attacker must be a drawn endorser for a slot where the victim is the block producer (bounds the attacker set → Medium).

## Why P0 / needs design discussion
Constraining accepted `endorsed_block` values touches endorsement-acceptance semantics (consensus-adjacent) and the fix must keep the two structures consistent under prune/finalize without changing valid endorsement handling. Related to AI-report **F118** (conflicting endorsements with arbitrary endorsed blocks accepted/gossiped).

## Acceptance criteria
- [ ] Prune removes entries from **both** structures atomically (no orphans).
- [ ] `get_block_endorsements` cannot panic on a missing/dropped id.
- [ ] Bound the number of accepted endorsements per `(slot, index)` regardless of `endorsed_block`.
- [ ] Regression test for the prune/finalize desync + the panic path.

peterjah:
…ements per draw

  * [ ] make tests compile
  * [ ] make tests pass 

### DIFF
--- a/massa-pool-worker/src/endorsement_pool.rs
+++ b/massa-pool-worker/src/endorsement_pool.rs
@@ -3,7 +3,7 @@
 use massa_models::{
     block_id::BlockId,
     endorsement::EndorsementId,
-    prehash::{CapacityAllocator, PreHashSet},
+    prehash::{CapacityAllocator, PreHashMap, PreHashSet},
     slot::Slot,
 };
 use massa_pool_exports::{PoolChannels, PoolConfig};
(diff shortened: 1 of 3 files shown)


massa-pool-worker/src/endorsement_pool.rs
```diff
@@ -3,30 +3,36 @@
...
     sync::Arc,
 };
 use tracing::{trace, warn};
 
+/// Maximum number of endorsements kept for a given `(slot, index)` pair, whatever the endorsed
+/// block is. Only one endorsement per `(slot, index)` can ever end up in a block, and an honest
+/// endorser signs exactly one. The endorsed block is not constrained at ingress, so without this
+/// bound an equivocating drawn endorser could mint arbitrarily many valid endorsements differing
+/// only by their endorsed block and grow the pool without limit.
+const MAX_ENDORSEMENTS_PER_SLOT_INDEX: usize = 1;
```
Leo-Besancon: If there is a speculative rollback, couldn't we have multiple valid `endorsed_blocks` for a given `(slot, index)`? I'm not sure if we clear the pool there or not, but I don't think so?

If that's the case, we could increase the bound a bit (to something like maybe between 4 and 16? wdyt?)

Leo-Besancon (commented):
Left a comment for the choice of the bounding constant.
Otherwise, the fix looks good, but I don't have an opinion on the implementation choice to unify `endorsement_indexed` with `endorsement_sorted`, as I'm not sur why it was built this way.

massa-pool-worker/src/endorsement_pool.rs
peterjah: Good catch on questioning the constant, though I think the rollback mechanism doesn't apply: the endorsement factory walks slots strictly monotonically (get_next_slot forces next_slot > prev_slot) and calls process_slot once per slot, so an honest endorser signs exactly one endorsement per (slot, index) and never re-endorses after a reorg. Since index binds to a single drawn address, >1 entry per (slot, index) means that address equivocated.

 That said, I agree the bound should be >1, for a different reason. get_block_endorsements looks up by our get_best_parents()[thread], so an endorsement is only usable if its endorsed_block matches the parent we pick. With MAX = 1 the map is first-write-wins, so an equivocating drawn endorser can deliberately send us the wrong-block variant first and permanently occupy the slot — denying our block an endorsement, which costs fitness (endorsements.len() + 1) and rewards. Under a genuine split, an equivocator hedging both tips signs both, and we want to keep whichever matches the tip we build on.

Leo-Besancon: Ok that make sense, thank you.
I don't think the scenario is very likely volontarily (if the block producer denies us from an endorsement he also gets less rewards), but I agree having the bound to 4 seems ok!

Leo-Besancon (dismissed):
LGTM

Leo-Besancon (approved):

PR #5179, opened 2026-08-19
[F144] document that massa-sdk's certificate_store setting is ignored…

### LINKED ISSUE #5093: [F144] SDK certificate-store setting is ignored while client constructors force plaintext transport

---

## Source
AI security-scan **Finding 144** (severity: **minor**). Tracked in `SUBMISSIONS.md` as F144.

**Affected file:** `massa-sdk/src/lib.rs`

## Summary
`massa-sdk` exposes a `certificate_store` setting in its HTTP and WebSocket configuration, but the built-in client constructors do not honor that setting. In `massa-main/massa-sdk/src/lib.rs`, `Client::new` and `RpcClientV2::from_url` hard-code plaintext `http://`, `ws://`, and `grpc://` endpoints, while the `certificate_store` selection logic in `http_client_from_url` and `ws_client_from_url` is commented out and explicitly unused. This confirms a real issue: the SDK presents transport-related configuration that is ineffective on these client paths. The original claim is overstated to the extent it frames this as a demonstrated certificate-based gRPC client trust failure driven by `gen_signed_cert`.

## Recommendation
Remove or clearly deprecate the exposed `certificate_store` setting on client construction paths where it is not applied, and align the SDK surface with the actual transport behavior. If secure transport is intended, update the built-in constructors to support explicit TLS-enabled endpoint schemes and to consistently honor the configured trust-store selection for HTTP and WebSocket clients. If plaintext transport is the intended behavior for these paths, document that behavior explicitly and avoid presenting certificate-related options that suggest otherwise. Ensure the transport configuration is applied consistently across `Client::new`, `RpcClientV2::from_url`, `http_client_from_url`, and `ws_client_from_url` so developers cannot mistakenly rely on ineffective security settings.

## Why this is an issue and not a direct PR
Making the exposed `certificate_store` setting effective requires **implementing TLS transport selection** in the SDK client constructors (`Client::new`, `RpcClientV2::from_url`) and switching hard-coded `http://` / `ws://` / `grpc://` schemes to TLS-capable ones. That is a feature with an **observable client-transport behavior change** (and potential tooling impact), so it needs design discussion rather than a quick robustness PR. See `SUBMISSIONS.md` Part D.

peterjah:
… by the plaintext client constructors

  * [ ] make tests compile
  * [ ] make tests pass 

### DIFF
--- a/massa-client/base_config/config.toml
+++ b/massa-client/base_config/config.toml
@@ -23,6 +23,8 @@ chain_id = 77658377
     # maximum number of outcoming connections allowed
     max_concurrent_requests = 100
     # certificate_store, `Native` or `WebPki`
+    # deprecated: this setting is currently ignored, the client connects to the node over
+    # plaintext http/ws/grpc. It is kept to avoid breaking existing configuration files.
     certificate_store = "Native"
     # JSON-RPC request object id data type, `String` or `Number`
     id_kind = "Number"
--- a/massa-sdk/src/config.rs
+++ b/massa-sdk/src/config.rs
@@ -13,6 +13,11 @@ pub struct ClientConfig {
     /// maximum concurrent requests.
     pub max_concurrent_requests: usize,
     /// certificate_store, `Native` or `WebPki`
+    ///
+    /// Deprecated: this setting is currently ignored. The underlying `jsonrpsee` client no
+    /// longer exposes a trust-store selection, and TLS is instead enabled by using an
+    /// `https`/`wss` url, validated against the platform trust store. It is kept to avoid
+    /// breaking existing configuration files.
     pub certificate_store: String,
     /// JSON-RPC request object id data type, `Number` or `String`
     pub id_kind: String,
--- a/massa-sdk/src/lib.rs
+++ b/massa-sdk/src/lib.rs
@@ -82,6 +82,11 @@ pub struct Client {
 
 impl Client {
     /// creates a new client
+    ///
+    /// Note: this constructor always uses plaintext transports (`http` for the JSON-RPC
+    /// APIs, `grpc` for the gRPC APIs) and cannot connect to a node whose gRPC server has
+    /// `enable_tls` set. For a TLS-enabled JSON-RPC endpoint, build the client with
+    /// [`RpcClient::from_url`] and an `https` url instead.
     pub async fn new(
         ip: IpAddr,
         public_port: u16,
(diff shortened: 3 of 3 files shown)


peterjah:
[F144] Documents the inert certificate_store setting; no behavior change.

massa-sdk exposed ClientConfig::certificate_store ("Native" / "WebPki"), surfaced to operators as [client] certificate_store in massa-client/base_config/config.toml, but the setting has been dead since the jsonrpsee migration in 4fd7a85 (#3776, Apr 2023) — the trust-store selection was left in place as a commented-out block referencing use_native_rustls() / use_webpki_rustls(), APIs that no longer exist. A reader of the config would reasonably conclude the client validates TLS certificates against a chosen trust store; it does not, because Client::new, ClientV2::new, and RpcClientV2::from_url hard-code plaintext http:// / ws:// / grpc://.

This PR only makes the SDK honest about that. It does not add TLS support.

- Removed the two commented-out match certificate_store blocks in http_client_from_url / ws_client_from_url. They read as "temporarily disabled" but reference a removed API, so nobody could uncomment them.
- Marked ClientConfig::certificate_store deprecated in its doc comment and in config.toml, stating it is ignored. The field is kept: removing it would break massa-client's struct literal and every deployed config.toml, so it needs a deprecation cycle.
- Documented on Client::new / ClientV2::new that they are plaintext-only, and on RpcClient::from_url that it selects the transport from the url scheme.

One clarification worth recording, since the original finding got it backwards: JSON-RPC TLS is not structurally missing. jsonrpsee 0.24 is pulled with default features, so hyper-rustls / rustls-platform-verifier are compiled in and HttpBackend picks its Https variant from the url scheme. RpcClient::from_url(url: &str, …) is public and already works with an https:// url today — only the convenience constructors block it, by not letting a scheme through. Likewise, gen_signed_cert is server-side only (massa-grpc/src/server.rs) and is not part of any client trust decision.
Not consensus-observable — client-side transport only, no MIP gate needed.Residual gaps, deliberately left open here:

- gRPC cannot do TLS at all. massa-sdk/Cargo.toml takes tonic with only the gzip feature,tls and the node exposes enable_tls / enable_mtls. A node operator who turns on gRPC TLSsilently loses massa-client gRPC connectivity. Fixing this means enabling tonic's tls feature and threading a ClientTlsConfig (CA root, plus client cert/key for mTLS — the node already writes these to  client_certificate_path).
- No way to request TLS through the convenience constructors. Client::new / ClientV2::new'd need to accept a scheme or a full url so https:// / wss:// reach the builders, whichalready handle it correctly.- "WebPki" no longer maps to anything. jsonrpsee 0.24's API is with_custom_cert_store(Cusult is the platform verifier (≈ the old Native). Any future trust-store option must bere-expressed in those terms rather than restoring the old strings.
- The gRPC failure is undiagnosable. Client::new swallows a gRPC connect error into None hich is why the TLS mismatch above surfaces as "gRPC just doesn't work" with no actionable
message.

peterjah:
@Leo-Besancon @damip 
I created a followup issue to enable TLS on grpc:
https://github.com/massalabs/massa/issues/5180

Leo-Besancon (approved):
Approving this small documentation PR, but the associated finding should potentially be kept opened while the follow-up isn't done

PR #5181, opened 2026-08-19
[F73] refuse to start the gRPC API when enable_mtls is set without en…

### LINKED ISSUE #5087: [F73] Private Admin gRPC Silently Downgrades to Plaintext When mTLS Is Enabled Without TLS

---

## Source
AI security-scan **Finding 73** (severity: **major**). Tracked in `SUBMISSIONS.md` as F73.

**Affected file:** `massa-grpc/src/server.rs`

## Summary
`massa-grpc::server::serve` sets up transport security only when `config.enable_tls` is true. If `config.enable_mtls` is true while `enable_tls` is false, the server still starts the private admin gRPC service without loading certificates or client CA settings. This creates a silent downgrade from the operator's expected *mutual TLS* protection to *plaintext* transport. As a result, privileged methods can be exposed under a weaker trust boundary than intended.

## Recommendation
Validate the configuration before startup and reject any deployment where `enable_mtls = true` and `enable_tls = false`. Treat mTLS as a strict extension of TLS, so enabling mTLS must always require TLS to be enabled as well.

In `configure_grpc`, enforce this dependency when building `GrpcConfig` so invalid flag combinations cannot propagate further. In `massa-grpc::server::serve`, add a defensive runtime check to fail closed if mTLS is requested without TLS, rather than starting the private admin service in plaintext.

Also ensure startup logs and operator-facing error messages clearly state that mTLS cannot be enabled without TLS, so misconfigurations are detected immediately and do not result in a silent downgrade of the private admin gRPC transport.

## Why P0 / breaking
The fix changes **API / gRPC behavior** (private admin gRPC transport). Per maintainer guidance, API/gRPC interface or observable-behavior changes are treated as **breaking** (they can break external tooling), so this is filed as an issue for design/versioning discussion rather than a direct PR. Note: the gRPC proto types are auto-generated and must not be hand-edited. See `SUBMISSIONS.md` Part D.

peterjah:
…able_tls

  * [ ] make tests compile
  * [ ] make tests pass 

### DIFF
--- a/massa-node/base_config/config.toml
+++ b/massa-node/base_config/config.toml
@@ -141,7 +141,7 @@
         # whether to enable TLS
         enable_tls = false
         # whether to enable mTLS (requires `enable_tls` to be true)
-        enable_mtls = true
+        enable_mtls = false
         # whether to generate a self-signed certificate if none is provided(ignored if `enable_tls` is false)
         generate_self_signed_certificates = true
         # list of subject alternative names for the server certificate(requires `generate_self_signed_certificates` to be true)
--- a/massa-grpc/src/error.rs
+++ b/massa-grpc/src/error.rs
@@ -43,6 +43,8 @@ pub enum GrpcError {
     InternalServerError(String),
     /// Invalid argument error: {0}
     InvalidArgument(String),
+    /// Invalid configuration error: {0}
+    ConfigError(String),
     /// Not implemented error: {0}
     Unimplemented(String),
 }
@@ -63,6 +65,7 @@ impl From<GrpcError> for tonic::Status {
             GrpcError::InternalServerError(e) => tonic::Status::internal(e),
             GrpcError::ReflectionError(e) => tonic::Status::internal(e.to_string()),
             GrpcError::InvalidArgument(e) => tonic::Status::invalid_argument(e),
+            GrpcError::ConfigError(e) => tonic::Status::internal(e),
             GrpcError::Unimplemented(e) => tonic::Status::unimplemented(e),
         }
     }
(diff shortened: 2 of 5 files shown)


peterjah:
@Leo-Besancon @damip 
This is startup-breaking for anyone who currently has enable_mtls = true, enable_tls = false in their own config.toml — which, given the shipped default, could be a fair number of operators. They were never actually getting mTLS, so nothing is lost, but the node will now refuse to boot until they set enable_mtls = false or enable TLS properly. That needs a changelog entry, and it's the one decision worth a second opinion before merge: fail closed (what we did, and what F73 recommends) versus warn-and-continue. I'd keep fail-closed — a warning in a startup log is exactly what nobody reads, and silently serving an unauthenticated admin API is the bug.

massa-node/base_config/config.toml
```diff
@@ -141,7 +141,7 @@
         # whether to enable TLS
         enable_tls = false
         # whether to enable mTLS (requires `enable_tls` to be true)
-        enable_mtls = true
+        enable_mtls = false
```
Leo-Besancon, line 144: That's weird it already shows as false in main for me? Why does the diff shows it was set to true?

Leo-Besancon (approved):
Good sanity check! I agree with your reasonning above @peterjah

massa-node/base_config/config.toml
peterjah: Its already false/false for public grpc section but false/true for private grpc

Leo-Besancon: ah my bad, thank you!

PR #5182, opened 2026-08-19
[F107] prune expired propagation entries before insert and warn on co…

### LINKED ISSUE #5070: [F107] Count-based propagation cache can drop block refs before the advertised propagation window ends

---

## Source
AI security-scan **Finding 107** (severity: **minor**). Tracked in `SUBMISSIONS.md` as F107.

**Affected file:** `massa-protocol-worker/src/handlers/block_handler/propagation.rs`

## Summary
`BlockHandlerPropagationThread::run` stores each integrated block in `stored_for_propagation`, but that structure is a count-bounded `LruMap` rather than a purely time-bounded retention queue. As a result, older entries can be evicted before `max_block_propagation_time` elapses, even though the stored `Storage` reference is intended to keep block data alive during the propagation window. The original finding is therefore valid as a *count-versus-time retention mismatch*. However, the broader claim that this causes a routine propagation-layer availability failure during ordinary bursts or realistic attacker floods is not supported by the protocol's effective block-rate limits and other retention paths.

## Root cause
In `massa-main/massa-protocol-worker/src/handlers/block_handler/propagation.rs`, `run` handles `BlockHandlerPropagationCommand::IntegratedBlock` by inserting a `BlockPropagationData` entry into `self.stored_for_propagation` with `time_added`, `_storage`, and `header`. The code comment states that this cache is meant to ensure the block and its dependencies are stored for the duration of propagation, but the container is bounded by *entry count*, not strictly by elapsed time.

Because `self.stored_for_propagation` is an `LruMap`, inserting enough blocks can evict older entries immediately, before their intended propagation-retention period has expired. The ruled-out portion of the original theory is that such eviction alone necessarily makes block operations unavailable to peers: recent active blocks may still be retained elsewhere by consensus with full storage, so propagation-cache eviction is not by itself a guaranteed loss of retrievable data.

## Affected code
```rust
fn run(&mut self) {
        let tick_interval = self.config.block_propagation_tick.to_duration();
        let mut deadline = Instant::now()
            .checked_add(tick_interval)
            .expect("could not get time of next propagation tick");
        loop {
            match self.receiver.recv_deadline(deadline) {
                Ok(command) => {
                    match command {
                        // Message: the block was integrated and should be propagated
                        BlockHandlerPropagationCommand::IntegratedBlock { block_id, storage } => {
                            debug!("received IntegratedBlock({})", block_id);

                            // get the block header
                            let header = match storage
                                .read_blocks()
                                .get(&block_id)
                                .map(|block| block.content.header.clone())
                            {
                                Some(h) => h,
                                None => {
                                    warn!(
                                        "claimed block {} absent from storage on propagation",
                                        block_id
                                    );
                                    continue;
                                }
                            };

                            // Add the block and its dependencies to the propagation LRU
                            // to ensure they are stored for the time of the propagation.
                            self.stored_for_propagation.insert(
                                block_id,
                                BlockPropagationData {
                                    time_added: Instant::now(),
                                    _storage: storage,
                                    header,
                                },
                            );

                            // propagate everything that needs to be propagated
                            self.perform_propagations();

                            // renew tick because propagation propagations were updated
                            deadline = Instant::now()
// … (truncated)
```

## Recommendation
Align the retention mechanism with the intended guarantee by enforcing propagation-entry expiry primarily by elapsed time rather than only by cache size. Ensure `stored_for_propagation` preserves each block’s `Storage` reference for at least `max_block_propagation_time`, or revise the documented guarantee and related comments if only best-effort retention is intended.

If a size limit must remain, add safeguards so count-based eviction cannot silently invalidate the advertised propagation window, such as separating time-based retention from memory-control logic and defining explicit behavior when capacity is exceeded. Also review dependent propagation paths to confirm they do not assume this cache provides a strict availability guarantee beyond what other storage-retention mechanisms actually ensure.

## Why P0 / breaking

peterjah:
…unt-based eviction

  * [ ] make tests compile
  * [ ] make tests pass 

### DIFF
--- a/massa-protocol-worker/src/handlers/block_handler/propagation.rs
+++ b/massa-protocol-worker/src/handlers/block_handler/propagation.rs
@@ -10,6 +10,9 @@
 //!
 //! Here we need to announce block headers to other nodes that haven't sene them,
 //! and keep the blocks alive long enough for our peers to be able to retrieve them from us.
+//! That retention is best-effort: blocks are kept for `max_block_propagation_time` unless
+//! the `max_blocks_kept_for_propagation` size cap forces an earlier eviction. Blocks that
+//! are still active in the graph remain retrievable through consensus storage anyway.
 
 use super::{
     cache::SharedBlockCache, commands_propagation::BlockHandlerPropagationCommand,
@@ -53,7 +56,8 @@ pub struct PropagationThread {
     config: ProtocolConfig,
     /// Shared access to the block cache
     cache: SharedBlockCache,
-    /// Blocks stored for propagation
+    /// Blocks stored for propagation, bounded both by `max_block_propagation_time`
+    /// and by `max_blocks_kept_for_propagation` entries
     stored_for_propagation: LruMap<BlockId, BlockPropagationData>,
     /// Shared access to the list of peers connected to us
     active_connections: Box<dyn ActiveConnectionsTrait>,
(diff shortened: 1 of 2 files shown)


massa-protocol-worker/src/handlers/block_handler/propagation.rs
```diff
@@ -10,6 +10,9 @@
 //!
 //! Here we need to announce block headers to other nodes that haven't sene them,
 //! and keep the blocks alive long enough for our peers to be able to retrieve them from us.
+//! That retention is best-effort: blocks are kept for `max_block_propagation_time` unless
+//! the `max_blocks_kept_for_propagation` size cap forces an earlier eviction. Blocks that
+//! are still active in the graph remain retrievable through consensus storage anyway.
```
Leo-Besancon, line 15: Should we warn (maybe a !warn() at startup if `max_block_propagation_time` is more than the time the consensus stores a block (through `force_keep_final_periods` I think)?

Because even if we warn below when we do evict early a block, a warn at startup that it could happen with current config may be easier to catch.

(before doing this, we should also check the default values and see if we want to change them or not)

Leo-Besancon (dismissed):
Left a comment I thought may be interesting, but don't hesitate to ignore if you think it's not really relevant. LGTM!

massa-protocol-worker/src/handlers/block_handler/propagation.rs
peterjah: warning added

Leo-Besancon (approved):
Thanks! LGTM

PR #5183, opened 2026-08-19
[F124] track each propagated operation once to keep propagation queue…

### LINKED ISSUE #5071: [F124] Duplicate propagation batch tracking causes premature operation eviction and stale storage references

---

## Source
AI security-scan **Finding 124** (severity: **major**). Tracked in `SUBMISSIONS.md` as F124.

**Affected file:** `massa-protocol-worker/src/handlers/operation_handler/propagation.rs`

## Summary
In `massa-main/massa-protocol-worker/src/handlers/operation_handler/propagation.rs`, `PropagationThread` keeps propagated operations in `stored_for_propagation`, a queue of batches, and also stores them in `op_storage`. The code allows the same `OperationId` to appear in multiple queued batches because `run(&mut self)` always appends a new batch and calls `self.op_storage.extend(operations)` without checking whether those IDs are already retained. Later, `prune_propagation_storage(&mut self)` prunes batches by queue membership but drops storage references with a deduplicated set of IDs. This mismatch can both evict unrelated honest operations too early and leave stale operation data pinned in `op_storage`.

## Root cause
The main issue is in `massa-main/massa-protocol-worker/src/handlers/operation_handler/propagation.rs` inside `PropagationThread::run(&mut self)`. Each `OperationHandlerPropagationCommand::PropagateOperations(operations)` unconditionally does:

- `self.stored_for_propagation.push_back((std::time::Instant::now(), new_ops.clone()))`
- `self.op_storage.extend(operations)`

There is no deduplication against existing `stored_for_propagation: VecDeque<(std::time::Instant, PreHashSet<OperationId>)>`, so the same `OperationId` can exist in several propagation batches at the same time.

A second root cause is in `PropagationThread::prune_propagation_storage(&mut self)`. Its size-pressure logic computes retained size with:

- `self.stored_for_propagation.iter().map(|(_, ops)| ops.len()).sum::<usize>().saturating_sub(self.config.max_ops_kept_for_propagation)`

This counts one `OperationId` once per batch, not once globally. If duplicate IDs exist across batches, `config.max_ops_kept_for_propagation` is enforced on summed batch lengths rather than on unique retained operations.

A third root cause is that `prune_propagation_storage(&mut self)` collects all removed batch entries into one deduplicated `PreHashSet<OperationId>` named `removed`, then calls:

- `self.op_storage.drop_operation_refs(&removed)`

This is not aligned with queue membership removal. If the same `OperationId` is removed from multiple popped batches in one prune pass, the repeated removals are collapsed before reaching storage.

The storage-side behavior reinforces this mismatch. `drop_operation_refs(&mut self, ids: &PreHashSet<OperationId>)` eventually calls:

- `Self::internal_drop_refs(ids, &mut owners, &mut self.local_used_ops)`

and then removes `orphaned_ids` from `self.operations`. Because the prune path p …

## Affected code
```rust
fn prune_propagation_storage(&mut self) {
        let mut removed = PreHashSet::default();

        // remove expired
        let max_op_prop_time = self.config.max_operations_propagation_time.to_duration();
        while let Some((t, _)) = self.stored_for_propagation.front() {
            if t.elapsed() > max_op_prop_time {
                let (_, op_ids) = self
                    .stored_for_propagation
                    .pop_front()
                    .expect("there should be at least one element, checked above");
                removed.extend(op_ids);
            } else {
                break;
            }
        }

        // Cap cache size
        // Note that we directly remove batches of operations, not individual operations
        // to favor simplicity and performance over precision.
        let mut excess_count = self
            .stored_for_propagation
            .iter()
            .map(|(_, ops)| ops.len())
            .sum::<usize>()
            .saturating_sub(self.config.max_ops_kept_for_propagation);
        while excess_count > 0 {
            if let Some((_t, op_ids)) = self.stored_for_propagation.pop_front() {
                excess_count = excess_count.saturating_sub(op_ids.len());
                removed.extend(op_ids);
            } else {
                break;
            }
        }

        // remove from storage
        self.op_storage.drop_operation_refs(&removed);
    }
```

## Recommendation
Prevent the same `OperationId` from being tracked in multiple propagation batches at the same time. `PropagationThread::run(&mut self)` should only add an operation to `stored_for_propagation` and increment its storage retention if that operation is not already present in the propagation cache. This keeps queue membership and `op_storage` references aligned and avoids artificial cache-pressure inflation.

If repeated propagation of an already tracked operation must remain supported, maintain explicit per-operation membership or reference counts and make `prune_propagation_storage(&mut self)` remove references with the same multiplicity as the queue entries being pruned. In particular, avoid deduplicating removed IDs before updating `op_storage` unless storage retention is also deduplicated at insertion time.

`config.max_ops_kept_for_propagation` should be enforced against the number of uniquely retained operations, not the sum of per-batch lengths. This prevents duplicate resubmissions from consuming the propagation budget multiple times and prematurely evicting unrelated operations.

After the change, ensure the following invariant always holds: an operation is present in `op_storage` if and only if it is still retained by at least one propagation batch. Add regression tests covering duplicate repropagation, pruning of overlapping batches in a single pass, and size-limit enforcement with repeated `OperationId` values.

## Why P0 / breaking

peterjah:
… and storage refs aligned

  * [ ] make tests compile
  * [ ] make tests pass 

### DIFF
--- a/massa-protocol-worker/src/handlers/operation_handler/propagation.rs
+++ b/massa-protocol-worker/src/handlers/operation_handler/propagation.rs
@@ -33,9 +33,11 @@ struct PropagationThread {
     active_connections: Box<dyn ActiveConnectionsTrait>,
     // times at which previous ops were announced
     stored_for_propagation: VecDeque<(std::time::Instant, PreHashSet<OperationId>)>,
-    // running total number of operations held in `stored_for_propagation`, maintained
-    // incrementally so pruning does not need to scan the whole queue on every call
-    stored_ops_count: usize,
+    // ids of the operations held in `stored_for_propagation`, maintained incrementally
+    // so pruning does not need to scan the whole queue on every call. The queue batches
+    // are disjoint, so this is both the set of operations referenced in `op_storage` and
+    // the number of operations the cache size is capped on.
+    stored_ops: PreHashSet<OperationId>,
     op_storage: Storage,
     next_batch: PreHashSet<OperationId>,
     config: ProtocolConfig,
@@ -64,10 +66,19 @@ impl PropagationThread {
 
                             // add to propagation storage
                             let new_ops = operations.get_op_refs().clone();
-                            self.stored_ops_count =
-                                self.stored_ops_count.saturating_add(new_ops.len());
-                            self.stored_for_propagation
-                                .push_back((std::time::Instant::now(), new_ops.clone()));
+                            // an operation can be propagated again while it is still
+                            // queued: only queue the ones that are not already tracked, so
+                            // that queue membership stays aligned with the `op_storage`
+                            // refs and duplicates do not consume the cache budget twice
+                            let newly_tracked: PreHashSet<OperationId> = new_ops
+                                .iter()
+                                .filter(|op_id| self.stored_ops.insert(**op_id))
+                                .copied()
+                                .collect();
+                            if !newly_tracked.is_empty() {
+                                self.stored_for_propagation
+                                    .push_back((std::time::Instant::now(), newly_tracked));
+                            }
                             self.op_storage.extend(operations);
                             self.prune_propagation_storage();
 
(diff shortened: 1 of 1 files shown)


Leo-Besancon (approved):
LGTM

PR #5184, opened 2026-08-20
[F122] enforce the operation announcement deadline under a continuous…

### LINKED ISSUE #5119: [F122] Operation propagation announcement path can be starved under channel flooding

---

**Severity:** minor
**Component:** `massa-protocol-worker` operation handler (`PropagationThread::run`)

## Summary

`PropagationThread::run` can starve its time-based announcement path when
`internal_receiver.recv_deadline(batch_deadline)` is fed by a continuously non-empty channel.
`announce_ops()` is only triggered on timeout or when `next_batch` reaches
`operation_announcement_buffer_capacity`; repeated `propagate_operations(...)` submissions can
keep the timeout path from firing while not materially growing `next_batch`, delaying
announcement.

The "propagation completely halted" and "peer-to-peer duplicate injection" framings are
overstated; the validated issue is timeout-path starvation under continuous local submission.

The fix changes the propagation thread's batching/deadline loop, which is timing-sensitive and
affects propagation latency. Needs review to avoid regressing batching efficiency or
introducing premature announcements.

## Suggested direction

Ensure the deadline/announce path makes progress even under a continuously non-empty channel
(e.g. track an absolute deadline and force `announce_ops()` when it passes regardless of new
messages, or bound the number of consecutive `recv` iterations before checking the deadline).

Tracking: finding F122.

peterjah:
…ly non-empty channel

  * [ ] make tests compile
  * [ ] make tests pass 

### DIFF
--- a/massa-protocol-worker/src/handlers/operation_handler/propagation.rs
+++ b/massa-protocol-worker/src/handlers/operation_handler/propagation.rs
@@ -45,10 +45,15 @@ struct PropagationThread {
 }
 
 impl PropagationThread {
-    fn run(&mut self) {
-        let mut batch_deadline = std::time::Instant::now()
+    /// Time at which the next announcement should happen if the batch is not full before that.
+    fn next_batch_deadline(&self) -> std::time::Instant {
+        std::time::Instant::now()
             .checked_add(self.config.operation_announcement_interval.to_duration())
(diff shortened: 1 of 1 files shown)


Leo-Besancon (approved):
I'm not very familiar with the propagation expected timings, but the fix looks good to me!

PR #5185, opened 2026-08-20
[F52] bound read-only request starvation behind slot execution backlog

### LINKED ISSUE #5115: [F52] Read-only execution requests can be starved indefinitely behind slot execution backlog

---

**Severity:** medium
**Component:** `massa-execution-worker/src/worker.rs` (`main_loop` / `execute_one_readonly_request`)

## Summary

`main_loop()` reaches `execute_one_readonly_request()` only when
`self.slot_sequencer.run_task_with(...)` returns `None`, while `execute_readonly_request()`
waits synchronously for the result. Under a sustained final/candidate slot backlog, slot work
is continuously available, so queued read-only execution requests can be starved indefinitely.

The priority ordering (slot work before read-only) is intentional; the validated issue is the
resulting read-only starvation/availability problem under sustained backlog. The claim of
universally indefinite blocking for all RPC surfaces is overstated.

Any fix changes the worker's execution-priority policy (e.g. fairness / budgeting between slot
work and read-only requests), which affects execution latency and throughput characteristics.
This is a deliberate policy area and needs maintainer design input, not a drive-by change.

## Suggested direction

Introduce a bounded fairness guarantee for read-only requests (e.g. serve at least one
read-only request per N slot tasks, or a small time budget), or a configurable read-only
deadline, while preserving slot-work priority overall.

Tracking: finding F52.

peterjah:
  * [ ] make tests compile
  * [ ] make tests pass 


massa-node/base_config/config.toml
```diff
@@ -208,6 +208,11 @@
     max_final_events = 10000
     # maximum length of the read-only execution requests queue
     readonly_queue_length = 100
+    # maximum time in milliseconds a queued read-only request may wait behind slot
+    # execution before being served, even if slot work is still pending.
+    # Slot execution keeps priority overall: this only bounds read-only starvation
+    # when the worker is lagging behind real time (e.g. catching up after bootstrap).
+    readonly_starvation_timeout = 200
```
Leo-Besancon: As we have a slot every 500ms, this timeout seems really low to me, I would expect it to be around 5 or 10 seconds. Otherwise, it just means that at least one readonly requests need to be handled every slot whatever the load, which seems disproportionate to me.
Although maybe the API timeouts are low?

Leo-Besancon (commented):
The implementation itself is good, I've left a comment for the chosen config. I'm still not 100% sure this needs to be resolved though, to me starving the readonly requests is not a big deal if this means we keep up with finality.

So I'd like a second opinion from @damip before we merge this :)

massa-node/base_config/config.toml
damip: yes let's make it 5 seconds

damip (dismissed):
check the config comment

massa-node/base_config/config.toml
peterjah: done

Leo-Besancon (approved):
Thanks, LGTM!

PR #5187, opened 2026-08-20
[F116] bound endorsement slots by an acceptance window instead of ban…

### LINKED ISSUE #5112: [F116] Endorsement retrieval ban policy conflates local selector unavailability with out-of-range (bannable) slots

---

**Severity:** minor (peer ban-policy correctness — both over-ban and under-ban risks)
**Component:** `massa-protocol-worker/src/handlers/endorsement_handler/retrieval.rs`
(`process_message` / `note_endorsements_from_peer`), interacting with
`massa-pos-worker/src/controller.rs` (`get_selection`).

## Original problem (F116)

`process_message` bans the sending peer whenever `note_endorsements_from_peer`
returns `Err`. During PoS-draw validation that function calls:

```rust
let selection = selector_controller
    .get_selection(endorsement.content.slot)?   // propagates PosError
    .endorsements;
```

`get_selection` returns `PosError::CycleUnavailable(cycle)` when the slot's cycle
draws are not present in the local selector cache. This happens for an **honest**
peer's valid endorsement when the local node is simply lagging / has not computed
that cycle's draws yet. Result: an honest peer can be banned due to a purely local
state condition, which harms network health (a lagging node bans its good peers).

## New dimension (raised in review of PR #5099)

The same `get_selection(slot)` call takes an **attacker-controlled** slot and returns
the **same** `PosError::CycleUnavailable(cycle)` for endorsements whose slot is out of
the acceptable range — far in the future, far in the past, or otherwise structurally
invalid (`draws.get(&slot)` is `None`). Such endorsements are invalid and **should**
lead to a ban.

Because the selector error does not distinguish:

- (a) an **in-window** slot whose draws are merely not available locally yet
  (honest lag → should **not** ban), from
- (b) an **out-of-window / structurally invalid** slot (peer offense → **should** ban),

neither the original policy nor the naive relaxation is correct:

- **Original** (ban on any error): over-bans honest lagging peers (the F116 report).
- **Naive relaxation** (never ban on `PosError`, as attempted in PR #5099): under-bans —
  an attacker can send endorsements with bogus out-of-range slots with no penalty, and
  it also short-circuits the producer-selection ban checks (`GeneralProtocolError`:
  "No selection on slot ...", "Invalid endorsement producer selection ...") that would
  otherwise run for in-window slots.

## Code references

- `note_endorsements_from_peer` (retrieval.rs): the `get_selection(..)?` PoS-draw loop;
  the subsequent `GeneralProtocolError` producer-selection checks are only reached for
  slots whose cycle is available in the selector cache.
- `get_selection` (pos-worker controller.rs): returns `CycleUnavailable(cycle)` whenever
  `cache.get(cycle)` or `draws.get(&slot)` is absent — no distinction between
  "not computed yet" and "invalid/out-of-range slot".

## Suggested direction (needs design)

Validate the endorsement slot against an explicit acceptance window **before /
independently of** selector availability:

1. **Ban** endorsements whose slot is too far in the future or too old relative to the
   current slot (peer offense). This needs a defined window — the past bound can reuse
   `max_endorsements_propagation_time`; the future bound needs a small bounded horizon
   consistent with endorsement inclusion rules.
2. For **in-window** slots where `get_selection` returns `CycleUnavailable` /
   `CycleUnfinished`, treat as **local** (log, no ban), optionally deferring/retrying.
3. Keep banning genuine peer-caused validation errors (bad signature, wrong producer
   selection).

This overlaps with existing endorsement-validation findings **F113** (missing
endorsement freshness validation) and **F117** (validation-order / stale-endorsement
replay) and should be designed together with them.

## Status

Attempted fix PR #5099 (which simply skipped banning on any `PosError`) has been
**closed** in favor of this issue, because it introduced the under-ban / ban-evasion
gap described above.

peterjah:
…ning peers on local selector errors

  * [ ] make tests compile
  * [ ] make tests pass 

### DIFF
--- a/massa-protocol-worker/src/handlers/endorsement_handler/retrieval.rs
+++ b/massa-protocol-worker/src/handlers/endorsement_handler/retrieval.rs
@@ -199,11 +199,13 @@ fn is_endorsement_fresh(
 ///
 /// Caches knowledge of valid ones.
 ///
-/// Does not ban if the endorsement is invalid
+/// Returns an error if the peer must be banned.
 ///
 /// Checks performed:
 /// - Valid signature.
-/// - The creator is the endorser drawn for the endorsement's `(slot, index)` pair.
+/// - Endorsement slot not too far in the future (peer offense if it is).
+/// - The creator is the endorser drawn for the endorsement's `(slot, index)` pair, when the draws
+///   of that slot are locally available.
 /// - At most `MAX_ENDORSEMENTS_PER_SLOT_INDEX` distinct endorsements are noted for a given draw,
 ///   so that an equivocating endorser cannot have us relay unlimited variants of its endorsement.
 #[allow(clippy::too_many_arguments)]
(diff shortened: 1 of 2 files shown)


massa-protocol-worker/src/handlers/endorsement_handler/retrieval.rs
```diff
@@ -210,15 +212,68 @@ pub(crate) fn note_endorsements_from_peer(
...
+    let now = MassaTime::now();
+
+    // Check the endorsement slots and the PoS draws.
+    //
+    // An endorsement is created at its own slot, so its slot can only be as far ahead of our clock
+    // as the propagation window, plus one period of tolerance for clock drift. Beyond that the
+    // endorsement is invalid whatever our local state is, and propagating it would amplify it, so we
+    // fail the whole batch and the caller bans the peer.
+    //
+    // Endorsements that are too old, and endorsements whose cycle draws are not available locally,
```
Leo-Besancon, line 266: In the original issue, it was suggested to also ban for endorsements too old:
> the past bound can reuse max_endorsements_propagation_time

Is there an issue with this? Even if we take a bigger margin?

peterjah: Future bound is safe to ban on. An endorsement is created at its own slot. No honest peer can hold one whose slot hasn't happened yet (beyond clock drift), so slot_timestamp > now + margin is an offense independent of local state.

Past bound is not. Three concrete false-positive paths, all verified:

1. The block-header path shares this function. note_endorsements_from_peer is called from massa-protocol-worker/src/handlers/block_handler/retrieval.rs:595, and any Err there becomes ProtocolError::InvalidBlock, which bans at retrieval.rs:452. note_header_from_peer deliberately has no lower bound on header slot age (retrieval.rs:532-545 only rejects genesis + version) — consensus decides staleness later. Endorsement slots equal the header slot (checked at line ~630). So a peer serving us an old block we asked for (catch-up after downtime, missing-parent resolution from the wishlist) would be banned for doing exactly what we requested. That's the F116 bug re-introduced through the other door.

2. The margin is tiny. max_endorsements_propagation_time = 32000 ms + T0 = 16000 ms = 48 s of tolerance. Any peer more than 48 s behind — reconnect, GC pause, congested retrieval queue, our own clock running fast — becomes bannable.

3. A bigger margin only shifts the threshold. For path (1) there is no correct finite value, because legitimate header exchange is unbounded in the past. You'd have to make the past-ban apply only to the standalone EndorsementMessage::Endorsements path, and even there it still punishes lagging honest peers for a message that costs us nothing.

Leo-Besancon (dismissed):
Thanks for the answer, LGTM then!

Leo-Besancon (dismissed):

Leo-Besancon (approved):

PR #5188, opened 2026-08-20
[F127][F131] time-bound per-peer operation ask state to dedupe deferr…

### LINKED ISSUE #5108: [F127+F131] asked_operations/op_batch_buffer: duplicate-announcement buffering vs. stale state across reconnects

---

**Severity:** medium (mempool synchronization reliability / liveness)
**Component:** `massa-protocol-worker/src/handlers/operation_handler/retrieval.rs`
(`on_operations_announcements_received`, `asked_operations`, `op_batch_buffer`)

These two findings concern the same `asked_operations` / `op_batch_buffer` state
machine and pull in **opposite** directions, so they should be designed together.

## F127 — duplicate delayed announcements can fill `op_batch_buffer`

When an `OperationsAnnouncement` arrives, prefixes that are not yet re-askable are
added to `future_set` and an `OperationBatchItem` is pushed into `op_batch_buffer`, but
the announcing peer is **not** recorded in `asked_operations` at that time. During the
cooldown window the same peer can re-send the same announcement repeatedly, creating
many duplicate buffered entries that can fill `op_batch_buffer` and degrade mempool
sync.

## F131 — stale `asked_operations` across reconnects can prevent re-asking

`asked_operations` records which peers were already queried for each announced prefix
and skips later announcements from the same peer (`previous_ask.1.contains(peer_id)`).
This peer tracking is **not** cleared on disconnect, so after a peer disconnects
(before answering `AskForOperations`) and reconnects, its repeated announcement is
ignored and no new `AskForOperations` is sent — some operations become temporarily
unavailable from that node, reducing liveness.

F127 argues for recording the peer earlier (to suppress duplicates), while F131 argues
for clearing peer state (to allow re-asking after reconnect). A correct fix must
reconcile both: dedupe abusive repeats **without** permanently suppressing legitimate
re-asks after reconnect. F131 additionally requires new plumbing — a disconnect signal
delivered to the retrieval thread to invalidate stale `asked_operations` entries — which
touches the peer-lifecycle/command wiring. This is a protocol-behavior design decision.

- Introduce a per-(prefix, peer) cooldown/entry that both suppresses duplicate buffered
  items (F127) and expires or is invalidated on peer disconnect (F131).
- Plumb a disconnect notification into the operation retrieval thread to clear stale
  `asked_operations` entries for that peer.
- Add tests for: duplicate announcements during cooldown do not grow the buffer
  unboundedly; a reconnecting peer's re-announcement triggers a fresh ask.

## Refs

Findings F127, F131.

peterjah:
…ed asks and allow re-asking

  * [ ] make tests compile
  * [ ] make tests pass 

### DIFF
--- a/massa-protocol-worker/src/handlers/operation_handler/retrieval.rs
+++ b/massa-protocol-worker/src/handlers/operation_handler/retrieval.rs
@@ -51,11 +51,52 @@ pub struct OperationBatchItem {
     pub operations_prefix_ids: OperationPrefixIds,
 }
 
+/// State of a peer with respect to an announced operation prefix.
+#[derive(Clone, Copy)]
+enum PeerAskState {
+    /// an `AskForOperations` was sent to that peer at that instant
+    Asked(Instant),
+    /// a deferred ask for that peer is waiting in `op_batch_buffer`
+    Buffered,
+}
+
+/// Ask state of an announced operation prefix.
+struct AskedOperation {
+    /// last instant at which that prefix was asked to any peer
+    last_ask: Instant,
+    /// state of the peers that announced that prefix.
+    /// Kept as a `Vec` because it holds at most one entry per connected peer.
+    peers: Vec<(PeerId, PeerAskState)>,
+}
+
+impl AskedOperation {
+    fn peer_state(&self, peer_id: &PeerId) -> Option<PeerAskState> {
+        self.peers
+            .iter()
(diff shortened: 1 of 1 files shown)


Leo-Besancon (approved):
LGTM, I'm not very familiar with that part of the codebase but the refactor makes sense and seems to align well with both findings goal!

PR #5189, opened 2026-08-20
[F42] propagate the configured datastore key query cap from RPC/gRPC …

### LINKED ISSUE #5059: [F42] Unbounded Datastore Key Enumeration Enables Underpriced DoS

---

## Source
AI security-scan **Finding 42** (severity: **major**). Tracked in `SUBMISSIONS.md` as F42.

**Affected file:** `massa-execution-worker/src/datastore_scan.rs`

## Summary
`scan_datastore` in `massa-execution-worker/src/datastore_scan.rs` can do work proportional to the full datastore key range when `count = None` or when the effective limit is not enforced. This happens on both the smart-contract path and the external query path. In some branches, the function collects or scans many matching keys before any truncation is applied. Because this work is underpriced or uncapped, an attacker can cause excessive CPU, memory, and query-processing load.

## Root cause
The main issue is that datastore key enumeration is not consistently bounded, and several entry points allow `scan_datastore(...)` to run with `count = None`.

- In `massa-execution-worker/src/datastore_scan.rs`, `scan_datastore(...)` performs unbounded or effectively unbounded work when `count` is missing.
  - In the `SpeculativeResetType::Set` branch, it builds `key_updates` with:
    - `v.datastore.range(k_range.clone()).map(|(k, _v)| (k.clone(), true)).collect()`
  - This clones and collects *all* matching speculative keys before any later truncation such as `filter_it.take(cnt as usize)`.
  - In the `SpeculativeResetType::None` merge path, `scan_datastore(...)` merges speculative updates with final-state keys. When `final_keys_queue` becomes empty, it repeatedly calls:
    - `get_datastore_keys(addr, prefix, Bound::Excluded(last_k), end_key.clone(), count)`
  - If `count = None`, this replenishment can continue until the whole matching datastore range is scanned. Even when `count` is present, extra final-state batches may still be fetched until enough non-deleted surviving keys are found.

- On the smart-contract path, the limit is not propagated.
  - `Context::get_keys` hardcodes `max_datastore_query = None`.
  - `InterfaceImpl::get_ds_keys_wasmv1` and the deprecated `get_keys*` wrappers pass no limit.
  - As a result, smart contracts can reach `scan_datastore(...)` with `count = None`.

- On the external query path, the configured cap can be bypassed.
  - In `massa-execution-worker/src/controller.rs`, `query_state(&self, req: ExecutionQueryRequest) -> ExecutionQueryResponse` forwards:
    - `ExecutionQueryRequestItem::AddressDatastoreKeysCandidate { address, prefix, start_key, end_key, count }`
    - `ExecutionQueryRequestItem::AddressDatastoreKeysFinal { address …

## Affected code
```rust
/// Gets a copy of a datastore keys for a given address
///
/// # Arguments
/// * `addr`: address to query
/// * `prefix`: prefix to filter keys
/// * `start_key`: start key of the range
/// * `end_key`: end key of the range
/// * `count`: maximum number of keys to return
///
/// # Returns
/// A tuple of two `Option<BTreeSet<Vec<u8>>>`:
/// `None` means that the address does not exist.
/// The first element is the final state keys, the second element is the speculative keys.
#[allow(clippy::type_complexity, clippy::too_many_arguments)]
pub fn scan_datastore(
    addr: &Address,
    prefix: &[u8],
    start_key: Bound<Vec<u8>>,
    end_key: Bound<Vec<u8>>,
    count: Option<u32>,
    final_state: Arc<RwLock<dyn FinalStateController>>,
    active_history: Arc<RwLock<ActiveHistory>>,
    added_changes: Option<&LedgerChanges>,
) -> (Option<BTreeSet<Vec<u8>>>, Option<BTreeSet<Vec<u8>>>) {
    // get final keys
    let final_keys = final_state.read().get_ledger().get_datastore_keys(
        addr,
        prefix,
        start_key.clone(),
        end_key.clone(),
        count,
    );

    // the iteration range is the intersection of the prefix range and the selection range
    let key_range = range_intersection(
        get_prefix_bounds(prefix),
        (start_key.clone(), end_key.clone()),
    );

    enum SpeculativeResetType {
        None,
        Set,
        Delete,
    }

// … (truncated)
```

## Recommendation
Enforce a strict upper bound on datastore key enumeration across all entry points, and reject or clamp any request that does not provide a valid limit. Ensure the effective capped value is propagated unchanged from RPC/gRPC parsing through `query_state(...)` to `scan_datastore(...)`, and apply the same bound to smart-contract calls instead of allowing `count = None`.

Refactor `scan_datastore(...)` so work is bounded before allocation and collection occur. Avoid branches that clone or collect the full matching key set prior to truncation, and ensure merge logic stops fetching additional final-state batches once the capped number of surviving keys has been reached. If deleted or filtered entries require over-fetching, enforce a hard maximum on total scanned entries as well as returned entries.

Align pricing with actual resource consumption. Smart-contract datastore key queries should charge based on the amount of work performed, including scanning, merging, sorting, and memory allocation, rather than relying only on a fixed host-call cost.

Additionally, impose practical protocol- or implementation-level limits on datastore enumeration size and response size to reduce CPU, memory, and lock-holding impact during query processing.

## Why P0 / breaking

peterjah:
…parsing to the scan

  * [ ] make tests compile
  * [ ] make tests pass 

### DIFF
--- a/massa-execution-worker/src/context.rs
+++ b/massa-execution-worker/src/context.rs
@@ -660,7 +660,7 @@ impl ExecutionContext {
         let max_datastore_query = None;
 
         // cleanup bounds
-        let (prefix, start_key, end_key) = cleanup_datastore_key_range_query(
+        let (prefix, start_key, end_key, count) = cleanup_datastore_key_range_query(
             prefix,
             start_key,
             end_key,
--- a/massa-execution-worker/src/datastore_scan.rs
+++ b/massa-execution-worker/src/datastore_scan.rs
@@ -8,7 +8,7 @@ use std::{
 };
 
 use massa_final_state::FinalStateController;
-use massa_ledger_exports::LedgerChanges;
+use massa_ledger_exports::{LedgerChanges, LedgerEntry};
 use massa_models::{
     address::Address,
     datastore::{get_prefix_bounds, range_intersection},
@@ -68,6 +68,7 @@ pub fn scan_datastore(
     let mut key_updates = BTreeMap::new();
     {
         let mut update_indices = VecDeque::new();
+        let mut reset_entry: Option<&LedgerEntry> = None;
         let history_lock = active_history.read();
 
         let it = history_lock
@@ -84,13 +85,9 @@ pub fn scan_datastore(
 
                 // address ledger entry being reset to an absolute new list of keys
                 Some(SetUpdateOrDelete::Set(v)) => {
-                    if let Some(k_range) = key_range.as_ref() {
-                        key_updates = v
-                            .datastore
-                            .range(k_range.clone())
-                            .map(|(k, _v)| (k.clone(), true))
-                            .collect();
-                    }
+                    // the entry is only scanned once the newer updates are known, so that
+                    // the scan can be bounded instead of materializing the whole range
+                    reset_entry = Some(v);
                     speculative_reset = SpeculativeResetType::Set;
                     break;
                 }
@@ -144,6 +141,27 @@ pub fn scan_datastore(
                 }
             } else {
                 panic!("unexpected state change");
+            }
+        }
+
+        // Scan the keys of a reset entry, now that the newer updates are known.
+        // Only the first `count + key_updates.len()` keys of the entry can matter: the
+        // updates already gathered are the only thing that can delete a key from that
+        // range, so any key past that bound is guaranteed not to reach the final result.
+        // Updates are newer than the reset, so they take precedence over the entry.
+        if let (Some(entry), Some(k_range)) = (reset_entry, key_range.as_ref()) {
+            let base_it = entry.datastore.range(k_range.clone()).map(|(k, _v)| k);
+            match count.map(|cnt| (cnt as usize).saturating_add(key_updates.len())) {
+                Some(limit) => {
+                    for k in base_it.take(limit) {
+                        key_updates.entry(k.clone()).or_insert(true);
+                    }
+                }
+                None => {
+                    for k in base_it {
+                        key_updates.entry(k.clone()).or_insert(true);
+                    }
+                }
             }
         }
     }
@@ -264,6 +282,12 @@ pub fn scan_datastore(
         if final_keys_queue.is_empty() {
             if let Some(last_k) = last_final_batch_key.take() {
                 // the last final item was consumed: replenish the queue by querying more
+                // only the keys still missing to reach `count` are worth fetching
+                let remaining = count.map(|cnt| {
+                    (cnt as u64)
+                        .saturating_sub(speculative_keys.len() as u64)
+                        .max(1) as u32
+                });
                 final_keys_queue = final_state
                     .read()
                     .get_ledger()
@@ -272,7 +296,7 @@ pub fn scan_datastore(
                         prefix,
                         std::ops::Bound::Excluded(last_k),
                         end_key.clone(),
-                        count,
+                        remaining,
                     )
                     .expect("address expected to exist in final state")
                     .iter()
--- a/massa-execution-worker/src/tests/tests_scan_datastore.rs
+++ b/massa-execution-worker/src/tests/tests_scan_datastore.rs
@@ -433,3 +433,141 @@ fn display_bound_human_readable(bound: Bound<Vec<u8>>) {
         Bound::Unbounded => dbg!("bound key Unbounded".to_string()),
     };
 }
+
+/// Bounding the speculative scan must not change what is returned: for any `count`,
+/// the result has to be exactly the first `count` keys of the unbounded result.
+///
+/// The reset (`Set`) path is the interesting one, because deletions applied on top of
+/// the reset entry mean the scan has to look past the first `count` entry keys to
+/// still produce `count` keys.
+#[test]
+fn test_scan_datastore_count_is_a_prefix_of_unbounded() {
+    let mut rng = thread_rng();
+    for _ in 0..20 {
+        scan_datastore_count_prefix_case(rng.gen_range(20..60));
+    }
+}
+
+fn scan_datastore_count_prefix_case(nb_keys: usize) {
+    let keypair = KeyPair::generate(0).unwrap();
+    let addr = Address::from_public_key(&keypair.get_public_key());
+
+    let mut foreign_controllers = ExecutionForeignControllers::new_with_mocks();
+    foreign_controllers
+        .ledger_controller
+        .set_expectations(|ledger_controller| {
+            ledger_controller
+                .expect_get_datastore_keys()
+                .returning(move |_, _, _, _, _| None);
+        });
+    foreign_controllers
+        .final_state
+        .write()
+        .expect_get_ledger()
+        .return_const(Box::new(foreign_controllers.ledger_controller.clone()));
+
+    let mut rng = thread_rng();
+
+    // the reset entry
+    let mut data = BTreeMap::new();
+    for i in 0..nb_keys {
+        // fixed-width keys so that byte order matches the generation order
+        let key = format!("key{:04}", i).into_bytes();
+        let value: Vec<u8> = (0..rng.gen_range(1..10))
+            .map(|_| rng.sample(Alphanumeric) as u8)
+            .collect();
+        data.insert(key, value);
+    }
+    let existing_keys: Vec<_> = data.keys().cloned().collect();
+
+    let mut changes = PreHashMap::default();
+    changes.insert(
+        addr,
+        massa_models::types::SetUpdateOrDelete::Set(LedgerEntry {
+            datastore: data.clone(),
+            ..Default::default()
+        }),
+    );
+
+    // updates newer than the reset: deletions concentrated on the lowest keys, so a
+    // naive `take(count)` on the entry would come up short, plus a few added keys
+    let mut datastore_updates = BTreeMap::new();
+    for key in existing_keys.iter().take(nb_keys / 3) {
+        datastore_updates.insert(key.clone(), massa_models::types::SetOrDelete::Delete);
+    }
+    for i in 0..5 {
+        // "add" sorts before "key", so these land at the front of the range
+        datastore_updates.insert(
+            format!("add{:04}", i).into_bytes(),
+            massa_models::types::SetOrDelete::Set(vec![1, 2, 3]),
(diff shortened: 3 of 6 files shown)


Leo-Besancon (dismissed):
Implementation looks good to me. @peterjah you know more about existing projects than me, even if this is the expected behaviour, do you think it could break any apps that rely on the no limit behaviour?
I assume if that's the case the fix is just to chunk the queries, probably easy to do?

Not that it's an issue, we just have to make a clear breaking changes list for builders like we've done before.

peterjah:
> Implementation looks good to me. @peterjah you know more about existing projects than me, even if this is the expected behaviour, do you think it could break any apps that rely on the no limit behaviour? I assume if that's the case the fix is just to chunk the queries, probably easy to do?
> 
> Not that it's an issue, we just have to make a clear breaking changes list for builders like we've done before.

Actually yes its a breaking change, and quite dangerous one because it breaks "silently", just truncating the 500th first elements...
I am actually afraid it could break deployed apps..
Not sure if it worth to merge this :/

Leo-Besancon:
> > Implementation looks good to me. @peterjah you know more about existing projects than me, even if this is the expected behaviour, do you think it could break any apps that rely on the no limit behaviour? I assume if that's the case the fix is just to chunk the queries, probably easy to do?
> > Not that it's an issue, we just have to make a clear breaking changes list for builders like we've done before.
> 
> Actually yes its a breaking change, and quite dangerous one because it breaks "silently", just truncating the 500th first elements... I am actually afraid it could break deployed apps.. Not sure if it worth to merge this :/

I agree it's a bit scary. Any maintained project should be able to update easily, but any unmaintained project using this behaviour may break. WDYT @damip?

damip:
I think it is worth the risk given the risk of DDoS.
The ones we need to absolutely check are:
- bridge
- MNS
- deweb
Anything else that comes to mind that might be listing massively without chunking?

massa-execution-worker/src/context.rs
```diff
@@ -660,7 +660,7 @@ impl ExecutionContext {
         let max_datastore_query = None;
```
damip, line 660: Verified: keeping `max_datastore_query = None` here makes the new destructured `count` equal to `count.or(None)` — an identity — so the smart-contract path's behavior is byte-identical to before. That's the right call for a PR targeting `main` (bounding the ABI would change execution outcomes and needs a deterministic constant + versioned rollout, as the TODO says).

This is also the residual half of #5059: smart contracts can still trigger unbounded enumeration through this path, including the full speculative-key `.collect()` inside `scan_datastore`. Please make sure this residual is tracked (follow-up issue or unlink #5059 from this PR) since merging will auto-close the issue.

massa-models/src/datastore.rs
```diff
@@ -246,7 +249,7 @@ pub fn cleanup_datastore_key_range_query(
     count: Option<u32>,
     max_datastore_key_length: u8,
     max_datastore_query_config: Option<u32>,
-) -> Result<(Vec<u8>, Bound<Vec<u8>>, Bound<Vec<u8>>), ModelsError> {
+) -> Result<(Vec<u8>, Bound<Vec<u8>>, Bound<Vec<u8>>, Option<u32>), ModelsError> {
     // check item count
     let count = count.or(max_datastore_query_config);
```
damip, line 254: This fallback is what makes the fix work — and it's also where the client-visible semantics change happens: `count: None` previously meant "return all keys" and now means "return up to `max_datastore_query_config` (default 500)", with no truncation indicator in the response. Clients that relied on the old behavior will silently get partial results.

Suggest (a) a changelog/API-docs note telling client authors to paginate via `start_key`/inclusive flags, and (b) mentioning in the function doc that when the operator config is `None` the returned count can still be `None` (unbounded), so the DoS bound relies on the config being set.

damip (changes requested):
## Review (F42 / #5059)

The fix is correct and consensus-safe for what it covers, and CI is fully green. I traced the wiring end to end: `cleanup_datastore_key_range_query` already computed the effective count (`count.or(max_datastore_query_config)`) for validation but discarded it, so JSON-RPC and gRPC forwarded the caller's `None` into `scan_datastore` for an unbounded enumeration. The PR propagates the effective value; both call sites pass the configured cap (`max_datastore_keys_query = 500` by default for API and gRPC), so a `count: None` request is now clamped to 500. A request with count > max already errored before (pre-existing check, unchanged). The smart-contract path stays byte-identical (see inline), which is required on `main`.

Requesting changes for one process item and one documentation item — the code itself looks good:

1. **This PR only fixes one of the three problems in #5059, but it is linked to auto-close it on merge.** F42 also covers (a) the smart-contract path (`Context::get_keys` hardcodes no limit, so SCs still reach `scan_datastore` with `count = None`), (b) the `scan_datastore` internals (the `SpeculativeResetType::Set` branch still `.collect()`s the full matching speculative key set before truncation, and the merge path can over-fetch final-state batches), and (c) work-based pricing. Those are execution-semantics changes that need a versioned rollout on `dev_breaking` — out of scope here, rightly so. Please either unlink the issue from this PR or open a follow-up issue for the consensus-gated residuals before merging, so they don't silently disappear when GitHub closes #5059.

2. **Silent truncation is a client-visible semantics change and should be documented.** `count: None` used to mean "return all keys"; it now means "return up to the configured max (default 500)" with no truncation indicator in the response. This is exactly the error-vs-silent-drop interface decision that routed F42 to an issue. Silent clamping is defensible since pagination exists (`start_key` + inclusive flags), but it needs an explicit changelog / API-docs note telling client authors to paginate — otherwise existing clients that relied on "None = everything" will silently see incomplete key sets.

Minor (no change needed, just noting): if an operator unsets `max_datastore_keys_query`, the fallback is `None` and the path is unbounded again — worth one line in the config comment saying the DoS bound relies on this setting.

peterjah:
> I think it is worth the risk given the risk of DDoS. The ones we need to absolutely check are:
> 
> * bridge
> * MNS
> * deweb
>   Anything else that comes to mind that might be listing massively without chunking?

i will check

peterjah:
@damip @Leo-Besancon 
i digged into it.
It seems the only app that will break is the block explorer MNS resolution feature. I will prepare a fix there.
Another thing that worth to be pointed is about the  MAX_DATASTORE_KEYS_QUERY = 500 
The value is configurable in node operator config but is not exposed afterwise in the API.
To make something clean, consumer (massa-web3) needs the configured value to use the endpoint properly.
As of today the value 500 is hardcoded in massa-web3... so if any node operator set a lower value, the endpoint called from massa-web3 will error

peterjah:
Followup issue created: https://github.com/massalabs/massa/issues/5241

Issue to expose the configured MAX_DATASTORE_KEYS value:
https://github.com/massalabs/massa/issues/5228

peterjah:
I also prepared a doc PR in massa docs
https://github.com/massalabs/docs/pull/461

Leo-Besancon (approved):
LGTM on this PR and the followup issues you created, but I'd be more confortable with a second opinion as it's quite impactful, so don't merge this yet please :).

bilboquet:
This PR looks good to me.
What I'm not sure about is if https://github.com/massalabs/massa/issues/5241 and https://github.com/massalabs/massa/issues/5228 are follows up or are sub tasks that have to be done and merge at the same as this PR.

bilboquet (approved):

damip (commented):
Checked the new commit (b8964548) since it touches `scan_datastore` which is shared with the SC path: on `main` this is only OK if the output is byte-identical. It is. The `Set` bounding is correct (among the first `count + |updates|` entry keys at most `|updates|` can be overridden, so the count-th surviving key is always within that bound, and updates still win via `or_insert`). With `count = None` the walk is unbounded so the SC path is unchanged. The replenish change is also output-preserving (sequential merge, refill happens as soon as the queue empties, `.max(1)` avoids a 0-size fetch being read as exhaustion). The prefix-of-unbounded test is the right property. Two nits inline.

@bilboquet #5241 and #5228 are follow-ups, not blockers. #5241 is consensus-observable and goes through a MIP on dev_breaking, that's why it was split out. #5228 is a nicety: massa-web3 already sends `count = 500` explicitly so it already errors on a node with a lower cap, this PR doesn't change that, and everyone runs the default anyway.

@peterjah before merging massalabs/docs#461, double-check the "MAIN.6.0 / DEVN.31.0" claim matches the release this actually ships in.

massa-execution-worker/src/datastore_scan.rs
```diff
@@ -264,6 +282,12 @@ pub fn scan_datastore(
         if final_keys_queue.is_empty() {
             if let Some(last_k) = last_final_batch_key.take() {
                 // the last final item was consumed: replenish the queue by querying more
+                // only the keys still missing to reach `count` are worth fetching
+                let remaining = count.map(|cnt| {
```
damip, line 286: Correct but low value: the old `count` batch was already bounded, so this saves at most one partial batch. And it has a downside: once `speculative_keys.len()` hits `count - 1`, every speculatively-deleted final key ahead costs one lock + RocksDB seek for a single key (up to ~500 single-key fetches per API call instead of one batch). Not a DoS (the deletes cost gas) but a regression in the adversarial case. I'd revert this part to `count` (also removes the need for a dedicated test). If you want to keep it, add a floor (e.g. `.max(64)`).

massa-execution-worker/src/tests/tests_scan_datastore.rs
```diff
@@ -433,3 +433,141 @@ fn display_bound_human_readable(bound: Bound<Vec<u8>>) {
...
 }
+
+/// Bounding the speculative scan must not change what is returned: for any `count`,
+/// the result has to be exactly the first `count` keys of the unbounded result.
+///
+/// The reset (`Set`) path is the interesting one, because deletions applied on top of
+/// the reset entry mean the scan has to look past the first `count` entry keys to
+/// still produce `count` keys.
+#[test]
+fn test_scan_datastore_count_is_a_prefix_of_unbounded() {
```
damip, line 444: Good test, exactly the property we need. It only covers the `Set` path though. If the `remaining` change in the replenish loop stays, add the same prefix-of-unbounded check for the merge path (mock `get_datastore_keys` honoring the start bound and the count, so the refill is actually exercised).

PR #5191, opened 2026-08-20
[F2] enforce max_arguments on the caller-supplied operation ids in ge…

### LINKED ISSUE #5047: [F2] Private JSON-RPC node bans bypass gRPC batch guards

---

## Source
AI security-scan **Finding 2** (severity: **minor**). Tracked in `SUBMISSIONS.md` as F2.

**Affected file:** `massa-api/src/private.rs`

## Summary
The confirmed issue is in the public JSON-RPC `get_operations` flow in `massa-main/massa-api/src/public.rs`. The handler applies `api_cfg.max_arguments` only after filtering out missing operation IDs, so an oversized request containing many syntactically valid but absent IDs can bypass the configured count cap while still triggering parsing and storage lookup work. The original claim about the private `node_ban_by_id` path is not retained as a security finding: that interface is admin-only, bound to localhost by default, and the cited empty-batch behavior is only a no-op mismatch with gRPC.

## Root cause
In `API<Public>::get_operations`, the JSON-RPC path builds the result set from storage first and only then checks the limit using the number of found operations, rather than the number of user-supplied IDs. Concretely, the code performs `operations_ids.iter().filter_map(|id| read_ops.get(id).cloned()).collect()` and derives `ops`, then enforces `if ops.len() as u64 > api_cfg.max_arguments`, which means absent IDs are discarded before the cap is evaluated.

Because of this ordering, the effective limit applies to returned matches instead of the raw request size. An oversized batch of syntactically valid but nonexistent operation IDs therefore avoids the intended argument-count restriction and still consumes request processing resources, subject to `max_request_body_size`.

The original root-cause theory for `node_ban_by_id` is ruled out as a meaningful vulnerability. Although `massa-main/massa-api/src/private.rs` does not mirror the gRPC empty-batch and length checks before calling `ban_peers`, that path is a privileged private interface, so the inconsistency does not create a material new exploit path beyond already authorized administrative access.

## Affected code
```rust
async fn node_ban_by_id(&self, ids: Vec<NodeId>) -> RpcResult<()> {
    let protocol_controller = self.0.protocol_controller.clone();
    let peer_ids = ids
        .into_iter()
        .map(|id| PeerId::from_public_key(id.get_public_key()))
        .collect();
    protocol_controller
        .ban_peers(peer_ids)
        .map_err(|e| ApiError::ProtocolError(e.to_string()).into())
}

pub(crate) fn ban_nodes_by_ids(
    grpc: &MassaPrivateGrpc,
    request: tonic::Request<grpc_api::BanNodesByIdsRequest>,
) -> Result<grpc_api::BanNodesByIdsResponse, GrpcError> {
    let node_ids = request.into_inner().node_ids;

    if node_ids.is_empty() {
        return Err(GrpcError::InvalidArgument(
            "no node id received".to_string(),
        ));
    }

    if node_ids.len() as u64 > grpc.grpc_config.max_arguments {
        return Err(GrpcError::InvalidArgument(format!(
            "too many node ids received. Only a maximum of {} node ids are accepted per request",
            grpc.grpc_config.max_arguments
        )));
    }

    let peer_ids = node_ids
        .into_iter()
        .map(|id| {
            NodeId::from_str(&id).map(|node_id| PeerId::from_public_key(node_id.get_public_key()))
        })
        .collect::<Result<Vec<_>, _>>()?;

    grpc.protocol_controller.ban_peers(peer_ids)?;

    Ok(grpc_api::BanNodesByIdsResponse {})
}

fn ban_peers(&self, peer_ids: Vec<PeerId>) -> Result<(), ProtocolError> {
    self.sender_peer_management_thread
        .as_ref()
// … (truncated)
```

## Recommendation
Enforce `api_cfg.max_arguments` against the number of user-supplied operation IDs before any storage access, filtering, or result construction in the public JSON-RPC `get_operations` flow. The request should be rejected based on the raw input size, not the number of operations found in storage. This aligns the effective limit with its intended purpose of bounding per-request processing cost and prevents oversized batches of valid-but-absent IDs from consuming unnecessary resources. Additionally, review JSON-RPC and gRPC batch-validation behavior to ensure consistent limits and input-handling semantics across interfaces.

## Why P0 / breaking

peterjah:
…t_operations

  * [ ] make tests compile
  * [ ] make tests pass 

### DIFF
--- a/massa-api/src/public.rs
+++ b/massa-api/src/public.rs
@@ -662,6 +662,12 @@ impl MassaRpcServer for API<Public> {
         &self,
         operations_ids: Vec<OperationId>,
     ) -> RpcResult<Vec<OperationInfo>> {
+        let api_cfg = self.0.api_settings.clone();
+        // bound the request size on the ids provided by the caller, before any storage lookup
+        if operations_ids.len() as u64 > api_cfg.max_arguments {
+            return Err(ApiError::BadRequest("too many arguments".into()).into());
+        }
+
         // get the operations and the list of blocks that contain them from storage
         let secure_share_operations: Vec<SecureShareOperation> = {
             let read_ops = self.0.storage.read_operations();
@@ -690,11 +696,6 @@ impl MassaRpcServer for API<Public> {
 
         // keep only the ops id (found in storage)
         let ops: Vec<OperationId> = storage_info.iter().map(|(op, _)| op.id).collect();
-
-        let api_cfg = self.0.api_settings.clone();
-        if ops.len() as u64 > api_cfg.max_arguments {
-            return Err(ApiError::BadRequest("too many arguments".into()).into());
-        }
 
         // ask pool whether it carries the operations
         let in_pool = self


Leo-Besancon (approved):
LGTM

PR #5198, opened 2026-08-21
[F9] feat: add max elements count for bootstrap batch

### LINKED ISSUE #5052: [F9] `BootstrapPart` new-element parsing can be weaponized into a heap-exhaustion DoS

---

## Source
AI security-scan **Finding 9** (severity: **minor**). Tracked in `SUBMISSIONS.md` as F9.

## Summary
In `massa-main/massa-bootstrap/src/messages.rs`, `MessageServerTypeId::FinalStatePart` deserializes both `state_part.new_elements` and `versioning_part.new_elements` with `length_value(..., many0(tuple((key, value))))` before converting the resulting vectors into maps. This creates a real denial-of-service condition because parsing is bounded only by raw byte length, not by entry count, and `VecU8Deserializer` permits zero-length keys and values. A malicious bootstrap server already trusted through the configured `bootstrap_list` can therefore send a single validly signed `BootstrapPart` containing tens of millions of tiny pairs and force extreme memory and CPU consumption during bootstrap. The original claim is only overstated insofar as it asserted that zero-length entries also produce a second comparably large `BTreeMap` allocation during `.collect()`.

## Root cause
The confirmed root cause is the unbounded element count in the `new_elements` parsing path for `MessageServerTypeId::FinalStatePart`. In `massa-main/massa-bootstrap/src/messages.rs`, both `state_part.new_elements` and `versioning_part.new_elements` are decoded with `length_value(..., many0(tuple((key, value))))`, so the parser eagerly materializes all parsed pairs into a `Vec` based only on byte-size caps of roughly `100 MB` and `10 MB`, rather than enforcing a maximum number of entries.

Because the key and value deserializers accept empty byte arrays, an attacker can pack roughly `55 million` tiny pairs into a single signed bootstrap message, causing `many0` to allocate and retain an enormous intermediate vector and consume substantial CPU while parsing. The ruled-out portion of the original root-cause theory is the claim that `.collect::<BTreeMap<_, _>>()` necessarily creates a second similarly sized allocation for the cited `2`-byte empty-key entries: when keys are all empty, those entries collapse into a single map entry instead of producing a second map proportional to the vector size.

## Affected code
```rust
fn deserialize<'a, E: nom::error::ParseError<&'a [u8]> + nom::error::ContextError<&'a [u8]>>(
        &self,
        buffer: &'a [u8],
    ) -> IResult<&'a [u8], BootstrapServerMessage, E> {
        context("Failed BootstrapServerMessage deserialization", |buffer| {
            let (input, id) = context("Failed id deserialization", |input| {
                self.message_id_deserializer.deserialize(input)
            })
            .map(|id| {
                MessageServerTypeId::try_from(id).map_err(|_| {
                        buffer,
                        nom::error::ErrorKind::Eof,
                    ))
                })
            })
            .parse(buffer)?;
            match id? {
                MessageServerTypeId::BootstrapTime => tuple((
                    context("Failed server_time deserialization", |input| {
                        self.time_deserializer.deserialize(input)
                    }),
                    context("Failed version deserialization", |input| {
                        self.version_deserializer.deserialize(input)
                    }),
                ))
                .map(
                    |(server_time, version)| BootstrapServerMessage::BootstrapTime {
                        server_time,
                        version,
                    },
                )
                .parse(input),
                MessageServerTypeId::Peers => context("Failed peers deserialization", |input| {
                    self.peers_deserializer.deserialize(input)
                })
                .map(|peers| BootstrapServerMessage::BootstrapPeers { peers })
                .parse(input),
                MessageServerTypeId::FinalStatePart => tuple((
                    }),
                    context(
                        "Failed state_part deserialization",
                        tuple((
// … (truncated)
```

## Recommendation
Enforce an explicit maximum number of entries for both `state_part.new_elements` and `versioning_part.new_elements` during deserialization, rather than relying solely on the total byte-length bound. Reject payloads that exceed this entry cap before materializing all pairs in memory, and validate that parsed elements make forward progress to avoid abuse with minimal-length entries. If empty keys or values are not required by the protocol, disallow them to further reduce the attack surface. Additionally, prefer a parsing approach that avoids building a large intermediate `Vec` before conversion, so resource usage remains bounded even when processing malformed or adversarial bootstrap messages from trusted peers in `bootstrap_list`.

## Why P0 / breaking

Leo-Besancon:
  * [ ] make tests compile
  * [ ] make tests pass 

Note: This PR does not solve the issue for `updates_on_previous_elements`, which shows the same issue, but are a lot trickier to solve with no false positive as the bootstrap server sends all the updates available at the time in the same batch.
The finding specifically targets `new_elements` for this reason.
Moreover, as the server is trusted by definition, the attack surface is minimal.


massa-bootstrap/src/tests/config.rs
```diff
@@ -61,6 +61,8 @@ impl Default for BootstrapConfig {
             max_bootstrap_error_length: MAX_BOOTSTRAP_ERROR_LENGTH,
             max_final_state_elements_size: MAX_BOOTSTRAP_FINAL_STATE_PARTS_SIZE,
             max_versioning_elements_size: MAX_BOOTSTRAP_VERSIONING_ELEMENTS_SIZE,
+            max_final_state_elements_count: 100_000,
```
bilboquet: why not a constant like other fields?

bilboquet (dismissed):

massa-bootstrap/src/tests/config.rs
Leo-Besancon: no reason :). Updated in the latest commit, thanks!

bilboquet (approved):

PR #5199, opened 2026-08-21
docs(execution): note that send_message admits async messages whose m…

### LINKED ISSUE #5005: [F48] send_message Allows Admission of Permanently Unexecutable Async Messages Due to Missing Maximum Gas Validation

---

## Source
AI security-scan **Finding 48** (severity: **minor**). Tracked in `SUBMISSIONS.md` as F48.

**Affected file:** `massa-execution-worker/src/interface_impl.rs`

## Summary
The async message flow accepts messages in `send_message`, then later tries to schedule them in `take_batch_to_execute`. A message is scheduled only if `message.max_gas + async_msg_cst_gas_cost` fits inside the async gas budget available for a slot. However, `send_message` checks only a *minimum* `max_gas` and does not reject values that are higher than any slot can ever schedule under `max_async_gas + max_gas_per_block`. As a result, some accepted messages are *permanently unexecutable*, while the sender still pays the async fee and the message keeps occupying async-pool capacity until expiration.

## Root cause
The main issue is a validation gap between message admission in `send_message` and execution scheduling in `take_batch_to_execute`.

- In `take_batch_to_execute`, a message is scheduled only when ``message.max_gas + async_msg_cst_gas_cost`` is within the slot's available async gas budget. That budget is globally bounded by ``max_async_gas + max_gas_per_block``, and it can be lower when deferred calls already consume part of ``max_async_gas``.
- In `send_message`, the code enforces only a *minimum* value for ``max_gas``. It does not reject messages whose ``max_gas`` is greater than the maximum gas any slot can ever allocate.
- Because of this mismatch, `send_message` can admit messages that `take_batch_to_execute` can never schedule.

## Affected code
```rust
fn send_message(
        &self,
        target_address: &str,
        target_function: &str,
        validity_start: (u64, u8),
        validity_end: (u64, u8),
        max_gas: u64,
        raw_fee: u64,
        raw_coins: u64,
        data: &[u8],
        filter: Option<(&str, Option<&[u8]>)>,
    ) -> Result<()> {
        if validity_start.1 >= self.config.thread_count {
            bail!("validity start thread exceeds the configuration thread count")
        }
        if validity_end.1 >= self.config.thread_count {
            bail!("validity end thread exceeds the configuration thread count")
        }

        let target_addr = Address::from_str(target_address).map_err(|e| e.to_string())?;

        // check that the target address is an SC address
        if !matches!(target_addr, Address::SC(..)) {
            bail!("target address is not a smart contract address")
        }

        // Length verifications
        if target_function.len() > self.config.max_function_length as usize {
            bail!("Function name is too large");
        }
        if data.len() > self.config.max_parameter_length as usize {
            bail!("Parameter size is too large");
        }

        let mut execution_context = context_guard!(self);
        let emission_slot = execution_context.slot;

        if max_gas < self.config.gas_costs.max_instance_cost {
            bail!("max gas is lower than the minimum instance cost")
        }
        if Slot::new(validity_end.0, validity_end.1) < Slot::new(validity_start.0, validity_start.1)
        {
            bail!("validity end is earlier than the validity start")
        }
        if Slot::new(validity_end.0, validity_end.1) < emission_slot {
// … (truncated)
```

## Recommendation
Validate `max_gas` in `send_message` against the maximum gas that can ever be scheduled by `take_batch_to_execute`, not only against the minimum bound. Reject any message whose `max_gas + async_msg_cst_gas_cost` exceeds the highest schedulable per-slot budget, taking into account the protocol limits derived from `max_async_gas + max_gas_per_block`. This ensures only executable async messages are admitted, preventing silent fee loss and unnecessary async-pool occupancy.

If deferred execution can reduce the effective async gas available for a slot, document and enforce a conservative upper bound that remains schedulable under all valid conditions, or otherwise align admission rules with the exact scheduling constraints so accepted messages cannot become permanently unexecutable.

## Why P0 / breaking

peterjah:
…ax_gas exceeds any schedulable slot budget

Status: documented, not fixed. No behavior change; no MIP required.
  
   Confirmed the finding is real: send_message bounds max_gas only from below, while take_batch_to_execute needs max_gas + async_msg_cst_gas_cost to fit a budget capped at max_async_gas + max_gas_per_block
   (≈5.29e9). Above that ceiling a message is admitted but never schedulable, and cancel_async_message refunds coins but not fee at expiry — so the fee is genuinely burned.
   
   Adding an max gas check would be breaking. So better keep it like it is imho

### DIFF
--- a/massa-execution-worker/src/context.rs
+++ b/massa-execution-worker/src/context.rs
@@ -1440,7 +1440,6 @@ impl ExecutionContext {
     }
 
     /// Returns `true` if the execution component version is at least `version`.
-    #[allow(dead_code)]
     pub fn is_execution_component_version_at_least(&self, version: u32) -> bool {
         self.execution_component_version >= version
     }
--- a/massa-execution-worker/src/interface_impl.rs
+++ b/massa-execution-worker/src/interface_impl.rs
@@ -55,6 +55,11 @@ use tracing::warn;
 ))]
 use massa_models::datastore::Datastore;
 
+/// Execution component version at which `send_message` starts rejecting messages whose
+/// `max_gas` is too large to ever be scheduled (see [`InterfaceImpl::send_message`]).
+/// Bundled into the same MIP as the WMAS patch (`MIP-0002-BugFix`).
+const ASYNC_MSG_MAX_GAS_CHECK_EXEC_VERSION: u32 = 2;
+
 /// helper for locking the context mutex
 macro_rules! context_guard {
     ($self:ident) => {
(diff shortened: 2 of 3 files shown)


Leo-Besancon:
> Adding an max gas check would be breaking. So better keep it like it is imho

I agree it's not super critical so we can skip it, but IMO even with versioning the fix should be very easy (`if version > 2 && message.gas > [...] { return Err(...)`) no?

peterjah:
> > Adding an max gas check would be breaking. So better keep it like it is imho
> 
> I agree it's not super critical so we can skip it, but IMO even with versioning the fix should be very easy (`if version > 2 && message.gas > [...] { return Err(...)`) no?

Lets do it !

Leo-Besancon (approved):
LGTM! just check the CI

PR #5200, opened 2026-08-24
[F8] bound bootstrap accepts per poll cycle and cap concurrent refusa…

### LINKED ISSUE #5051: [F8] The listener accepts the full backlog before enforcing any admission limits

---

## Source
AI security-scan **Finding 8** (severity: **medium**). Tracked in `SUBMISSIONS.md` as F8.

**Affected file:** `massa-bootstrap/src/listener.rs`

## Summary
`listener::poll` in `massa-bootstrap/src/listener.rs` accepts every pending connection in a loop until `WouldBlock`, converts each accepted socket into a blocking `TcpStream`, and returns the full batch before `server::event_loop` applies admission controls. This creates a real resource-exhaustion condition because a backlog flood can temporarily move many sockets into process-owned file descriptors and trigger one refusal thread per rejected connection through `close_and_send_error`. The original claim is only partially supported, however, because this behavior does *not* remove the `max_bootstraps` limit on active bootstrap sessions. The additional claim that this helps trigger an "infinite-error condition" is not supported by the verified code.

## Root cause
The confirmed issue is caused by `listener::poll` accepting connections in the `NEW_CONNECTION` branch with a `loop` around `self.server.accept()`, pushing each `(stream, remote_addr)` into `results`, and stopping only when `ErrorKind::WouldBlock` is reached. Each accepted socket is converted with `mio_stream_to_std(stream)` and `stream.set_nonblocking(false)?` before any whitelist, blacklist, per-IP throttling, or `max_bootstraps` enforcement occurs in `server::event_loop`.

Because admission checks are deferred until after the entire pending backlog has been accepted, an attacker can force the process to briefly own a large burst of sockets even when many or all of them will later be rejected. On those rejection paths, `close_and_send_error` creates an additional thread per refused connection, which compounds resource usage.

The ruled-out portion of the original theory is the statement that this behavior "bypasses the intended concurrency cap." The verified code still enforces `max_bootstraps` for active bootstrap sessions; the flaw is the uncapped burst of accepted and then rejected connections around that cap, not unlimited concurrent bootstrap sessions.

## Affected code
```rust
pub fn poll(&mut self) -> Result<PollEvent, BootstrapError> {
        self.poll.poll(&mut self.events, None).unwrap();

        let mut results = Vec::with_capacity(self.events.iter().count());

        // Process each event.
        for event in self.events.iter() {
            match event.token() {
                NEW_CONNECTION => loop {
                    match self.server.accept() {
                        Ok((mut stream, remote_addr)) => {
                            let _ = self.poll.registry().deregister(&mut stream);
                            let stream: std::net::TcpStream = mio_stream_to_std(stream);
                            stream.set_nonblocking(false)?;
                            results.push((stream, remote_addr));
                        }
                        Err(ref e) if e.kind() == ErrorKind::WouldBlock => {
                            break;
                        }
                        Err(e) => {
                            warn!("Error accepting connection in bootstrap: {:?}", e);
                            continue;
                        }
                    }
                },
                STOP_LISTENER => {
                    return Ok(PollEvent::Stop);
                }
                _ => unreachable!(),
            }
        }

        Ok(PollEvent::NewConnections(results))
    }
```

## Recommendation
Enforce admission limits before draining the entire pending backlog. Bound the number of connections accepted per poll cycle, and apply early filtering such as whitelist, blacklist, per-IP throttling, and `max_bootstraps` checks as close to `accept()` as possible so excess connections are rejected before they are converted and accumulated in process-owned resources.

Reduce the cost of rejection paths as well. Avoid spawning one thread per refused connection in `close_and_send_error`, and use a bounded or asynchronous mechanism for refusals so a flood of rejected sockets cannot amplify resource consumption. Also consider keeping accepted sockets non-blocking until they pass admission checks, and add explicit limits, backpressure, and monitoring for file descriptors, pending accepts, and rejection handling to make backlog floods degrade gracefully.

## Why P0 / breaking

peterjah:
…l threads

  * [ ] make tests compile
  * [ ] make tests pass 

### DIFF
--- a/massa-bootstrap/src/bindings/server.rs
+++ b/massa-bootstrap/src/bindings/server.rs
@@ -19,6 +19,7 @@ use massa_serialization::{DeserializeError, Deserializer, Serializer};
 use massa_signature::KeyPair;
 use massa_time::MassaTime;
 use std::io;
+use std::sync::atomic::{AtomicUsize, Ordering};
 use std::time::Instant;
 use std::{
     convert::TryInto,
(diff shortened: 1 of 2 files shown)


Leo-Besancon (approved):
LGTM, similarly to an other PR we could potentially put the constants in config.toml, but as the limits are large on purpose and don't have a lot of impact (at most the other connections will be captured at the next loop), I don't mind!

PR #5201, opened 2026-08-24
[F10] validate the network restart metadata sent by a bootstrap serve…

### LINKED ISSUE #5053: [F10] Insufficient Validation of Bootstrap Restart Metadata Enables Remote Denial of Service

---

## Source
AI security-scan **Finding 10** (severity: **medium**). Tracked in `SUBMISSIONS.md` as F10.

## Summary
The `BootstrapPart` branch accepts `last_start_period: Option<u64>` and `last_slot_before_downtime: Option<Option<Slot>>` and only checks their primitive types. It does not validate whether these two fields are consistent with later slot and timestamp logic. `stream_final_state_and_consensus` stores the received values as-is, and later startup code uses them with `unwrap()` and `expect(...)`. A malicious bootstrap server can therefore send invalid values and make the client panic after an apparently successful bootstrap, causing a reliable *remote denial of service*.

## Root cause
The main issue is missing validation in the `BootstrapPart` branch for the fields `last_start_period: Option<u64>` and `last_slot_before_downtime: Option<Option<Slot>>`. These values are accepted with only primitive type bounds, but the code does not check important cross-field invariants before they are used later.

- In `stream_final_state_and_consensus`, the received `last_start_period` and `last_slot_before_downtime` are stored verbatim. There is no validation that `last_start_period` can be safely represented in later timestamp computations.

- The code also does not validate that `last_slot_before_downtime.get_next_slot(thread_count)` exists, or that the downtime interval is coherent before later startup logic relies on it.

- Because these fields are not validated early, later startup code reaches unsafe assumptions and calls `get_next_slot(...).unwrap()`, `get_prev_slot(...).unwrap()`, `get_block_slot_timestamp(...).expect(...)`, and shutdown-period consistency checks. If the bootstrap data is inconsistent or impossible, these calls panic.

- The input explicitly allows dangerous values such as `last_start_period = u64::MAX` or an impossible `last_slot_before_downtime`, which can pass through bootstrap handling and only fail later during startup.

## Affected code
```rust
fn deserialize<'a, E: nom::error::ParseError<&'a [u8]> + nom::error::ContextError<&'a [u8]>>(
        &self,
        buffer: &'a [u8],
    ) -> IResult<&'a [u8], BootstrapServerMessage, E> {
        context("Failed BootstrapServerMessage deserialization", |buffer| {
            let (input, id) = context("Failed id deserialization", |input| {
                self.message_id_deserializer.deserialize(input)
            })
            .map(|id| {
                MessageServerTypeId::try_from(id).map_err(|_| {
                        buffer,
                        nom::error::ErrorKind::Eof,
                    ))
                })
            })
            .parse(buffer)?;
            match id? {
                MessageServerTypeId::BootstrapTime => tuple((
                    context("Failed server_time deserialization", |input| {
                        self.time_deserializer.deserialize(input)
                    }),
                    context("Failed version deserialization", |input| {
                        self.version_deserializer.deserialize(input)
                    }),
                ))
                .map(
                    |(server_time, version)| BootstrapServerMessage::BootstrapTime {
                        server_time,
                        version,
                    },
                )
                .parse(input),
                MessageServerTypeId::Peers => context("Failed peers deserialization", |input| {
                    self.peers_deserializer.deserialize(input)
                })
                .map(|peers| BootstrapServerMessage::BootstrapPeers { peers })
                .parse(input),
                MessageServerTypeId::FinalStatePart => tuple((
                    }),
                    context(
                        "Failed state_part deserialization",
                        tuple((
// … (truncated)
```

## Recommendation
Validate `last_start_period` and `last_slot_before_downtime` during bootstrap deserialization or immediately before storing them in `stream_final_state_and_consensus`. Enforce all cross-field invariants required by subsequent startup logic, including that the restart metadata is internally consistent, any referenced next or previous slot exists for the configured `thread_count`, and any derived timestamp or period computation remains within valid bounds. Reject bootstrap data that violates these constraints and treat it as a bootstrap failure rather than allowing it to reach later startup stages. Additionally, remove reliance on `unwrap()` and `expect(...)` for this metadata path and replace those assumptions with explicit error handling so malformed restart metadata cannot trigger a panic.

## Why P0 / breaking

peterjah:
…r before storing it

  * [ ] make tests compile
  * [ ] make tests pass 


massa-bootstrap/src/client.rs
```diff
@@ -161,6 +170,33 @@ pub(crate) fn stream_final_state_and_consensus(
...
+                    if let Some(last_slot_before_downtime) = *guard.get_last_slot_before_downtime()
+                    {
+                        let (shutdown_start, shutdown_end) = shutdown_range(
+                            cfg,
+                            guard.get_last_start_period(),
+                            last_slot_before_downtime,
+                        )?;
+                        guard
+                            .get_mip_store()
+                            .is_consistent_with_shutdown_period(
```
Leo-Besancon, line 185: Note that this is also checked in main (L. 458), I think we should remove the one in main (and potentially add a comment somewhere to say it has already been checked during bootstrap).

massa-bootstrap/src/client.rs
```diff
@@ -337,6 +373,83 @@ pub(crate) fn bootstrap_from_server(
...
+            last_slot_before_downtime, last_start_period
+        )));
+    }
+
+    shutdown_range(cfg, last_start_period, last_slot_before_downtime).map(|_| ())
+}
+
+/// Computes the slot range of the last network shutdown, as startup does: from the slot right
+/// after the last one executed before the downtime, to the slot right before the restart.
+fn shutdown_range(
```
Leo-Besancon, line 429: Use this function in main to ensure consistency and avoiding code duplication

Leo-Besancon (commented):
Two small nitpicks, otherwise the logic LGTM

massa-bootstrap/src/client.rs
peterjah: done!

Leo-Besancon (approved):
Thanks! LGTM

PR #5202, opened 2026-08-24
[F136] reject future-dated peer announcements instead of treating the…

### LINKED ISSUE #5045: [F136] Future-dated announcements bypass the 3-day freshness policy indefinitely

---

## Source
AI security-scan **Finding 136** (severity: **medium**). Tracked in `SUBMISSIONS.md` as F136.

**Affected file:** `massa-protocol-worker/src/handlers/peer_handler/models.rs`

## Summary
`get_rand_peers_to_send` in `massa-main/massa-protocol-worker/src/handlers/peer_handler/models.rs` treats `last_announce.timestamp` as the peer freshness signal and only excludes entries older than `now - THREE_DAYS_MS`. Because peer announcements carry an attacker-controlled timestamp that is accepted during `perform_handshake` without a sanity check against local time, a far-future announcement remains eligible for peer sharing and bootstrap responses for an arbitrarily long time. The original finding is correct on this freshness-bypass behavior, but it overstated persistence by claiming the poisoned value cannot be replaced through later handshake updates.

## Root cause
The confirmed issue is the combination of two behaviors:

- `perform_handshake` in `massa-main/massa-protocol-worker/src/handlers/peer_handler/mod.rs` accepts a remote peer's signed `Announcement` and stores it as `last_announce` without validating whether `announcement.timestamp` is near current wall-clock time or excessively future-dated.
- `get_rand_peers_to_send` only checks whether `last_announce.timestamp < now - THREE_DAYS_MS`, so any timestamp far in the future is treated as fresh and its `listeners` remain eligible for propagation.

The original claim about the replacement mechanism was only partially correct. The strictly-greater timestamp rule exists in the tester path in `tester.rs`, but it does *not* generally apply to handshake updates: `perform_handshake` overwrites `info.last_announce` on a later successful handshake, so a future-dated record is not inherently pinned across all update paths.

## Affected code
```rust
fn get_rand_peers_to_send(
        &self,
        nb_peers: usize,
    ) -> Vec<(PeerId, HashMap<SocketAddr, TransportType>)> {
        //TODO: Add ourself
        let now = MassaTime::now().as_millis();

        let min_time = now - THREE_DAYS_MS;

        let mut keys = self.peers.keys().cloned().collect::<Vec<_>>();
        let mut rng = rand::thread_rng();
        keys.shuffle(&mut rng);

        let mut result = Vec::new();

        for key in keys {
            if result.len() >= nb_peers {
                break;
            }
            if let Some(peer) = self.peers.get(&key) {
                // skip old peers
                if let Some(last_announce) = &peer.last_announce {
                    if last_announce.timestamp < min_time {
                        continue;
                    }
                    let listeners: HashMap<SocketAddr, TransportType> =
                        last_announce.listeners.clone().into_iter().collect();
                    if listeners.is_empty() {
                        continue;
                    }
                    result.push((key, listeners));
                }
            }
        }

        result
    }
```

## Recommendation
Validate announcement timestamps during `perform_handshake` against the local clock and reject or clamp announcements that are excessively future-dated. Enforce a bounded acceptable clock-skew window in both directions so only reasonably recent announcements are stored as `last_announce`. As a defense in depth measure, update peer-selection and bootstrap-sharing logic to treat timestamps later than `now + allowed_skew` as invalid or ineligible, rather than relying solely on the check against `now - THREE_DAYS_MS`. Ensure the same time-validity rules are applied consistently wherever peer announcements are accepted, refreshed, or propagated.

## Why P0 / breaking

peterjah:
…m as fresh

  * [ ] make tests compile
  * [ ] make tests pass 

### DIFF
--- a/massa-protocol-worker/src/handlers/peer_handler/announcement.rs
+++ b/massa-protocol-worker/src/handlers/peer_handler/announcement.rs
@@ -24,6 +24,12 @@ use massa_serialization::{
     U64VarIntSerializer,
 };
 
+/// Maximum tolerated clock skew between our local clock and the timestamp carried by a
+/// remote peer's announcement. The timestamp is attacker-controlled and is used as the peer
+/// freshness signal, so an announcement dated arbitrarily far in the future would otherwise
+/// stay "fresh" (and thus eligible for peer sharing and bootstrap responses) forever.
+pub const MAX_ANNOUNCEMENT_CLOCK_SKEW_MS: u64 = 5 * 60 * 1_000;
+
 #[derive(Clone, Debug, PartialEq, Eq)]
 pub struct Announcement {
     /// Listeners
@@ -170,6 +176,12 @@ impl Deserializer<Announcement> for AnnouncementDeserializer {
 }
 
 impl Announcement {
+    /// Check the announcement timestamp against the local clock: a peer cannot legitimately
+    /// announce more than `MAX_ANNOUNCEMENT_CLOCK_SKEW_MS` ahead of us.
+    pub fn is_future_dated(&self, now_millis: u64) -> bool {
+        self.timestamp > now_millis.saturating_add(MAX_ANNOUNCEMENT_CLOCK_SKEW_MS)
+    }
+
     pub fn new(
         listeners: HashMap<SocketAddr, TransportType>,
         routable_ip: Option<IpAddr>,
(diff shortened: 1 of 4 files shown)


Leo-Besancon (dismissed):
The fix looks good to me, the only comment I have is wondering if `MAX_ANNOUNCEMENT_CLOCK_SKEW_MS` should stay as a constant or maybe be set in the config.toml. But IMO as this is only used to ensure "announcement" freshness, it seems it's not very critical so can stay like that if we don't have a good reason to put it in config.

Do you agree @peterjah?

peterjah:
Yes, i don't see any benefits of put this in the config

Leo-Besancon (approved):

PR #5203, opened 2026-08-24
[F134] filter non-routable announced peer endpoints before probing an…

### LINKED ISSUE #5044: [F134] Signed announcements can advertise arbitrary third-party endpoints

---

## Source
AI security-scan **Finding 134** (severity: **medium**). Tracked in `SUBMISSIONS.md` as F134.

**Affected file:** `massa-protocol-worker/src/handlers/peer_handler/announcement.rs`

## Summary
`perform_handshake` accepts a peer's signed `Announcement.listeners` and treats those addresses as trusted peer data without applying any address-policy filtering. Those listener addresses are stored in `last_announce`, forwarded through `PeerManagementMessage::NewPeerConnected`, later re-queued from `PeerManagementMessage::ListPeers`, and eventually probed by `tester.rs` and re-gossiped to other peers. The confirmed issue is therefore the lack of validation for advertised endpoint classes before probing and redistribution. The original claim that the contacted socket must appear in, or be cryptographically bound to, `announcement.listeners` is not supported because the protocol intentionally allows announced addresses to differ from the actual socket tuple.

## Root cause
The root cause is that deserialized listener addresses are accepted as signed metadata and propagated through the peer-management pipeline without checking whether they are acceptable to probe or re-advertise. In `massa-main/massa-protocol-worker/src/handlers/peer_handler/mod.rs`, `perform_handshake` verifies only the signature over the announcement, then stores `announcement` in trusted peer state and forwards `announcement.listeners` through `PeerManagementMessage::NewPeerConnected`.

`PeerManagementMessage::ListPeers` entries are likewise forwarded to `test_sender`, and `tester.rs` later attempts outbound connections to queued addresses while excluding only already-connected peers, exact configured listeners, and `config.routable_ip`. This leaves private, loopback, link-local, unspecified, and other non-global addresses eligible for probing.

The ruled-out portion of the original theory is the assertion that `announcement.listeners` must include or match the socket reached during the handshake. That is not a valid protocol invariant here because `Announcement::new` intentionally rewrites advertised listener IPs to `config.routable_ip`, so divergence between the contacted socket tuple and advertised listeners is part of the design.

## Affected code
```rust
        &self,
        buffer: &'a [u8],
    ) -> IResult<&'a [u8], Announcement, E> {
        let (rest, (listeners, timestamp)) = context(
            "Failed announcement deserialization",
            tuple((
                length_count(
                    context("Failed listeners deserialization", |buffer| {
                        self.length_listeners_deserializer.deserialize(buffer)
                    }),
                    context("Failed listener deserialization", |buffer: &'a [u8]| {
                        tuple((
                            tuple((
                                context("Failed ip deserialization", |buffer| {
                                    self.ip_addr_deserializer.deserialize(buffer)
                                }),
                                context("Failed port deserialization", |buffer| {
                                    nom::number::complete::be_u16(buffer)
                                }),
                            ))
                            .map(|(addr, ip)| SocketAddr::new(addr, ip)),
                            context("Failed transport deserialization", |buffer| {
                                let (rest, id) = nom::number::complete::be_u8(buffer)?;
                                match id {
                                    0 => Ok((rest, TransportType::Tcp)),
                                    1 => Ok((rest, TransportType::Quic)),
                                    _ => Err(nom::Err::Error(ParseError::from_error_kind(
                                        buffer,
                                        nom::error::ErrorKind::MapRes,
                                    ))),
                                }
                            }),
                        ))(buffer)
                    }),
                ),
                context("Failed timestamp deserialization", |buffer: &'a [u8]| {
                    let timestamp = u64::from_be_bytes(
                        buffer
                            .get(..8)
                            .ok_or(nom::Err::Error(ParseError::from_error_kind(
                                buffer,
                                nom::error::ErrorKind::LengthValue,
                            )))?
                            .try_into()
// … (truncated)
```

## Recommendation
Validate `Announcement.listeners` against a strict endpoint policy before storing, probing, or redistributing them. Reject or drop any listener address that is not explicitly allowed for peer discovery, including loopback, private, link-local, unspecified, multicast, and other non-global or reserved address ranges. Apply the same validation consistently across all ingress and propagation paths, including handshake processing, peer-list handling, persistence in `last_announce`, outbound test queues, and gossip responses, so that only policy-compliant endpoints can enter or move through the peer-management pipeline. If certain non-public addresses must be supported in specific deployments, gate that behavior behind explicit configuration rather than accepting such endpoints by default.

## Why P0 / breaking

peterjah:
…d re-advertising them

  * [ ] make tests compile
  * [ ] make tests pass 

### DIFF
--- a/massa-protocol-exports/src/settings.rs
+++ b/massa-protocol-exports/src/settings.rs
@@ -179,3 +179,18 @@ pub struct ProtocolConfig {
     /// Chain id
     pub chain_id: u64,
 }
+
+impl ProtocolConfig {
+    /// Whether the node is configured to deal with peers that do not have a
+    /// globally routable address, i.e. local networks. Mirrors the
+    /// `allow_local_peers` category setting: as soon as one category accepts
+    /// local peers, non global endpoints are kept in the peer management
+    /// pipeline, the per category check being done when connecting out.
+    pub fn allow_local_peers(&self) -> bool {
+        self.default_category_info.allow_local_peers
+            || self
+                .peers_categories
+                .values()
+                .any(|category| category.allow_local_peers)
+    }
+}
--- a/massa-protocol-worker/src/handlers/peer_handler/mod.rs
+++ b/massa-protocol-worker/src/handlers/peer_handler/mod.rs
@@ -27,6 +27,7 @@ use tracing::{debug, error, info, warn};
 
 use crate::context::Context;
 use crate::handlers::peer_handler::models::PeerState;
+use crate::ip::filter_routable_listeners;
 use crate::messages::{Message, MessagesHandler, MessagesSerializer};
 use crate::wrap_network::ActiveConnectionsTrait;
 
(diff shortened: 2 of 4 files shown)


Leo-Besancon (approved):
LGTM

PR #5204, opened 2026-08-24
[F121] cap the aggregate serialized size of operation batches on both…

### LINKED ISSUE #5042: [F121] Operations messages are deserialized without an aggregate size cap

---

## Source
AI security-scan **Finding 121** (severity: **medium**). Tracked in `SUBMISSIONS.md` as F121.

**Affected file:** `massa-protocol-worker/src/handlers/operation_handler/messages.rs`

## Summary
`OperationMessageDeserializer::deserialize()` accepts `MessageTypeId::Operations` by delegating directly to `self.operations_deserializer.deserialize(input)` in `massa-main/massa-protocol-worker/src/handlers/operation_handler/messages.rs`. This path enforces the number of operations and per-operation field limits, but it does *not* enforce an aggregate serialized-size cap for the entire `Operations` batch. As a result, a peer can cause the node to fully deserialize, copy, and hash a batch whose total size exceeds `max_serialized_operations_size_per_block`. The original claim is narrowed in two respects: batch signature verification is *not* guaranteed to occur after an oversized single operation is detected, and the available code does *not* establish `ProtocolConfig.max_message_size` as the effective transport-level limit.

## Root cause
The confirmed root cause is a *missing batch-level size check* on the `MessageTypeId::Operations` message path. In `OperationMessageDeserializer::deserialize()`, the `MessageTypeId::Operations` branch passes the payload directly to `self.operations_deserializer.deserialize(input)`, which validates count and per-item bounds only.

Downstream, `note_operations_from_peer(...)` narrows acceptance based on `serialized_size()` *per operation*, not on the aggregate serialized size of the full `Vec<SecureShareOperation>`. This means the batch can already have been fully deserialized and processed before any size-based rejection tied to `max_serialized_operations_size_per_block` is applied.

The original root-cause theory is overstated to the extent it implies that later processing always continues into `verify_sigs_batch` for oversized content. When a *single* operation exceeds `max_serialized_operations_size_per_block`, `note_operations_from_peer(...)` returns before batch signature verification.

## Affected code
```rust
        &self,
        buffer: &'a [u8],
    ) -> IResult<&'a [u8], OperationMessage, E> {
        context("Failed OperationMessage deserialization", |buffer| {
            let (buffer, raw_id) = self.id_deserializer.deserialize(buffer)?;
            let id = MessageTypeId::try_from(raw_id).map_err(|_| {
                    buffer,
                    nom::error::ErrorKind::Eof,
                ))
            })?;
            match id {
                MessageTypeId::AskForOperations => {
                    context("Failed AskForOperations deserialization", |input| {
                        self.operation_prefix_ids_deserializer.deserialize(input)
                    })
                    .map(OperationMessage::AskForOperations)
                    .parse(buffer)
                }
                MessageTypeId::OperationsAnnouncement => {
                    context("Failed OperationsAnnouncement deserialization", |input| {
                        self.operation_prefix_ids_deserializer.deserialize(input)
                    })
                    .map(OperationMessage::OperationsAnnouncement)
                    .parse(buffer)
                }
                MessageTypeId::Operations => {
                    context("Failed Operations deserialization", |input| {
                        self.operations_deserializer.deserialize(input)
                    })
                    .map(OperationMessage::Operations)
                    .parse(buffer)
                }
            }
        })
        .parse(buffer)
    }
```

## Recommendation
Enforce a batch-level serialized-size limit for `MessageTypeId::Operations` during deserialization, before allocating, copying, hashing, or otherwise processing the full payload. The limit should apply to the aggregate size of the entire operations vector, not just the number of operations or per-operation bounds, and should be aligned with `max_serialized_operations_size_per_block` or another explicitly defined protocol-level maximum.

Reject oversized batches as early as possible in `OperationMessageDeserializer::deserialize()` so that malformed or excessive messages cannot consume unnecessary CPU or memory before downstream validation. Also ensure the same aggregate-size invariant is consistently enforced across all operation-ingestion paths to avoid relying on later per-operation checks that do not protect against oversized batches.

## Why P0 / breaking

peterjah:
… send and receive

  * [ ] make tests compile
  * [ ] make tests pass 

### DIFF
--- a/massa-protocol-worker/src/handlers/peer_handler/announcement.rs
+++ b/massa-protocol-worker/src/handlers/peer_handler/announcement.rs
@@ -24,6 +24,12 @@ use massa_serialization::{
     U64VarIntSerializer,
 };
 
+/// Maximum tolerated clock skew between our local clock and the timestamp carried by a
+/// remote peer's announcement. The timestamp is attacker-controlled and is used as the peer
+/// freshness signal, so an announcement dated arbitrarily far in the future would otherwise
+/// stay "fresh" (and thus eligible for peer sharing and bootstrap responses) forever.
+pub const MAX_ANNOUNCEMENT_CLOCK_SKEW_MS: u64 = 5 * 60 * 1_000;
+
 #[derive(Clone, Debug, PartialEq, Eq)]
 pub struct Announcement {
     /// Listeners
@@ -170,6 +176,12 @@ impl Deserializer<Announcement> for AnnouncementDeserializer {
 }
 
 impl Announcement {
+    /// Check the announcement timestamp against the local clock: a peer cannot legitimately
+    /// announce more than `MAX_ANNOUNCEMENT_CLOCK_SKEW_MS` ahead of us.
+    pub fn is_future_dated(&self, now_millis: u64) -> bool {
+        self.timestamp > now_millis.saturating_add(MAX_ANNOUNCEMENT_CLOCK_SKEW_MS)
+    }
+
     pub fn new(
         listeners: HashMap<SocketAddr, TransportType>,
         routable_ip: Option<IpAddr>,
--- a/massa-protocol-worker/src/handlers/peer_handler/mod.rs
+++ b/massa-protocol-worker/src/handlers/peer_handler/mod.rs
@@ -13,6 +13,7 @@ use massa_protocol_exports::{
 };
 use massa_serialization::{DeserializeError, Deserializer, Serializer};
 use massa_signature::Signature;
+use massa_time::MassaTime;
 use peernet::context::Context as _;
 use peernet::messages::MessagesSerializer as _;
 use rand::{rngs::StdRng, RngCore, SeedableRng};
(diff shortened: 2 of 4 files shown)


Leo-Besancon (approved):
LGTM

massa-protocol-worker/src/handlers/operation_handler/retrieval.rs
```diff
@@ -468,6 +477,43 @@ impl RetrievalThread {
     }
 }
 
+/// Maximum length of the `u32` varint that prefixes the operations vector of an `Operations`
+/// message. Deducted from the size budget of a batch so that the message we build stays within
+/// the limit the receiving peer enforces on the whole payload.
+const OPERATIONS_LENGTH_PREFIX_MAX_LEN: usize = 5;
```
Leo-Besancon:
```suggestion
///
/// A varint encodes 7 bits per byte, so a `u32` takes at most `32.div_ceil(7) == 5` bytes.
const OPERATIONS_LENGTH_PREFIX_MAX_LEN: usize = 5;
```

Leo-Besancon (dismissed):
Added a suggestion for a more explicit computation and avoid magic numbers, otherwise LGTM!

Leo-Besancon (approved):
Thanks! LGTM

PR #5205, opened 2026-08-24
[F119] wire the ExecuteSC bytecode bound, not the op-datastore value …

### LINKED ISSUE #5041: [F119] P2P operation deserialization applies the datastore-value limit to smart-contract bytecode

---

## Source
AI security-scan **Finding 119** (severity: **info**). Tracked in `SUBMISSIONS.md` as F119.

**Affected file:** `massa-protocol-worker/src/handlers/operation_handler/messages.rs`

## Summary
The `OperationMessage::Operations` deserialization path applies `OperationMessageDeserializerArgs.max_datastore_value_length` to `ExecuteSC.data`, causing the P2P operation-message parser to treat the datastore-value limit as a smart-contract bytecode limit. In `massa-protocol-worker/src/handlers/operation_handler/retrieval.rs`, this value is populated from `config.max_op_datastore_value_length`, and `massa-models/src/operation.rs` then uses it to bound `ExecuteSC.data`. This parameter wiring mismatch is a real issue in the protocol message path. However, the original claim that this creates a practical censorship or DoS vector against otherwise valid `500 KB`-`10 MB` deployments is not supported.

## Root cause
The confirmed root cause is a *parameter mismatch* between the protocol-layer deserializer configuration and the operation model deserializer behavior.

`massa-protocol-worker/src/handlers/operation_handler/retrieval.rs` sets `OperationMessageDeserializerArgs.max_datastore_value_length` from `config.max_op_datastore_value_length`, and the `OperationMessage::Operations` branch in `massa-protocol-worker/src/handlers/operation_handler/messages.rs` forwards parsing to `self.operations_deserializer.deserialize(input)`. Downstream, `massa-models/src/operation.rs` uses `max_datastore_value_length` to enforce the size bound for `ExecuteSC.data`, even though that field contains smart-contract bytecode rather than a datastore value.

The ruled-out portion of the original theory is that this mismatch lets the network reject a meaningful set of otherwise acceptable large deployment operations. That does not hold because later protocol and pool checks independently enforce the block-size ceiling.

## Affected code
```rust
        &self,
        buffer: &'a [u8],
    ) -> IResult<&'a [u8], OperationMessage, E> {
        context("Failed OperationMessage deserialization", |buffer| {
            let (buffer, raw_id) = self.id_deserializer.deserialize(buffer)?;
            let id = MessageTypeId::try_from(raw_id).map_err(|_| {
                    buffer,
                    nom::error::ErrorKind::Eof,
                ))
            })?;
            match id {
                MessageTypeId::AskForOperations => {
                    context("Failed AskForOperations deserialization", |input| {
                        self.operation_prefix_ids_deserializer.deserialize(input)
                    })
                    .map(OperationMessage::AskForOperations)
                    .parse(buffer)
                }
                MessageTypeId::OperationsAnnouncement => {
                    context("Failed OperationsAnnouncement deserialization", |input| {
                        self.operation_prefix_ids_deserializer.deserialize(input)
                    })
                    .map(OperationMessage::OperationsAnnouncement)
                    .parse(buffer)
                }
                MessageTypeId::Operations => {
                    context("Failed Operations deserialization", |input| {
                        self.operations_deserializer.deserialize(input)
                    })
                    .map(OperationMessage::Operations)
                    .parse(buffer)
                }
            }
        })
        .parse(buffer)
    }
```

## Recommendation
Introduce a dedicated deserialization limit for `ExecuteSC.data` and stop reusing `OperationMessageDeserializerArgs.max_datastore_value_length` for smart-contract bytecode. The P2P operation-message configuration in `massa-protocol-worker/src/handlers/operation_handler/retrieval.rs` should pass a bytecode-specific bound, and `massa-models/src/operation.rs` should enforce that bound only for `ExecuteSC.data`, while keeping datastore limits scoped to datastore fields.

Align all related validation paths so that operation parsing, protocol checks, and pool filtering use clearly separated and semantically correct size constraints. If oversized `ExecuteSC` operations are intentionally disallowed because of `MAX_BLOCK_SIZE`, make that restriction explicit in configuration and validation logic to avoid misleading or inconsistent parameter wiring.

## Why P0 / breaking

peterjah:
…bound, into the operation message deserializer

  * [ ] make tests compile
  * [ ] make tests pass 


Leo-Besancon (changes requested):
For me the names of the fields `max_datastore_value_length` and `max_size_value_datastore` this PR updates should be a lot more explicit, something like `max_bytecode_size`. We potentially need to introduce a new field for this.

massa-api-exports/src/config.rs
```diff
@@ -49,8 +49,8 @@ pub struct APIConfig {
     pub enable_http: bool,
     /// whether to enable WS.
     pub enable_ws: bool,
-    /// max datastore value length
-    pub max_datastore_value_length: u64,
```
Leo-Besancon, line 53: Was `max_datastore_value_length` only used for the bytecode limit?
I assumed it was used elsewhere and a rename would just lead to the reverse issue (using `max_bytecode_size` somewhere where we want to limit datastore values)

massa-api/src/tests/mock.rs
```diff
@@ -237,7 +237,7 @@ pub(crate) fn start_public_api(addr: SocketAddr) -> (API<Public>, APIConfig) {
             max_size_channel_commands_peers: 300,
             max_message_size: MAX_MESSAGE_SIZE as usize,
             endorsement_count: ENDORSEMENT_COUNT,
-            max_size_value_datastore: 1_000_000,
```
Leo-Besancon, line 240: Same question here for `max_size_value_datastore`

massa-api-exports/src/config.rs
peterjah: checked every occurrence at the pre-rename commit: APIConfig.max_datastore_value_length had exactly one read, massa-api/src/public.rs:1573, which is the first arg of OperationDeserializer::new — the
 ExecuteSC.data bound. The remaining occurrences are the declaration and two test initializers. Datastore values in this crate are bounded by the separate max_op_datastore_value_length field (used at public.rs:286 for DatastoreDeserializer), so the reverse mix-up you're describing has no field to happen through. max_datastore_value_length is untouched in bootstrap / final-state / ledger, where it does mean a datastore value.

massa-api/src/tests/mock.rs
peterjah: Same check for ProtocolConfig.max_size_value_datastore: 6 occurrences total, and the only two reads are block_handler/retrieval.rs:124 and operation_handler/retrieval.rs:117 — both feeding the bytecode slot of the message-deserializer args. Note this PR also fixes the second one, which was previously fed max_op_datastore_value_length (500 KB) instead, which is the actual F119 bug. Operation-datastore values in ProtocolConfig are bounded by max_op_datastore_value_length, which stays as-is.

peterjah:
One genuine instance of the pattern does exist, but outside this PR: LedgerConfig.max_datastore_value_length feeds both the datastore bound and BytecodeDeserializer::new (massa-ledger-exports/src/ledger_entry.rs:109,113; ledger_db.rs:97). One field, two meanings — latent only because MAX_DATASTORE_VALUE_LENGTH == MAX_BYTECODE_LENGTH == 10_000_000. 
Splitting it needs a new LedgerConfig field threaded through bootstrap. i ll takle it in this PR

massa-api-exports/src/config.rs
Leo-Besancon: great, thanks for checking

Leo-Besancon (approved):
LGTM

PR #5206, opened 2026-08-24
[F112] mark blocks invalid when their committed operations fail conte…

### LINKED ISSUE #5036: [F112] Block-Intrinsic Operation Validation Failures Do Not Mark Blocks Invalid

---

## Source
AI security-scan **Finding 112** (severity: **medium**). Tracked in `SUBMISSIONS.md` as F112.

## Summary
The issue is in how `on_block_full_operations_received` handles failures from `note_operations_from_peer`. Some failures come from the block content itself, such as an operation whose `serialized_size()` is greater than `max_serialized_operations_size_per_block`. In that case, the block is permanently invalid because the operation IDs already commit to that invalid content. However, the code treats the failure as if only the sending peer misbehaved, so the bad block is not marked invalid and may stay in the wishlist. This can make nodes keep retrying retrieval of a block that can never be fetched successfully.

## Root cause
In `on_block_full_operations_received`, every validation error returned by `note_operations_from_peer` is handled as *sender-only misconduct*. The function bans the current peer and returns, but it does not call `mark_block_as_invalid` and does not remove the block from the wishlist.

Some validation failures are not peer-specific. The input specifically mentions operation content that is already committed by the block's operation IDs, including an operation whose `serialized_size()` exceeds `max_serialized_operations_size_per_block`. For those cases, another peer cannot provide a different valid operation with the same ID, so the block itself is unrecoverably invalid.

Because `on_block_full_operations_received` does not distinguish these *block-intrinsic* failures from ordinary peer misbehavior, an impossible block remains tracked as retrievable even though no peer can ever satisfy it.

## Affected code
```rust
fn on_block_full_operations_received(
        &mut self,
        from_peer_id: PeerId,
        block_id: BlockId,
    ) {
        debug!(
            "received full operations for block {} from {}",
            block_id, &from_peer_id
        );

        // Ensure that we were looking for that data.
        let wishlist_info = if let Some(info) = self
            .block_wishlist
            .get_mut(&block_id)
            .filter(|i| i.header.is_some() && i.operation_ids.is_some())
        {
            info
        } else {
            // we were not looking for this data
            debug!(
                "Peer id {} sent us full operations for block id {} but we were not looking for it",
                from_peer_id, block_id
            );
            // still mark the sender as knowing the block and operations
            self.cache
                .write()
                .insert_peer_known_block(&from_peer_id, &[block_id], true);
            self.operation_cache.write().insert_peer_known_ops(
                &from_peer_id,
                &operations
                    .into_iter()
                    .map(|op| op.id.prefix())
                    .collect::<Vec<_>>(),
            );
            return;
        };

        // Move the ops into a hashmap
        let mut operations: PreHashMap<OperationId, SecureShareOperation> =
            operations.into_iter().map(|op| (op.id, op)).collect();

        // Make a set of all the block ops for fast lookup and deduplication
        let block_ops_set = wishlist_info
            .operation_ids
// … (truncated)
```

## Recommendation
Differentiate peer-specific delivery errors from block-intrinsic validation failures returned by `note_operations_from_peer`. When the failure is determined solely by operation content already committed by the block’s operation IDs, treat the block as permanently invalid: call `mark_block_as_invalid`, remove it from the wishlist, and stop further retrieval attempts. Reserve peer sanctions for cases where the sender could have provided different valid data. If the error classification is not currently explicit, introduce a clear distinction between recoverable sender misconduct and unrecoverable block invalidity so invalid blocks are not retried indefinitely.

## Why P0 / breaking

peterjah:
…nt-level checks

  * [ ] make tests compile
  * [ ] make tests pass 

### DIFF
--- a/massa-protocol-worker/src/handlers/block_handler/retrieval.rs
+++ b/massa-protocol-worker/src/handlers/block_handler/retrieval.rs
@@ -52,7 +52,7 @@ use rand::{seq::SliceRandom, Rng};
 use tracing::{debug, info, warn};
 
 use super::{
-    super::operation_handler::note_operations_from_peer,
+    super::operation_handler::{is_block_intrinsic_operation_failure, note_operations_from_peer},
     cache::{BlockDataKind, SharedBlockCache},
     commands_propagation::BlockHandlerPropagationCommand,
     commands_retrieval::BlockHandlerRetrievalCommand,
@@ -900,9 +900,6 @@ impl RetrievalThread {
         // Here we know that we were looking for that block's operations and that the sender node sent us some of the missing ones.
 
         // Check the validity of the received operations.
-        // TODO: in the future if the validiy check fails for something non-malleable (eg. not sig verif),
-        //       we should stop retrieving the block and ban everyone who knows it
-        //       because we know for sure that this op's ID belongs to the block.
         if let Err(err) = note_operations_from_peer(
             &self.storage,
             &mut self.operation_cache,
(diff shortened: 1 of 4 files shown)


Leo-Besancon (approved):
LGTM

PR #5208, opened 2026-08-24
[F118] bound conflicting endorsements per draw on the protocol retrie…

### LINKED ISSUE #5040: [F118] Conflicting endorsements with arbitrary endorsed blocks are accepted and gossiped

---

## Source
AI security-scan **Finding 118** (severity: **minor**). Tracked in `SUBMISSIONS.md` as F118.

**Affected file:** `massa-protocol-worker/src/handlers/endorsement_handler/retrieval.rs`

## Summary
`note_endorsements_from_peer` in `massa-main/massa-protocol-worker/src/handlers/endorsement_handler/retrieval.rs` accepts standalone endorsements after checking signature validity and whether the creator matches the PoS-selected endorser for the relevant `(slot, index)`. Because deduplication is keyed only by `endorsement.id`, a selected endorser can submit multiple conflicting endorsements for the same `(slot, index)` by changing `endorsed_block`, and each variant is treated as fresh data and propagated. The retrieval path also does not verify that `endorsed_block` is known or contextually plausible at this stage. However, the original claim that this automatically pollutes every node's endorsement pool and broadly starves legitimate pool entries is overstated.

## Root cause
The confirmed issue is that `note_endorsements_from_peer` validates standalone endorsements too narrowly on the retrieval path. It verifies signatures and checks that the endorsement creator matches `selector_controller.get_selection(slot).endorsements[index]`, but it does not enforce uniqueness for the same endorser on the same `(slot, index)` when only `endorsed_block` changes.

Because `endorsement.id` is derived from the full signed content, each distinct `endorsed_block` produces a different `endorsement.id`. The cache therefore deduplicates only exact duplicates through `checked_endorsements` and does not detect conflicts at the `(slot, index, creator)` level, allowing multiple conflicting variants from the same valid endorser to be accepted and forwarded.

The retrieval path also does not reject endorsements merely because `endorsed_block` is unknown or implausible. That contextual validation occurs later when endorsements are processed inside a block header, so the original theory that these standalone endorsements are already constrained by an equivalent upstream plausibility check does not hold. By contrast, the broader root-cause claim that all accepted variants necessarily contaminate every node's local pool is not supported, because `EndorsementPool::add_endorsements` retains endorsements only under the wallet-ownership condition identified by the validator.

## Affected code
```rust
/// Process incoming message
fn process_message(&mut self, peer_id: PeerId, message: Vec<u8>) {
    let (rest, message) = match self
        .endorsement_message_deserializer
        .deserialize::<DeserializeError>(&message)
    {
        Ok((rest, message)) => (rest, message),
        Err(err) => {
            debug!(
                "Error while deserializing message from peer {} err: {:?}",
                peer_id, err
            );
            return;
        }
    };
    if !rest.is_empty() {
        debug!("Message not fully consumed");
        return;
    }
    match message {
        EndorsementMessage::Endorsements(endorsements) => {
            debug!("Received endorsement message: Endorsement from {}", peer_id);
            if let Err(err) = note_endorsements_from_peer(
                endorsements,
                &peer_id,
                &self.cache,
                self.selector_controller.as_ref(),
                &self.storage,
                &self.config,
                &self.internal_sender,
                self.pool_controller.as_mut(),
            ) {
                warn!(
                    "peer {} sent us critically incorrect endorsements, \\n                    which may be an attack attempt by the remote node or a \\n                    loss of sync between us and the remote node. Err = {}",
                    peer_id, err
                );
                if let Err(err) = self.ban_peer(&peer_id) {
                    warn!("Error while banning peer {} err: {:?}", peer_id, err);
                }
            }
        }
    }
}
```

## Recommendation
Enforce conflict detection for standalone endorsements at the `(slot, index, creator)` level rather than deduplicating only by `endorsement.id`. Reject and avoid propagating any additional endorsement from the same selected endorser for the same `(slot, index)` once one variant has already been accepted, regardless of differences in `endorsed_block`.

In addition, apply lightweight contextual validation on the retrieval path so that endorsements referencing unknown, unavailable, or clearly implausible `endorsed_block` values are not accepted or gossiped prematurely. If full validation is only possible later during block processing, treat such endorsements as non-propagable until the referenced block becomes known and the endorsement can be validated consistently.

Document the intended handling of conflicting endorsements explicitly, including whether they should be retained only for denunciation purposes, and ensure the gossip path follows that policy to prevent unnecessary bandwidth consumption and propagation of unusable data.

## Why P0 / breaking

peterjah:
…val path

  * [ ] make tests compile
  * [ ] make tests pass 

### DIFF
--- a/massa-protocol-worker/src/handlers/endorsement_handler/cache.rs
+++ b/massa-protocol-worker/src/handlers/endorsement_handler/cache.rs
@@ -3,22 +3,13 @@ use std::{
     sync::Arc,
 };
 
-use massa_models::{endorsement::EndorsementId, prehash::PreHashSet, slot::Slot};
+use massa_models::{
+    config::MAX_ENDORSEMENTS_PER_SLOT_INDEX, endorsement::EndorsementId, prehash::PreHashSet,
+    slot::Slot,
+};
 use massa_protocol_exports::PeerId;
 use parking_lot::RwLock;
 use schnellru::{ByLength, LruMap};
-
-/// Maximum number of distinct endorsements we accept and propagate for a given `(slot, index)`
-/// endorsement draw. An honest drawn endorser signs exactly one endorsement per draw, so several
-/// distinct ones mean it equivocated.
-///
-/// The bound is kept above 1 for two reasons: the denunciation pool needs to see two conflicting
-/// endorsements to build a denunciation, and dropping everything after the first variant would let
-/// an equivocating endorser pick which variant we keep, denying us the endorsement slot when it
-/// endorses a block that is not the parent our block factory settles on. It matches the pool-side
-/// bound (`MAX_ENDORSEMENTS_PER_SLOT_INDEX`) so that the propagation path never becomes the tighter
-/// of the two limits.
-pub const MAX_ENDORSEMENTS_PER_DRAW: usize = 4;
 
 /// Cache of endorsements
 pub struct EndorsementCache {
@@ -38,11 +29,11 @@ impl EndorsementCache {
     pub fn new(max_known_endorsements: u32, max_known_endorsements_by_peer: u32) -> Self {
         Self {
             checked_endorsements: LruMap::new(ByLength::new(max_known_endorsements)),
-            // one entry per draw holds up to `MAX_ENDORSEMENTS_PER_DRAW` ids, so scaling the
+            // one entry per draw holds up to `MAX_ENDORSEMENTS_PER_SLOT_INDEX` ids, so scaling the
             // capacity down keeps this index's memory in line with `checked_endorsements` while
             // covering the same horizon of endorsements
             endorsements_by_draw: LruMap::new(ByLength::new(std::cmp::max(
-                max_known_endorsements / MAX_ENDORSEMENTS_PER_DRAW as u32,
+                max_known_endorsements / MAX_ENDORSEMENTS_PER_SLOT_INDEX as u32,
                 1,
             ))),
             endorsements_known_by_peer: HashMap::new(),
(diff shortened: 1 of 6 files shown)


massa-protocol-worker/src/handlers/endorsement_handler/cache.rs
```diff
@@ -3,15 +3,30 @@ use std::{
...
+/// Maximum number of distinct endorsements we accept and propagate for a given `(slot, index)`
+/// endorsement draw. An honest drawn endorser signs exactly one endorsement per draw, so several
+/// distinct ones mean it equivocated.
+///
+/// The bound is kept above 1 for two reasons: the denunciation pool needs to see two conflicting
+/// endorsements to build a denunciation, and dropping everything after the first variant would let
+/// an equivocating endorser pick which variant we keep, denying us the endorsement slot when it
+/// endorses a block that is not the parent our block factory settles on. It matches the pool-side
+/// bound (`MAX_ENDORSEMENTS_PER_SLOT_INDEX`) so that the propagation path never becomes the tighter
+/// of the two limits.
```
Leo-Besancon: If there isn't any reason for this one to be greater than `MAX_ENDORSEMENTS_PER_SLOT_INDEX` I would unify them as a single constant in constant.rs file.

Leo-Besancon (changes requested):
One small comment, looks good otherwise!

massa-protocol-worker/src/handlers/endorsement_handler/cache.rs
peterjah: done!

Leo-Besancon (dismissed):
Thanks, LGTM!

Leo-Besancon (approved):

PR #5209, opened 2026-08-24
[F117] drop stale endorsements before signature and PoS draw verifica…

### LINKED ISSUE #5039: [F117] Improper Validation Order Allows DoS via Replay of Stale Endorsements

---

## Source
AI security-scan **Finding 117** (severity: **medium**). Tracked in `SUBMISSIONS.md` as F117.

**Affected file:** `massa-protocol-worker/src/handlers/endorsement_handler/retrieval.rs`

## Summary
In `massa-main/massa-protocol-worker/src/handlers/endorsement_handler/retrieval.rs`, `note_endorsements_from_peer(...)` verifies endorsements before checking whether they are already too old to be useful. It only excludes IDs already present in `cache.checked_endorsements`, then runs `verify_sigs_batch(...)` and `selector_controller.get_selection(...)` on all remaining endorsements. After that, it applies the freshness test with `get_block_slot_timestamp(...)` and `config.max_endorsements_propagation_time`, so expired but otherwise valid historical endorsements consume CPU and selector lookup work before being discarded. Because `checked_endorsements` is a bounded LRU cache, a remote peer can rotate old endorsement IDs and keep forcing this wasted work.

## Root cause
The main issue is in `massa-main/massa-protocol-worker/src/handlers/endorsement_handler/retrieval.rs` inside `note_endorsements_from_peer(...)`. The function builds `new_endorsements` by checking only whether each `endorsement_id` is absent from `cache_read.checked_endorsements.peek(&endorsement_id)`. This cache gate does not check whether the endorsement is still fresh.

After that early cache check, `note_endorsements_from_peer(...)` immediately performs expensive validation on every item in `new_endorsements`:
- It runs batch signature verification through `verify_sigs_batch(...)`.
- It performs PoS producer validation through `selector_controller.get_selection(endorsement.content.slot)?.endorsements`, then checks `selection.get(endorsement.content.index as usize)` against `endorsement.content_creator_address`.

The freshness filter is applied only later in the same function. It computes the endorsement slot time with `get_block_slot_timestamp(config.thread_count, config.t0, config.genesis_timestamp, endorsement.content.slot)` and keeps only endorsements satisfying `t.saturating_add(config.max_endorsements_propagation_time) >= now`. This means stale endorsements are fully verified first and discarded afterward.

The replay cost can be sustained because the cache used by `note_endorsements_from_peer(...)` is bounded. The finding notes `checked_endorsements` has capacity `2048`, while message batches can be `1024` endorsements and `process_message(...)` passes `EndorsementMessage::Endorsements` batches directly into `note_endorsements_from_peer(...)`. By rotating through more unique historical endorsement IDs than the LRU can retain, an attacker can trigger eviction and cause repeated execution of the same expensive path. The relevant cache update points are `cache_wri …

## Affected code
```rust
/// Process incoming message
fn process_message(&mut self, peer_id: PeerId, message: Vec<u8>) {
    let (rest, message) = match self
        .endorsement_message_deserializer
        .deserialize::<DeserializeError>(&message)
    {
        Ok((rest, message)) => (rest, message),
        Err(err) => {
            debug!(
                "Error while deserializing message from peer {} err: {:?}",
                peer_id, err
            );
            return;
        }
    };
    if !rest.is_empty() {
        debug!("Message not fully consumed");
        return;
    }
    match message {
        EndorsementMessage::Endorsements(endorsements) => {
            debug!("Received endorsement message: Endorsement from {}", peer_id);
            if let Err(err) = note_endorsements_from_peer(
                endorsements,
                &peer_id,
                &self.cache,
                self.selector_controller.as_ref(),
                &self.storage,
                &self.config,
                &self.internal_sender,
                self.pool_controller.as_mut(),
            ) {
                warn!(
                    "peer {} sent us critically incorrect endorsements, \\n                    which may be an attack attempt by the remote node or a \\n                    loss of sync between us and the remote node. Err = {}",
                    peer_id, err
                );
                if let Err(err) = self.ban_peer(&peer_id) {
                    warn!("Error while banning peer {} err: {:?}", peer_id, err);
                }
            }
        }
    }
}
```

## Recommendation
Reject stale endorsements as early as possible in `note_endorsements_from_peer(...)`, before any expensive validation such as `verify_sigs_batch(...)` or `selector_controller.get_selection(...)`. Use the endorsement slot to compute freshness first and discard any endorsement that already exceeds `config.max_endorsements_propagation_time`, so expired data cannot consume signature-verification or selector resources.

To prevent repeated processing of replayed stale endorsements, ensure the deduplication strategy also covers expired items, rather than relying only on `checked_endorsements` after costly validation. Consider maintaining lightweight tracking or rate-limiting for stale endorsement replays so LRU eviction cannot be abused to re-trigger the same work indefinitely.

Additionally, review cache sizing and peer-level limits for endorsement batches to reduce cache churn and limit the effectiveness of replay-based resource exhaustion.

## Why P0 / breaking

peterjah:
…tion

  * [ ] make tests compile
  * [ ] make tests pass 

### DIFF
--- a/massa-protocol-worker/src/handlers/endorsement_handler/retrieval.rs
+++ b/massa-protocol-worker/src/handlers/endorsement_handler/retrieval.rs
@@ -122,6 +122,19 @@ impl RetrievalThread {
         match message {
             EndorsementMessage::Endorsements(endorsements) => {
                 debug!("Received endorsement message: Endorsement from {}", peer_id);
+                // Discard endorsements that are already too old to be useful before doing
+                // any expensive work on them (signature verification, PoS draw lookup):
+                // otherwise a peer could replay arbitrary amounts of old endorsements
+                // to make us burn CPU, cycling through IDs faster than the
+                // `checked_endorsements` cache can remember them.
+                let now = MassaTime::now();
+                let endorsements: Vec<_> = endorsements
+                    .into_iter()
+                    .filter(|endorsement| is_endorsement_fresh(endorsement, &self.config, now))
+                    .collect();
+                if endorsements.is_empty() {
+                    return;
+                }
                 if let Err(err) = note_endorsements_from_peer(
                     endorsements,
                     &peer_id,
@@ -152,6 +165,24 @@ impl RetrievalThread {
         self.peer_cmd_sender
             .try_send(PeerManagementCmd::Ban(vec![*peer_id]))
             .map_err(|err| ProtocolError::SendError(err.to_string()))
+    }
+}
+
+/// Returns true if the inclusion slot of the endorsement is recent enough for the
+/// endorsement to still be worth processing (see `max_endorsements_propagation_time`).
+fn is_endorsement_fresh(
+    endorsement: &SecureShareEndorsement,
+    config: &ProtocolConfig,
+    now: MassaTime,
+) -> bool {
+    match get_block_slot_timestamp(
+        config.thread_count,
+        config.t0,
+        config.genesis_timestamp,
+        endorsement.content.slot,
+    ) {
+        Ok(t) => t.saturating_add(config.max_endorsements_propagation_time) >= now,
+        Err(_) => false,
     }
 }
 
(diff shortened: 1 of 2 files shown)


Leo-Besancon (approved):
LGTM!

PR #5210, opened 2026-08-25
fix(protocol): ban peers sending messages with trailing bytes

### LINKED ISSUE #5147: Followup for F120 / F128: Ban peers sending messages with trailing bytes
In PR https://github.com/massalabs/massa/pull/5105, we fixed an issue regarding trailing bytes on message received.
We kept the PR minimal, so didn't ban peers that sent us malformed messages. This should be done in a separate PR.

0xB19:
## Summary

Follow-up of #5105, closes #5147. That PR stopped messages with trailing
bytes from tearing down the shared retrieval threads, but deliberately
kept the change minimal and did not sanction the sender. This PR adds the
ban.

## Change

The four deserialization sites that detect trailing bytes (block,
operation, endorsement and peer management handlers) now warn with the
peer id and trailing byte count, then ban the sender, following the
existing "critically incorrect payload → warn + ban" pattern already used
for invalid signatures. Block, operation and endorsement handlers reuse
their existing ban helpers; the peer management loop bans directly since
it owns `peer_db` and `active_connections`.

## Testing

- Four new tests in `ban_nodes_scenarios` (one per handler), using a new
  test-universe helper that appends junk bytes after a valid serialized
  message. Each test asserts the peer is disconnected and banned, and
  that the message content is not processed (`.never()` on the downstream
  controllers).
- Verified the tests fail (hang on the ban waitpoint) without the fix.
- Full `massa_protocol_worker` suite passes (64 tests), clippy and fmt
  clean.
- Full CI suite green on this PR (including the `full` test job).

## Risk

Low: compliant peers never send trailing bytes, and the path is only
reachable for messages that deserialize successfully with leftover bytes.
One point for reviewers: with this change, extending an existing message
type by appending fields would make old nodes ban newer peers; message
evolution must keep going through new message type ids instead.

---

  * [x] make tests compile
  * [x] make tests pass

### DIFF
--- a/massa-protocol-worker/src/handlers/block_handler/retrieval.rs
+++ b/massa-protocol-worker/src/handlers/block_handler/retrieval.rs
@@ -151,12 +151,16 @@ impl RetrievalThread {
                                 // A valid message prefix followed by trailing bytes must not
                                 // tear down this long-lived shared retrieval thread (doing so
                                 // would let a single peer deny block handling for everyone).
-                                // Skip the malformed message and keep serving other peers.
+                                // A compliant peer never sends trailing bytes, so ban the
+                                // sender and keep serving other peers.
                                 warn!(
-                                    "peer {} sent a block message with {} unexpected trailing byte(s); ignoring it",
+                                    "peer {} sent a block message with {} unexpected trailing byte(s); banning it",
                                     peer_id,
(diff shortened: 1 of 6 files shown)


Leo-Besancon (approved):
LGTM, thank you!

PR #5211, opened 2026-08-25
ci: re-enable check_gas_cost_definitions xtask

### LINKED ISSUE #4943: re enable CI task "check_gas_cost_definitions"
With the release of MAIN.4.0 some code that supported this task was removed.
Hence, the task failed.
Hence, the task was commented.

This is an important task that must be re-enabled.

0xB19:
## Summary

Closes #4943. The `check_gas_cost_definitions` xtask compared the ABI
gas cost keys of the shipped `abi_gas_costs.json` against
`massa-sc-runtime`. The runtime update shipped with MAIN.4.0 replaced
the internal ABI-costs map (and its accessor) with named struct fields,
the task stopped compiling, and the `misc` CI job was commented out.
This PR reimplements the check on the current `GasCosts` API and
re-enables the job.

## Change

- `massa-xtask/src/check_gas_cost_definitions.rs`: new implementation.
  - Missing definitions: loading the shipped file with `GasCosts::new`
    fails if any cost required by the runtime is absent (the node would
    not start), and fails the task. Same failure semantics as the old
    check.
  - Stale definitions: the runtime no longer exposes its expected key
    set, so the task probes `GasCosts::new` with each key removed in
    turn; a key whose removal does not break loading is unused. Warned
    about but not failing, as before. This replaces the old hardcoded
    exclude list and cannot drift.
- The task is no longer gated behind the `gas_calibration` feature: it
  no longer needs `GasCosts::default()`. It still works through the
  `cargo xtask` alias, which enables that feature.
- `.github/workflows/ci.yml`: the `misc` job is uncommented, unchanged.

## Testing

- `cargo xtask check_gas_cost_definitions` passes on the current file
  and reports no stale key.
- Removing a key from `abi_gas_costs.json` makes the task fail with
  "assembly_script_abort cost not found in ABI gas cost file".
- Adding a bogus key makes the task print
  "Found in json but not used by the runtime: bogus_stale_key" and
  still succeed.
- clippy and fmt clean on the crate.
- Full CI suite green on this PR, including the re-enabled `misc` job
  running `cargo xtask check_gas_cost_definitions` on itself.

## Risk

None at runtime: the change only touches the xtask and CI. The CI job
keeps `continue-on-error: true`, as before.

---

* [ ] try in sandbox /simulation/labnet (not applicable, CI-only change)
* [ ] unit tests on the added/changed features (validated by running the
  task against altered inputs, see Testing)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -116,23 +116,23 @@ jobs:
           submodules: "recursive"
       - uses: crate-ci/typos@master
 
-  # misc:
-  #   if: github.ref != 'refs/heads/staging'
-  #   needs: sanity
-  #   runs-on: ubuntu-latest
-  #   continue-on-error: true
-  #   steps:
-  #     - uses: actions/checkout@v4
-  #       with:
-  #         submodules: "recursive"
-  #     - uses: dtolnay/rust-toolchain@master
-  #       with:
-  #         toolchain: 1.81.0
-  #     - uses: Swatinem/rust-cache@v2
-  #       with:
-  #         shared-key: "clippy"
-  #         save-if: ${{ github.ref_name == 'main' }}
-  #     - run: cargo xtask check_gas_cost_definitions
+  misc:
+    if: github.ref != 'refs/heads/staging'
+    needs: sanity
+    runs-on: ubuntu-latest
+    continue-on-error: true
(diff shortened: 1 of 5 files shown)


Leo-Besancon (approved):
LGTM.

PR #5212, opened 2026-08-25
feat(sdk): TLS and mTLS support for the gRPC client

### LINKED ISSUE #5180: Enable TLS (and mTLS) for the massa-sdk gRPC client
Follow-up to F144 #5179 . The doc-only PR for F144 recorded that the SDK's client constructors are plaintext-only; this issue tracks actually making the gRPC client able to talk to a TLS-enabled node.

Problem

The node's gRPC server fully supports TLS and mTLS — enable_tls / enable_mtls / generate_self_signed_certificates in massa-grpc/src/config.rs, wired up at massa-grpc/src/server.rs:197-230 via ServerTlsConfig::new().identity(…) and .client_ca_root(…). The SDK client cannot connect to it:

- massa-sdk/Cargo.toml pulls tonic with only the gzip feature. massa-grpc/Cargo.toml enables tls. So the SDK has no ClientTlsConfig available at all.
- Client::new (massa-sdk/src/lib.rs:84) hard-codes grpc://{addr} and calls Channel::from_shared(...).connect() with no TLS configuration.
- When the node has enable_tls = true, that connect fails, and the failure is swallowed: grpc_public / grpc_private are set to None behind a tracing::warn!("unable to connect to grpc server {}"). The user sees gRPC commands fail with "no gRPC client" rather than a TLS diagnostic.

The gap is especially visible in the self-signed path: generate_self_signed_certificates (massa-grpc/src/server.rs:326) writes client material to client_certificate_path / client_private_key_path, which default to ../massa-client/config/tls_public_client.pem and .key (massa-node/base_config/config.toml:123-125). The node provisions certificates into massa-client's config directory that no client code ever reads — massa-client has no TLS settings whatsoever.

Scope

1. Enable tonic's tls feature in massa-sdk/Cargo.toml.
2. Let the caller supply gRPC TLS material — CA root for server verification, plus client certificate + private key for the mTLS case — and apply it via ClientTlsConfig on the Channel builder, using https:// instead of the hard-coded grpc:// when TLS is on.
3. Add the corresponding massa-client settings and config.toml keys, defaulting to the paths the node already writes to, so the self-signed flow works end to end without manual copying.
4. Surface the connect failure. Even with TLS support, a cert/SAN mismatch is currently invisible; Client::new returning None behind a warn! should at minimum log the underlying error at a level users see, and it's worth discussing whether it should be a hard error.

Design questions

- Constructor shape. Client::new takes IpAddr + four ports and builds every URL itself. TLS needs either a scheme/hostname parameter or a shift to full URLs. Note that TLS certificate verification is hostname-based, while the current API is IP-based — the self-signed certs rely on subject_alt_names, so the client needs to connect by a name present in the SAN list, not by bare IP. This is the main API decision.
- Config surface. Whether the new keys are per-endpoint (public vs private gRPC can have different CAs, and do in the defaults: tls_public_ca.pem vs tls_private_ca.pem) or a single shared block.
- Relationship to certificate_store. The existing setting is inert and its "Native" / "WebPki" values no longer map to anything in jsonrpsee 0.24. It should not be repurposed for gRPC; if a trust-store choice is wanted for JSON-RPC too, that's a separate decision expressed via jsonrpsee's CustomCertStore.

Out of scope

JSON-RPC TLS already works — jsonrpsee 0.24 comes in with default features, so hyper-rustls is compiled in and HttpBackend selects its Https variant from the URL scheme. RpcClient::from_url(url: &str, …) accepts an https:// URL today. Only the convenience constructors (Client::new, ClientV2::new, RpcClientV2::from_url) prevent passing one, and if item 1 above reshapes those constructors it should fix the JSON-RPC schemes in the same pass.

Notes

- Not consensus-observable — client-side transport only, no MIP / versioning gate required.
- This is a breaking change to the massa-sdk public constructor signatures and adds massa-client config keys, so it wants a version bump and a changelog entry.

0xB19:
## Summary

Closes #5180 (follow-up of the F144 documentation PR #5179). The node's
gRPC server fully supports TLS and mTLS, but the SDK client could not
connect to it: `massa-sdk` compiled tonic without its `tls` feature,
`Client::new` hard-coded a plaintext `grpc://` channel, and the failed
connection was swallowed behind a `warn!`, leaving the user with "no
gRPC client" and no diagnostic.

## Change

**API decision.** The issue asked whether the constructors should move
to full URLs or grow a hostname parameter. Neither: tonic's
`ClientTlsConfig::domain_name` lets the client keep dialing the
configured IP while verifying the server certificate under a configured
name (also used for SNI). The constructors keep their IP + ports shape;
the only signature change is one optional TLS configuration per gRPC
endpoint.

- `massa-sdk`: tonic `tls` feature enabled. New `GrpcTlsConfig`
  (server name, CA root path, optional client certificate + key for
  mTLS) and a public `connect_grpc_channel` helper. `Client::new`
  takes `Option<&GrpcTlsConfig>` for the public and private gRPC
  endpoints.
- Error surfacing, split by nature: an invalid TLS configuration
  (unreadable certificate material, client certificate without its
  key) is a hard error from `Client::new`; a connection failure still
  degrades to `None` so the JSON-RPC clients stay usable when the node
  is down, but is now logged at `error!` level with the underlying
  cause and the target address.
- `massa-client`: new `[client.grpc.public]` and
  `[client.grpc.private]` settings blocks (`enable_tls`, `enable_mtls`,
  `server_name`, CA and client material paths). Defaults point to the
  paths the node writes its self-signed material to
  (`../massa-node/config/tls_*_ca.pem` for the CA it generates,
  `config/tls_*_client.pem|.key` which the node already provisions
  into massa-client's config directory), so the self-signed flow works
  end to end without manual copying when both run from sibling
  directories. `server_name` defaults to "localhost", which the node's
  generated certificates always include in their subject alternative
  names.
- JSON-RPC transports are untouched (TLS already reachable through
  `RpcClient::from_url` with an `https` url), `ClientV2` untouched.
  The inert `certificate_store` setting is left as is (deprecated).

## Testing

New integration tests in `massa-sdk/tests/grpc_tls.rs`, against a real
TLS-enabled tonic health server using the same certificate generation
path as the node (`cert_manager`):

- TLS connect succeeds with the right CA and server name;
- an untrusted CA is rejected;
- mTLS: an RPC succeeds with the client identity, fails without it
  (the reject only surfaces at the first RPC with TLS 1.3, which the
  test documents);
- incomplete client identity is a configuration error before dialing;
- plaintext regression: `None` still yields a working plaintext
  channel.

All 6 pass; clippy and fmt clean; `massa-client`, `massa_grpc` and
`massa-node` still build; the massa-client settings test loads the new
config keys.

Not done here: a full massa-node sandbox run with `enable_tls = true`.
The transport mechanics are covered by the integration tests above, but
a sandbox pass would additionally validate the default relative paths.

## Risk

Low at runtime (TLS stays opt-in, defaults unchanged: plaintext). This
is a **breaking change to the `massa-sdk` public API** (`Client::new`
signature) and adds massa-client config keys: it wants a version bump
and a changelog entry. The repo versions all crates together at release
time via `cargo xtask update_package_versions` and carries no changelog
file, so flagging it here for the maintainers instead.

---

* [ ] try in sandbox /simulation/labnet (not done, see Testing)
  * [x] make tests compile
  * [x] make tests pass
* [ ] if the API has changed, update the API specification (gRPC API
  itself unchanged; SDK/client configuration documented in config.toml)


Leo-Besancon:
Do we agree that with this PR, massa-client still never uses the GRPC part of the massa-sdk?
I've tested giving a wrong client .pem file with mtls enabled and it still works well, I expected the server to reject my requests.

I assume from the issue's description that this basically fix connecting to a server that enabled tls/mtls, so it's still a useful PR, but wondering if we need to expand on it or not?

Is it expected with the issues's scope @peterjah @0xB19?

Leo-Besancon:
Additional note: @peterjah added the following check on the grpc side, I think it would be great to also call it in massa-client
```rust
/// mTLS is a strict extension of TLS: reject the combination rather than silently serving in
/// plaintext a service the operator meant to protect with client certificates.
pub(crate) fn check_mtls_requires_tls(config: &GrpcConfig) -> Result<(), GrpcError> {
    if config.enable_mtls && !config.enable_tls {
        return Err(GrpcError::ConfigError(format!(
            "gRPC {:?} API: `enable_mtls` requires `enable_tls` to be enabled, refusing to start in plaintext",
            config.name
        )));
    }

    Ok(())
}
```

0xB19:
> Additional note: @peterjah added the following check on the grpc side, I think it would be great to also call it in massa-client
> 
> ```rust
> /// mTLS is a strict extension of TLS: reject the combination rather than silently serving in
> /// plaintext a service the operator meant to protect with client certificates.
> pub(crate) fn check_mtls_requires_tls(config: &GrpcConfig) -> Result<(), GrpcError> {
>     if config.enable_mtls && !config.enable_tls {
>         return Err(GrpcError::ConfigError(format!(
>             "gRPC {:?} API: `enable_mtls` requires `enable_tls` to be enabled, refusing to start in plaintext",
>             config.name
>         )));
>     }
> 
>     Ok(())
> }
> ```

Can we refactor this by moving the check to massa-sdk? Both massa-grpc and massa-client already depend on it, so the invariant would live in one place.

Not sure it's really worth it, the logic is just two booleans, duplicating it on the client side with its own test would be fine too. No strong opinion

0xB19:
> Do we agree that with this PR, massa-client still never uses the GRPC part of the massa-sdk? I've tested giving a wrong client .pem file with mtls enabled and it still works well, I expected the server to reject my requests.
> 
> I assume from the issue's description that this basically fix connecting to a server that enabled tls/mtls, so it's still a useful PR, but wondering if we need to expand on it or not?
> 
> Is it expected with the issues's scope @peterjah @0xB19?

This PR only fixes the transport: massa-client still never sends anything over gRPC. 

I think it's worth splitting the work though: keep this PR transport-only (the issue's scope), and open a dedicated issue for wiring client commands to gRPC because that part is bigger. 

A cheap first step could be moving a single command (e.g. get_status) to gRPC so the mTLS path gets actually exercised.

Leo-Besancon:
> > Additional note: @peterjah added the following check on the grpc side, I think it would be great to also call it in massa-client
> > ```rust
> > /// mTLS is a strict extension of TLS: reject the combination rather than silently serving in
> > /// plaintext a service the operator meant to protect with client certificates.
> > pub(crate) fn check_mtls_requires_tls(config: &GrpcConfig) -> Result<(), GrpcError> {
> >     if config.enable_mtls && !config.enable_tls {
> >         return Err(GrpcError::ConfigError(format!(
> >             "gRPC {:?} API: `enable_mtls` requires `enable_tls` to be enabled, refusing to start in plaintext",
> >             config.name
> >         )));
> >     }
> > 
> >     Ok(())
> > }
> > ```
> 
> Can we refactor this by moving the check to massa-sdk? Both massa-grpc and massa-client already depend on it, so the invariant would live in one place.
> 
> Not sure it's really worth it, the logic is just two booleans, duplicating it on the client side with its own test would be fine too. No strong opinion

I think it makes sense yes, as massa-sdk already handles certificates for mTLS / TLS, it can also include the check as a helper that can be called in both massa-grpc and massa-client.

0xB19:
Opened #5224 to track wiring the client commands to gRPC. Keeping this PR transport-only as agreed.

Leo-Besancon (approved):
LGTM!

PR #5213, opened 2026-08-25
fix(execution): cancel incoming requests when the read-only queue is full

### LINKED ISSUE #5114: [F49] Accepted read-only requests can be silently dropped during full-queue handoff

---

**Severity:** medium
**Component:** `massa-execution-worker/src/request_queue.rs` (`RequestQueue::extend`) + read-only handoff in `worker.rs`

## Summary

`RequestQueue::extend` is meant to append another queue and cancel any excess requests when
capacity is exceeded. But when the destination queue is **already full**, it returns early and
drops the incoming queue **without explicitly cancelling** its requests. In the read-only flow
this can happen during the handoff from the shared input queue to the worker-owned
`readonly_requests` queue, so a request can be accepted first and then silently lost — causing
timing-dependent failures / DoS for read-only requests.

Closely related to F51 (worker-local readonly cancellation on shutdown, PR #5110). The fix
(cancel dropped requests in the already-full early-return path) is small, but it interacts with
the queue capacity/eviction semantics and the read-only handoff ordering, so it should be
reviewed together with the read-only queue behaviour rather than patched blindly.

## Suggested direction

In `RequestQueue::extend`, when the destination is already full, cancel the incoming queue's
requests (with a clear "queue full" error) instead of dropping them silently; add a unit test
for the full-destination handoff path.

Tracking: finding F49 (related: F51 / PR #5110).

0xB19:
## Summary

Closes #5114 (finding F49, related to F51 / #5110). `RequestQueue::extend`
cancels excess requests when the incoming queue only partially fits, but
when the destination queue is **already full** it returned early and
dropped the whole incoming queue without cancelling it. `RequestQueue`
does not cancel on drop, so during the handoff from the shared input
queue to the worker-owned `readonly_requests` queue, an already-accepted
read-only request could see its response channel close with a generic
channel error instead of a clean capacity error.

## Change

In the full-destination early return of `RequestQueue::extend`, cancel
the incoming queue with the same "maximal request queue capacity
reached" error already used by the partial-overflow path and by `push`.
This is the direction suggested in the issue.

Blast radius: `extend` has a single call site, the read-only handoff in
`worker.rs` (`update_readonly_requests`), so no other queue behavior
changes. Requesters now always get an explicit answer: executed,
cancelled with a capacity error, or cancelled on worker shutdown
(#5110).

## Testing

- Two new unit tests in `request_queue.rs`:
  - full destination: every incoming request receives the capacity
    error (this test fails against the previous code, verified);
  - partial overflow: the fitting request is kept untouched, only the
    excess is cancelled (guards the pre-existing behavior).
- `massa_execution_worker` clippy and fmt clean.

## Risk

Low: the change only turns a silent drop into the existing explicit
cancellation, on a path where the requests were already lost. Not
consensus-observable (read-only request plumbing), so `main` is the
right target.

---

* [ ] try in sandbox /simulation/labnet (not applicable, covered by unit tests)
  * [x] make tests compile
  * [x] make tests pass
* [x] add logs allowing easy debugging in case the changes caused problems (the error message is the log)

### DIFF
--- a/massa-execution-worker/src/request_queue.rs
+++ b/massa-execution-worker/src/request_queue.rs
@@ -232,8 +232,8 @@ mod tests {
 
         destination.extend(incoming);
 
-        // the first incoming request fits and must not have received anything
-        assert!(rx1.try_recv().is_err());
+        // the first incoming request fits: nothing received, sender still alive
+        assert!(rx1.try_recv().unwrap_err().is_empty());
         // the second one is in excess and must be cancelled with a clean error
         assert_cancelled_for_capacity(rx2);
         assert!(destination.is_full());
--- a/massa-execution-worker/src/request_queue.rs
+++ b/massa-execution-worker/src/request_queue.rs
@@ -82,20 +82,13 @@ impl<T, R> RequestQueue<T, R> {
         // compute the number of available item slots
         let free_slots = self.max_items.saturating_sub(self.queue.len());
 
-        // If there are no available slots remaining, cancel the whole incoming queue.
-        // `RequestQueue` does not cancel on drop, so returning without cancelling would
-        // drop the response senders silently: already-accepted requests would fail with
-        // a generic channel error instead of this clean capacity error.
-        if free_slots == 0 {
-            other.cancel(ExecutionError::ChannelError(
-                "maximal request queue capacity reached".into(),
-            ));
-            return;
-        }
-
         // if there are not enough available slots to fit the entire incoming queue
         if free_slots < other.queue.len() {
-            // truncate the incoming queue to the size that fits, cancelling excess items
+            // Truncate the incoming queue to the size that fits, cancelling excess items.
+            // When the destination is already full this cancels the whole incoming queue:
+            // `RequestQueue` does not cancel on drop, so dropping it instead would close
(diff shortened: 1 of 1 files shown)


massa-execution-worker/src/request_queue.rs
```diff
@@ -82,8 +82,14 @@ impl<T, R> RequestQueue<T, R> {
...
+        // If there are no available slots remaining, cancel the whole incoming queue.
+        // `RequestQueue` does not cancel on drop, so returning without cancelling would
+        // drop the response senders silently: already-accepted requests would fail with
+        // a generic channel error instead of this clean capacity error.
         if free_slots == 0 {
+            other.cancel(ExecutionError::ChannelError(
+                "maximal request queue capacity reached".into(),
+            ));
             return;
         }
```
Leo-Besancon: this `if free_slots == 0` block is actually fully handled by the `.drain` and `.extend` calls below.
I would personally suggest to remove this branch and early return entirely to avoid (a small) code duplication.

Leo-Besancon (commented):
I just have a small suggestion you can consider, otherwise LGTM!

Leo-Besancon (approved):
LGTM, thanks!

PR #5097, opened 2026-07-21
fix(bootstrap): canonicalize IPs on white/black list mutations

damip:
## Summary
`SharedWhiteBlackList::add_ips_to_blacklist` / `add_ips_to_whitelist` inserted caller-supplied `IpAddr` values verbatim, while `is_ip_allowed` checks membership against `to_canonical(remote_addr.ip())` (and `load_list` canonicalizes entries on reload). Adding a non-canonical entry such as an IPv4-mapped IPv6 address (`::ffff:a.b.c.d`) through the private API therefore **failed to block the equivalent IPv4 peer** until the next file reload canonicalized the stored entry.

Addresses AI-report **Finding 17** (severity: minor).

## Fix
Canonicalize incoming IPs with `to_canonical` on insert (and on removal, for symmetry) so the live in-memory white/black lists always match the canonical form used at admission time.

## Scope / safety
Node-local ACL bookkeeping only. No consensus, wire-format, JSON-RPC, or gRPC surface is touched. Reload behavior already canonicalized, so on-disk semantics are unchanged.

## Testing
- New test `blacklisting_mapped_ipv6_blocks_equivalent_ipv4_immediately`: blacklisting `::ffff:127.0.0.2` immediately blocks a plain `127.0.0.2` peer (returns `BlackListed`), without waiting for a reload.
- `cargo test -p massa_bootstrap --lib`: pass. `cargo clippy --all-targets`: clean. `cargo fmt`: clean.

## Checklist
* [x] document all added functions
* [ ] try in sandbox /simulation/labnet
* [x] unit tests on the added/changed features
  * [x] make tests compile
  * [x] make tests pass
* [x] add logs allowing easy debugging (n/a)
* [x] if the API has changed, update the API specification (n/a)

Made with [Cursor](https://cursor.com)

PR #5214, opened 2026-08-25
fix(bootstrap): keep the loaded access lists when their file becomes unreadable

### LINKED ISSUE #5122: [PDF#13] Bootstrap whitelist updater fails open when the whitelist file disappears/unreadable

---

**Severity:** medium (security — bootstrap access control fail-open)
**Component:** `massa-bootstrap/src/white_black_list.rs` (`load_list`, `is_ip_allowed`)
**Source:** PDF submission #13 (reporter jokermicrolab@gmail.com); partial overlap with F16/F17.

## Summary

On non-init refreshes, a whitelist/blacklist read failure is treated as `Ok(None)`
(`load_list(..., is_init=false)` returns `Ok(None)` on `read_to_string` error, ~line 197). A
`None` `white_list` makes `is_ip_allowed` (~line 140) allow all IPs. So if the whitelist file
disappears or becomes unreadable at runtime, bootstrap access control **fails open**: access is
silently opened until the file is restored and reloaded.

## Relationship to existing work

- **F16** (PR #5096): malformed-JSON reload no longer kills the refresher (keeps stale list).
- **F17** (PR #5097): canonicalizes IPs on ACL mutations.
- This PDF#13 nuance — *fail-open on a missing/unreadable file* (as opposed to malformed JSON) —
  is distinct from both and not addressed by those PRs.

The natural fix (fail **closed**, or retain the previous in-memory list on read failure instead
of clearing it) changes bootstrap admission behaviour when the file is absent/unreadable. This
is a security-relevant policy change that could affect operators who intentionally remove the
file, and must be coordinated with the F16 "keep stale list on reload error" behaviour. Needs
maintainer decision on the desired failure mode.

## Suggested direction

On a non-init refresh read failure, do **not** collapse to `None` (fail-open). Either retain the
last successfully-loaded list (consistent with F16) or fail closed, per maintainer policy; add a
test that a disappearing/unreadable whitelist file does not open access.

Tracking: PDF submission #13 (related: F16 #5096, F17 #5097).

0xB19:
## Summary

Closes #5122 (PDF submission #13, related to F16 #5096 and F17 #5097).
On a periodic refresh, a whitelist/blacklist file read failure was
collapsed to `None`, and a `None` whitelist allows all IPs: if the
configured whitelist file became unreadable at runtime, bootstrap
access control silently **failed open** until the file came back.

## Change

Design settled in review (thanks @damip, @Leo-Besancon): distinguish
the two failure modes instead of treating them alike.

- **Missing file** (`NotFound`): an explicit operator choice, the list
  is disabled on the next refresh, matching the pre-existing operator
  workflow. A warning states that the list is being disabled when one
  was loaded; a node with no list files (the common case) stays silent,
  since a missing file is its normal state.
- **Unreadable file** (exists, but reading fails: permissions, IO
  error): keep the previously loaded list and warn. Failing open here
  would silently disable bootstrap access control on a filesystem
  hiccup.
- `load_list` (startup): unchanged behaviour, a missing/unreadable file
  disables the feature with a warning.
- Parsing is shared and unchanged (malformed JSON still errors, and the
  refresher keeps the previous lists, as per F16). The parse error
  message now names the right list instead of hardcoding "whitelist".
- The list files are now read before taking the inner lock, so a slow
  or hung filesystem read cannot stall `is_ip_allowed` callers.

## Testing

- New unit tests: deleting a list file disables it on the next refresh
  (whitelist and blacklist), an unreadable-but-existing path (a
  directory) keeps the loaded whitelist instead of failing open, a node
  without configured lists stays open and silent across refreshes, and
  a recreated file is picked up again by the next refresh.
- Full `massa_bootstrap` suite passes, clippy and fmt clean.

## Risk

Low: the refresh path is node-local policy, not consensus-observable,
so `main` is the right target. Operator-visible nuance: while a list
file is unreadable (not deleted), the warn fires at every refresh on
purpose, as a standing reminder that the on-disk state and the
in-memory list diverge.

---

* [ ] try in sandbox /simulation/labnet (not applicable, covered by unit tests)
  * [x] make tests compile
  * [x] make tests pass


Leo-Besancon (commented):
The implem for the fix looks good, but in the issue I had [a comment](https://github.com/massalabs/massa/issues/5122#issuecomment-5262905241) on whether we want this or not.

This makes it impossible to open up a node without restart, which may be needed in some emergency cases (like we're the only node alive). I'd be waiting for @damip thoughts on this before merging (on my side I would lean towards not merging this).

We could potentially add an explicit allow_all file to bypass the behaviour introduced here, but it's not the cleanest.

massa-bootstrap/src/white_black_list.rs
```diff
@@ -170,59 +178,73 @@ impl SharedWhiteBlackList<'_> {
...
+            Ok(list) => Ok(Some(Self::parse_list(&list)?)),
         }
     }
+
+    /// Parse the JSON content of a list file into a canonicalized IP set.
+    fn parse_list(list: &str) -> Result<HashSet<IpAddr>, BootstrapError> {
+        Ok(serde_json::from_str::<HashSet<IpAddr>>(list)
+            .map_err(|e| {
+                BootstrapError::InitListError(format!(
+                    "Failed to parse bootstrap whitelist : {}",
```
damip: Nit: this error message hardcodes "whitelist" even when the blacklist fails to parse. Pre-existing wording, but since `refresh_list` now has a `list_kind` parameter, threading it into `parse_list` would make the error accurate for free (and `load_list` could pass a generic label or gain the same parameter).

massa-bootstrap/src/white_black_list.rs
```diff
@@ -124,8 +124,16 @@ impl SharedWhiteBlackList<'_> {
     /// Creates a new list, and replaces the old one in a write-lock
     pub(crate) fn update(&mut self) -> Result<(), BootstrapError> {
         let read_lock = self.inner.read();
```
damip, line 130: Note (pre-existing, non-blocking): the inner read lock is held across the file I/O in the two `refresh_list` calls below, so a slow/hung filesystem read would stall `is_ip_allowed` callers for the duration. Harmless in practice since the updater is a single thread and the reads are tiny, just flagging it since this function was reshaped anyway. Reading the files before taking the lock would remove the coupling if anyone ever cares.

damip (commented):
The fix looks correct to me: `refresh_list` retaining the previously loaded list on read failure closes the fail-open, the startup/refresh asymmetry is well justified (startup cannot distinguish "not configured" from "unreadable", refresh can, thanks to the previously loaded list), parse errors still flow into the refresher's existing keep-and-retry path, and the tests cover the actual attack scenario plus recovery. Two non-blocking nits inline.

damip:
Maybe the trick could be to explicitly check for file existence. If the file actually does not exist (it was not a fialure-to-open) we consider it disabled. Otherwise we keep the previously read value as intended here. WDYT @Leo-Besancon @0xB19  ? it's basically the current behavior but clean and explicit

Leo-Besancon:
> Maybe the trick could be to explicitly check for file existence. If the file actually does not exist (it was not a fialure-to-open) we consider it disabled. Otherwise we keep the previously read value as intended here. WDYT @Leo-Besancon @0xB19 ? it's basically the current behavior but clean and explicit

Yes I think this would be better on my side!

Leo-Besancon (approved):
LGTM with the latest changes!

bilboquet (approved):

PR #5215, opened 2026-08-25
[F65] feat: prevent using stale production_stats during network-restart

### LINKED ISSUE #5014: [F65] Stale Production Statistics Are Propagated Across Interpolated Restart Cycles
> ⚠️ **Warning — This finding was routed to an issue rather than a PR because it affects persisted final-state / ledger / DB recovery and bootstrap-restart handling.** A fix changes on-disk or state-recovery behavior that must stay consistent across nodes and restarts; a wrong change can corrupt or diverge state, so it needs design review and careful (possibly coordinated) rollout.

---

## Source
AI security-scan **Finding 65** (severity: **medium**). Tracked in `SUBMISSIONS.md` as F65.

**Affected file:** `massa-final-state/src/final_state.rs`

## Summary
`create_new_cycle_from_last` initializes a new cycle by copying `last_cycle_info.production_stats`. During a multi-cycle restart, `interpolate_multiple_cycles` reuses the same `latest_snapshot_cycle_info` for every interpolated cycle between the snapshot and the restart slot. As a result, cycles that happened fully during downtime are created with stale production statistics instead of empty statistics. If the restart stops inside one of those interpolated cycles, later settlement can count old successes or misses again and may trigger unjustified auto-sell of rolls, while also exposing incorrect accountability data for the affected interpolated cycles.

## Root cause
The issue comes from `create_new_cycle_from_last`, which copies `last_cycle_info.production_stats` into each newly created cycle. This behavior carries over production data from the previous cycle instead of resetting it for a newly initialized cycle.

During restart, `interpolate_multiple_cycles` uses the same `latest_snapshot_cycle_info` for all cycles created between the snapshot and the restart slot. Because that reused source already contains `production_stats`, every interpolated cycle inherits the production statistics of the pre-shutdown snapshot cycle.

This becomes especially problematic when the restart ends inside one of those interpolated future cycles. In that case, the final state for that cycle already contains stale `production_stats`, and post-restart settlement later reads and combines them with fresh statistics collected after restart.

## Affected code
```rust
/// Initializes a `FinalState` from a snapshot.
    ///
    /// # Arguments
    /// * `db`: A type that implements the `MassaDBController` trait. Used to read and write to the database.
    /// * `config`: the configuration of the final state to use for initialization
    /// * `ledger`: the instance of the ledger on disk. Used to apply changes to the ledger.
    /// * `selector`: the pos selector. Used to send draw inputs when a new cycle is completed.
    /// * `last_start_period`: at what period we should attach the final_state
    pub fn new_derived_from_snapshot(
        db: ShareableMassaDBController,
        config: FinalStateConfig,
        ledger: Box<dyn LedgerController>,
        selector: Box<dyn SelectorController>,
        mip_store: MipStore,
        last_start_period: u64,
    ) -> Result<Self, FinalStateError> {
        info!("Restarting from snapshot");

        let mut final_state =
            FinalState::new(db, config.clone(), ledger, selector, mip_store, false)?;

        let recovered_slot =
            final_state.db.read().get_change_id().map_err(|_| {
                FinalStateError::InvalidSlot(String::from("Could not get slot in db"))
            })?;

        // This is needed for `test_bootstrap_server` to work
        if cfg!(feature = "test-exports") {
            let mut batch = DBBatch::new();
            final_state.pos_state.create_initial_cycle(&mut batch);
            final_state
                .db
                .write()
                .write_batch(batch, Default::default(), Some(recovered_slot));
        }

        final_state.last_slot_before_downtime = Some(recovered_slot);

        // Check that MIP store is consistent with the network shutdown time range
        // Assume that the final state has been edited during network shutdown
        let shutdown_start = recovered_slot
            .get_next_slot(config.thread_count)
            .map_err(|e| {
                FinalStateError::InvalidSlot(format!(
                    "Unable to get next slot from recovered slot: {:?}",
// … (truncated)
```

## Recommendation
Initialize `production_stats` to an empty state whenever a new cycle is created during restart interpolation, rather than copying it from `last_cycle_info`. Ensure each interpolated cycle is treated as a fresh cycle with no inherited production results, especially when `interpolate_multiple_cycles` derives several cycles from the same `latest_snapshot_cycle_info`. Also verify that the partially resumed cycle at the restart slot cannot retain stale statistics from the pre-shutdown snapshot, so post-restart production data is recorded and settled only against statistics generated within that cycle.

## Why P0 / breaking

Leo-Besancon:
  * [ ] make tests compile
  * [ ] make tests pass 

### DIFF
--- a/massa-pos-exports/src/pos_final_state.rs
+++ b/massa-pos-exports/src/pos_final_state.rs
@@ -344,9 +344,19 @@ impl PoSFinalState {
         );
     }
 
-    /// Create the a cycle based off of another cycle_info.
+    /// Create a cycle based off of another cycle_info.
     ///
     /// Used for downtime interpolation, when restarting from a snapshot.
+    /// Roll counts are always carried over from `last_cycle_info`.
+    ///
+    /// `rng_seed` and `production_stats` are kept only when continuing the same
+    /// incomplete cycle (`last_cycle_info.cycle` matches the target cycle and
+    /// `first_slot` is not the first slot of that cycle). Otherwise the seed
+    /// starts fresh and `production_stats` are empty, so interpolated future
+    /// cycles do not inherit stale production results from the snapshot.
+    ///
+    /// In all cases, `rng_seed` is then extended with `false` bits for every
+    /// interpolated slot in `[first_slot, last_slot]`.
     pub fn create_new_cycle_from_last(
         &mut self,
         last_cycle_info: &CycleInfo,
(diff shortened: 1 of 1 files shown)


peterjah (dismissed):

bilboquet:
LGTM.
I added a test (proposed by AI), feel free to keep or kill it.

bilboquet (dismissed):

Leo-Besancon:
> LGTM. I added a test (proposed by AI), feel free to keep or kill it.

Thanks! Keeping it, I've rebased on main and fixed the CI.

bilboquet (approved):

PR #5218, opened 2026-08-26
[F98] feat: document allowed malleability for deferred credits serialization

### LINKED ISSUE #5029: [F98] Serialization Malleability via Unenforced Key Ordering

---

## Source
AI security-scan **Finding 98** (severity: **info**). Tracked in `SUBMISSIONS.md` as F98.

**Affected file:** `massa-pos-exports/src/deferred_credits.rs`

## Summary
`DeferredCreditsDeserializer::deserialize` in `massa-main/massa-pos-exports/src/deferred_credits.rs` accepts `DeferredCredits` entries in arbitrary `Slot` order and then normalizes them by collecting into a `BTreeMap`. As a result, multiple distinct serialized byte sequences with the same set of unique `Slot` keys deserialize to the same in-memory `DeferredCredits` value. The original claim that this behavior leads to signature verification failures or consensus forks is not supported by the codebase.

## Root cause
The deserializer uses `length_count(...)` to read a sequence of `(Slot, Credits)` pairs and then constructs `DeferredCredits` with `credits: elements.into_iter().collect()`. Because the target container is a `BTreeMap`, insertion canonicalizes ordering by sorting on `Slot`, but `deserialize` does not verify that the incoming serialized entries were already strictly increasing or otherwise canonical.

This creates a *serialization malleability* issue at the object level: distinct non-canonical byte representations can map to the same `DeferredCredits` instance. However, the broader root-cause theory that this normalization affects signed network payloads or consensus-critical hashes does not hold here, because the identified network paths verify signatures on raw bytes before deserialization, and the aggregate `DeferredCredits` binary format does not appear to be used as a consensus-critical wire format.

## Affected code
```rust
        &self,
        buffer: &'a [u8],
    ) -> IResult<&'a [u8], DeferredCredits, E> {
        context(
            "Failed DeferredCredits deserialization",
            length_count(
                    self.u64_deserializer.deserialize(input)
                }),
                tuple((
                    }),
                    context("Failed credit deserialization", |input| {
                        self.credit_deserializer.deserialize(input)
                    }),
                )),
            ),
        )
        .map(|elements| DeferredCredits {
            credits: elements.into_iter().collect(),
        })
        .parse(buffer)
    }
```

## Recommendation
Enforce canonical ordering during `DeferredCreditsDeserializer::deserialize` by rejecting inputs whose `Slot` entries are not strictly increasing in the serialized stream. This ensures a one-to-one mapping between serialized bytes and the resulting `DeferredCredits` value, prevents acceptance of non-canonical encodings, and preserves stable round-trip serialization. Additionally, document the required ordering as part of the format specification and add validation tests to confirm that out-of-order or otherwise non-canonical inputs are rejected.

## Why P0 / breaking

Leo-Besancon:
  * [ ] make tests compile
  * [ ] make tests pass 

### DIFF
--- a/massa-pos-exports/src/deferred_credits.rs
+++ b/massa-pos-exports/src/deferred_credits.rs
@@ -155,6 +155,9 @@ impl DeferredCreditsSerializer {
 }
 
 impl Serializer<DeferredCredits> for DeferredCreditsSerializer {
+    /// Serializes deferred credits with slots in ascending `BTreeMap` order.
+    ///
+    /// The matching deserializer requires that same strictly increasing slot order.
     fn serialize(
         &self,
         value: &DeferredCredits,
@@ -163,7 +166,7 @@ impl Serializer<DeferredCredits> for DeferredCreditsSerializer {
         // deferred credits length
         self.u64_ser
             .serialize(&(value.credits.len() as u64), buffer)?;
-        // deferred credits
+        // deferred credits (BTreeMap iteration is sorted by Slot)
         for (slot, credits) in &value.credits {
             // slot
             self.slot_ser.serialize(slot, buffer)?;
@@ -201,13 +204,17 @@ impl DeferredCreditsDeserializer {
 }
 
 impl Deserializer<DeferredCredits> for DeferredCreditsDeserializer {
+    /// Deserializes deferred credits.
+    ///
+    /// Slot entries must appear in strictly increasing order (the order produced by
+    /// [`DeferredCreditsSerializer`]). Out-of-order or duplicate slots are rejected so
+    /// each valid byte sequence maps to exactly one `DeferredCredits` value.
     fn deserialize<'a, E: ParseError<&'a [u8]> + ContextError<&'a [u8]>>(
         &self,
         buffer: &'a [u8],
     ) -> IResult<&'a [u8], DeferredCredits, E> {
-        context(
-            "Failed DeferredCredits deserialization",
-            length_count(
+        context("Failed DeferredCredits deserialization", |input| {
+            let (rest, elements) = length_count(
                 context("Failed length deserialization", |input| {
                     self.u64_deserializer.deserialize(input)
                 }),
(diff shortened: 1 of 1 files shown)


peterjah (dismissed):

Leo-Besancon:
@damip I don't think this would cause as many issues as https://github.com/massalabs/massa/pull/5239 or https://github.com/massalabs/massa/pull/5220, but do you want that we also just document the allowed malleability regarding deferred credits?

damip:
Same situation as F35 (#5239). I checked: DeferredCreditsDeserializer::deserialize has NO production caller. pos_final_state reads deferred credits per key with the sub deserializers, and the aggregate is only reached through PoSChangesDeserializer / StateChangesDeserializer, which are only built in cfg(test) (bootstrap streams DB batches, not StateChanges). So this hardens a parser that nothing can feed.

Let's be consistent with #5239 and #5220: drop the check and just leave a comment:

```rust
// Note: unsorted slots, and duplicate slots (last occurrence wins), on the wire still deserialize
// to a normalized BTreeMap. This deserializer is only reached through PoSChangesDeserializer /
// StateChangesDeserializer, which are test-only: production reads deferred credits per key from
// the DB and bootstrap streams DB batches. Massa is malleability-resistant by construction, this
// is not exploitable.
```

Longer term we should probably just delete the dead aggregate (de)serializers for F35 and F98 in a cleanup PR, a parser that does not exist cannot be malleable. WDYT @Leo-Besancon ?

Leo-Besancon:
> Same situation as F35 (#5239). I checked: DeferredCreditsDeserializer::deserialize has NO production caller. pos_final_state reads deferred credits per key with the sub deserializers, and the aggregate is only reached through PoSChangesDeserializer / StateChangesDeserializer, which are only built in cfg(test) (bootstrap streams DB batches, not StateChanges). So this hardens a parser that nothing can feed.
> 
> Let's be consistent with #5239 and #5220: drop the check and just leave a comment:
> 
> ```rust
> // Note: unsorted slots, and duplicate slots (last occurrence wins), on the wire still deserialize
> // to a normalized BTreeMap. This deserializer is only reached through PoSChangesDeserializer /
> // StateChangesDeserializer, which are test-only: production reads deferred credits per key from
> // the DB and bootstrap streams DB batches. Massa is malleability-resistant by construction, this
> // is not exploitable.
> ```
> 
> Longer term we should probably just delete the dead aggregate (de)serializers for F35 and F98 in a cleanup PR, a parser that does not exist cannot be malleable. WDYT @Leo-Besancon ?

Good idea, created the followup issue here: https://github.com/massalabs/massa/issues/5254

bilboquet (approved):

PR #5219, opened 2026-08-26
[F115] validate the PoS endorsement draw on the public gRPC send_endo…

### LINKED ISSUE #5038: [F115] Propagation thread rebroadcasts public endorsements without PoS producer validation

---

## Source
AI security-scan **Finding 115** (severity: **medium**). Tracked in `SUBMISSIONS.md` as F115.

**Affected file:** `massa-protocol-worker/src/handlers/endorsement_handler/propagation.rs`

## Summary
The confirmed issue is an *externally reachable validation bypass* in the public gRPC `send_endorsements` path. External endorsements that are only deserialized and signature-checked can be forwarded into `ProtocolController::propagate_endorsements`, which then reaches the propagation worker in `massa-main/massa-protocol-worker/src/handlers/endorsement_handler/propagation.rs` and is rebroadcast without the producer-selection check enforced elsewhere. The original finding is only partially correct because the propagation worker's trust-in-caller behavior is an *intended internal fast path* for endorsements coming "from pool," not an inherently unsafe design by itself.

## Root cause
The root cause is that the *public gRPC `send_endorsements` entrypoint* accepts endorsements from untrusted remote clients, only deserializes them and verifies signatures, and then forwards them into `protocol_controller.propagate_endorsements()` without applying the proof-of-stake producer-selection validation used in `note_endorsements_from_peer` and the endorsement pool.

Inside the propagation worker, `EndorsementHandlerPropagationCommand::PropagateEndorsements` is intentionally handled as a trusted internal path: `run()` batches incoming propagation requests and calls `self.propagate_endorsements(endorsements)`. That internal trust model is not itself the bug. The ruled-out portion of the original finding is the claim that the propagation thread is inherently unsafe because it does not revalidate caller-supplied endorsements; according to the validated analysis, this omission is *intended design* for internal callers and becomes unsafe only because an *untrusted public entrypoint* uses that path.

## Affected code
```rust
fn run(&mut self) {
        let mut next_message = None;
        loop {
            // get the next message to process
            let msg = match next_message.take() {
                Some(msg) => msg,
                None => match self.receiver.recv() {
                    Ok(msg) => msg,
                    Err(_) => {
                        info!("Stop endorsement propagation thread");
                        return;
                    }
                },
            };

            match msg {
                // endorsements to propagate
                EndorsementHandlerPropagationCommand::PropagateEndorsements(mut endorsements) => {
                    // also drain any remaining propagation messages that might have accumulated
                    while let Ok(msg) = self.receiver.try_recv() {
                        match msg {
                            // we got more endorsements to propagate: extend the buffer
                            EndorsementHandlerPropagationCommand::PropagateEndorsements(
                                new_endorsements,
                            ) => {
                                endorsements.extend(new_endorsements);
                            }
                            // we grabbed a message that is not a propagation message, mark it for processing
                            other_msg => {
                                next_message = Some(other_msg);
                                break;
                            }
                        }
                    }
                    // propagate the endorsements
                    self.propagate_endorsements(endorsements);
                }
                // stop the handler
                EndorsementHandlerPropagationCommand::Stop => {
                    info!("Stop endorsement propagation thread");
                    return;
                }
            }
        }
    }
```

## Recommendation
Apply the proof-of-stake producer-selection validation before any endorsement received through the public gRPC `send_endorsements` endpoint is forwarded to `protocol_controller.propagate_endorsements()`. Align this entrypoint with the validation performed in `note_endorsements_from_peer` and in the endorsement pool so that only endorsements valid for the corresponding `(slot, index)` producer assignment can enter the trusted propagation flow.

If `propagate_endorsements` is intended to remain an internal fast path, explicitly restrict its callers to trusted, already-validated sources and document that requirement. As a defense-in-depth measure, consider enforcing the same producer-selection invariant at the protocol-controller boundary or rejecting propagation requests that have not passed full endorsement validation.

## Why P0 / breaking

peterjah:
…rsements path

  * [ ] make tests compile
  * [ ] make tests pass 

### DIFF
--- a/massa-grpc/src/stream/send_endorsements.rs
+++ b/massa-grpc/src/stream/send_endorsements.rs
@@ -5,6 +5,7 @@ use crate::server::MassaPublicGrpc;
 use futures_util::StreamExt;
 use massa_models::endorsement::{EndorsementDeserializer, SecureShareEndorsement};
 use massa_models::secure_share::SecureShareDeserializer;
+use massa_pos_exports::SelectorController;
 use massa_proto_rs::massa::api::v1 as grpc_api;
 use massa_proto_rs::massa::model::v1 as grpc_model;
 use massa_serialization::{DeserializeError, Deserializer};
@@ -31,6 +32,7 @@ pub(crate) async fn send_endorsements(
 ) -> Result<SendEndorsementsStreamType, GrpcError> {
     let mut pool_command_sender = grpc.pool_controller.clone();
     let protocol_command_sender = grpc.protocol_controller.clone();
+    let selector_controller = grpc.selector_controller.clone();
     let config = grpc.grpc_config.clone();
     let storage = grpc.storage.clone_without_refs();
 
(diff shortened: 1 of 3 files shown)


Leo-Besancon (approved):
LGTM

PR #5220, opened 2026-08-26
[F89] docs: add comments regarding not rejecting unsorted datastore keys in operations

### LINKED ISSUE #5025: [F89] Payload Malleability via Unenforced Key Ordering and Duplicates in Datastore Deserialization

---

## Source
AI security-scan **Finding 89** (severity: **medium**). Tracked in `SUBMISSIONS.md` as F89.

**Affected file:** `massa-models/src/datastore.rs`

## Summary
`DatastoreDeserializer::deserialize` in `massa-main/massa-models/src/datastore.rs` accepts a length-prefixed sequence of datastore entries and converts it directly into a `BTreeMap` with `.collect()`. Because this path does not enforce *strict key ordering* or *key uniqueness*, distinct serialized byte sequences can deserialize into the same logical datastore. This becomes a real issue for `ExecuteSC` operations because arbitrary `serialized_content` bytes are accepted, while `OperationId` and signature verification are computed from those exact bytes. The original finding's broader claims about third-party malleability, replay, and reserialization-induced signature breakage are not supported.

## Root cause
The confirmed root cause is that `DatastoreDeserializer::deserialize` parses `(key, value)` pairs and then executes `elements.into_iter().collect()` into a `BTreeMap`. In Rust, this causes *out-of-order keys* to be sorted and *duplicate keys* to be collapsed, so multiple non-canonical encodings map to the same datastore state.

This behavior is practically reachable because operation submission accepts arbitrary `serialized_content` bytes in `massa-main/massa-api-exports/src/operation.rs` and `massa-main/massa-api/src/public.rs`, rather than requiring a canonical datastore encoding before hashing and signing. As a result, semantically equivalent `ExecuteSC` payloads can be represented by different byte streams.

The ruled-out portion of the original theory is the claim that this enables post-signature modification by a third party. Each byte-level variant must be independently signed by the sender because `OperationId` and signature checks are performed over the original serialized bytes. The additional claim that nodes canonicalize and reserialize the datastore during gossip in a way that invalidates signatures is also incorrect, because `SecureShareSerializer` re-emits the original `serialized_data` bytes.

## Affected code
```rust
        &self,
        buffer: &'a [u8],
    ) -> IResult<&'a [u8], BTreeMap<Vec<u8>, Vec<u8>>, E> {
        context(
            "Failed Datastore deserialization",
            length_count(
                    self.length_deserializer.deserialize(input)
                }),
                tuple((
                    context("Failed key deserialization", |input| {
                        self.key_deserializer.deserialize(input)
                    }),
                    context("Failed value deserialization", |input| {
                        self.value_deserializer.deserialize(input)
                    }),
                )),
            ),
        )
        .map(|elements| elements.into_iter().collect())
        .parse(buffer)
    }
```

## Recommendation
Enforce a canonical datastore encoding at deserialization time by rejecting any payload that contains duplicate keys or keys that are not strictly ordered according to the datastore’s canonical sort order. This ensures that each logical datastore maps to exactly one valid byte representation.

In addition, validate canonical form before hashing, signing, accepting, or propagating `ExecuteSC` operations so that non-canonical `serialized_content` cannot produce distinct `OperationId` values for equivalent datastore states. If arbitrary pre-serialized content must remain supported for compatibility, add explicit validation that fails on non-canonical encodings rather than normalizing them implicitly during collection into a `BTreeMap`.

Review operation pool and storage deduplication assumptions to ensure they do not rely solely on `OperationId` uniqueness when semantically equivalent but differently encoded payloads may still be admitted by legacy paths.

## Why P0 / breaking

Leo-Besancon:
  * [ ] make tests compile
  * [ ] make tests pass 

Note: following the discussion below, we're only keeping a small comment in the deserializer and rebased on top of main, and not implementing the finding's recommended changes as they don't improve security.

### DIFF
--- a/massa-models/src/datastore.rs
+++ b/massa-models/src/datastore.rs
@@ -139,6 +139,10 @@ impl Deserializer<Datastore> for DatastoreDeserializer {
                 )),
             ),
         )
+        // Note: Unsorted key/value pairs on the wire still deserialize to the same BTreeMap.
+        // That can yield different raw bytes / OperationIds for the same logical action.
+        // This is intentional malleability that Massa tolerates by construction and is not exploitable.
+        // As updating this behavior would be a breaking change, we do not reject it.
         .map(|elements| elements.into_iter().collect())
         .parse(buffer)
     }
--- a/massa-models/src/datastore.rs
+++ b/massa-models/src/datastore.rs
@@ -139,10 +139,12 @@ impl Deserializer<Datastore> for DatastoreDeserializer {
                 )),
             ),
         )
-        // Note: Unsorted key/value pairs on the wire still deserialize to the same BTreeMap.
-        // That can yield different raw bytes / OperationIds for the same logical action.
-        // This is intentional malleability that Massa tolerates by construction and is not exploitable.
-        // As updating this behavior would be a breaking change, we do not reject it.
+        // Note: unsorted key/value pairs, and duplicate keys (last occurrence wins), on the wire
+        // still deserialize to the same normalized BTreeMap. So different raw bytes / OperationIds can
+        // encode the same logical action. Massa is malleability-resistant by construction: nothing
+        // relies on canonical encodings, and only the signer can produce such variants (who can already
+        // produce distinct ids for the same action anyway), so this is not exploitable.
+        // Rejecting non-canonical encodings would be a breaking change for no security gain.
         .map(|elements| elements.into_iter().collect())
         .parse(buffer)
     }


Leo-Besancon:
@damip would be great if you could review that the impacts (outlined in the PR comment above) are sane regarding the versioning

damip:
Massa is designed to be malleability resistant.
This is one example of malleability and contrary to what the issue suggests, it cant be exploited at all.
It's OK by construction: different raw transactions ARE allowed to represent the same action, it does not lead to any exploitable issues.
Not worth a versioning/network upgrade for this given the risks, also it's potentially breaking for some apps /APIs.
Worth writing a comment in the code to explain this

damip:
Good, that's the right resolution. Two small things:

- the comment should also mention duplicate keys and that the LAST occurrence wins. That is the one implicit rule an alternative implementation would have to replicate, so let's pin it. Proposed:

```rust
// Note: unsorted key/value pairs, and duplicate keys (last occurrence wins), on the wire
// still deserialize to the same normalized BTreeMap. So different raw bytes / OperationIds can
// encode the same logical action. Massa is malleability-resistant by construction: nothing
// relies on canonical encodings, and only the signer can produce such variants (who can already
// produce distinct ids for the same action anyway), so this is not exploitable.
// Rejecting non-canonical encodings would be a breaking change for no security gain.
```

- please rewrite the PR description, it still describes the old MIP-gated implementation (validity_start gating, API/GRPC rejection, cleanup in #5127) which is not what the PR does anymore.

Then let's merge.

Leo-Besancon:
> Good, that's the right resolution. Two small things:
> 
>     * the comment should also mention duplicate keys and that the LAST occurrence wins. That is the one implicit rule an alternative implementation would have to replicate, so let's pin it. Proposed:
> 
> 
> ```rust
> // Note: unsorted key/value pairs, and duplicate keys (last occurrence wins), on the wire
> // still deserialize to the same normalized BTreeMap. So different raw bytes / OperationIds can
> // encode the same logical action. Massa is malleability-resistant by construction: nothing
> // relies on canonical encodings, and only the signer can produce such variants (who can already
> // produce distinct ids for the same action anyway), so this is not exploitable.
> // Rejecting non-canonical encodings would be a breaking change for no security gain.
> ```
> 
>     * please rewrite the PR description, it still describes the old MIP-gated implementation (validity_start gating, API/GRPC rejection, cleanup in [Cleanup outdated code after breaking update release #5127](https://github.com/massalabs/massa/issues/5127)) which is not what the PR does anymore.
> 
> 
> Then let's merge.

Done both suggestions!

bilboquet (approved):

Issue #4982, opened 2026-07-21
[F19] Silent Dropping of Consensus Commands via `try_send` Channel Saturation


damip:
> ⚠️ **Warning — This finding was routed to an issue rather than a PR because it touches consensus / block-graph processing or block/operation validity rules.** Changing what a node accepts, rejects, or how it advances state can make patched and unpatched nodes diverge (a network split), so it requires careful design and a coordinated, version-gated (MIP / hard-fork) rollout rather than a self-contained PR.

---

## Source
AI security-scan **Finding 19** (severity: **medium**). Tracked in `SUBMISSIONS.md` as F19.

External bug-bounty PDF #12 (try_send command dropping).

**Affected file:** `massa-consensus-worker/src/controller.rs`

## Summary
`register_block_header` in `massa-consensus-worker/src/controller.rs` forwards `ConsensusCommand::RegisterBlockHeader` through a bounded `command_sender` using `try_send`. When that queue is saturated, the function only emits a warning and drops the header instead of preserving or retrying it. This creates a real risk of consensus message loss under overload. The original claim is overstated, however, because a dropped header does not by itself guarantee permanent desynchronization or network divergence.

## Root cause
The confirmed issue is the use of non-blocking `try_send` in `register_block_header` for the consensus `command_sender` queue:

- `self.command_sender.try_send(ConsensusCommand::RegisterBlockHeader(block_id, header))` fails immediately when the bounded channel is full.
- On failure, the function only logs `warn!("error trying to register a block header: {}", err);` and does not retry, block, propagate the error, or otherwise retain the dropped header.

This means valid headers can be discarded during queue saturation. It is also supported that peers can help create this condition, because new signed headers are forwarded toward consensus before consensus verifies block-producer eligibility, allowing many bogus-but-well-formed headers to consume queue capacity. The ruled-out portion of the original theory is that any single dropped legitimate header necessarily causes lasting consensus failure; that conclusion does not hold because later dependency handling can re-request missing data.

## Affected code
```rust
fn register_block_header(&self, block_id: BlockId, header: SecureShare<BlockHeader, BlockId>) {
        if self.broadcast_enabled {
            if let Err(err) = self.broadcasts.block_header_sender.send(header.clone()) {
                trace!(
                    "error, failed to broadcast block header with block id {}: {}",
                    block_id,
                    err
                );
            }
        }

        if let Err(err) = self
            .command_sender
            .try_send(ConsensusCommand::RegisterBlockHeader(block_id, header))
[...]

PR #5221, opened 2026-08-26
[F109] forward wishlisted block headers to consensus and gate content…

### LINKED ISSUE #5033: [F109] Delayed Consensus Header Validation for Wishlisted Blocks Enables Bandwidth and CPU DoS

---

## Source
AI security-scan **Finding 109** (severity: **major**). Tracked in `SUBMISSIONS.md` as F109.

## Summary
In `block_wishlist`, if a block ID is already tracked, a header accepted by `note_header_from_peer` is kept only locally and is *not* forwarded to consensus at the header stage. Because of this, consensus does not run its normal header-only validation early. Those checks are delayed until `fully_gathered_block` submits the full block. This can let an attacker make the node download operation IDs and possibly the full block for headers that consensus would later reject, creating a practical *bandwidth* and *CPU denial of service* risk.

## Root cause
The issue starts when a block ID already exists in `block_wishlist`. In that case, a header that passes `note_header_from_peer` is stored locally, but it is *not sent to consensus* during the header stage.

Because consensus does not receive that header early, it cannot perform its header-only checks at that time. The delayed checks include:
- *producer selection*
- *stale/future handling*
- *parent compatibility*
- *dependency topology*

Instead, those validations only happen later, when `fully_gathered_block` submits the complete block to consensus.

This ordering creates a gap. A peer can first populate `block_wishlist` with valid child headers that reference missing parents. Then the peer can respond to the missing-parent requests with parent headers that are structurally valid for `note_header_from_peer` but still *consensus-invalid*. Since consensus has not yet rejected them at the header stage, the protocol may continue requesting operation IDs and downloading more data before the block is finally rejected.

## Affected code
```rust
/// Performs validity checks on a block header,
    /// and if valid update the node's view of its surrounding peers.
    ///
    /// Returns a boolean indicating whether the header is new.
    ///
    /// Does not ban the source node if the header is invalid.
    ///
    /// Checks performed on Header:
    /// - Not genesis
    /// - Compatible version
    /// - Can compute a `BlockId`
    /// - Valid signature
    /// - All endorsement are valid
    /// - Endorsements have unique indices
    /// - Endorsement slots match that of the block
    /// - Endorsed blocks match the same-thread parent of the header
    pub(crate) fn note_header_from_peer(
        &mut self,
        header: &SecuredHeader,
        from_peer_id: &PeerId,
    ) -> Result<bool, ProtocolError> {
        // refuse genesis blocks
        if header.content.slot.period == 0 || header.content.parents.is_empty() {
            return Err(ProtocolError::InvalidBlock("block is genesis".to_string()));
        }

        // Check that our node supports the block version
        self.check_network_version_compatibility(header)?;

        let block_id = header.id;

        // check if the header has not been seen before (is_new == true)
        let is_new;
        {
            let mut cache_write = self.cache.write();
            is_new = cache_write.checked_headers.get(&block_id).is_none();
            if !is_new {
                // the header was previously verified

                // mark the sender peer as knowing the block and its parents
                cache_write.insert_peer_known_block(
                    from_peer_id,
                    &[&[block_id], header.content.parents.as_slice()].concat(),
                    true,
                );
// … (truncated)
```

## Recommendation
Forward every newly received header that passes `note_header_from_peer` to consensus immediately, even when the corresponding block ID is already present in `block_wishlist`. Ensure consensus performs its header-only validation at the earliest stage so structurally valid but consensus-invalid headers are rejected before any operation ID requests or full block downloads are triggered.

Also align the block wishlist flow with an explicit validation state machine so a block cannot progress to dependency resolution, operation retrieval, or full block assembly unless its header has already been accepted by consensus. If immediate forwarding is not possible, add an equivalent prefetch gate in the protocol layer that applies the same consensus-critical checks before requesting additional data.

## Why P0 / breaking

peterjah:
…s retrieval on its approval

  * [ ] make tests compile
  * [ ] make tests pass 


Leo-Besancon (commented):
The implementation looks good to me, but as it touches consensus (and impact full block retrieval by adding latency), I'd like @damip to sanity check that this impact is acceptable.

Otherwise we could forward block headers early to consensus and still ask for the full blocks (dismissing them if the checks reveal it was a bad header), but it would be a lot weaker.

damip (dismissed):
Thanks for the work on this — the implementation is careful (the consensus-side unlock via the `None -> Some(header)` wishlist transition is verified correct, and the integration test is good). But after weighing the actual exploitability against the risks this handshake introduces, I think we should **not** do this, and instead document the accepted trade-off with a comment in the code.

## Why the attack is weaker than the issue suggests

The wishlisted-headerless state only exists after consensus **accepted the child header**, which includes the producer-draw check. So the attacker cannot be an arbitrary peer: they must hold rolls and win a real production draw. Moreover, the missing parent IDs are committed in the child header (block id = header hash), so the only party able to serve "structurally valid but consensus-invalid" parents matching those IDs is the child's producer, who crafted them in advance. Each attack instance therefore burns one legitimate production slot of a staker (their child can never become active), and costs the victim at most ~`thread_count` bounded parent downloads. That is a low-yield, high-cost attack — a bounded bandwidth nuisance, not a practical DoS.

## Why the fix is riskier than the bug

1. **Liveness: a dropped command permanently stalls a block.** The unlock depends on a single `register_block_header` call from the retrieval thread. On `main` this is a `try_send` that silently drops when the consensus command channel is saturated (#4982), and the forwarding is one-shot: after the first reception `info.header` is `Some`, so re-received headers no longer re-register, and `update_block_retrieval` will never ask anything for that block again. One lost command = that block stuck in the wishlist until node restart, with no retry path. Channel saturation is exactly the condition an adversary can influence, so the fix for one DoS finding hands out a lever for a worse one (stalled dependency resolution on the live cliques).
2. **A latency tax on the honest path.** Every missing-dependency retrieval now waits a full round-trip through the consensus worker (command -> header check -> `block_db_changed` -> wishlist delta -> retrieval thread) before any content is asked. This couples retrieval throughput to consensus-worker responsiveness, and bites hardest during catch-up after a desync — precisely when the wishlist mechanism matters most and the consensus worker is busiest.
3. **Complexity where bugs are expensive.** This adds state-machine invariants to the retrieval path, where a mistake means stalled sync or banning honest peers, to mitigate a bounded nuisance.

## Proposal

Close this PR and #5033 as "accepted risk, documented", and add a comment in `retrieval.rs` in the branch where a header is received for a wishlisted block that had none (`note_header_from_peer` caller, the `info.header.is_none()` case). Suggested text:

```rust
// SECURITY NOTE (see issue #5033): this header has only passed the protocol-level
// structural checks (signature, endorsements, format), not the consensus-level ones
// (producer draw, staleness, parent compatibility). We deliberately proceed to
// retrieve the block contents without waiting for consensus to validate the header:
//
// - Blocks only enter the wishlist headerless as missing dependencies of a header
//   that consensus already accepted, which includes the producer-draw check. The
//   parent IDs are committed in that accepted child header (block id = header hash),
//   so only the child's producer — a staker who won a real draw and burned that slot
//   on a block that can never become active — can make us fetch consensus-invalid
//   contents here, and the waste is bounded by the number of parents and max block size.
//
// - Gating content retrieval on a consensus round-trip would couple retrieval
//   liveness and latency to the consensus worker (busy during catch-up, bounded
//   command channel), which is a worse failure mode than the bounded bandwidth
//   waste it would prevent. Consensus still fully validates the header when the
//   assembled block is submitted via register_block.
```

If we want a cheap hardening on top of the comment, the ungated subset of this PR is safe: forward the header to consensus here *without* blocking retrieval on the answer. Retrieval proceeds in parallel as today, and if consensus rejects the header, the parent drops out of the wishlist and the existing remove-delta already cancels the retrieval mid-flight. A dropped command then costs nothing (we just fall back to today's behaviour). No new flag, no unlock handshake, no stall mode.

peterjah:
done. updated to doc only PR

Leo-Besancon (approved):
LGTM

PR #5222, opened 2026-08-26
[F113] enforce endorsement freshness in the propagation thread so eve…

### LINKED ISSUE #5037: [F113] Missing endorsement freshness validation in propagate_endorsements allows rebroadcast of stale endorsements

---

## Source
AI security-scan **Finding 113** (severity: **medium**). Tracked in `SUBMISSIONS.md` as F113.

**Affected file:** `massa-protocol-worker/src/handlers/endorsement_handler/propagation.rs`

## Summary
`ProtocolController::propagate_endorsements` forwards endorsements for network propagation, but it does not check whether they are too old. In contrast, the retrieval path filters out endorsements older than `max_endorsements_propagation_time` before sending them to propagation. Because of this mismatch, any direct caller, including the public gRPC `send_endorsements` endpoint, can make the node rebroadcast obsolete endorsements. This wastes propagation capacity and can weaken duplicate filtering by filling `checked_endorsements` with stale entries.

## Root cause
The main root cause is inconsistent freshness validation between the retrieval flow and the propagation flow.

- On the retrieval side, endorsements older than `max_endorsements_propagation_time` are explicitly filtered out before they are forwarded for propagation.

- In `ProtocolController::propagate_endorsements`, the propagation thread does *not* apply the same age check. As a result, endorsements can enter propagation even when their inclusion slot is already obsolete.

- This gap affects any code path that calls `ProtocolController::propagate_endorsements` directly. The input explicitly mentions the public gRPC `send_endorsements` endpoint as one such caller.

- Because stale endorsements still reach propagation, they are also stored in `checked_endorsements`, even though they should no longer be eligible for useful rebroadcast.

## Affected code
```rust
fn run(&mut self) {
        let mut next_message = None;
        loop {
            // get the next message to process
            let msg = match next_message.take() {
                Some(msg) => msg,
                None => match self.receiver.recv() {
                    Ok(msg) => msg,
                    Err(_) => {
                        info!("Stop endorsement propagation thread");
                        return;
                    }
                },
            };

            match msg {
                // endorsements to propagate
                EndorsementHandlerPropagationCommand::PropagateEndorsements(mut endorsements) => {
                    // also drain any remaining propagation messages that might have accumulated
                    while let Ok(msg) = self.receiver.try_recv() {
                        match msg {
                            // we got more endorsements to propagate: extend the buffer
                            EndorsementHandlerPropagationCommand::PropagateEndorsements(
                                new_endorsements,
                            ) => {
                                endorsements.extend(new_endorsements);
                            }
                            // we grabbed a message that is not a propagation message, mark it for processing
                            other_msg => {
                                next_message = Some(other_msg);
                                break;
                            }
                        }
                    }
                    // propagate the endorsements
                    self.propagate_endorsements(endorsements);
                }
                // stop the handler
                EndorsementHandlerPropagationCommand::Stop => {
                    info!("Stop endorsement propagation thread");
                    return;
                }
            }
        }
    }
```

## Recommendation
Enforce endorsement freshness validation inside `ProtocolController::propagate_endorsements` so that every caller is subject to the same `max_endorsements_propagation_time` policy, including the public gRPC `send_endorsements` path. Reject or drop endorsements whose inclusion slot is older than the allowed propagation window before they enter the propagation pipeline or are inserted into `checked_endorsements`. To prevent similar inconsistencies, centralize this validation in a single shared check and apply it uniformly across both retrieval and direct propagation flows.

## Why P0 / breaking

peterjah:
…ry caller is subject to max_endorsements_propagation_time

  * [ ] make tests compile
  * [ ] make tests pass 

### DIFF
--- a/massa-protocol-worker/src/handlers/endorsement_handler/mod.rs
+++ b/massa-protocol-worker/src/handlers/endorsement_handler/mod.rs
@@ -2,10 +2,12 @@ use std::thread::JoinHandle;
 
 use massa_channel::{receiver::MassaReceiver, sender::MassaSender};
 use massa_metrics::MassaMetrics;
+use massa_models::{endorsement::SecureShareEndorsement, timeslots::get_block_slot_timestamp};
 use massa_pool_exports::PoolController;
 use massa_pos_exports::SelectorController;
 use massa_protocol_exports::ProtocolConfig;
 use massa_storage::Storage;
+use massa_time::MassaTime;
 
 use crate::wrap_network::ActiveConnectionsTrait;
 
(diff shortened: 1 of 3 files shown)


Leo-Besancon (dismissed):
LGTM

Leo-Besancon (approved):

PR #5223, opened 2026-08-26
[F111] only mark a peer as knowing a block on request-correlated or v…

### LINKED ISSUE #5035: [F111] Unrequested BlockInfoReply::Operations Can Poison Peer Block-Knowledge State

---

## Source
AI security-scan **Finding 111** (severity: **medium**). Tracked in `SUBMISSIONS.md` as F111.

## Summary
The code handling `BlockInfoReply::Operations` accepts replies even when the node is *not* waiting for full operations for that `block_id`. In that path, it still marks `from_peer_id` as knowing the block by calling `insert_peer_known_block(..., true)`. This can let a malicious peer claim knowledge of arbitrary known block IDs without proving it has valid block data. As a result, later retrieval logic in `update_block_retrieval` may prefer the attacker and waste requests and timeout windows before trying honest peers.

## Root cause
The issue is in the handling of `BlockInfoReply::Operations`. When the node is not currently waiting for full operations for a given `block_id`, the code goes into the `else` branch instead of rejecting or ignoring the message.

In that `else` branch, the code unconditionally calls `insert_peer_known_block(..., true)` for `from_peer_id`. This happens even when there was *no outstanding request* for that `block_id`.

The same logic also marks the peer as knowing the block even if the reply payload is *empty* or *unrelated to the real block contents*. Because of this, the peer knowledge state is updated based only on receipt of the message, not on a valid pending request or verified contents.

## Affected code
```rust
fn on_block_full_operations_received(
        &mut self,
        from_peer_id: PeerId,
        block_id: BlockId,
    ) {
        debug!(
            "received full operations for block {} from {}",
            block_id, &from_peer_id
        );

        // Ensure that we were looking for that data.
        let wishlist_info = if let Some(info) = self
            .block_wishlist
            .get_mut(&block_id)
            .filter(|i| i.header.is_some() && i.operation_ids.is_some())
        {
            info
        } else {
            // we were not looking for this data
            debug!(
                "Peer id {} sent us full operations for block id {} but we were not looking for it",
                from_peer_id, block_id
            );
            // still mark the sender as knowing the block and operations
            self.cache
                .write()
                .insert_peer_known_block(&from_peer_id, &[block_id], true);
            self.operation_cache.write().insert_peer_known_ops(
                &from_peer_id,
                &operations
                    .into_iter()
                    .map(|op| op.id.prefix())
                    .collect::<Vec<_>>(),
            );
            return;
        };

        // Move the ops into a hashmap
        let mut operations: PreHashMap<OperationId, SecureShareOperation> =
            operations.into_iter().map(|op| (op.id, op)).collect();

        // Make a set of all the block ops for fast lookup and deduplication
        let block_ops_set = wishlist_info
            .operation_ids
// … (truncated)
```

## Recommendation
Only update peer block-knowledge state after receiving a `BlockInfoReply::Operations` that matches an outstanding request for the same `block_id` and passes all expected validation checks. If no full-operations request is pending, ignore the reply and do not call `insert_peer_known_block(..., true)`. Likewise, do not mark a peer as knowing a block based on empty, malformed, or otherwise unverifiable operation data. To further reduce poisoning risk, ensure peer selection for later retrieval relies on request-response correlation and validated data rather than unsolicited replies alone.

## Why P0 / breaking

peterjah:
…alidated block info replies

  * [ ] make tests compile
  * [ ] make tests pass 

### DIFF
--- a/massa-protocol-worker/src/handlers/block_handler/retrieval.rs
+++ b/massa-protocol-worker/src/handlers/block_handler/retrieval.rs
@@ -692,6 +692,19 @@ impl RetrievalThread {
         }
     }
 
+    /// Whether we currently have an outstanding request for that block's data towards that peer.
+    ///
+    /// Used to correlate an incoming `BlockInfoReply` with a request we actually made:
+    /// an unsolicited reply proves nothing about what the sender holds, so it must not be
+    /// used to mark the sender as knowing the block. Otherwise any peer could claim knowledge
+    /// of arbitrary block ids and get prioritized by `update_block_retrieval`, wasting an
+    /// ask timeout window before honest peers are tried.
+    fn has_pending_block_ask(&self, peer_id: &PeerId, block_id: &BlockId) -> bool {
+        self.asked_blocks
+            .get(peer_id)
+            .is_some_and(|asked| asked.contains_key(block_id))
+    }
+
     /// Mark a block as invalid
     fn mark_block_as_invalid(&mut self, block_id: &BlockId) {
         // stop retrieving the block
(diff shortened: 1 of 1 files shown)


Leo-Besancon (approved):
LGTM

PR #5225, opened 2026-08-26
5006 f53 trigger based async messages can be armed outside their validity window and miss execution on validity end

### LINKED ISSUE #5006: [F53] Trigger-Based Async Messages Can Be Armed Outside Their Validity Window and Miss Execution on `validity_end`

---

## Source
AI security-scan **Finding 53** (severity: **medium**). Tracked in `SUBMISSIONS.md` as F53.

**Affected file:** `massa-execution-worker/src/speculative_async_pool.rs`

## Summary
In `massa-main/massa-execution-worker/src/speculative_async_pool.rs`, trigger-based async messages are activated at the wrong point relative to their validity window. `take_batch_to_execute(...)` only selects a message when it is in its validity range and `message.can_be_executed` is already `true`, but `settle_slot(...)` sets `msg.can_be_executed = true` later based only on `is_triggered(filter, ledger_changes)`. This mismatch lets a trigger that happens *before* `validity_start` permanently arm the message, and it also prevents a trigger that first happens on `validity_end` from ever being executed. The result is incorrect async message behavior, including unintended execution and denial of execution.

## Root cause
The main issue is in `massa-main/massa-execution-worker/src/speculative_async_pool.rs`, where trigger activation is decoupled from the validity checks used for execution.

- In `take_batch_to_execute(...)`, a message is selected only if all of the following are true:
  - `available_gas >= corrected_max_gas`
  - `Self::is_message_ready_to_execute(&slot, &message.validity_start, &message.validity_end)`
  - `message.can_be_executed`

  This means execution for the current slot requires `message.can_be_executed` to already be set before selection happens.

- In `settle_slot(...)`, trigger-based activation happens only after execution selection for that slot. The code activates a message with:
  - `if let Some(filter) = &msg.trigger {`
  - `if is_triggered(filter, ledger_changes) {`
  - `msg.can_be_executed = true;`
  - `self.pool_changes.push_activate(*id);`
  - `}`

  Because this runs later, a trigger that first becomes true during `validity_end` is activated too late for `take_batch_to_execute(validity_end, ...)` to pick it.

- `settle_slot(...)` does not verify that the current slot is within `[validity_start, validity_end]` before setting `msg.can_be_executed = true`. It only checks `is_triggered(filter, ledger_changes)`. As a result, a trigger that occurs before `msg.validity_start` can still set `msg.can_be_executed = true`.

- The activation is persisted through `self.pool_changes.push_activate(*id)`. This turns the early activation into a latched state, so a pre-window trigger can remain effective until the message later enters its validity range.

- This behavior conflicts with the documented interval semantics in `settle_slot(...)`, where the comment states that `validity_end` is included in the message validity interval, while the actual sequencing prevents firs …

## Affected code
```rust
/// Settle a slot.
    /// Consume newly emitted messages into `self.async_pool`, recording changes into `self.settled_changes`.
    ///
    /// # Arguments
    /// * slot: slot that is being settled
    /// * ledger_changes: ledger changes for that slot, used to see if we can activate some messages
    ///
    /// # Returns
    /// the list of deleted `(message_id, message)`, used for reimbursement
    pub fn settle_slot(
        &mut self,
        slot: &Slot,
        ledger_changes: &LedgerChanges,
    ) -> Vec<(AsyncMessageId, AsyncMessage)> {
        // Update eliminated_msgs: remove messages that should be removed
        // Filter out all messages for which the validity end is expired.
        // Note: that the validity_end bound is included in the validity interval of the message.

        let mut eliminated_msgs = Vec::new();

        self.message_cache.retain(|id, msg| {
            if Self::is_message_expired(slot, &msg.validity_end) {
                eliminated_msgs.push((*id, msg.clone()));
                false
            } else {
                true
            }
        });

        let mut eliminated_new_messages = Vec::new();
        self.pool_changes.0.retain(|k, v| match v {
            SetUpdateOrDelete::Set(message) => {
                if Self::is_message_expired(slot, &message.validity_end) {
                    eliminated_new_messages.push((*k, v.clone()));
                    false
                } else {
                    true
                }
            }
            SetUpdateOrDelete::Update(_v) => true,
            SetUpdateOrDelete::Delete => true,
        });

        eliminated_msgs.extend(eliminated_new_messages.iter().filter_map(|(k, v)| match v {
            SetUpdateOrDelete::Set(v) => Some((*k, v.clone())),
// … (truncated)
```

## Recommendation
Ensure trigger activation is evaluated within the message validity window and before execution selection for the same slot. Specifically, only set `msg.can_be_executed = true` when the current slot is within `[validity_start, validity_end]`, and prevent triggers observed before `validity_start` from latching into a persistent executable state. In addition, align the ordering between `settle_slot(...)` and `take_batch_to_execute(...)` so that a trigger first observed on `validity_end` can still be executed during that slot, consistent with the documented inclusive validity interval. If reordering is not feasible, update the activation logic so eligibility for execution is derived atomically from both trigger satisfaction and slot validity at selection time, rather than from a previously persisted activation flag alone.

## Why P0 / breaking

peterjah:
- A trigger first observed at validity_end still never executes. Within slot S, take_async_batch(S) (execution.rs:1906) runs before settle_slot(S) (execution.rs:1945), so activation from slot S's ledger changes is only usable from S+1; a message triggered at validity_end is armed and then eliminated at the next settle_slot. The coins are reimbursed via cancel_async_message (context.rs:1051), so this is a semantics gap, not a fund loss. Fixing it means evaluating the trigger at selection time against the partial speculative-ledger changes accumulated so far in the slot, which makes trigger firing sensitive to intra-slot ordering — a larger execution-semantics change deserving its own MIP and review.

### DIFF
--- a/massa-execution-worker/src/speculative_async_pool.rs
+++ b/massa-execution-worker/src/speculative_async_pool.rs
@@ -211,6 +211,20 @@ impl SpeculativeAsyncPool {
         }
 
         // Activate the messages that can be activated (triggered)
+        //
+        // Note: arming is intentionally not restricted to the message validity interval. A trigger
+        //       observed before `validity_start` arms the message for good, and it then executes at
+        //       the first slot of its validity interval that has gas available. This is wanted: a
+        //       message is armed by an event that happened, and consistently with the fact that an
+        //       armed message can never be disarmed, that event is not forgotten just because it
+        //       came early. The validity interval bounds execution, not observation of the trigger.
+        //
+        // Note: activation happens here, i.e. after `take_batch_to_execute` already selected the
+        //       messages to execute at this slot, so a message armed at slot S is executable from
+        //       S+1 on. A trigger first observed at `validity_end` therefore never executes: the
+        //       message is armed and then expires at the next slot, its coins being reimbursed by
+        //       `cancel_async_message`. This cannot be avoided, since the ledger writes that arm
+        //       the message are only known once the slot has been executed.
         for (id, msg) in self.message_cache.iter_mut() {
             if let Some(filter) = &msg.trigger {
                 if is_triggered(filter, ledger_changes) {
@@ -268,7 +282,6 @@ fn is_triggered(filter: &AsyncMessageTrigger, ledger_changes: &LedgerChanges) ->
 #[cfg(test)]
 mod tests {
     use super::*;
-
     // Test if is_message_expired & is_message_ready_to_execute are consistent
     #[test]
     fn test_validity() {
--- a/massa-models/src/async_msg.rs
+++ b/massa-models/src/async_msg.rs
@@ -149,7 +149,7 @@ pub struct AsyncMessage {
     /// Slot at which the message starts being valid (bound included in the validity range)
     pub validity_start: Slot,
 
-    /// Slot at which the message stops being valid (bound not included in the validity range)
+    /// Slot at which the message stops being valid (bound included in the validity range)
     pub validity_end: Slot,
 
     /// Raw payload parameters to call the function with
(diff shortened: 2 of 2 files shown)


Leo-Besancon (commented):
Two points/questions:
- I don't think this PR fixes the issue of a trigger done exactly at validity_end.
- I'm not sure the fact that ASCs could be armed outside their validity range wasn't expected behaviour. Depending on the use case, the legacy behaviour could be wanted, no? It would also align with the fact that once a message is triggered, it cannot be untriggered. What did you have in mind @damip?

damip:
I agree with both statements @Leo-Besancon . Let's leave a documenting comment and update the docs to make this explicit. Let's not change it now. The validity_end thing is completely non implementable and a non-issue.

peterjah:
PR is now Docs only

Leo-Besancon (approved):
Thanks, LGTM

PR #5226, opened 2026-08-26
[F19] apply bounded backpressure instead of dropping consensus comman…

### LINKED ISSUE #4982: [F19] Silent Dropping of Consensus Commands via `try_send` Channel Saturation

peterjah:
…ds when the command channel is saturated

  * [ ] make tests compile
  * [ ] make tests pass 

### DIFF
--- a/massa-consensus-worker/src/controller.rs
+++ b/massa-consensus-worker/src/controller.rs
@@ -31,6 +31,24 @@ use crate::{commands::ConsensusCommand, state::ConsensusState};
 /// - still be able to read the current state of the graph as processed so far (for this we need a shared state)
 ///
 /// Note that sending commands and reading the state is done from different, mutually-asynchronous tasks and they can have data that are not sync yet.
+///
+/// The command channel is bounded and the sends below use `try_send`: when it is full, the command
+/// is dropped rather than waited on. This is deliberate, and waiting instead would be worse:
+///
+/// - the consensus worker blocks on the protocol block-retrieval channel (`send_wishlist_delta`)
+///   while processing these very commands, and the block-retrieval thread is what both drains that
+///   channel and calls `register_block_header`, so waiting here without a deadline deadlocks the
+///   two threads against each other;
+/// - `register_block` is called by the block factory, which does not propagate the block we just
+///   produced until the command has gone through, so waiting would delay the announcement of our
+///   own block precisely when the node is already struggling.
+///
+/// Dropping is acceptable because consensus is already its own retry mechanism: a block whose
+/// dependencies are missing is held in `WaitingForDependencies` and re-requested from peers through
+/// the wishlist. A header that was merely announced to us is therefore fetched again if it turns
+/// out to matter, so layering another retry queue on top of the graph would duplicate that
+/// mechanism rather than add robustness. Channel saturation is visible through the
+/// `consensus_command_channel_actual_size` metric.
 #[derive(Clone)]
 pub struct ConsensusControllerImpl {
     command_sender: MassaSender<ConsensusCommand>,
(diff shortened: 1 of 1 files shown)


Leo-Besancon (commented):
The only worry I have is the potential impact on the block factory: it seems that we do not propagate our created block until the register_block command returns. Here, we block it for 500ms (even if it's during a specific overload event).

We also block protocol retrieval, but I'm more inclined on agreeing with this as protocol would feed an overloaded consensus in that case, so blocking shouldn't do too much damage IMO.

WDYT @damip?

damip:
Honestly I agree, we should NOT ship this PR but write a clean response: the queue is ALREADY a history system, adding another retry queue on top does not help. In case of congestion, dropping the header that was announced to us is ACCEPTABLE and the system is robust to it. Worth a comment in the code

peterjah:
@Leo-Besancon  PR updated to doc only

Leo-Besancon (approved):
LGTM

PR #5227, opened 2026-08-26
[F88] return the sanitized datastore key query count so omitted API l…

### LINKED ISSUE #5024: [F88] Unreturned Cleaned Count Parameter Allows Bypassing Query Limits and DoS

---

## Source
AI security-scan **Finding 88** (severity: **medium**). Tracked in `SUBMISSIONS.md` as F88.

**Affected file:** `massa-models/src/datastore.rs`

## Summary
`cleanup_datastore_key_range_query` in `massa-models/src/datastore.rs` locally normalizes an omitted `count` with `count.or(max_datastore_query_config)`, but it does not return that sanitized value to its callers. As a result, upstream JSON-RPC and gRPC paths can still forward the original `value.count` or `value.limit` as `None` into `ExecutionQueryRequestItem`, leaving the datastore key-range query unbounded. The confirmed issue is therefore a real bypass of the configured `max_datastore_keys_query` limit. The original description overstated scope and impact: the query is limited to datastore keys for a single address, not the entire node datastore, and it collects keys rather than loading all datastore values.

## Root cause
The root cause is that `cleanup_datastore_key_range_query(prefix, start_bound, end_bound, count, max_datastore_key_length, max_datastore_query_config)` computes a cleaned local `count` with `count.or(max_datastore_query_config)` and validates it, but its return type is only `Result<(Vec<u8>, Bound<Vec<u8>>, Bound<Vec<u8>>), ModelsError>`. The sanitized `count` is therefore discarded instead of being propagated back to the caller.

Because the JSON-RPC and gRPC callers continue building `ExecutionQueryRequestItem` with the original `value.count` or `value.limit`, an omitted limit remains `None` downstream. In `massa-ledger-worker/src/ledger_db.rs` and `massa-execution-worker/src/datastore_scan.rs`, early termination only occurs when `count` is `Some`, so the configured maximum is not enforced for omitted counts.

The ruled-out portion of the original root-cause theory is any implication that this bug exposes an unrestricted scan of all datastore contents across the node. The validated issue is specifically the loss of the cleaned per-query limit, not a failure of address scoping.

## Affected code
```rust
pub fn cleanup_datastore_key_range_query(
    prefix: &[u8],
    start_bound: Bound<Vec<u8>>,
    end_bound: Bound<Vec<u8>>,
    count: Option<u32>,
    max_datastore_key_length: u8,
    max_datastore_query_config: Option<u32>,
) -> Result<(Vec<u8>, Bound<Vec<u8>>, Bound<Vec<u8>>), ModelsError> {
    // check item count
    let count = count.or(max_datastore_query_config);
    if let (Some(cnt), Some(max_cnt)) = (count.as_ref(), max_datastore_query_config.as_ref()) {
        if cnt > max_cnt {
            return Err(ModelsError::ErrorRaised(format!(
                "max item count in datastore key query is {} but {} items were queried",
                max_cnt, cnt
            )));
        }
    }

    // check prefix length
    let prefix = if prefix.len() > max_datastore_key_length as usize {
        // prefix is too long: it won't match anything. Adjust bounds to reflect this.
        return Ok((
            Vec::new(),
            std::ops::Bound::Excluded(Vec::new()),
            std::ops::Bound::Excluded(Vec::new()),
        ));
    } else {
        prefix.to_vec()
    };

    // If the key is longer than the max possible length
    // it will be by definition excluded
    // and since its truncation is before, it will also be excluded
    let start_bound = match start_bound {
        std::ops::Bound::Unbounded => std::ops::Bound::Unbounded,
        std::ops::Bound::Excluded(mut k) => {
            k.truncate(max_datastore_key_length as usize);
            Bound::Excluded(k)
        }
        std::ops::Bound::Included(mut k) => {
            if k.len() > max_datastore_key_length as usize {
                k.truncate(max_datastore_key_length as usize);
                Bound::Excluded(k)
            } else {
// … (truncated)
```

## Recommendation
Ensure the effective datastore query limit is propagated back to the caller and used consistently when constructing `ExecutionQueryRequestItem`. If `count` or `limit` is omitted, replace it with the configured `max_datastore_keys_query` value before any downstream processing, rather than only normalizing it locally inside `cleanup_datastore_key_range_query`. This sanitized limit should then be enforced uniformly across both JSON-RPC and gRPC paths and at the ledger or execution scan layer to prevent unbounded key enumeration for a single address. Additionally, add validation and regression tests covering omitted limits to confirm the configured maximum is always applied.

## Why P0 / breaking

peterjah:
…imits stay bounded

 Breaking in effect:
  - A client that omits count/limit previously received every key for the address; it now receives at most max_datastore_keys_query (500 in base_config/config.toml:46 and :127, for JSON-RPC and gRPC
  respectively).
  - GetAddressDatastoreKeysResponse has only {address, is_final, keys} — no truncation flag. So such a client can't distinguish "that's all the keys" from "that's the first 500". It silently gets a short
  answer. That's the real sharp edge: a client with >500 keys under the queried prefix would need to switch to paginating via start_key, and nothing in the response tells it to.
  
  Because we build the getDatastoreKeys endpoint recently, I am pretty confident that pagination is well supported by clients

### DIFF
--- a/massa-api/src/public.rs
+++ b/massa-api/src/public.rs
@@ -1636,7 +1636,7 @@ fn get_address_datastore_keys_to_state_query_item(
         (Some(k), false) => std::ops::Bound::Excluded(k),
     };
 
-    let (prefix, start_key, end_key) = cleanup_datastore_key_range_query(
+    let (prefix, start_key, end_key, count) = cleanup_datastore_key_range_query(
         &value.prefix,
         start_key,
         end_key,
@@ -1657,7 +1657,7 @@ fn get_address_datastore_keys_to_state_query_item(
             prefix,
             start_key,
             end_key,
-            count: value.count,
+            count,
         })
     } else {
         Ok(ExecutionQueryRequestItem::AddressDatastoreKeysCandidate {
@@ -1665,7 +1665,7 @@ fn get_address_datastore_keys_to_state_query_item(
             prefix,
             start_key,
             end_key,
-            count: value.count,
+            count,
         })
     }
 }
--- a/massa-execution-exports/src/mapping_grpc.rs
+++ b/massa-execution-exports/src/mapping_grpc.rs
@@ -71,7 +71,7 @@ pub fn to_querystate_filter(
                     (Some(k), false) => std::ops::Bound::Excluded(k),
                 };
 
-                let (prefix, start_key, end_key) = cleanup_datastore_key_range_query(
+                let (prefix, start_key, end_key, count) = cleanup_datastore_key_range_query(
                     &value.prefix,
                     start_key,
                     end_key,
@@ -85,7 +85,7 @@ pub fn to_querystate_filter(
                     prefix,
                     start_key,
                     end_key,
-                    count: value.limit,
+                    count,
                 })
             }
             exec::RequestItem::AddressDatastoreKeysFinal(value) => {
@@ -102,7 +102,7 @@ pub fn to_querystate_filter(
                     (Some(k), false) => std::ops::Bound::Excluded(k),
                 };
 
-                let (prefix, start_key, end_key) = cleanup_datastore_key_range_query(
+                let (prefix, start_key, end_key, count) = cleanup_datastore_key_range_query(
                     &value.prefix,
                     start_key,
                     end_key,
@@ -116,7 +116,7 @@ pub fn to_querystate_filter(
                     prefix,
                     start_key,
                     end_key,
-                    count: value.limit,
+                    count,
                 })
             }
             exec::RequestItem::AddressDatastoreValueCandidate(value) => {
--- a/massa-execution-worker/src/context.rs
+++ b/massa-execution-worker/src/context.rs
@@ -660,7 +660,7 @@ impl ExecutionContext {
         let max_datastore_query = None;
 
         // cleanup bounds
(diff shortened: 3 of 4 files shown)


peterjah:
It appears massa-web3 Grpc does not implements the pagination.. but i dont think its used widely.
I will fix on massa-web3

Leo-Besancon:
> * A client that omits count/limit previously received every key for the address; it now receives at most max_datastore_keys_query (500 in base_config/config.toml:46 and :127, for JSON-RPC and gRPC
>       respectively).

For this point, isn't it the same as https://github.com/massalabs/massa/pull/5189? What's the difference?

peterjah:
> > * A client that omits count/limit previously received every key for the address; it now receives at most max_datastore_keys_query (500 in base_config/config.toml:46 and :127, for JSON-RPC and gRPC
> >   respectively).
> 
> For this point, isn't it the same as #5189? What's the difference?

Exact same thing it seems 😸 

The thing i didnt pointed last time is that its conerns the newly created getDatastoreKey, that was build especially to replace the old get_addresses that was not paginated.
Pagination has been implemented day 1 on massa web3 (at least for the jsonRPC side...)
So it increase my level of confidence of pushing it

peterjah:
As this PR seems a duplicate of https://github.com/massalabs/massa/pull/5189 i think i will report my comments there and close this one

PR #5230, opened 2026-08-27
[F54] feat: reorder ASCs priority based on real gas usage

### LINKED ISSUE #5007: [F54] Async Message Priority Misprices Gas by Omitting Fixed Per-Message Surcharge

---

## Source
AI security-scan **Finding 54** (severity: **medium**). Tracked in `SUBMISSIONS.md` as F54.

**Affected file:** `massa-execution-worker/src/speculative_async_pool.rs`

## Summary
The function `take_batch_to_execute` selects async messages using an order derived from `AsyncMessage::compute_id()`. That order ranks messages by `fee / max(max_gas, 1)`, but execution cost is charged as `message.max_gas + async_msg_cst_gas_cost`. Because the fixed cost `async_msg_cst_gas_cost` is not included in priority, low-`max_gas` messages can look cheaper than they really are. This can let an attacker push ahead of better-priced honest messages and reduce or monopolize async execution throughput.

## Root cause
In `take_batch_to_execute`, each selected message consumes slot budget based on `message.max_gas + async_msg_cst_gas_cost`. This means every executed message pays both its variable gas cost and the fixed cost `async_msg_cst_gas_cost`.

However, the pool order used by `take_batch_to_execute` comes from `AsyncMessage::compute_id()`, which ranks messages only by `fee / max(max_gas, 1)`. The fixed execution cost `async_msg_cst_gas_cost` is not part of this ranking.

Because the selection order and the actual budget charge use different cost models, messages with small `max_gas` can appear highly priced in `AsyncMessage::compute_id()` even though they still consume the fixed `async_msg_cst_gas_cost` when executed in `take_batch_to_execute`.

## Affected code
```rust
pub fn take_batch_to_execute(
        &mut self,
        slot: Slot,
        max_gas: u64,
        async_msg_cst_gas_cost: u64,
    ) -> Vec<(AsyncMessageId, AsyncMessage)> {
        let mut available_gas = max_gas;

        // Choose which messages to take based on the message_cache
        // (all messages are considered: finals, in active_history and in speculative)

        let mut wanted_ids = Vec::new();
        for (message_id, message) in self.message_cache.iter() {
            let corrected_max_gas = message.max_gas.saturating_add(async_msg_cst_gas_cost);
            // Note: SecureShareOperation.get_validity_range(...) returns RangeInclusive
            //       so to be consistent here, use >= & <= checks
            if available_gas >= corrected_max_gas
                && Self::is_message_ready_to_execute(
                    &slot,
                    &message.validity_start,
                    &message.validity_end,
                )
                && message.can_be_executed
            {
                available_gas -= corrected_max_gas;

                wanted_ids.push(*message_id);
            }
        }

        // Remove the messages_info of the taken messages, and push their deletion in the pool changes
        let mut taken_msgs = Vec::with_capacity(wanted_ids.len());
        for msg_id in &wanted_ids {
            taken_msgs.push((
                *msg_id,
                self.message_cache.remove(msg_id).unwrap(), // won't panic, items were listed above
            ));
        }
        self.delete_messages(wanted_ids);

        taken_msgs
    }
```

## Recommendation
Align message prioritization with the actual execution cost charged in `take_batch_to_execute`. Specifically, rank async messages using an effective gas cost that includes both `message.max_gas` and `async_msg_cst_gas_cost`, so priority reflects the true fee paid per unit of consumed batch budget. Ensure the same cost model is used consistently for ordering and execution accounting to prevent low-`max_gas` messages from receiving artificially favorable priority. Revalidate batch selection behavior and edge cases after the change to confirm that ordering remains fair, deterministic, and resistant to throughput-griefing.

## Why P0 / breaking

Leo-Besancon:
The current PR reorders the ASC priority when `take_async_batch` is called, without changing the AsyncMessageId type to ensure smooth transition.
I see three options:
1. Either we consider that this PR is too risky to merge for the benefits (I don't expect the orders to change too much in practice?), and we just add comments in code
2. Or we merge this implementation (it adds a sorting step but it should not impact performances too much)
3. Or we update AsyncMessageId to implement a new `Ord` implem, keeping the BTreeMap sort. However, we need to add, at version activation, a migration step to update all legacy ids to the new ones, which may be more prone to errors.

What do you think is best @damip? Personnaly I would lean towards 1.

Going with option 2 as discussed below.

As outlined below the impact is bounded, which is why this is not urgent: send_message REJECTS max_gas < gas_costs.max_instance_cost (2.1M gas), so the skew (max_gas + cst) / max_gas is at most 1 + 1M/2.1M = ~1.48x, and emitting costs 40M gas of ABI per message (~107 per block max, by taking the WHOLE block). So a 48% priority advantage at most, for someone paying full blocks to get it. Still a mispricing and the fix is cheap, so let's align selection with what we actually charge.

### DIFF
--- a/massa-async-pool/src/lib.rs
+++ b/massa-async-pool/src/lib.rs
@@ -42,6 +42,9 @@
 //! ## How is the `AsyncPool` handled
 //! ```md
 //! * In the AsyncPool, Messages are kept sorted by `priority = AsyncMessageId(rev(Ratio(msg.fee, max(msg.max_gas,1))), rev(msg.slot), rev(msg.emission_index))`
+//!   From execution component version 2 (MIP-0002), batch selection and overflow eviction
+//!   re-rank with `fee / max(max_gas + async_msg_cst_gas_cost, 1)` so priority matches the
+//!   gas actually charged when taking a batch.
 //!
 //! * when an AsyncMessage is added to the AsyncPool:
 //!   * if the AsyncPool length has exceeded config.max_async_pool_length:
--- a/massa-execution-worker/src/context.rs
+++ b/massa-execution-worker/src/context.rs
@@ -8,7 +8,9 @@
 //! and does not write anything persistent to the consensus state.
 
 use crate::active_history::HistorySearchResult;
-use crate::speculative_async_pool::SpeculativeAsyncPool;
+use crate::speculative_async_pool::{
+    SpeculativeAsyncPool, ASYNC_MSG_EFFECTIVE_GAS_PRIORITY_EXEC_VERSION,
+};
 use crate::speculative_deferred_calls::{
     SpeculativeDeferredCallRegistry, DEFERRED_CALL_INDEX_FIX_EXEC_VERSION,
 };
@@ -482,6 +484,9 @@ impl ExecutionContext {
             self.slot,
             max_gas,
             async_msg_cst_gas_cost,
+            self.is_execution_component_version_at_least(
+                ASYNC_MSG_EFFECTIVE_GAS_PRIORITY_EXEC_VERSION,
+            ),
         )
     }
 
@@ -1066,9 +1071,14 @@ impl ExecutionContext {
         let deferred_credits_transfers = self.execute_deferred_credits(&slot);
 
         // settle emitted async messages and reimburse the senders of deleted messages
-        let deleted_messages = self
-            .speculative_async_pool
-            .settle_slot(&slot, &self.speculative_ledger.added_changes);
+        let deleted_messages = self.speculative_async_pool.settle_slot(
+            &slot,
+            &self.speculative_ledger.added_changes,
+            self.config.async_msg_cst_gas_cost,
+            self.is_execution_component_version_at_least(
+                ASYNC_MSG_EFFECTIVE_GAS_PRIORITY_EXEC_VERSION,
+            ),
+        );
 
         let mut cancel_async_message_transfers = vec![];
         for (_msg_id, msg) in deleted_messages {
--- a/massa-execution-worker/src/speculative_async_pool.rs
+++ b/massa-execution-worker/src/speculative_async_pool.rs
@@ -417,13 +417,21 @@ mod tests {
 
     // Low-max_gas messages look better under fee/max_gas, but worse once the fixed
     // async_msg_cst_gas_cost is included — the MIP-0002 ranking must prefer the latter.
+    //
+    // Values are ABI-reachable: `send_message` rejects `max_gas < max_instance_cost` (2.1M).
+    // Old ratio 25_000/2.1M ≈ 0.0119 vs 1_000_000/100M = 0.0100 (attacker first);
+    // new ratio 25_000/3.1M ≈ 0.00806 vs 1_000_000/101M ≈ 0.0099 (honest first).
+    const MAX_INSTANCE_COST: u64 = 2_100_000;
+    const ATTACKER_MAX_GAS: u64 = MAX_INSTANCE_COST;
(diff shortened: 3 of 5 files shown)


damip:
I went through it carefully. The code is fine: the same cst is threaded to selection and eviction, the sort is stable with the stored id as tie breaker, stored ids are untouched so no migration, and it is gated on the execution version. Let's go with option 2 and merge it. MIP-0002 is happening anyway, and an attacker willing to burn money on griefing does not care about being rational.

For context the impact is bounded, which is why this is not urgent: send_message REJECTS max_gas < gas_costs.max_instance_cost (2.1M gas), so the skew (max_gas + cst) / max_gas is at most 1 + 1M/2.1M = ~1.48x, and emitting costs 40M gas of ABI per message (~107 per block max, by taking the WHOLE block). So a 48% priority advantage at most, for someone paying full blocks to get it. Still a mispricing and the fix is cheap, so let's align selection with what we actually charge.

Two things before merging:

- the tests use max_gas = 1, which cannot happen through the ABI. Please use max_gas >= max_instance_cost so they reflect a reachable case. The flip still happens with realistic values, e.g. honest: 100M gas, fee 1_000_000 / attacker: 2.1M gas, fee 25_000. Old ratio 0.0119 vs 0.0100 (attacker first), new ratio 0.00806 vs 0.0099 (honest first).
- yet another hardcoded 2: 5173, 5159, 5220, 5225 and this one each define their own constant equal to 2 for MIP-0002. Let's have ONE constant for the execution version of MIP-0002 and reference it everywhere. Can be a separate small PR on dev_breaking but before the release.

Also please put the bounded impact explanation above in the PR description and in a short comment on compute_id, so the next reader knows why the ratio differs from the stored id and why the old formula was tolerable. WDYT @Leo-Besancon ?

Leo-Besancon:
> I went through it carefully. The code is fine: the same cst is threaded to selection and eviction, the sort is stable with the stored id as tie breaker, stored ids are untouched so no migration, and it is gated on the execution version. Let's go with option 2 and merge it. MIP-0002 is happening anyway, and an attacker willing to burn money on griefing does not care about being rational.
> 
> For context the impact is bounded, which is why this is not urgent: send_message REJECTS max_gas < gas_costs.max_instance_cost (2.1M gas), so the skew (max_gas + cst) / max_gas is at most 1 + 1M/2.1M = ~1.48x, and emitting costs 40M gas of ABI per message (~107 per block max, by taking the WHOLE block). So a 48% priority advantage at most, for someone paying full blocks to get it. Still a mispricing and the fix is cheap, so let's align selection with what we actually charge.
> 
> Two things before merging:
> 
>     * the tests use max_gas = 1, which cannot happen through the ABI. Please use max_gas >= max_instance_cost so they reflect a reachable case. The flip still happens with realistic values, e.g. honest: 100M gas, fee 1_000_000 / attacker: 2.1M gas, fee 25_000. Old ratio 0.0119 vs 0.0100 (attacker first), new ratio 0.00806 vs 0.0099 (honest first).
> 
>     * yet another hardcoded 2: 5173, 5159, 5220, 5225 and this one each define their own constant equal to 2 for MIP-0002. Let's have ONE constant for the execution version of MIP-0002 and reference it everywhere. Can be a separate small PR on dev_breaking but before the release.
> 
> 
> Also please put the bounded impact explanation above in the PR description and in a short comment on compute_id, so the next reader knows why the ratio differs from the stored id and why the old formula was tolerable. WDYT @Leo-Besancon ?

Thanks for the review! Looks good to me, made the followup PR https://github.com/massalabs/massa/issues/5252, updated the compute_id comment and the tests as suggested, and rebased on top of `dev_breaking` to resolve conflicts

bilboquet (approved):

PR #5233, opened 2026-08-27
[F97] feat: keep speculative ops in pool, but hard skip them

### LINKED ISSUE #5028: [F97] Speculative Execution Marking Causes Permanent Operation Pool Eviction After Reorg

---

## Source
AI security-scan **Finding 97** (severity: **medium**). Tracked in `SUBMISSIONS.md` as F97.

**Affected file:** `massa-pool-worker/src/operation_pool.rs`

## Summary
`OperationPool::get_execution_statuses` computes whether operations look executed, and `OperationPool::prefilter_ops` removes operations from the pool when they have any execution marker. The problem is that an operation marked as executed only in *candidate history* is treated like a positively executed operation and is dropped during pool refresh. If that candidate history is later removed during a rollback, the execution mark disappears, but the operation is not put back into the pool. This can keep a victim transaction out of the pool until manual resubmission and can affect ordering, censorship resistance, and MEV.

## Root cause
The main root cause is in `OperationPool::get_execution_statuses`. It collapses the state `(Some(speculative_status), None)` into a positive execution marker. This treats an operation that was executed only in *candidate history* as executed, even though it is not final.

A second root cause is in `OperationPool::prefilter_ops`. This function drops every operation that has *any* execution marker at all. Because of that behavior, operations marked only by speculative candidate execution are removed from the pool during refresh.

A third root cause is the missing recovery path after rollback. When candidate history is later truncated on a rollback, the speculative execution mark disappears, but in the scoped code there is *no reinsertion path* after `OperationPool::prefilter_ops` has already dropped the operation references.

## Affected code
```rust
fn get_execution_statuses(&self) -> PreHashMap<OperationId, bool> {
    let op_ids: Vec<OperationId> = self.sorted_ops.iter().map(|op_info| op_info.id).collect();
    self.channels
        .execution_controller
        .get_ops_exec_status(&op_ids)
        .into_iter()
        .zip(op_ids)
        .filter_map(
            |((spec_status, final_status), op_id)| match (spec_status, final_status) {
                (Some(_), Some(_)) => Some((op_id, true)),
                (Some(_), None) => Some((op_id, false)),
                _ => None,
            },
        )
        .collect()
}
...
// filter out ops that have been executed in final or candidate slots
// TODO: in the re-execution followup, we should only filter out final-executed ops here (exec_status == Some(true))
if retain {
    retain = !exec_statuses.contains_key(&op_info.id);
}
```

## Recommendation
Distinguish speculative candidate execution from finalized or active blockclique execution when computing operation execution status. `OperationPool::get_execution_statuses` should not treat `(Some(speculative_status), None)` as a definitive execution result, and `OperationPool::prefilter_ops` should avoid evicting operations based solely on candidate-history markers.

Only remove operations from the pool when execution is confirmed by a durable source, such as finalized history or another state that is guaranteed to survive rollback. If speculative execution information must be tracked, keep it separate from pool-eviction logic and use it only for temporary scheduling or deduplication decisions.

Add a rollback-safe recovery path so that operations removed due to non-final execution state are reinserted or revalidated when candidate history is truncated. Ensure pool refresh logic is reorg-aware and can reconstruct eligibility from canonical state after rollback, preventing operations from remaining permanently excluded unless they were actually finalized or otherwise invalid.

## Why P0 / breaking

Leo-Besancon:
  * [ ] make tests compile
  * [ ] make tests pass 

Note: In this version I hard skip speculative ops (and kept the 1/1000° that was there in commented code to drop them in priority if the pool is full). See more details in the discussion below

### DIFF
--- a/massa-pool-worker/src/operation_pool.rs
+++ b/massa-pool-worker/src/operation_pool.rs
@@ -115,7 +115,14 @@ impl OperationPool {
         pos_draws
     }
 
-    /// Returns the list of executed ops with a boolean indicating whether they are executed as final.
+    /// Returns execution markers for ops currently in the pool.
+    ///
+    /// Map value semantics:
+    /// - `true`: executed in final history (durable; safe to evict from the pool)
+    /// - `false`: executed only in speculative/candidate history (can disappear on rollback;
+    ///   must not drive pool eviction)
+    ///
+    /// Ops with no execution record are omitted.
     fn get_execution_statuses(&self) -> PreHashMap<OperationId, bool> {
         let op_ids: Vec<OperationId> = self.sorted_ops.iter().map(|op_info| op_info.id).collect();
         self.channels
@@ -125,7 +132,9 @@ impl OperationPool {
             .zip(op_ids)
             .filter_map(
                 |((spec_status, final_status), op_id)| match (spec_status, final_status) {
-                    (Some(_), Some(_)) => Some((op_id, true)),
+                    // Final execution is durable (execution layer also surfaces it as speculative).
+                    (_, Some(_)) => Some((op_id, true)),
+                    // Candidate-history only: keep as non-final for optional scoring, never as eviction.
                     (Some(_), None) => Some((op_id, false)),
                     _ => None,
                 },
@@ -179,10 +188,11 @@ impl OperationPool {
                 retain = op_info.fee.checked_sub(self.config.minimal_fees).is_some();
             }
 
-            // filter out ops that have been executed in final or candidate slots
-            // TODO: in the re-execution followup, we should only filter out final-executed ops here (exec_status == Some(true))
+            // Filter out ops whose execution is final/durable only.
+            // Speculative/candidate-only markers (exec_status == false) must not evict:
+            // they can vanish on rollback, and there is no reinsertion path after drop.
             if retain {
-                retain = !exec_statuses.contains_key(&op_info.id);
+                retain = exec_statuses.get(&op_info.id) != Some(&true);
             }
 
             // filter out ops that spend more than the sender's balance
@@ -253,7 +263,7 @@ impl OperationPool {
     /// Score the operations
     fn score_operations(
         &self,
-        _exec_statuses: &PreHashMap<OperationId, bool>,
+        exec_statuses: &PreHashMap<OperationId, bool>,
         pos_draws: &BTreeSet<Slot>,
     ) -> PreHashMap<OperationId, f32> {
         let now = MassaTime::now();
@@ -315,23 +325,19 @@ impl OperationPool {
                     0.0
                 };
 
-            /* TODO: re-execution followup
-            // If the op was executed previously, there is still an exponentially decaying chance of its block being cancelled
-            // so that it can be reincluded.
-            // We approximate it with a constant factor for simplicity since we don't have the inclusion slot for now.
-            let reexecution_penalty = 1.0 / 1000.0; // re-execution penalty factor
+            // If the op was executed previously, there is still a chance of its block being
+            // cancelled so that it can be reincluded. Keep it in the pool (for rollback recovery)
+            // but score it far below non-executed ops. Block production also skips currently
+            // executed ops to avoid wasting gas while the mark is live.
+            let reexecution_penalty = 1.0 / 1000.0;
             let reexecution_factor = if exec_statuses.contains_key(&op_info.id) {
-                // executed previously
                 reexecution_penalty
             } else {
-                // not executed previously => score 1
                 1.0
             };
-            */
 
             // compute the score as being the product of all the factors and the fee
-            let score = fee_factor * resource_factor * inclusion_factor;
-            //  * reexecution_factor; // TODO: re-execution followup
+            let score = fee_factor * resource_factor * inclusion_factor * reexecution_factor;
 
             // store the score
             scores.insert(op_info.id, score);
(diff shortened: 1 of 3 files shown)


bilboquet (dismissed):

Leo-Besancon:
@damip can you take a quick look and see if it's consistant with what you wanted for the reexecution followup (see the main PR comment)?

massa-pool-worker/src/operation_pool.rs
```diff
@@ -470,7 +476,13 @@ impl OperationPool {
     /// Searches the available operations, and selects the sub-set of operations that:
     /// - fit inside the block
     /// - is the most profitable for block producer
+    /// - are not currently marked executed (speculative or final)
     pub fn get_block_operations(&self, slot: &Slot) -> (Vec<OperationId>, Storage) {
+        // Live execution marks: skip while present so we do not waste block gas/space.
+        // Speculative marks can disappear on rollback; those ops stay in the pool and
+        // become selectable again once the mark is gone (without needing reinsertion).
+        let exec_statuses = self.get_execution_statuses();
```
damip: This is the blocker. Factory thread, inside the pool read guard, at slot time, and it goes through `execution_state.read()` which waits for the execution worker to finish the slot it is executing. Block production must never wait on execution. Replace with a flag on `OperationInfo` set in `refresh()` and skip on the flag here.

massa-pool-worker/src/operation_pool.rs
```diff
@@ -179,10 +188,11 @@ impl OperationPool {
...
-            // TODO: in the re-execution followup, we should only filter out final-executed ops here (exec_status == Some(true))
+            // Filter out ops whose execution is final/durable only.
+            // Speculative/candidate-only markers (exec_status == false) must not evict:
+            // they can vanish on rollback, and there is no reinsertion path after drop.
             if retain {
-                retain = !exec_statuses.contains_key(&op_info.id);
+                retain = exec_statuses.get(&op_info.id) != Some(&true);
             }
 
             // filter out ops that spend more than the sender's balance
```
damip: Candidate balance already includes the speculative execution, so a marked op that spent most of the sender's balance fails this check and is evicted before any rollback. Skip this check (and `eliminate_balance_overflows`) for ops with a live mark: they are not selectable anyway and their spend is already in the balance.

massa-pool-worker/src/tests/operation_pool_tests.rs
```diff
@@ -63,6 +78,100 @@ fn create_recursive_selector_for_ops(addr: Address) -> MockSelectorController {
...
+            1 => vec![(Some(true), None); ops.len()],
+            2 => vec![(Some(true), Some(true)); ops.len()],
+            _ => vec![(None, None); ops.len()],
+        });
+    execution_controller
+        .expect_get_final_and_candidate_balance()
+        .returning(|addrs| {
+            vec![
+                (
+                    Some(Amount::const_init(1_000_000_000, 0)),
```
damip: 1e9 hides problem 2. Use a balance close to the op's `max_spending` and check the op survives refresh while marked, then is selectable again once the mark is gone.

damip (dismissed):
Good direction, two things to change before merge. Rule for both: block production must be fast and must never block on anything. Dropping or duplicating a few tx is always preferable to slowing down block production.

1. `get_block_operations` must not call execution. It runs in the factory thread, inside the pool read guard, right at slot time. The PR adds `get_execution_statuses()` there: one `get_ops_exec_status` on every pool op id (up to 600k) behind `execution_state.read()`, while the execution worker holds the write lock for the whole of `execute_candidate_slot` / `execute_final_slot` / read-only requests. And `active_history.get_ops_exec_status` is O(pool x non-final slots). `block_opt_channel_timeout` only bounds `try_read_for` on the pool lock, not this. Today the factory path never touches execution (refresh runs on the buffer in the pool worker) and it must stay that way.
   Fix: an `executed: bool` on `OperationInfo`, set in `refresh()` from the statuses already fetched there, and `get_block_operations` just does `continue` on the flag. Zero cost on the production path, travels through `replace_with` for free. Staleness is at most 5 s and the next block of the same thread is 16 s away, so no duplicate in practice. If one slips in, execution skips it. Fine.

2. The balance filters undo the fix in the common case. `get_sender_balances` uses the candidate balance, which already contains the speculative execution. Alice has 10, sends an op spending 9: candidate is 1 after speculative execution, next refresh `max_spending <= 1` fails in `prefilter_ops`, op evicted before any rollback. `eliminate_balance_overflows` does the same when the sender has other pending ops (marked ops score last, so they are cut first). Only ops small relative to what is left survive. The test does not see it because the mock balance is 1e9.
   Fix: skip both balance checks for ops with a live mark (not selectable anyway, their spend is already in the candidate balance). When the mark disappears the balance is back and the next refresh evaluates them normally. Add a test with a realistic balance: op spending most of it, speculative mark, rollback, selectable again.

@Leo-Besancon on your question: hard skip is right, keep it. Scoring down only would still put every speculatively executed op in the next 2-3 blocks of its thread until the including block is final, each copy ignored by execution, zero fee, pure waste for everyone. The 1/1000 factor is only useful as truncation priority when the pool is full, which is what you did.

Not consensus, node-local, no MIP: execution ignores re-included ops (`IncludeOperationError("operation was executed previously")`, errors do not interrupt the slot), block validity is unchanged. Close #5028 on merge.

massa-pool-worker/src/operation_pool.rs
Leo-Besancon: Good catch! Addressed in e93ccdb7d051ec5992892d562b5285f72c1ec729

massa-pool-worker/src/operation_pool.rs
Leo-Besancon: Addressed in c3d8ebdef64154f49b3e35355aa87d7ae19fb0d2

massa-pool-worker/src/tests/operation_pool_tests.rs
Leo-Besancon: Tested the described behaviour in c3d8ebdef64154f49b3e35355aa87d7ae19fb0d2

bilboquet (approved):
looks ok

Issue #5057, opened 2026-07-21
[F40] Unbounded `query_state` Batch Processing Can Cause Execution Lock Contention and Node DoS


damip:
> ⚠️ **Warning — This finding was routed to an issue rather than a PR because it changes a JSON-RPC / gRPC interface or its observable behavior** (accepted inputs, returned results, error-vs-silent-drop, or transport/security posture). Per maintainer guidance this is treated as **breaking**, because it can break external tooling and clients; the gRPC proto types are additionally **auto-generated** (external `massa_proto_rs`) and must not be hand-edited. It needs an interface-compatibility decision, not a drop-in PR.

---

## Source
AI security-scan **Finding 40** (severity: **major**). Tracked in `SUBMISSIONS.md` as F40.

**Affected file:** `massa-execution-worker/src/controller.rs`

## Summary
The `query_state` logic takes a read lock on `self.execution_state.read()` before it starts processing a batch. It keeps that lock for the full batch, including expensive query items and serial evaluation of all responses. Because the execution worker needs `self.execution_state.write()` to process final slots, candidate slots, and queued read-only executions, a long or heavy query batch can block execution progress for the whole node. A remote caller can abuse the public API batch limit of up to `128` items to slow the node, make it fall behind, or make it unresponsive.

## Root cause
The main root cause is that `query_state` acquires `self.execution_state.read()` too early and holds it for too long. The lock remains held until every request item has been evaluated and appended to the response, instead of being limited to the small part that actually needs shared access.

A second root cause is that the execution worker depends on `self.execution_state.write()` to execute final slots, candidate slots, and queued read-only executions. Since the read lock from `query_state` is held across the full batch, this write access cannot proceed while the batch is still running.

A third root cause is that public API wrappers still allow batches of up to `128` query items, and expensive items such as `Events`, deferred-credit lookups, and repeated datastore scans are executed serially under the same read lock. This increases the time that `self.execution_state.read()` stays held and makes lock contention externally triggerable.

## Affected code
```rust
/// Atomically query the execution state with multiple requests
fn query_state(&self, req: ExecutionQueryRequest) -> ExecutionQueryResponse {
    let execution_lock = self.execution_state.read();
    let mut resp: ExecutionQueryResponse = ExecutionQueryResponse {
        responses: Vec::with_capacity(req.requests.len()),
        candidate_cursor: execution_lock.active_cursor,
        final_cursor: execution_lock.final_cursor,
[...]

PR #5236, opened 2026-08-27
[F41] feat: implem cumulative budget for query_state

### LINKED ISSUE #5058: [F41] Large bytecode and datastore-value queries have no aggregate size cap, enabling memory-exhaustion DoS

---

## Source
AI security-scan **Finding 41** (severity: **medium**). Tracked in `SUBMISSIONS.md` as F41.

**Affected file:** `massa-execution-worker/src/controller.rs`

## Summary
`query_state` in `massa-execution-worker/src/controller.rs` batches `ExecutionQueryResponseItem::Bytecode` and `ExecutionQueryResponseItem::DatastoreValue` results into `resp.responses` without any aggregate response-size limit. Because public query entrypoints still allow up to `128` request items and the underlying bytecode and datastore-value blobs can each reach `10_000_000` bytes, repeated requests for the same large object can force the node to allocate and retain very large in-memory responses before the API or gRPC layer rejects the reply. The original finding is valid as a memory-pressure denial-of-service issue, but the claim that response serialization pauses occur while the global execution read lock is held is not supported.

## Root cause
The confirmed root cause is that `query_state` acquires `self.execution_state.read()` and then iterates over all requested items, pushing each result into `resp.responses` with no cumulative byte budget for large payload-bearing variants such as `ExecutionQueryRequestItem::AddressBytecodeCandidate`, `ExecutionQueryRequestItem::AddressBytecodeFinal`, `ExecutionQueryRequestItem::AddressDatastoreValueCandidate`, and `ExecutionQueryRequestItem::AddressDatastoreValueFinal`.

For these request types, the underlying getters return owned `Bytecode` or `Vec<u8>` values, which are then stored in the response vector rather than referenced lazily. This allows a single batched request to accumulate many large copies in memory before any downstream response-size enforcement occurs.

The ruled-out portion of the original theory is the lock-scope claim about transport work. Although the cloning and response assembly inside `query_state` occur while the execution read lock is held, the later API or gRPC serialization and encoding happen only after `query_state` returns, when that lock has already been released.

## Affected code
```rust
/// Atomically query the execution state with multiple requests
fn query_state(&self, req: ExecutionQueryRequest) -> ExecutionQueryResponse {
    let execution_lock = self.execution_state.read();
    let mut resp: ExecutionQueryResponse = ExecutionQueryResponse {
        responses: Vec::with_capacity(req.requests.len()),
        candidate_cursor: execution_lock.active_cursor,
        final_cursor: execution_lock.final_cursor,
        final_state_fingerprint: execution_lock.get_final_state_fingerprint(),
    };
    for req_item in req.requests {
        let resp_item = match req_item {
            ExecutionQueryRequestItem::AddressExistsCandidate(addr) => {
                Ok(ExecutionQueryResponseItem::Boolean(
                    execution_lock
                        .get_final_and_candidate_balance(&addr)
                        .1
                        .is_some(),
                ))
            }
            ExecutionQueryRequestItem::AddressExistsFinal(addr) => {
                Ok(ExecutionQueryResponseItem::Boolean(
                    execution_lock
                        .get_final_and_candidate_balance(&addr)
                        .0
                        .is_some(),
                ))
            }
            ExecutionQueryRequestItem::AddressBalanceCandidate(addr) => {
                let (_final_v, speculative_v) =
                    execution_lock.get_final_and_candidate_balance(&addr);
                match speculative_v {
                    Some(balance) => Ok(ExecutionQueryResponseItem::Amount(balance)),
                    None => Err(ExecutionQueryError::NotFound(format!("Account {}", addr))),
                }
            }
            ExecutionQueryRequestItem::AddressBalanceFinal(addr) => {
                let (final_v, _speculative_v) =
                    execution_lock.get_final_and_candidate_balance(&addr);
                match final_v {
                    Some(balance) => Ok(ExecutionQueryResponseItem::Amount(balance)),
                    None => Err(ExecutionQueryError::NotFound(format!("Account {}", addr))),
                }
            }
            ExecutionQueryRequestItem::AddressBytecodeCandidate(addr) => {
                let (_final_v, speculative_v) =
// … (truncated)
```

## Recommendation
Enforce a cumulative response-size limit in `query_state` for payload-bearing query types, especially bytecode and datastore-value results, and reject or truncate requests once the aggregate budget is exceeded. Apply this limit before appending items to `resp.responses` so memory usage remains bounded during response construction, rather than relying on downstream API or gRPC size checks after allocation has already occurred. To further reduce abuse potential, consider using stricter per-request limits for large-object queries and preventing repeated retrieval of the same large value within a single batch.

## Why P0 / breaking

Leo-Besancon:
  * [ ] make tests compile
  * [ ] make tests pass 

This PR volontarily only takes Bytecode and Datastore queries into account for the budget (for performances, if the total size exceeds the budget it will be caught later anyway).
It does not impact API / GRPC users as we just catch earlier big responses that would have error out later.

### DIFF
--- a/massa-api/src/public.rs
+++ b/massa-api/src/public.rs
@@ -28,7 +28,7 @@ use massa_api_exports::{
 use massa_consensus_exports::block_status::DiscardReason;
 use massa_consensus_exports::ConsensusController;
 use massa_execution_exports::{
-    ExecutionController, ExecutionQueryRequest, ExecutionQueryRequestItem,
+    ExecutionController, ExecutionQueryRequest, ExecutionQueryRequestItem, ExecutionQueryResponse,
     ExecutionQueryResponseItem, ExecutionStackElement, ReadOnlyExecutionRequest,
     ReadOnlyExecutionTarget,
 };
@@ -95,6 +95,15 @@ impl API<Public> {
             storage,
             keypair_factory: KeyPairFactory { mip_store },
         })
+    }
+
+    fn exec_query(&self, requests: Vec<ExecutionQueryRequestItem>) -> ExecutionQueryResponse {
+        self.0
+            .execution_controller
+            .query_state(ExecutionQueryRequest {
+                requests,
+                max_response_size: self.0.api_settings.max_response_body_size as usize,
+            })
     }
 }
 
@@ -1034,13 +1043,7 @@ impl MassaRpcServer for API<Public> {
 
         let mut result: Vec<GetAddressDatastoreKeysResponse> = Vec::with_capacity(requests.len());
 
-        let response = self
-            .0
-            .execution_controller
-            .query_state(ExecutionQueryRequest {
-                requests,
-                max_response_size: self.0.api_settings.max_response_body_size as usize,
-            });
+        let response = self.exec_query(requests);
 
         for response_item in response.responses {
             match response_item {
@@ -1256,14 +1259,7 @@ impl MassaRpcServer for API<Public> {
             return Err(ApiError::BadRequest(format!("too many arguments received. Only a maximum of {} arguments are accepted per request", self.0.api_settings.max_arguments)).into());
         }
 
-        let responses = self
-            .0
-            .execution_controller
-            .query_state(ExecutionQueryRequest {
-                requests: queries,
-                max_response_size: self.0.api_settings.max_response_body_size as usize,
-            })
-            .responses;
+        let responses = self.exec_query(queries).responses;
 
         let res: Result<Vec<Vec<u8>>, ApiError> = responses
             .into_iter()
@@ -1299,12 +1295,7 @@ impl MassaRpcServer for API<Public> {
             .collect();
 
         let result = self
-            .0
-            .execution_controller
-            .query_state(ExecutionQueryRequest {
-                requests: queries,
-                max_response_size: self.0.api_settings.max_response_body_size as usize,
-            })
+            .exec_query(queries)
             .responses
             .into_iter()
             .map(|response| match response {
@@ -1347,12 +1338,7 @@ impl MassaRpcServer for API<Public> {
             .collect::<Result<_, _>>()?;
 
         let result: Vec<DeferredCallResponse> = self
-            .0
-            .execution_controller
-            .query_state(ExecutionQueryRequest {
-                requests,
-                max_response_size: self.0.api_settings.max_response_body_size as usize,
-            })
+            .exec_query(requests)
             .responses
             .into_iter()
             .map(|exec| match exec {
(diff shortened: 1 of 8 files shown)


massa-api/src/public.rs
```diff
@@ -1037,7 +1037,10 @@ impl MassaRpcServer for API<Public> {
         let response = self
             .0
             .execution_controller
-            .query_state(ExecutionQueryRequest { requests });
+            .query_state(ExecutionQueryRequest {
```
bilboquet: I suggest to factorize code this way:

add
```rs
impl API<Public> {
    fn exec_query(
        &self,
        requests: Vec<ExecutionQueryRequestItem>,
    ) -> ExecutionQueryResponse {
        self.0.execution_controller.query_state(ExecutionQueryRequest {
            requests,
            max_response_size: self.0.api_settings.max_response_body_size as usize,
        })
    }
}
```

then replace every
```rs
let response = self
    .0
    .execution_controller
    .query_state(ExecutionQueryRequest {
        requests,
        max_response_size: self.0.api_settings.max_response_body_size as usize,
    });
```
by
```rs
let response = self.exec_query(requests);
```


WDYT?

bilboquet (dismissed):

damip (commented):
This is the right pattern for bounding `query_state` (see #5057): the lock is kept for the whole batch, over-budget items become per-item errors, and the cursors/fingerprint stay valid for everything returned. Two things since you're already editing the doc comment of `query_state`:

1. State the rule explicitly in that doc comment so nobody splits the lock later (F40's recommendation does exactly that and is rejected):

```rust
/// Atomically query the execution state with multiple requests.
///
/// INVARIANT: all items MUST be computed under the single `execution_state` read lock taken below,
/// because the returned cursors and fingerprint describe the whole batch and clients rely on this
/// to read a consistent state. Never release the lock between items to reduce contention:
/// bound the cost of a batch via budgets (per-item errors) instead.
```

2. Follow-up (can be another PR): apply the same cumulative budget to `Events` (128 items x `max_event_per_query` is still 900k events under one lock) and to datastore key queries.

Also agree with @bilboquet's factorization suggestion.

Leo-Besancon:
> This is the right pattern for bounding `query_state` (see #5057): the lock is kept for the whole batch, over-budget items become per-item errors, and the cursors/fingerprint stay valid for everything returned. Two things since you're already editing the doc comment of `query_state`:
> 
>     1. State the rule explicitly in that doc comment so nobody splits the lock later (F40's recommendation does exactly that and is rejected):
> 
> 
> ```rust
> /// Atomically query the execution state with multiple requests.
> ///
> /// INVARIANT: all items MUST be computed under the single `execution_state` read lock taken below,
> /// because the returned cursors and fingerprint describe the whole batch and clients rely on this
> /// to read a consistent state. Never release the lock between items to reduce contention:
> /// bound the cost of a batch via budgets (per-item errors) instead.
> ```
> 
>     2. Follow-up (can be another PR): apply the same cumulative budget to `Events` (128 items x `max_event_per_query` is still 900k events under one lock) and to datastore key queries.
> 
> 
> Also agree with @bilboquet's factorization suggestion.

Thanks! In the latest commits I've taken into account the factorization suggestion as well as improved the doc comment.
I've created a followup issue for your second point: https://github.com/massalabs/massa/issues/5255.

massa-api/src/public.rs
Leo-Besancon: Thanks, it's a lot cleaner! Added in 8d1712a4d4c760fbe2c28872db3f47ae258895b8

bilboquet (approved):

PR #5237, opened 2026-08-27
[F34] feat: align deferred calls ser and deser config checks

### LINKED ISSUE #4997: [F34] Missing Gas Limit Validation Causes Deserialization Denial of Service (DoS)

---

## Source
AI security-scan **Finding 34** (severity: **info**). Tracked in `SUBMISSIONS.md` as F34.

**Affected file:** `massa-deferred-calls/src/slot_changes.rs`

## Summary
`massa-main/massa-deferred-calls/src/slot_changes.rs` contains a real encode/decode asymmetry in `DeferredRegistrySlotChanges`: serialization accepts any `effective_slot_gas` value and any `calls.len()`, while `DeferredRegistrySlotChangesDeserializer::new(config)` enforces `config.max_gas` and `config.max_pool_changes`. This inconsistency means the serializer and deserializer do not implement the same bounds for the same structure. However, the original finding overstated the practical exploit path: the claimed network-broadcast or persisted-state "poison pill" denial of service is not supported by the runtime code paths.

## Root cause
The confirmed issue is a validation mismatch inside the custom `DeferredRegistrySlotChanges` codec in `massa-main/massa-deferred-calls/src/slot_changes.rs`.

- `set_effective_slot_gas` writes any `u64` directly into `effective_slot_gas` without checking `DeferredCallsConfig.max_gas`.
- `DeferredRegistrySlotChangesSerializer` uses unbounded encoders for both `effective_slot_gas` and `calls.len()`, so it can serialize values above the limits later enforced by the paired deserializer.
- `DeferredRegistrySlotChangesDeserializer::new(config)` instead builds bounded deserializers for gas and map length, enforcing `Included(config.max_gas)` and `Included(config.max_pool_changes)`.

The ruled-out portion is the claim that this mismatch leads to a practical production poison-pill path for deferred-call state. According to the validated analysis, this custom `DeferredRegistrySlotChanges` serialization path is used only in tests, while runtime finalization stores deferred-call data directly in database fields and reads slot gas back with an unbounded `u64` deserializer. The original gas-reachability theory is also narrowed: normal deferred-call registration checks `slot_occupancy + gas_request <= config.max_gas` before `set_effective_slot_gas` is called.

## Affected code
```rust
pub fn set_effective_slot_gas(&mut self, gas: u64) {
        self.effective_slot_gas = DeferredRegistryGasChange::Set(gas);
    }
```

## Recommendation
Align serialization and deserialization invariants for `DeferredRegistrySlotChanges` so both enforce the same bounds under the same `DeferredCallsConfig`. In particular, validate `effective_slot_gas` against `config.max_gas` and ensure the number of entries in `calls` does not exceed `config.max_pool_changes` before serialization or when mutating the structure. If these values are intended to be unrestricted in some contexts, remove or centralize the mismatch by defining a single source of truth for the allowed bounds and applying it consistently across constructors, mutators, serializers, and deserializers. Add regression tests to confirm that any value accepted for serialization is also accepted by the paired deserializer under the same configuration.

## Why P0 / breaking

Leo-Besancon:
  * [ ] make tests compile
  * [ ] make tests pass 

### DIFF
--- a/massa-final-state/src/state_changes.rs
+++ b/massa-final-state/src/state_changes.rs
@@ -71,7 +71,9 @@ impl StateChangesSerializer {
         Self {
             ledger_changes_serializer: LedgerChangesSerializer::new(),
             async_pool_changes_serializer: AsyncPoolChangesSerializer::new(),
-            deferred_call_changes_serializer: DeferredRegistryChangesSerializer::new(),
+            deferred_call_changes_serializer: DeferredRegistryChangesSerializer::new(
+                DeferredCallsConfig::default(),
+            ),
             pos_changes_serializer: PoSChangesSerializer::new(),
             ops_changes_serializer: ExecutedOpsChangesSerializer::new(),
             de_changes_serializer: ExecutedDenunciationsChangesSerializer::new(),
--- a/massa-final-state/src/state_changes.rs
+++ b/massa-final-state/src/state_changes.rs
@@ -59,20 +59,14 @@ pub struct StateChangesSerializer {
     execution_trail_hash_change_serializer: SetOrKeepSerializer<massa_hash::Hash, HashSerializer>,
 }
 
-impl Default for StateChangesSerializer {
-    fn default() -> Self {
-        Self::new()
-    }
-}
-
 impl StateChangesSerializer {
     /// Creates a `StateChangesSerializer`
-    pub fn new() -> Self {
+    pub fn new(deferred_calls_config: DeferredCallsConfig) -> Self {
         Self {
             ledger_changes_serializer: LedgerChangesSerializer::new(),
             async_pool_changes_serializer: AsyncPoolChangesSerializer::new(),
             deferred_call_changes_serializer: DeferredRegistryChangesSerializer::new(
-                DeferredCallsConfig::default(),
+                deferred_calls_config,
             ),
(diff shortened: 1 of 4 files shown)


massa-final-state/src/state_changes.rs
```diff
@@ -71,7 +71,9 @@ impl StateChangesSerializer {
         Self {
             ledger_changes_serializer: LedgerChangesSerializer::new(),
             async_pool_changes_serializer: AsyncPoolChangesSerializer::new(),
-            deferred_call_changes_serializer: DeferredRegistryChangesSerializer::new(),
+            deferred_call_changes_serializer: DeferredRegistryChangesSerializer::new(
```
bilboquet, line 68: here we're using DeferredCallsConfig::default() while in the StateChangesDeserializer the real config is used. It feels still asymmetric to me.

I'd rather drop DeferredCallsConfig::default()

bilboquet (changes requested):

massa-final-state/src/state_changes.rs
Leo-Besancon: Good catch! Addressed in 6ce4f8433e08764e8d055038caaf46e8c4279b02

bilboquet (approved):

PR #5238, opened 2026-08-28
[F11] feat: check blocks validity against last_start_period and parents mismatch

### LINKED ISSUE #4980: [F11] Bootstrap graph parsing disables genesis parent validation, enabling crashable malformed final blocks

---

## Source
AI security-scan **Finding 11** (severity: **medium**). Tracked in `SUBMISSIONS.md` as F11.

## Summary
`BootstrapServerMessageDeserializer::new` in `massa-main/massa-bootstrap/src/messages.rs` constructs the bootstrap graph deserializer from `(&args).into()`, which sets `last_start_period` to `None`. This causes bootstrap block headers to bypass the normal genesis versus non-genesis parent-count validation, so malformed headers can deserialize during bootstrap. However, the originally cited consensus panic path is not caused by the header `parents` field alone: the later crashable condition depends on `ExportActiveBlock.parents`, which is deserialized separately and is not cross-checked against the header.

## Root cause
The confirmed issue is that `BootstrapServerMessageDeserializer::new` passes bootstrap graph deserialization arguments derived from `(&args).into()`, and that conversion hardcodes `last_start_period: None`. As a result, `BlockDeserializerArgs` does not enforce the slot-dependent parent-count invariant that normally distinguishes genesis blocks from ordinary blocks, allowing malformed bootstrapped block headers to deserialize.

The ruled-out portion of the original root-cause theory is the claim that this header-level validation bypass is itself what later drives the `parent.parents[gp_i]` panic path. That downstream path operates on `ActiveBlock.parents`, populated from the separate `ExportActiveBlock.parents` field, and that field is independently deserialized without a slot-based invariant or a consistency check against the header `parents`. Therefore, correcting `last_start_period` alone would not eliminate the malformed active-block state that can later break consensus processing.

The original example involving a genesis-era block with a full parent set is also not the supported crash mechanism for the cited path, because grandparent indexing is skipped when `parent_period == last_start_period`.

## Affected code
```rust
/// Creates a new `BootstrapServerMessageSerializer`
pub fn new() -> Self {
    Self {
        u32_serializer: U32VarIntSerializer::new(),
        u64_serializer: U64VarIntSerializer::new(),
        time_serializer: MassaTimeSerializer::new(),
        version_serializer: VersionSerializer::new(),
        peers_serializer: BootstrapPeersSerializer::new(),
        bootstrapable_graph_serializer: BootstrapableGraphSerializer::new(),
        block_id_set_serializer: PreHashSetSerializer::new(BlockIdSerializer::new()),
        vec_u8_serializer: VecU8Serializer::new(),
        opt_vec_u8_serializer: OptionSerializer::new(VecU8Serializer::new()),
        slot_serializer: SlotSerializer::new(),
        opt_last_start_period_serializer: OptionSerializer::new(U64VarIntSerializer::new()),
        opt_last_slot_before_downtime_serializer: OptionSerializer::new(OptionSerializer::new(
            SlotSerializer::new(),
        )),
    }
}

/// Creates a new `BootstrapServerMessageDeserializer`
pub fn new(args: BootstrapServerMessageDeserializerArgs) -> Self {
    Self {
        message_id_deserializer: U32VarIntDeserializer::new(Included(0), Included(u32::MAX)),
        time_deserializer: MassaTimeDeserializer::new((
            Included(MassaTime::from_millis(0)),
            Included(MassaTime::from_millis(u64::MAX)),
        )),
        version_deserializer: VersionDeserializer::new(),
        peers_deserializer: BootstrapPeersDeserializer::new(
            args.max_advertise_length,
            args.max_listeners_per_peer,
        ),
        datastore_key_deserializer: VecU8Deserializer::new(
            Included(0),
            Included(args.max_datastore_key_length.into()),
        ),
        datastore_val_deserializer: VecU8Deserializer::new(
            Included(0),
            Included(args.max_datastore_value_length),
        ),
        opt_vec_u8_deserializer: OptionDeserializer::new(VecU8Deserializer::new(
            Included(0),
            Included(args.max_datastore_value_length),
        )),
// … (truncated)
```

## Recommendation
Pass the correct `last_start_period` into bootstrap graph deserialization instead of deriving arguments with `(&args).into()` when that conversion leaves `last_start_period` unset. This ensures bootstrap block headers enforce the same genesis and non-genesis parent-count rules as normal deserialization.

In addition, validate `ExportActiveBlock.parents` independently during bootstrap and enforce consistency between the active block’s parent list and the header `parents` field before inserting bootstrapped data into the graph. Reject any bootstrapped block whose slot, header parent count, or exported active-block parent list violates the expected thread-based invariants or contains mismatched parent data.

## Why P0 / breaking

Leo-Besancon:
  * [ ] make tests compile
  * [ ] make tests pass 

### DIFF
--- a/massa-bootstrap/src/client.rs
+++ b/massa-bootstrap/src/client.rs
@@ -79,14 +79,20 @@ pub(crate) fn stream_final_state_and_consensus(
         ..
     } = &next_bootstrap_message
     {
-        // Reconnect / continuation: server will not resend last_start_period, but we
-        // already have it locally — enable header parent-count checks while parsing.
+        // Continuation / reconnect: metadata was received on the first part (empty
+        // consensus); seed the binder for block header validation while parsing.
         client.set_last_start_period(Some(
             global_bootstrap_state
                 .final_state
                 .read()
                 .get_last_start_period(),
         ));
+    } else if let BootstrapClientMessage::AskBootstrapPart {
+        send_last_start_period: true,
+        ..
+    } = &next_bootstrap_message
+    {
+        client.set_last_start_period(None);
     }
 
     if let BootstrapClientMessage::AskBootstrapPart { .. } = &next_bootstrap_message {
@@ -115,16 +121,6 @@ pub(crate) fn stream_final_state_and_consensus(
                     // successful bootstrap.
                     check_restart_metadata(cfg, &last_start_period, &last_slot_before_downtime)?;
 
-                    // Enforce genesis / non-genesis parent-count rules using the server
-                    // last_start_period when present (first parts), otherwise the value
-                    // already stored locally (subsequent parts / reconnect).
-                    let effective_last_start_period = last_start_period
-                        .unwrap_or_else(|| write_final_state.get_last_start_period());
-                    consensus_part
-                        .check_parent_invariants(cfg.thread_count, effective_last_start_period)
-                        .map_err(BootstrapError::GeneralError)?;
-
-                    // We only need to receive the initial_state once
                     if let Some(last_start_period) = last_start_period {
                         write_final_state.set_last_start_period(last_start_period);
                         client.set_last_start_period(Some(last_start_period));
@@ -240,7 +236,12 @@ pub(crate) fn stream_final_state_and_consensus(
                     };
                     let mut write_final_state = global_bootstrap_state.final_state.write();
                     write_final_state.reset();
+                    // `reset()` does not clear restart metadata; drop values from the
+                    // aborted server so the next one defines them from the wire again.
+                    write_final_state.set_last_start_period(0);
+                    write_final_state.set_last_slot_before_downtime(None);
                     drop(write_final_state);
+                    client.set_last_start_period(None);
                     // The cursor above restarts the consensus stream from `Started`, for which the
                     // server reports no outdated ids: blocks kept from the aborted attempt would
                     // never be pruned and would be merged into the next attempt's graph.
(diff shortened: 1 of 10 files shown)


massa-bootstrap/src/messages.rs
```diff
@@ -338,6 +339,19 @@ pub struct BootstrapServerMessageDeserializer {
 impl BootstrapServerMessageDeserializer {
     /// Creates a new `BootstrapServerMessageDeserializer`
     pub fn new(args: BootstrapServerMessageDeserializerArgs) -> Self {
```
bilboquet: drop to prevent future miss use?

Leo-Besancon: Good idea! Removed `::new` and replaced it with an explicit `with_last_start_period(.., None)` for existing test / doctest callsites in the latest commit.

massa-bootstrap/src/client.rs
```diff
@@ -74,6 +74,21 @@ pub(crate) fn stream_final_state_and_consensus(
     next_bootstrap_message: &mut BootstrapClientMessage,
     global_bootstrap_state: &mut GlobalBootstrapState,
 ) -> Result<(), BootstrapError> {
+    if let BootstrapClientMessage::AskBootstrapPart {
+        send_last_start_period: false,
+        ..
+    } = &next_bootstrap_message
+    {
+        // Reconnect / continuation: server will not resend last_start_period, but we
```
bilboquet: can't the server always send the send_last_start_period?
if it does, I guess we can simplify on the client side (and simplify a bit globally)?

Leo-Besancon: It's a good idea! I think an issue blocking the implementation a bit is that updated clients should be able to bootstrap on non-updated servers. This means we need to keep e.g. the send_last_start_period in the client message (always set to true). I'll try someting out

Leo-Besancon: What I did in the latest commit is keep the same order (only send the last_start_period only at start) but worked with the assumption (always verified) that consensus blocks never stream on the 1st message. This let me simplify a bit the workflow (no need for the client to verify the assumtions). Is it better?

bilboquet: well done!
To me, same functionality with less code, all good :)

bilboquet (approved):

PR #5239, opened 2026-08-28
[F35] feat: document allowed malleability for deferred calls serialization

### LINKED ISSUE #4998: [F35] Serialization Malleability via Unordered and Duplicate Keys in BTreeMap Deserialization

---

## Source
AI security-scan **Finding 35** (severity: **info**). Tracked in `SUBMISSIONS.md` as F35.

**Affected file:** `massa-deferred-calls/src/slot_changes.rs`

## Summary
`DeferredRegistrySlotChangesDeserializer::deserialize` in `massa-deferred-calls/src/slot_changes.rs` and `DeferredRegistryChangesDeserializer::deserialize` in `massa-deferred-calls/src/registry_changes.rs` both parse length-prefixed `(key, value)` sequences and then normalize them with `collect::<BTreeMap<_, _>>()`. This confirms a real *deserialization malleability* issue: duplicate keys are silently resolved with last-write-wins semantics, and out-of-order keys are accepted and reordered into the same in-memory map. However, the original claim overstated the downstream risk because these full map serializers and deserializers are effectively test-only, and production deferred-call processing does not rely on these serialized map blobs for hashing, networking, or deduplication.

## Root cause
The confirmed root cause is that the affected deserializers accept arbitrary key order and do not reject duplicate keys before constructing a `BTreeMap`.

In `massa-deferred-calls/src/slot_changes.rs`, `DeferredRegistrySlotChangesDeserializer::deserialize` parses a vector of `(DeferredCallId, DeferredRegistryCallChange)` entries and then executes `let calls = vec.into_iter().collect::<BTreeMap<_, _>>();`. In `massa-deferred-calls/src/registry_changes.rs`, `DeferredRegistryChangesDeserializer::deserialize` similarly parses `(Slot, DeferredRegistrySlotChanges)` entries and converts them with `slots_change: changes.into_iter().collect::<BTreeMap<_, _>>()`.

Because `BTreeMap` insertion normalizes ordering and overwrites earlier values for duplicate keys, multiple distinct byte sequences can deserialize into the same logical `DeferredRegistrySlotChanges` or `DeferredCallRegistryChanges` value. The ruled-out portion of the original theory is the assertion that this parser behavior directly affects production state hashing, signature validation, consensus, gossip, or mempool handling; the validated code paths do not support that conclusion.

## Affected code
```rust
        &self,
        buffer: &'a [u8],
    ) -> IResult<&'a [u8], DeferredRegistrySlotChanges, E> {
        context(
            "Failed DeferredRegistrySlotChanges deserialization",
            tuple((
                length_count(
                        self.deferred_registry_slot_changes_length
                            .deserialize(input)
                    }),
                    |input: &'a [u8]| {
                        tuple((
                            context("Failed id deserialization", |input| {
                                self.call_id_deserializer.deserialize(input)
                            }),
                            context(
                                "Failed set_update_or_delete_message deserialization",
                                |input| self.calls_set_or_delete_deserializer.deserialize(input),
                            ),
                        ))(input)
                    },
                ),
                context("Failed gas deserialization", |input| {
                    self.gas_deserializer.deserialize(input)
                }),
                context("Failed base fee deserialize", |input| {
                    self.base_fee_deserializer.deserialize(input)
                }),
            )),
        )
        .map(|(vec, gas, base_fee)| {
            let calls = vec.into_iter().collect::<BTreeMap<_, _>>();

            DeferredRegistrySlotChanges {
                calls,
                effective_slot_gas: gas,
                base_fee,
            }
        })
        .parse(buffer)
    }
```

## Recommendation
Require these deserializers to enforce a canonical map encoding during parsing. Reject inputs with duplicate keys and reject entries that are not strictly ordered by key, rather than normalizing them through `collect::<BTreeMap<_, _>>()`. This ensures each logical map has a unique valid byte representation and prevents malformed inputs from being silently accepted.

If these helpers are intended to remain test-only, document that scope explicitly and still consider adding strict validation to avoid incorrect assumptions or unsafe reuse in future production code. Also add negative tests covering duplicate keys and unordered entries to verify that non-canonical encodings are rejected.

## Why P0 / breaking

Leo-Besancon:
  * [ ] make tests compile
  * [ ] make tests pass 

Note: only documented the behaviour, no fix, as outlined in https://github.com/massalabs/massa/pull/5220

### DIFF
--- a/massa-deferred-calls/src/registry_changes.rs
+++ b/massa-deferred-calls/src/registry_changes.rs
@@ -233,6 +233,11 @@ impl Deserializer<DeferredCallRegistryChanges> for DeferredRegistryChangesDeseri
                 )),
             )),
         )
+        // Note: Unsorted or duplicate (Slot, DeferredRegistrySlotChanges) pairs on the wire still
+        // deserialize to the same BTreeMap. This serializer/deserializer pair is effectively test-only;
+        // production deferred-call processing does not rely on these serialized map blobs for hashing,
+        // networking, or deduplication. This is intentional malleability that Massa tolerates by
+        // construction and is not exploitable.
         .map(
             |(changes, total_gas, exec_stats)| DeferredCallRegistryChanges {
                 slots_change: changes.into_iter().collect::<BTreeMap<_, _>>(),
--- a/massa-deferred-calls/src/registry_changes.rs
+++ b/massa-deferred-calls/src/registry_changes.rs
@@ -233,11 +233,10 @@ impl Deserializer<DeferredCallRegistryChanges> for DeferredRegistryChangesDeseri
                 )),
             )),
         )
-        // Note: Unsorted or duplicate (Slot, DeferredRegistrySlotChanges) pairs on the wire still
-        // deserialize to the same BTreeMap. This serializer/deserializer pair is effectively test-only;
-        // production deferred-call processing does not rely on these serialized map blobs for hashing,
-        // networking, or deduplication. This is intentional malleability that Massa tolerates by
-        // construction and is not exploitable.
+        // Note: unsorted pairs, and duplicate keys (last occurrence wins), on the wire still deserialize
+        // to a normalized BTreeMap. This serializer/deserializer pair is only used in tests: production
+        // deferred-call processing reads per-key values from the DB and never parses these map blobs.
+        // Massa is malleability-resistant by construction, this is not exploitable.
         .map(
             |(changes, total_gas, exec_stats)| DeferredCallRegistryChanges {
                 slots_change: changes.into_iter().collect::<BTreeMap<_, _>>(),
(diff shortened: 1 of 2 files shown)


damip:
LGTM, exactly the treatment we want for these. I checked: massa-deferred-calls/src/lib.rs only uses the sub deserializers on per-key DB values, the aggregate map parsers are only built in cfg(test).

Optional precision: "duplicate pairs still deserialize to the same BTreeMap" is slightly off, duplicates normalize with last occurrence wins. Could be:

```rust
// Note: unsorted pairs, and duplicate keys (last occurrence wins), on the wire still deserialize
// to a normalized BTreeMap. This serializer/deserializer pair is only used in tests: production
// deferred-call processing reads per-key values from the DB and never parses these map blobs.
// Massa is malleability-resistant by construction, this is not exploitable.
```

Same wording for the slot_changes one. Not blocking.

bilboquet (approved):

PR #5240, opened 2026-08-28
[F46] feat: lock mip_store during finalization

### LINKED ISSUE #5003: [F46] Shared MIP stats advance before final-state finalization commits

---

## Source
AI security-scan **Finding 46** (severity: **info**). Tracked in `SUBMISSIONS.md` as F46.

**Affected file:** `massa-execution-worker/src/execution.rs`

## Summary
`ExecutionState::apply_final_execution_output` in `massa-main/massa-execution-worker/src/execution.rs` updates shared MIP versioning statistics before the slot is durably finalized. The call to `update_versioning_stats` occurs before `final_state.finalize(...)`, so the in-memory `MipStore` can temporarily reflect a newer state than the persisted final state if `finalize` panics. However, the original claim that this skew remains observable through a still-running node is not supported in the real runtime, because the node exits on panic.

## Root cause
The confirmed issue is the ordering inside `ExecutionState::apply_final_execution_output`: `self.update_versioning_stats(&exec_out.block_info, &exec_out.slot);` executes before `self.final_state.write().finalize(exec_out.slot, exec_out.state_changes);` and before `self.final_cursor = exec_out.slot;`. Because `MipStore` is shared through `Arc<RwLock<_>>`, the mutation performed by `update_versioning_stats` becomes immediately visible in memory even though final-state persistence has not yet succeeded.

If `FinalState::finalize` panics, the function aborts before advancing `final_cursor` or committing the slot to the database, leaving the in-memory MIP stats ahead of durable final-state progress at the moment of failure. The ruled-out portion of the original theory is the claim that this inconsistency continues to be exposed to external callers during normal runtime after the panic; `massa-node/src/main.rs` installs a panic hook that terminates the process with `std::process::exit(1)`.

## Affected code
```rust
pub fn apply_final_execution_output(&mut self, mut exec_out: ExecutionOutput) {
    if self.final_cursor >= exec_out.slot {
        panic!("attempting to apply a final execution output at or before the current final_cursor");
    }

    self.update_versioning_stats(&exec_out.block_info, &exec_out.slot);

    self.final_state
        .write()
        .finalize(exec_out.slot, exec_out.state_changes);

    self.final_cursor = exec_out.slot;
}

fn finalize(&mut self, slot: Slot, changes: StateChanges) {
    self._finalize(slot, changes).unwrap()
}

pub struct MipStore(pub Arc<RwLock<MipStoreRaw>>);

pub fn update_network_version_stats(
    &mut self,
    slot_timestamp: MassaTime,
    network_versions: Option<(u32, Option<u32>)>,
) {
    let mut lock = self.0.write();
    lock.update_network_version_stats(slot_timestamp, network_versions);
}
```

## Recommendation
Update the operation order in `ExecutionState::apply_final_execution_output` so that shared MIP versioning statistics are only advanced after `final_state.finalize(...)` succeeds and the slot is considered durably finalized. This ensures the in-memory `MipStore`, persisted final state, and `final_cursor` progress atomically from the caller’s perspective.

If the current design requires precomputation of versioning effects, defer publishing those changes to the shared `MipStore` until after finalization completes successfully. Additionally, avoid panic-driven partial application in this path where possible, and treat finalization failures as recoverable errors so state updates cannot be left in a partially applied order.

## Why P0 / breaking

Leo-Besancon:
  * [ ] make tests compile
  * [ ] make tests pass 

Note: The reorder recommended in the finding's issue has correctness issue.
We have a choice:
- Option 1: add a write lock for in-memory MIP (lock the in-memory MIP, update it, finalize, and release the lock)
- Option 2: have a two-step stagin process and a lock (first, stage the wanted MIP changes without updating the in-memory view, then lock the in-memory MIP, finalize, update the in-memory MIP once finalized, and release the lock).

Both options have similar impact (readers don't see a desync view, disk changes have no impact (same as without this PR)). I went with option 1 as it fixes the issue completely, it both is simpler and a smaller refactor.

### DIFF
--- a/massa-execution-worker/src/execution.rs
+++ b/massa-execution-worker/src/execution.rs
@@ -34,7 +34,6 @@ use massa_models::execution::EventFilter;
 use massa_models::output_event::SCOutputEvent;
 use massa_models::prehash::PreHashSet;
 use massa_models::stats::ExecutionStats;
-use massa_models::timeslots::get_block_slot_timestamp;
 use massa_models::{
     address::Address,
     block_id::BlockId,
@@ -283,11 +282,6 @@ impl ExecutionState {
             );
         }
 
-        // Update versioning stats
-        // This will update the MIP store and must be called before final state write
-        // as it will also write the MIP store on disk
-        self.update_versioning_stats(&exec_out.block_info, &exec_out.slot);
-
         let exec_out_2 = exec_out.clone();
         #[cfg(feature = "slot-replayer")]
         {
@@ -296,9 +290,13 @@ impl ExecutionState {
             println!("<<<");
         }
         // apply state changes to the final ledger
+        let network_versions = exec_out
+            .block_info
+            .as_ref()
+            .map(|i| (i.current_version, i.announced_version));
         self.final_state
             .write()
-            .finalize(exec_out.slot, exec_out.state_changes);
+            .finalize(exec_out.slot, exec_out.state_changes, network_versions);
 
         // update the final ledger's slot
         self.final_cursor = exec_out.slot;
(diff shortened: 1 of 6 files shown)


bilboquet (approved):
LGTM

PR #5242, opened 2026-08-28
[F47] feat: use min_gas_cost as fallback for deferred calls uninitialized slots

### LINKED ISSUE #5004: [F47] Uninitialized Slot Base Fees Allow Near-Zero-Cost Deferred Call Reservation DoS

---

## Source
AI security-scan **Finding 47** (severity: **major**). Tracked in `SUBMISSIONS.md` as F47.

**Affected file:** `massa-execution-worker/src/interface_impl.rs`

## Summary
The deferred call booking flow uses `get_deferred_call_quote` to quote a fee, and `deferred_call_register` trusts that quote. However, `compute_call_fee` uses `get_slot_base_fee(target_slot)`, which returns `0` if that slot has not been initialized. When the deferred-call registry starts empty, most of the first `max_future_slots` booking horizon can remain uninitialized, so attackers can reserve deferred-call capacity at almost no net cost. This can block legitimate users and create a practical denial of service.

## Root cause
The main issue is that `deferred_call_register` accepts the fee returned by `get_deferred_call_quote` without protecting against uninitialized slot pricing. The quote is derived from `compute_call_fee`, which calls `get_slot_base_fee(target_slot)`. If no base fee was initialized for `target_slot`, `get_slot_base_fee(target_slot)` returns `0`, so the booking can be priced at zero.

A second root cause is how slot fees are initialized. `advance_slot` initializes only the single slot that newly enters the booking window. If the deferred-call registry starts empty, such as at genesis or after a reset, this leaves almost all slots in the initial `max_future_slots` horizon uninitialized.

A third root cause is that the pricing model does not add a meaningful fallback charge in this state. While usage remains below `TARGET_BOOKING`, which is `50%` occupancy, both overbooking penalties are also zero. This lets an attacker spread registrations across many uninitialized slots and reserve up to half of the deferred-call gas capacity without paying a non-refundable fee.

The storage charge does not prevent this abuse. `execute_deferred_call` refunds the storage cost even if the deferred call later fails or was cancelled. As a result, bogus low-cost or zero-cost registrations can occupy slot capacity with essentially zero net cost.

## Affected code
```rust
fn deferred_call_register(
        &self,
        target_addr: &str,
        target_func: &str,
        target_slot: (u64, u8),
        max_gas: u64,
        params: &[u8],
        coins: u64,
    ) -> Result<String> {
        // This function spends coins + deferred_call_quote(target_slot, max_gas).unwrap() from the caller, fails if the balance is insufficient or if the quote would return None.

        if target_slot.1 >= self.config.thread_count {
            bail!("target slot thread exceeds the configuration thread count")
        }

        let target_addr = Address::from_str(target_addr).map_err(|e| e.to_string())?;

        // check that the target address is an SC address
        if !matches!(target_addr, Address::SC(..)) {
            bail!("target address is not a smart contract address")
        }

        // Length verifications
        if target_func.len() > self.config.max_function_length as usize {
            bail!("Function name is too large");
        }
        if params.len() > self.config.max_parameter_length as usize {
            bail!("Parameter size is too large");
        }

        // check fee, slot, gas
        let (available, fee_raw) =
            self.get_deferred_call_quote(target_slot, max_gas, params.len() as u64)?;
        if !available {
            bail!("The Deferred call cannot be registered. Ensure that the target slot is not before/at the current slot nor too far in the future, and that it has at least max_gas available gas.");
        }
        let fee = Amount::from_raw(fee_raw);
        let coins = Amount::from_raw(coins);

        // write-lock context

        // get caller address
        let sender_address = context.get_current_address().map_err(|e| e.to_string())?;

// … (truncated)
```

## Recommendation
Initialize base fees for the full `max_future_slots` booking horizon before allowing deferred-call registrations, including at genesis and after any registry reset. Do not allow `get_slot_base_fee(target_slot)` to resolve to `0` for bookable slots; instead, require every reservable slot to have an initialized fee or reject quoting and registration for uninitialized slots.

Add a non-zero minimum non-refundable booking charge so reservations cannot be made at negligible cost, even when slot usage is below `TARGET_BOOKING`. Review the refund logic as well to ensure storage-cost refunds do not reduce failed or cancelled reservations to near-zero net cost.

As a defense in depth measure, make `deferred_call_register` independently validate that the quoted fee was derived from a properly initialized slot fee, rather than relying solely on `get_deferred_call_quote`.

## Why P0 / breaking

Leo-Besancon:
  * [ ] make tests compile
  * [ ] make tests pass 

Note: I don't remember exactly the expected formula for the deferred call quotes, and if this first uninitialized window was expected / left on purpose or not. If you have something to add don't hesitate @damip @Soulthym @modship

### DIFF
--- a/massa-deferred-calls/src/lib.rs
+++ b/massa-deferred-calls/src/lib.rs
@@ -48,6 +48,7 @@ pub struct DeferredCallRegistry {
     call_id_deserializer: DeferredCallIdDeserializer,
     registry_changes_deserializer: DeferredRegistryChangesDeserializer,
     registry_changes_serializer: DeferredRegistryChangesSerializer,
+    min_gas_cost: u64,
 }
 
 impl DeferredCallRegistry {
@@ -68,6 +69,7 @@ impl DeferredCallRegistry {
             call_id_deserializer: DeferredCallIdDeserializer::new(),
             registry_changes_deserializer: DeferredRegistryChangesDeserializer::new(config),
             registry_changes_serializer: DeferredRegistryChangesSerializer::new(config),
+            min_gas_cost: config.min_gas_cost,
         }
     }
 
@@ -151,10 +153,12 @@ impl DeferredCallRegistry {
         }
     }
 
-    /// Returns the base fee for a slot
+    /// Returns the base fee for a slot.
+    /// Uninitialized slots (missing from storage or set to zero) fall back to
+    /// `min_gas_cost` so bookable slots are never priced at zero.
     pub fn get_slot_base_fee(&self, slot: &Slot) -> Amount {
         let key = deferred_call_slot_base_fee_key!(slot.to_bytes_key());
-        match self.db.read().get_cf(STATE_CF, key) {
+        let base_fee = match self.db.read().get_cf(STATE_CF, key) {
             Ok(Some(v)) => {
                 self.call_deserializer
                     .amount_deserializer
@@ -163,6 +167,12 @@ impl DeferredCallRegistry {
                     .1
             }
             _ => Amount::zero(),
+        };
+
+        if base_fee.is_zero() {
+            Amount::from_raw(self.min_gas_cost)
+        } else {
+            base_fee
         }
     }
 
--- a/massa-deferred-calls/src/lib.rs
+++ b/massa-deferred-calls/src/lib.rs
@@ -154,8 +154,11 @@ impl DeferredCallRegistry {
     }
 
     /// Returns the base fee for a slot.
-    /// Uninitialized slots (missing from storage or set to zero) fall back to
-    /// `min_gas_cost` so bookable slots are never priced at zero.
+    /// Uninitialized slots (no entry, or zero) fall back to `min_gas_cost`: this is exactly
+    /// what `advance_slot` would have written for them from an empty registry (0 booked gas
+    /// puts the controller in its "decrease" branch, which floors at `min_gas_cost`), so the
+    /// fallback is the missing initialization, not a different pricing rule. On a registry
+    /// that has been advancing for more than `max_future_slots` this never fires.
     pub fn get_slot_base_fee(&self, slot: &Slot) -> Amount {
         let key = deferred_call_slot_base_fee_key!(slot.to_bytes_key());
         let base_fee = match self.db.read().get_cf(STATE_CF, key) {
--- a/massa-execution-worker/src/speculative_deferred_calls.rs
+++ b/massa-execution-worker/src/speculative_deferred_calls.rs
@@ -165,14 +165,8 @@ impl SpeculativeDeferredCallRegistry {
             .get_prev_slot(self.config.thread_count)
             .expect("cannot get prev slot");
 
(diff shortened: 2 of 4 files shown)


bilboquet (dismissed):

massa-execution-worker/src/speculative_deferred_calls.rs
bilboquet: with this PR, I guess this if will always return false.

bilboquet (dismissed):
One comment for I think is dead code (harmless).
So approved.

massa-execution-worker/src/speculative_deferred_calls.rs
Leo-Besancon: Indeed, refactored in the latest commit (5f36d0ee70acc5aca49e522985c991978c5028c3)

bilboquet (dismissed):

massa-deferred-calls/src/lib.rs
```diff
@@ -151,10 +153,12 @@ impl DeferredCallRegistry {
         }
     }
 
-    /// Returns the base fee for a slot
+    /// Returns the base fee for a slot.
+    /// Uninitialized slots (missing from storage or set to zero) fall back to
+    /// `min_gas_cost` so bookable slots are never priced at zero.
```
damip: Please say why this value in particular, so nobody "fixes" it later:

```rust
    /// Returns the base fee for a slot.
    /// Uninitialized slots (no entry, or zero) fall back to `min_gas_cost`: this is exactly
    /// what `advance_slot` would have written for them from an empty registry (0 booked gas
    /// puts the controller in its "decrease" branch, which floors at `min_gas_cost`), so the
    /// fallback is the missing initialization, not a different pricing rule. On a registry
    /// that has been advancing for more than `max_future_slots` this never fires.
```

damip (commented):
Fine to merge, no MIP. To answer the question in the description: the uninitialized window was not intentional. The controller was meant to start at `min_gas_cost` and stay floored there, which is what `advance_slot` already did for its own `prev_slot` lookup. The fallback gives exactly the value the controller would have written for those slots from an empty registry (`avg_booked_gas = 0` -> `Less` -> `max(.., min_gas_cost)`), so this is the missing initialization, not a new pricing rule.

Why no MIP: on mainnet no bookable slot can have a zero base fee. `execute_slot` calls `deferred_calls_advance_slot` before any operation runs, so the newest bookable slot (`current + max_future_slots`, the inclusive bound in `compute_call_fee`) is initialized before anything can book it, and all three branches of `advance_slot` produce >= `min_gas_cost`. No zero in changes, history or final state, the fallback never fires, quotes do not change. Test deltas check out (`10 * 200_000` and `10 * 1_050_000` gas).

Only relevant for networks starting from an empty registry (buildnet/labnet resets, future networks), where the first 7 days were free to book up to 50%.

One small ask inline.

massa-deferred-calls/src/lib.rs
Leo-Besancon: Added your comment in the latest commit.

bilboquet (approved):

PR #5249, opened 2026-09-02
[F62] feat: add metrics to detect underfllled blocks

### LINKED ISSUE #5011: [F62] Pool query timeouts silently produce underfilled blocks

---

## Source
AI security-scan **Finding 62** (severity: **minor**). Tracked in `SUBMISSIONS.md` as F62.

**Affected file:** `massa-factory-worker/src/block_factory.rs`

## Summary
`BlockFactoryWorker::process_slot` in `massa-main/massa-factory-worker/src/block_factory.rs` treats pool query `LockTimeout` errors as soft failures and proceeds to build and register a block with empty operations, endorsements, or denunciations. This behavior is confirmed by the implementation and by `massa-factory-worker/src/tests/block_scenarios.rs`, which explicitly covers block production after such timeouts. The confirmed issue is therefore silent production of underfilled blocks when local pool access times out. The original claim is narrower than stated, however: the code does not support a reliable external attack path that can force these local lock timeouts or generally extract ordering value without affecting pool contents.

## Root cause
In `BlockFactoryWorker::process_slot`, failures from `self.channels.pool.get_block_endorsements(...)`, `self.channels.pool.get_block_operations(...)`, and `self.channels.pool.get_block_denunciations(...)` are caught and replaced with empty fallback values instead of aborting block production. Specifically, the function substitutes `vec![None; self.cfg.endorsement_count as usize]`, `vec![]`, and `vec![]` respectively, then still continues through block creation and calls `self.channels.consensus.register_block(block_id, slot, block_storage, true)`.

The confirmed root cause is this fail-open handling of local pool `RwLock` acquisition timeouts inside block assembly. The broader original theory that these timeouts can be induced as generic remote pool-query failures by a network attacker is not supported by the validator's analysis.

## Affected code
```rust
fn process_slot(&mut self, slot: Slot) {
    let channel_timeout = self.cfg.block_opt_channel_timeout;
    ...
    let (endorsements_ids, endo_storage) = match self.channels.pool.get_block_endorsements(
        &same_thread_parent_id,
        &slot,
        Some(channel_timeout),
    ) {
        Ok(result) => result,
        Err(e) => {
            warn!("Channel timeout: get_block_endorsements took >{}ms for slot {}: {:?}. Proceeding with empty endorsements.", channel_timeout.as_millis(), slot, e);
            (vec![None; self.cfg.endorsement_count as usize], self.channels.storage.clone_without_refs())
        }
    };
    ...
    let (op_ids, op_storage) = match self.channels.pool.get_block_operations(&slot, Some(channel_timeout)) {
        Ok(result) => result,
        Err(e) => {
            warn!("Channel timeout: get_block_operations took >{}ms for slot {}: {:?}. Proceeding without operations.", channel_timeout.as_millis(), slot, e);
            (vec![], massa_storage::Storage::create_root())
        }
    };
    ...
    let denunciations = match self.channels.pool.get_block_denunciations(&slot, Some(channel_timeout)) {
        Ok(result) => result,
        Err(e) => {
            warn!("Channel timeout: get_block_denunciations took >{}ms for slot {}: {:?}. Proceeding without denunciations.", channel_timeout.as_millis(), slot, e);
            vec![]
        }
    };
    ...
    self.channels.consensus.register_block(block_id, slot, block_storage, true);
}
```

## Recommendation
Treat pool query `LockTimeout` conditions during block assembly as hard failures rather than silently substituting empty operations, endorsements, or denunciations. If any required pool read times out, abort block construction for that slot, emit a clear error, and avoid registering a partially populated block. This fail-closed behavior prevents silent degradation of block contents and makes local contention visible to operators.

Additionally, review lock scope, contention hotspots, and timeout configuration in the pool access path to reduce the likelihood of transient internal lock starvation during `process_slot`. Ensure monitoring and alerting distinguish pool timeouts from ordinary empty-pool conditions so underfilled blocks cannot be produced without explicit operational visibility.

## Why P0 / breaking

Leo-Besancon:
  * [ ] make tests compile
  * [ ] make tests pass 

Adds metrics on:
- the count of specific pool timeouts underfilling a block
- the count of produced blocks
- the count of underfilled produced blocks

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3338,6 +3338,7 @@ dependencies = [
  "massa_consensus_exports",
  "massa_factory_exports",
  "massa_hash",
+ "massa_metrics",
  "massa_models",
  "massa_pool_exports",
  "massa_pos_exports",
--- a/massa-factory-worker/Cargo.toml
+++ b/massa-factory-worker/Cargo.toml
@@ -22,6 +22,7 @@ massa_wallet = {workspace = true}
 massa_pos_exports = {workspace = true}
 massa_pool_exports = {workspace = true}
 massa_versioning = {workspace = true}
+massa_metrics = {workspace = true}
 
 [dev-dependencies]
 num = {workspace = true}
--- a/massa-factory-worker/src/block_factory.rs
+++ b/massa-factory-worker/src/block_factory.rs
@@ -124,6 +124,7 @@ impl BlockFactoryWorker {
     fn process_slot(&mut self, slot: Slot) {
         let mut timings = Vec::new();
         let channel_timeout = self.cfg.block_opt_channel_timeout;
+        let mut had_pool_timeout = false;
 
         // get block producer address for that slot
         timings.push(("get_producer START", MassaTime::now()));
@@ -204,6 +205,8 @@ impl BlockFactoryWorker {
         ) {
             Ok(result) => result,
             Err(e) => {
+                had_pool_timeout = true;
+                massa_metrics::inc_factory_pool_timeout_endorsements();
                 warn!(
                     "Channel timeout: get_block_endorsements took >{}ms for slot {}: {:?}. Proceeding with empty endorsements.",
                     channel_timeout.as_millis(), slot, e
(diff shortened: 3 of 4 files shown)


bilboquet (dismissed):

damip (commented):
LGTM, this is the right tool for F62 (node-local visibility for the node runner, who is the one losing fees). See my comment on #5011 for the full reasoning. Since this PR touches exactly the fallback branches, please also add the code comment proposed in #5011 above the first fallback in `process_slot`, so we can close the issue with this PR.

bilboquet (approved):

PR #5253, opened 2026-09-02
[F29] unbounded command flood can indefinitely delay slot ticks and consensus maintenance

### LINKED ISSUE #4992: [F29] Unbounded Command Flood Can Indefinitely Delay Slot Ticks and Consensus Maintenance

---

## Source
AI security-scan **Finding 29** (severity: **major**). Tracked in `SUBMISSIONS.md` as F29.

**Affected file:** `massa-consensus-worker/src/worker/main_loop.rs`

## Summary
The `run` loop performs per-slot maintenance only when `wait_slot_or_command(self.next_instant)` returns `WaitingStatus::Ended`. If commands keep arriving, the call returns `Interrupted`, and the loop immediately waits again using the same absolute `next_instant`. This can prevent the timeout path from ever running, even after the slot deadline has passed. As a result, maintenance such as `slot_tick`, stats and desync updates, and pruning can be delayed indefinitely, which can stall consensus progress and increase memory usage.

## Root cause
In `run`, per-slot work is gated on `wait_slot_or_command(self.next_instant)` returning `WaitingStatus::Ended`. When the consensus command channel receives any command, the function returns `Interrupted` instead.

After an `Interrupted` result, `run` immediately calls `recv_deadline` again with the same absolute `next_instant`. Because `next_instant` is not advanced before re-waiting, a continuously non-empty command channel can keep the loop busy processing commands and stop execution from reaching the timeout branch.

This behavior affects the maintenance executed from the timeout path in `run`, including `slot_tick`, stats updates, desync updates, and pruning. During this delay, future blocks can continue accumulating in `WaitingForSlot`.

## Affected code
```rust
pub fn run(&mut self) {
        let mut last_prune = Instant::now();
        loop {
            match self.wait_slot_or_command(self.next_instant) {
                // When we reached the instant of the next slot
                WaitingStatus::Ended => {
                    if let Some(end) = self.config.end_timestamp {
                        // The testnet has ended. Will be removed for mainnet.
                        if self.next_instant > end.estimate_instant().unwrap() {
                            info!("This episode has come to an end, please get the latest testnet node version to continue");
                            let _ = self
                                .shared_state
                                .read()
                                .channels
                                .controller_event_tx
                                .send(ConsensusEvent::Stop);
                            break;
                        }
                    }
                    let previous_cycle = self
                        .previous_slot
                        .map(|s| s.get_cycle(self.config.periods_per_cycle));
                    let observed_cycle = self.next_slot.get_cycle(self.config.periods_per_cycle);
                    if previous_cycle.is_none() {
                        // first cycle observed
                        info!("Massa network has started ! 🎉")
                    }
                    if previous_cycle < Some(observed_cycle) {
                        info!("Started cycle {}", observed_cycle);
                    }
                    // Execute all operations and checks that should be performed at each slot
                    {
                        let mut write_shared_state = self.shared_state.write();
                        if let Err(err) = write_shared_state.slot_tick(self.next_slot) {
                            warn!("Error while processing block tick: {}", err);
                        }
                    };
                    if last_prune.elapsed().as_millis()
                        > self.config.block_db_prune_interval.as_millis() as u128
                    {
                        self.shared_state
                            .write()
                            .prune()
                            .expect("Error while pruning");
                        last_prune = Instant::now();
// … (truncated)
```

## Recommendation
Ensure slot-based maintenance is not solely dependent on `wait_slot_or_command(self.next_instant)` returning `WaitingStatus::Ended`. Once `next_instant` has been reached or exceeded, execute the overdue per-slot maintenance before processing additional commands, even if the wait was interrupted.

Update the loop so that repeated `Interrupted` results cannot indefinitely defer `slot_tick`, stats updates, desync handling, or pruning. In particular, re-check the current time after each interruption and trigger any missed slot work immediately, advancing `next_instant` as needed to catch up.

Also consider bounding or prioritizing command handling so sustained inbound traffic cannot starve time-driven consensus maintenance. This will preserve liveness and prevent unbounded growth of `WaitingForSlot`.

## Why P0 / breaking

peterjah:
  * [ ] make tests compile
  * [ ] make tests pass 

### DIFF
--- a/massa-consensus-worker/src/worker/main_loop.rs
+++ b/massa-consensus-worker/src/worker/main_loop.rs
@@ -1,6 +1,6 @@
 use std::time::Instant;
 
-use massa_consensus_exports::{error::ConsensusError, events::ConsensusEvent};
+use massa_consensus_exports::error::ConsensusError;
 use massa_models::{
     slot::Slot,
     timeslots::{get_block_slot_timestamp, get_closest_slot_to_timestamp},
@@ -110,28 +110,12 @@ impl ConsensusWorker {
         (next_slot, next_instant)
     }
 
-    /// Performs all per-slot maintenance: episode-end check, cycle logging,
-    /// `slot_tick`, pruning, and advancing to the next slot.
+    /// Performs all per-slot maintenance: cycle logging, `slot_tick`, pruning,
+    /// and advancing to the next slot.
     ///
     /// # Arguments
     /// * `last_prune`: instant when the block DB was last pruned
-    ///
-    /// # Returns
-    /// `true` if the worker should keep running, `false` if the episode ended.
-    fn process_slot_tick(&mut self, last_prune: &mut Instant) -> bool {
-        if let Some(end) = self.config.end_timestamp {
-            // The testnet has ended. Will be removed for mainnet.
-            if self.next_instant > end.estimate_instant().unwrap() {
-                info!("This episode has come to an end, please get the latest testnet node version to continue");
-                let _ = self
-                    .shared_state
-                    .read()
-                    .channels
-                    .controller_event_tx
-                    .send(ConsensusEvent::Stop);
-                return false;
-            }
-        }
+    fn process_slot_tick(&mut self, last_prune: &mut Instant) {
         let previous_cycle = self
             .previous_slot
             .map(|s| s.get_cycle(self.config.periods_per_cycle));
@@ -161,7 +145,6 @@ impl ConsensusWorker {
         }
         self.previous_slot = Some(self.next_slot);
         (self.next_slot, self.next_instant) = self.get_next_slot(Some(self.next_slot));
-        true
     }
 
     /// Runs in loop forever. This loop must stop every slot to perform operations on stats and graph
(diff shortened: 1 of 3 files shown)


massa-consensus-worker/src/worker/main_loop.rs
```diff
@@ -110,6 +110,60 @@ impl ConsensusWorker {
...
+    ///
+    /// # Arguments
+    /// * `last_prune`: instant when the block DB was last pruned
+    ///
+    /// # Returns
+    /// `true` if the worker should keep running, `false` if the episode ended.
+    fn process_slot_tick(&mut self, last_prune: &mut Instant) -> bool {
+        if let Some(end) = self.config.end_timestamp {
+            // The testnet has ended. Will be removed for mainnet.
+            if self.next_instant > end.estimate_instant().unwrap() {
```
bilboquet: maybe an opportunity to drop this if?
and simplify the code a bit.
maybe as as follow up if out of the scope of this pr.

bilboquet (dismissed):
leaved a comment, please take it into account.
I've also approved.

Leo-Besancon:
Agree with the comment above, and removing this deprecated branch let's you get rid of the returned bool too!
Aside from that and the clippy error, LGTM

massa-consensus-worker/src/worker/main_loop.rs
peterjah: done

Leo-Besancon (approved):
LGTM

bilboquet (approved):

PR #5260, opened 2026-09-04
[F30] fix(consensus-worker): preserve side-effect queues on downstrea…

### LINKED ISSUE #4993: [F30] Dropped Consensus Commands and Drained Side Effects on Channel/Processing Errors

---

## Source
AI security-scan **Finding 30** (severity: **medium**). Tracked in `SUBMISSIONS.md` as F30.

**Affected file:** `massa-consensus-worker/src/worker/main_loop.rs`

## Summary
`massa-consensus-worker` handles consensus commands and emits side effects such as block propagation and attack notifications. Several paths treat send or processing failures as *best effort*: they only log a warning and continue. Because of this, commands and notifications can be *silently dropped* when channels are full or downstream delivery fails. This can cause permanent loss of invalidation requests and consensus side effects, and can be abused for denial of service.

## Root cause
The issue comes from two related patterns in `massa-consensus-worker`.

- In `massa-consensus-worker/src/controller.rs`, `ConsensusControllerImpl::mark_invalid_block` sends `ConsensusCommand::MarkInvalidBlock(block_id, header)` with `command_sender.try_send(...)`. If `try_send` returns `Err(TrySendError<_>)`, the function only logs `warn!("error trying to mark block as invalid: {}")` and returns. There is no retry, no blocking wait, and no durable fallback, so the invalidation request is lost.

- The same non-blocking, drop-on-failure pattern is also used in `massa-consensus-worker/src/controller.rs` for `register_block` and `register_block_header`, which call `try_send(ConsensusCommand::RegisterBlock(...))` and `try_send(ConsensusCommand::RegisterBlockHeader(...))`.

- In `massa-consensus-worker/src/worker/main_loop.rs`, the worker loop does not treat processing failures as fatal. In `wait_slot_or_command`, any `manage_command` error is only logged and then ignored. In `Worker::run`, the timeout path calls `shared_state.write().slot_tick(self.next_slot)`. If `slot_tick` returns `Err`, the code only logs `warn!("Error while processing block tick: {}", err)` and still updates `self.previous_slot` and computes the next slot and instant.

- This becomes more severe because `block_db_changed` drains internal side-effect queues with `mem::take` *before* downstream delivery is confirmed. The drained queues include `to_propagate`, `attack_attempts`, `new_final_blocks`, and `new_stale_blocks`. If downstream calls such as `protocol_controller.integrated_block()` or `notify_block_attack()` fail, the error can bubble up into `manage_command` or `slot_tick`, but the worker loop only logs the error and continues. The already-drained items are not restored or retried.

- The architec …

## Affected code
```rust
pub fn run(&mut self) {
        let mut last_prune = Instant::now();
        loop {
            match self.wait_slot_or_command(self.next_instant) {
                // When we reached the instant of the next slot
                WaitingStatus::Ended => {
                    if let Some(end) = self.config.end_timestamp {
                        // The testnet has ended. Will be removed for mainnet.
                        if self.next_instant > end.estimate_instant().unwrap() {
                            info!("This episode has come to an end, please get the latest testnet node version to continue");
                            let _ = self
                                .shared_state
                                .read()
                                .channels
                                .controller_event_tx
                                .send(ConsensusEvent::Stop);
                            break;
                        }
                    }
                    let previous_cycle = self
                        .previous_slot
                        .map(|s| s.get_cycle(self.config.periods_per_cycle));
                    let observed_cycle = self.next_slot.get_cycle(self.config.periods_per_cycle);
                    if previous_cycle.is_none() {
                        // first cycle observed
                        info!("Massa network has started ! 🎉")
                    }
                    if previous_cycle < Some(observed_cycle) {
                        info!("Started cycle {}", observed_cycle);
                    }
                    // Execute all operations and checks that should be performed at each slot
                    {
                        let mut write_shared_state = self.shared_state.write();
                        if let Err(err) = write_shared_state.slot_tick(self.next_slot) {
                            warn!("Error while processing block tick: {}", err);
                        }
                    };
                    if last_prune.elapsed().as_millis()
                        > self.config.block_db_prune_interval.as_millis() as u128
                    {
                        self.shared_state
                            .write()
                            .prune()
                            .expect("Error while pruning");
                        last_prune = Instant::now();
// … (truncated)
```

## Recommendation
Treat consensus command submission and side-effect delivery as reliability-critical rather than best-effort operations. `ConsensusCommand::MarkInvalidBlock`, `RegisterBlock`, and `RegisterBlockHeader` should not be silently dropped when `try_send` fails. Replace drop-on-full behavior with a strategy that preserves delivery guarantees, such as bounded retry with backpressure handling, blocking or asynchronous waiting where appropriate, or a durable queue. If delivery cannot be completed, surface the failure to the caller and transition to a safe degraded state instead of only logging a warning.

Apply the same principle to worker-side processing. Errors returned by `manage_command` and `slot_tick` should not be ignored while slot progression continues. Ensure the worker either retries, restores the affected state, or stops advancing until the failure is resolved, so consensus state cannot move forward while associated side effects were not delivered.

In `block_db_changed`, avoid draining `to_propagate`, `attack_attempts`, `new_final_blocks`, and `new_stale_blocks` before downstream delivery is confirmed. Side effects should remain recoverable until successful dispatch is acknowledged. If dispatch fails, requeue or preserve the pending items so they can be retried without loss.

Define explicit semantics for backpressure and downstream failure handling across the consensus pipeline, including retry limits, observability, and fail-safe behavior. This should guarantee that invalidation requests, propagation events, and attack notifications are either delivered exactly as intended or fail in a visible and recoverable manner.

## Why P0 / breaking

peterjah:
…m failure and stop slot advancement on tick errors

### Problem

`ConsensusState::block_db_changed` drained its pending side-effect queues with `mem::take` and then propagated them with `?`:

```rust
for (block_id, storage) in mem::take(&mut self.to_propagate).into_iter() {
    self.channels.protocol_controller.integrated_block(block_id, storage)?;
}
```

The queues were emptied *before* the sends. `integrated_block` and `notify_block_attack` are `try_send` on a bounded channel, so they fail as soon as protocol is saturated — and the `?` then returned with **the entire remaining batch already gone**. One full channel dropped every pending item, including, potentially, the propagation of our own freshly produced block. The same shape applied to `new_final_blocks` / `new_stale_blocks`, which were taken before `notify_execution`.

Because the failure is channel-level rather than per-item, the dropped blocks were not retried anywhere: they were simply never propagated.

### Fix

**Drain incrementally, keep what wasn't delivered.** Each queue is processed item by item; on the first send failure the failed item and everything still unprocessed are put back into the pending set, and the error is returned:

```rust
let mut to_propagate: Vec<(BlockId, Storage)> =
    std::mem::take(&mut self.to_propagate).into_iter().collect();
let mut propagate_error = None;
while let Some((block_id, storage)) = to_propagate.pop() {
    if let Err(e) = self
        .channels
        .protocol_controller
        .integrated_block(block_id, storage.clone())
    {
        self.to_propagate.insert(block_id, storage);
        propagate_error = Some(e);
        break;
    }
}
for (block_id, storage) in to_propagate {
    self.to_propagate.insert(block_id, storage);
}
```

Same treatment for `attack_attempts`. `new_final_blocks` / `new_stale_blocks` are now only cleared *after* `notify_execution` has actually run.

**Keep advancing slots on tick errors.** The first iteration of this fix also stopped slot advancement when `slot_tick` returned an error. That was wrong — halting consensus is far worse than a delayed propagation, and it is unnecessary once the side effects survive the failure. The main loop logs and moves on:

```rust
if let Err(err) = write_shared_state.slot_tick(self.next_slot) {
    warn!("Error while processing block tick: {}", err);
    // Side effects are preserved in `to_propagate`, `attack_attempts`,
    // `new_final_blocks` and `new_stale_blocks`, so they will be
    // retried on the next tick. Keep advancing slots to avoid
    // halting consensus while downstream is under transient pressure.
}
```

### Design: no blocking, no retry queue

The delivery policy is deliberately **best effort**, and is now stated in the `block_db_changed` doc comment:

- the calls stay `try_send`, we never block the consensus worker on a saturated protocol channel;
- **no new retry queue** is introduced — the items stay in the pending sets they already lived in (`to_propagate`, `attack_attempts`) and are retried at the next `block_db_changed`, i.e. the next consensus command or slot tick;
- the failure is channel-level, not per-item, so nothing needs per-item expiry: everything flushes as soon as protocol drains. A closed channel means the block handler is gone and the node is shutting down anyway;
- growth is bounded by the block integration rate for as long as protocol stays saturated. `to_propagate` is a `PreHashMap<BlockId, Storage>`, so re-inserting the same block is idempotent, and `Storage` is a refcounted handle rather than a copy of the block.

### Observability

Both failure paths now `warn!` with the size of the pending set, so a lasting protocol saturation shows up in the logs instead of being a bare `ConsensusError` with no size information:

```
error propagating blocks to protocol: {e}. {n} block(s) kept pending for retry.
error notifying protocol of attack attempts: {e}. {n} attempt(s) kept pending for retry.
```

An F62-style `MassaMetrics` gauge was considered and left out as out of scope for this fix.

### Drive-by cleanup

The intermediate `VecDeque<MassaTime>` built for the stale-block stats (pre-existing, not introduced here) is gone, along with a redundant second `MassaTime::now()`. The count is captured before the queue is cleared:

```rust
let stale_block_count = self.new_stale_blocks.len();
// ...
self.stale_block_stats
    .extend((0..stale_block_count).map(|_| timestamp));
```

### Known trade-off

`integrated_block` takes `storage: Storage` **by value** and the implementation discards it on failure (`try_send(...).map_err(|_| ProtocolError::ChannelError(...))` throws away the `TrySendError` that still owns the command). To be able to re-insert into `to_propagate` on error we must still own a handle at call time, hence the `storage.clone()` before the call — a happy-path cost that `Storage::clone` does not make entirely free, since it takes write locks on the op/block/endorsement owner maps (`massa-storage/src/lib.rs:65`).

Removing it means changing the trait to take `&Storage`, or having `integrated_block` hand the storage back on error. Both ripple through every implementation and the mocks, so they are left for a follow-up rather than widening this PR.

### Scope

Not consensus-observable: no change to block validity rules, serialization, hashing, final-state or ledger hash, VM semantics, or PoS draws. This only changes *when* already-valid blocks are handed to protocol locally, and the retry is invisible to peers. **No MIP gating required.** No public API surface changes — JSON-RPC and gRPC are untouched.

### DIFF
--- a/massa-consensus-worker/src/state/process.rs
+++ b/massa-consensus-worker/src/state/process.rs
@@ -20,7 +20,7 @@ use massa_models::{
 use massa_signature::PublicKey;
 use massa_storage::Storage;
 use massa_time::MassaTime;
-use tracing::{debug, info};
+use tracing::{debug, info, warn};
 
 use crate::state::{
     clique_computation::compute_max_cliques,
@@ -709,6 +709,27 @@ impl ConsensusState {
     /// 9. notify protocol of block wish list
     /// 10. note new latest final periods (prune graph if changed)
     /// 11. add stale blocks to stats
+    ///
+    /// Delivery to protocol (`integrated_block`, `notify_block_attack`) is best effort: both go
+    /// through `try_send` on the bounded block propagation channel and fail when it is full. On
+    /// failure we neither block nor add a retry queue, and this is deliberate:
+    ///
+    /// - blocking would stall the consensus worker on a saturated protocol, and the first casualty
+    ///   would be the propagation of the block we just produced ourselves, precisely when the node
+    ///   is already struggling;
+    /// - a dedicated retry queue would duplicate state that already exists: on failure the items
+    ///   simply stay in the pending sets they were drained from (`to_propagate`,
+    ///   `attack_attempts`) and are retried at the next `block_db_changed`, i.e. the next consensus
+    ///   command or slot tick.
+    ///
+    /// No per-item expiry is needed either, because the failure is channel-level rather than
+    /// per-item: it says nothing about the block being sent, so everything flushes as soon as
+    /// protocol drains. A closed channel means the block handler is gone and the node is shutting
+    /// down anyway. Growth while protocol stays saturated is bounded by the block integration
+    /// rate: `to_propagate` is keyed by block id, so re-inserting the same block is idempotent,
+    /// and `Storage` is a refcounted handle rather than a copy of the block. Channel saturation is
+    /// visible through the `blocks_propagation_ext_channel_actual_size` metric, and the size of
+    /// the pending sets is logged on each failure.
     pub fn block_db_changed(&mut self) -> Result<(), ConsensusError> {
         massa_trace!("consensus.consensus_worker.block_db_changed", {});
 
@@ -727,6 +748,7 @@ impl ConsensusState {
                 .protocol_controller
                 .integrated_block(block_id, storage.clone())
             {
+                // keeping the undelivered items pending is intended, see the note on this method
                 self.to_propagate.insert(block_id, storage);
                 propagate_error = Some(e);
                 break;
@@ -736,6 +758,11 @@ impl ConsensusState {
             self.to_propagate.insert(block_id, storage);
         }
         if let Some(e) = propagate_error {
+            warn!(
+                "error propagating blocks to protocol: {}. {} block(s) kept pending for retry.",
+                e,
+                self.to_propagate.len()
+            );
             return Err(e.into());
         }
 
@@ -745,6 +772,7 @@ impl ConsensusState {
         let mut attack_error = None;
         while let Some(hash) = attack_attempts.pop() {
             if let Err(e) = self.channels.protocol_controller.notify_block_attack(hash) {
+                // keeping the undelivered items pending is intended, see the note on this method
                 attack_attempts.push(hash);
                 attack_error = Some(e);
                 break;
@@ -755,6 +783,11 @@ impl ConsensusState {
         }
         self.attack_attempts.extend(attack_attempts);
         if let Some(e) = attack_error {
+            warn!(
+                "error notifying protocol of attack attempts: {}. {} attempt(s) kept pending for retry.",
+                e,
+                self.attack_attempts.len()
+            );
             return Err(e.into());
         }
 
@@ -779,12 +812,7 @@ impl ConsensusState {
                 ));
             }
         }
-        // add stale blocks to stats
-        let timestamp = MassaTime::now();
-        let mut stale_block_stats = VecDeque::with_capacity(self.new_stale_blocks.len());
-        for _ in 0..self.new_stale_blocks.len() {
-            stale_block_stats.push_back(timestamp);
-        }
+        let stale_block_count = self.new_stale_blocks.len();
 
         // notify execution
         self.notify_execution(final_block_slots);
@@ -793,7 +821,9 @@ impl ConsensusState {
         self.new_final_blocks.clear();
         self.new_stale_blocks.clear();
         self.final_block_stats.extend(final_block_stats);
-        self.stale_block_stats.extend(stale_block_stats);
+        // add stale blocks to stats
+        self.stale_block_stats
+            .extend((0..stale_block_count).map(|_| timestamp));
 
         // notify protocol of block wishlist
         let new_wishlist = self.get_block_wishlist()?;
--- a/massa-consensus-worker/src/state/process.rs
+++ b/massa-consensus-worker/src/state/process.rs
@@ -1,7 +1,4 @@
-use std::{
-    collections::{BTreeSet, HashMap, VecDeque},
-    mem,
-};
+use std::collections::{BTreeSet, HashMap, VecDeque};
 
 use massa_consensus_exports::{
     block_status::{BlockStatus, DiscardReason, HeaderOrBlock, StorageOrBlock},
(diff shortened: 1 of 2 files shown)


massa-consensus-worker/src/worker/main_loop.rs
```diff
@@ -132,6 +132,9 @@ impl ConsensusWorker {
             let mut write_shared_state = self.shared_state.write();
             if let Err(err) = write_shared_state.slot_tick(self.next_slot) {
                 warn!("Error while processing block tick: {}", err);
+                // Do not advance to the next slot while the tick failed, so that
+                // consensus cannot move forward with undelivered side effects.
+                return;
```
Leo-Besancon, line 137: I'm not sure we want this behaviour, it changes timings in the loop and may retry the same slot many times, adding pressure at a time when the node is struggling.

It also means that while the pressure is there, consensus basically halts processing entirely waiting for it to be resolved.

I would personally be more confortable with an handling in the form of "we keep advancing, but retry the unprocessed side effects" if it's possible and stays consistant.

Leo-Besancon (commented):
Outlined a worry I have in the comment below. I'm not entirely sure the finding itself is a real problem that needs to be addressed, considering https://github.com/massalabs/massa/commit/abc688703c (F19).

To me the change making it possible to retry protocol / attacks notification may be wanted, but I'd be worried changing the timings otherwise.

Do you agree @damip?

massa-consensus-worker/src/worker/main_loop.rs
peterjah: i removed the return

Leo-Besancon (dismissed):
LGTM, I'd still wait for @damip's opinion before merging but I don't have other comments

massa-consensus-worker/src/state/process.rs
```diff
@@ -700,65 +697,91 @@ impl ConsensusState {
...
+        while let Some((block_id, storage)) = to_propagate.pop() {
+            massa_trace!("consensus.consensus_worker.block_db_changed.integrated", {
+                "block_id": block_id
+            });
+            if let Err(e) = self
+                .channels
+                .protocol_controller
+                .integrated_block(block_id, storage.clone())
+            {
+                self.to_propagate.insert(block_id, storage);
```
bilboquet, line 752: if this block always fails, will it be dropped at some point?

massa-consensus-worker/src/state/process.rs
```diff
@@ -700,65 +697,91 @@ impl ConsensusState {
...
+                    a_block.creator_address,
+                    block_is_from_protocol,
+                ));
             }
-            final_block_slots
-        };
+        }
+        // add stale blocks to stats
+        let timestamp = MassaTime::now();
+        let mut stale_block_stats = VecDeque::with_capacity(self.new_stale_blocks.len());
```
bilboquet: I don't know why we create a vector of timestramp (it was already the case). 
but as `self.notify_execution(final_block_slots);` below can't fail, this bindings can be dropped and the
line 783 `self.stale_block_stats.extend(stale_block_stats);` can be replaced by `self.stale_block_stats.extend((0..self.new_stale_blocks.len()).map(|_| timestamp));`

massa-consensus-worker/src/state/process.rs
```diff
@@ -700,65 +697,91 @@ impl ConsensusState {
     /// 10. note new latest final periods (prune graph if changed)
     /// 11. add stale blocks to stats
     pub fn block_db_changed(&mut self) -> Result<(), ConsensusError> {
```
damip, line 733: Please state the policy in the doc comment, same style as the F19 note on `ConsensusControllerImpl`, otherwise this comes back as a finding:

```rust
/// Delivery to protocol (`integrated_block`, `notify_block_attack`) is best effort: the calls are
/// `try_send` on a bounded channel and fail when it is full. On failure we do NOT block and we do
/// NOT add a retry queue: the items simply stay in the existing pending sets (`to_propagate`,
/// `attack_attempts`) and are retried at the next `block_db_changed` (next command or slot tick).
/// The failure is channel-level, not per-item, so everything flushes once protocol drains.
/// Growth is bounded by the block integration rate while protocol is saturated (`to_propagate`
/// is keyed by block id and `Storage` is a refcounted handle).
```

massa-consensus-worker/src/state/process.rs
```diff
@@ -700,65 +697,91 @@ impl ConsensusState {
...
+            {
+                self.to_propagate.insert(block_id, storage);
+                propagate_error = Some(e);
+                break;
             }
+        }
+        for (block_id, storage) in to_propagate {
+            self.to_propagate.insert(block_id, storage);
+        }
+        if let Some(e) = propagate_error {
```
damip, line 760: Add a `warn!` with `self.to_propagate.len()` here (and same for `attack_attempts` below) so protocol saturation is visible in the logs, or a metric like we did for F62. Today it will just come back as a `ConsensusError` log with no size info.

massa-consensus-worker/src/state/process.rs
```diff
@@ -700,65 +697,91 @@ impl ConsensusState {
...
+            std::mem::take(&mut self.to_propagate).into_iter().collect();
+        let mut propagate_error = None;
+        while let Some((block_id, storage)) = to_propagate.pop() {
+            massa_trace!("consensus.consensus_worker.block_db_changed.integrated", {
+                "block_id": block_id
+            });
+            if let Err(e) = self
+                .channels
+                .protocol_controller
+                .integrated_block(block_id, storage.clone())
```
damip, line 749: Nit: the `storage.clone()` bumps the refcount for every block on the happy path just to be able to reinsert on error. Move it and only clone in the error branch, or leave it, it's negligible.

massa-consensus-worker/src/state/process.rs
```diff
@@ -700,65 +697,91 @@ impl ConsensusState {
...
+        }
 
         // notify execution
         self.notify_execution(final_block_slots);
 
+        // Downstream delivery confirmed: drain notification queues and commit stats.
+        self.new_final_blocks.clear();
+        self.new_stale_blocks.clear();
+        self.final_block_stats.extend(final_block_stats);
+        self.stale_block_stats.extend(stale_block_stats);
```
damip: Agree with @bilboquet: `notify_execution` can't fail so the intermediate `VecDeque` is useless. `self.stale_block_stats.extend((0..self.new_stale_blocks.len()).map(|_| timestamp));`

damip (commented):
The fix is the right shape (see my comment on #4993): still `try_send`, no blocking, no new queue, it just stops throwing away the rest of the pending batch when one send fails. Today one full channel drops the whole batch, including possibly the propagation of our own block. Good that the `return` in the main loop is gone, Leo was right, halting slot advancement is the one thing we don't want.

@bilboquet on "will it be dropped at some point": the failure is channel-level, not per-item. Once protocol drains, everything flushes at the next `block_db_changed` (next command or tick). If the channel is closed the block handler is gone and the node is shutting down anyway. No per-item expiry needed. Growth is bounded by the block integration rate while protocol is saturated, and `to_propagate` is keyed by block id.

A few things before merging, inline.

peterjah:
@bilboquet @damip @Leo-Besancon  — all points addressed. Summary of the changes, all in `massa-consensus-worker/src/state/process.rs`.

### 1. Policy documented on `block_db_changed` (@damip)

Added your note verbatim, only reflowed to the file's line width:

```rust
/// Delivery to protocol (`integrated_block`, `notify_block_attack`) is best effort: the calls
/// are `try_send` on a bounded channel and fail when it is full. On failure we do NOT block and
/// we do NOT add a retry queue: the items simply stay in the existing pending sets
/// (`to_propagate`, `attack_attempts`) and are retried at the next `block_db_changed` (next
/// command or slot tick). The failure is channel-level, not per-item, so everything flushes
/// once protocol drains. Growth is bounded by the block integration rate while protocol is
/// saturated (`to_propagate` is keyed by block id and `Storage` is a refcounted handle).
```

### 2. `warn!` on saturation (@damip)

Both error branches now log the error plus the size of the pending set, measured *after* the leftovers have been restored:

```rust
if let Some(e) = propagate_error {
    warn!(
        "error propagating blocks to protocol: {}. {} block(s) kept pending for retry.",
        e,
        self.to_propagate.len()
    );
    return Err(e.into());
}
```

Same for `self.attack_attempts.len()`. I went with `warn!` rather than an F62-style metric — the gauge would need a new entry in `MassaMetrics`, which felt out of scope for this fix. Say the word if you'd rather have it here.

### 3. Stale-block stats simplified — with one correction (@bilboquet, @damip)

Agreed the intermediate `VecDeque` is useless, removed. But the one-liner as proposed would be a **bug** at that location: `self.new_stale_blocks.clear()` runs *before* the `extend` on line 783, so `(0..self.new_stale_blocks.len())` would be empty and the stale-block stats silently dropped. Capturing the count before the clear instead:

```rust
let stale_block_count = self.new_stale_blocks.len();

// notify execution
self.notify_execution(final_block_slots);

// Downstream delivery confirmed: drain notification queues and commit stats.
self.new_final_blocks.clear();
self.new_stale_blocks.clear();
self.final_block_stats.extend(final_block_stats);
// add stale blocks to stats
self.stale_block_stats
    .extend((0..stale_block_count).map(|_| timestamp));
```

This also drops the redundant second `let timestamp = MassaTime::now();` — stale and final stats now share the single timestamp taken at the top of the finalisation block, a few microseconds earlier, with no observable effect on the stats window.

### 4. `storage.clone()` nit — not applied, here's why (@damip)

Moving the clone into the error branch isn't possible without changing the trait:

```rust
// massa-protocol-exports/src/controller_trait.rs:30
fn integrated_block(&self, block_id: BlockId, storage: Storage) -> Result<(), ProtocolError>;
```

`storage` is taken **by value** and the impl drops it on failure — `try_send(...).map_err(|_| ProtocolError::ChannelError(...))` discards the `TrySendError` that still owns the command, so the `Storage` never comes back. To re-insert into `to_propagate` on error we must still own a handle at call time, hence the clone before the call.

Leaving it, but flagging honestly that it's a bit more than negligible: `Storage::clone` is not a bare refcount bump, it takes write locks on the op/block/endorsement owner maps (`massa-storage/src/lib.rs:65`). The fixes would be to take `&Storage` in the trait, or to return the storage back on error — both ripple through every impl and the mocks, so I'd rather do that as a follow-up than widen this PR.

### 5. "If this block always fails, will it be dropped at some point?" (@bilboquet)

No, and that's intentional — @damip's read is confirmed by the source. The only error `integrated_block` can produce is `ProtocolError::ChannelError` (channel full or closed); it carries no information about *this particular block*, so there is no per-item failure to expire on.

- **Channel full** → protocol is saturated; blocks stay in `to_propagate` and are retried at the next `block_db_changed` (next command or slot tick). Everything flushes once protocol drains.
- **Channel closed** → the block handler is gone, i.e. the node is shutting down. Nothing left to retry against, and dropping wouldn't help.

Dropping on repeated failure would be strictly worse: the likeliest victim is our own freshly produced block, which is exactly what F30 is about. On growth, `to_propagate` is a `PreHashMap<BlockId, Storage>`, so re-inserting the same block is idempotent and `Storage` is a handle, not a copy of the block. The new `warn!` makes a lasting saturation visible instead of silent.

massa-consensus-worker/src/state/process.rs
```diff
@@ -780,7 +808,9 @@ impl ConsensusState {
         self.new_final_blocks.clear();
         self.new_stale_blocks.clear();
         self.final_block_stats.extend(final_block_stats);
-        self.stale_block_stats.extend(stale_block_stats);
+        // add stale blocks to stats
+        self.stale_block_stats
+            .extend((0..stale_block_count).map(|_| timestamp));
```
Leo-Besancon: I saw your reasoning (`3. Stale-block stats simplified — with one correction`), but reordering this call above ` self.new_stale_blocks.clear();` would work and avoid the binding entirely, no?

bilboquet: apart from that (minor) point it looks good to me.

peterjah: suggestion implemented!

Leo-Besancon (approved):
Thanks, LGTM

Issue #4988, opened 2026-07-21
[F25] Premature Future-Slot Header Processing Enables Multistake Poisoning and Denunciation Loss


damip:
> ⚠️ **Warning — This finding was routed to an issue rather than a PR because it touches consensus / block-graph processing or block/operation validity rules.** Changing what a node accepts, rejects, or how it advances state can make patched and unpatched nodes diverge (a network split), so it requires careful design and a coordinated, version-gated (MIP / hard-fork) rollout rather than a self-contained PR.

---

## Source
AI security-scan **Finding 25** (severity: **medium**). Tracked in `SUBMISSIONS.md` as F25.

Corresponds to external bug-bounty PDF #7 (far-future block headers / OOM).

**Affected file:** `massa-consensus-worker/src/state/process_commands.rs`

## Summary
`register_block_header` in `massa-main/massa-consensus-worker/src/state/process_commands.rs` processes block headers for future slots before the selector data for that slot is available. In this state, the header is not rejected because `check_header` returns `WaitForSlot`, but two side effects still happen too early: denunciation precursors are sent immediately, and multistake tracking is updated. This can both poison `nonfinal_active_blocks_per_slot` with unvalidated future-slot headers and cause `DenunciationPrecursor` data to be dropped without retry. As a result, a victim node can miss the real block for that slot and fail to emit denunciation evidence for observed equivocation.

## Root cause
The main issue starts in `massa-main/massa-consensus-worker/src/state/process_commands.rs` inside `register_block_header`. This function immediately creates a `DenunciationPrecursor` from the incoming `SecuredHeader` with `DenunciationPrecursor::from(&header)` and forwards it through `self.channels.pool_controller.add_denunciation_precursor(de_p)` before consensus knows whether the slot can be validated.

The same function also inserts the header into block state through `self.blocks_state.transition_map(...)`, storing `BlockStatus::Incoming(HeaderOrBlock::Header(header.clone()))`, and then continues processing with `self.rec_process(to_ack, current_slot)?`. This means the header enters state and downstream processing even when validation cannot yet confirm the slot producer.

[...]

PR #5261, opened 2026-09-07
[F26] fix(consensus-worker): preserve pruned future/dep-waiting block…

### LINKED ISSUE #4989: [F26] Queue overflow silently drops dependency waiters and cancels retrieval of their missing blocks

---

## Source
AI security-scan **Finding 26** (severity: **medium**). Tracked in `SUBMISSIONS.md` as F26.

**Affected file:** `massa-consensus-worker/src/state/prune.rs`

## Summary
`prune_waiting_for_dependencies` in `massa-consensus-worker/src/state/prune.rs` can evict over-limit `BlockStatus::WaitingForDependencies` entries with `to_discard.insert(hash, None)`. In the final transition, those entries are removed from `blocks_state` instead of being preserved as `BlockStatus::Discarded`, which can temporarily blackhole valid headers that were already recorded in protocol-side `checked_headers`. A narrower crash condition also remains: if protocol later issues `mark_invalid_block` for a stale wishlist entry that consensus has already dropped, `transition_map` can panic on a `None -> Discarded` transition. The original report overstated this as a general replay or malformed-block driven denial of service, and silent drop is not unique to dependency pruning because `prune_slot_waiting` uses the same `None`-drop pattern.

## Root cause
The confirmed root cause is the size-limiting branch in `prune_waiting_for_dependencies`, where an excess entry is selected and added to `to_discard` with `None` rather than `Some(DiscardReason)`. During the final `transition_map` pass, a `None` reason returns `None`, so the `WaitingForDependencies` entry is deleted from `self.blocks_state` instead of being retained as a durable `BlockStatus::Discarded` record.

This becomes observable because `note_header_from_peer` records verified headers in protocol-side `checked_headers`, while `on_block_header_received` only forwards a header back to consensus when it is considered new. As a result, a valid header that was silently dropped from consensus can be ignored on reannouncement until the protocol LRU entry expires.

The panic theory is valid only in the narrower stale-invalidation path. `mark_invalid_block` can request a `Discarded` transition for a block that consensus has already removed, and `transition_map` rejects that `None -> Discarded` transition with a panic. The broader claim that any later malformed block data generally triggers this crash is not supported by the validated mechanism.

## Affected code
```rust
fn prune_waiting_for_dependencies(&mut self) -> Result<(), ConsensusError> {
        let mut to_discard: PreHashMap<BlockId, Option<DiscardReason>> = PreHashMap::default();
        let mut to_keep: PreHashMap<BlockId, (u64, Slot)> = PreHashMap::default();

        // list items that are older than the latest final blocks in their threads or have deps that are discarded
        {
            for block_id in self.blocks_state.waiting_for_dependencies_blocks().iter() {
                if let Some(BlockStatus::WaitingForDependencies {
                    header_or_block,
                    unsatisfied_dependencies,
                    sequence_number,
                }) = self.blocks_state.get(block_id)
                {
                    // has already discarded dependencies => discard (choose worst reason)
                    let mut discard_reason = None;
                    let mut discarded_dep_found = false;
                    for dep in unsatisfied_dependencies.iter() {
                        if let Some(BlockStatus::Discarded { reason, .. }) =
                            self.blocks_state.get(dep)
                        {
                            discarded_dep_found = true;
                            match reason {
                                DiscardReason::Invalid(reason) => {
                                    discard_reason = Some(DiscardReason::Invalid(format!("discarded because depend on block:{} that has discard reason:{}", block_id, reason)));
                                    break;
                                }
                                DiscardReason::Stale => discard_reason = Some(DiscardReason::Stale),
                                DiscardReason::Final => discard_reason = Some(DiscardReason::Stale),
                            }
                        }
                    }
                    if discarded_dep_found {
                        to_discard.insert(*block_id, discard_reason);
                        continue;
                    }

                    // is at least as old as the latest final block in its thread => discard as stale
                    let slot = header_or_block.get_slot();
                    if slot.period <= self.latest_final_blocks_periods[slot.thread as usize].1 {
                        to_discard.insert(*block_id, Some(DiscardReason::Stale));
                        continue;
                    }

                    // otherwise, mark as to_keep
                    to_keep.insert(*block_id, (*sequence_number, header_or_block.get_slot()));
// … (truncated)
```

## Recommendation
When pruning `BlockStatus::WaitingForDependencies` entries, do not delete them from `blocks_state` with a `None` discard reason. Instead, preserve evicted entries as `BlockStatus::Discarded` with an explicit discard reason so consensus retains a durable record of the block and avoids temporarily blackholing valid headers that are still present in protocol-side `checked_headers`.

Apply the same principle consistently to all pruning paths that currently use the `None`-drop pattern, including `prune_slot_waiting`, so over-limit entries are discarded explicitly rather than silently removed. This keeps consensus and protocol state aligned and prevents valid reannouncements from being ignored until protocol cache expiration.

Harden the invalidation path as well: `mark_invalid_block` and the associated transition logic should handle blocks that are no longer present in `blocks_state` without panicking. Missing entries should be treated as a no-op or converted into a safe discard flow, rather than attempting an invalid `None -> Discarded` transition.

## Why P0 / breaking

peterjah:
…s as Discarded and harden mark_invalid_block

  * [ ] make tests compile
  * [ ] make tests pass 

### DIFF
--- a/massa-consensus-worker/src/state/process_commands.rs
+++ b/massa-consensus-worker/src/state/process_commands.rs
@@ -148,14 +148,19 @@ impl ConsensusState {
         self.maybe_note_attack_attempt(&reason, block_id);
         massa_trace!("consensus.block_graph.process.invalid_block", {"block_id": block_id, "reason": reason});
         let sequence_number = self.blocks_state.sequence_counter();
-        self.blocks_state.transition_map(block_id, |_, _| {
-            Some(BlockStatus::Discarded {
-                slot: header.content.slot,
-                creator: header.content_creator_address,
-                parents: header.content.parents,
-                reason,
-                sequence_number,
-            })
-        });
+        self.blocks_state
+            .transition_map(block_id, |block_status, _| {
+                // If the block has already been pruned from consensus, there is nothing
+                // left to mark invalid. Avoid a None -> Discarded panic by treating it as
+                // a no-op; the attack attempt was already recorded above.
+                block_status.as_ref()?;
+                Some(BlockStatus::Discarded {
+                    slot: header.content.slot,
+                    creator: header.content_creator_address,
+                    parents: header.content.parents,
+                    reason,
+                    sequence_number,
+                })
+            });
     }
 }
--- a/massa-consensus-worker/src/state/process_commands.rs
+++ b/massa-consensus-worker/src/state/process_commands.rs
@@ -150,9 +150,11 @@ impl ConsensusState {
         let sequence_number = self.blocks_state.sequence_counter();
         self.blocks_state
             .transition_map(block_id, |block_status, _| {
-                // If the block has already been pruned from consensus, there is nothing
-                // left to mark invalid. Avoid a None -> Discarded panic by treating it as
-                // a no-op; the attack attempt was already recorded above.
+                // The block may be unknown to consensus: a wishlisted parent is never inserted in
+                // `blocks_state` until its full block arrives, but protocol can decide it is invalid
+                // before that (committed ops exceed the max block size). `None -> Discarded` is a
+                // forbidden transition (panic), so do nothing: the attack attempt is already noted
+                // above, and the waiters depending on it will be pruned as stale.
                 block_status.as_ref()?;
                 Some(BlockStatus::Discarded {
                     slot: header.content.slot,
--- a/massa-consensus-worker/src/state/prune.rs
+++ b/massa-consensus-worker/src/state/prune.rs
@@ -104,7 +104,8 @@ impl ConsensusState {
     }
 
     // Keep only a certain (`config.max_future_processing_blocks`) number of blocks that have slots in the future
-    // to avoid high memory consumption
+    // to avoid high memory consumption. Over-limit entries are preserved as `Discarded` rather than
+    // silently removed, so protocol-side `checked_headers` stays aligned with consensus state.
     fn prune_slot_waiting(&mut self) {
         if self.blocks_state.waiting_for_slot_blocks().len()
             <= self.config.max_future_processing_blocks
@@ -128,7 +129,32 @@ impl ConsensusState {
         let len_slot_waiting = slot_waiting.len();
         (self.config.max_future_processing_blocks..len_slot_waiting).for_each(|idx| {
             let (_slot, block_id) = &slot_waiting[idx];
-            self.blocks_state.transition_map(block_id, |_, _| None);
+            let sequence_number = self.blocks_state.sequence_counter();
+            self.blocks_state
+                .transition_map(block_id, |block_status, _| {
+                    if let Some(BlockStatus::WaitingForSlot(header_or_block)) = block_status {
+                        let header = match header_or_block {
+                            HeaderOrBlock::Header(h) => h,
+                            HeaderOrBlock::Block { id, .. } => self
+                                .storage
+                                .read_blocks()
+                                .get(&id)
+                                .unwrap_or_else(|| panic!("block {} should be in storage", id))
+                                .content
+                                .header
+                                .clone(),
+                        };
+                        Some(BlockStatus::Discarded {
+                            slot: header.content.slot,
+                            creator: header.content_creator_address,
+                            parents: header.content.parents,
+                            reason: DiscardReason::Stale,
+                            sequence_number,
+                        })
+                    } else {
+                        panic!("block {} should be in WaitingForSlot state", block_id);
+                    }
+                });
         });
     }
 
@@ -267,7 +293,8 @@ impl ConsensusState {
                     .min();
                 if let Some((_seq_num, _slot, hash)) = remove_elt {
                     to_keep.remove(&hash);
-                    to_discard.insert(hash, None);
+                    // Preserve the over-limit entry as discarded instead of silently deleting it.
+                    to_discard.insert(hash, Some(DiscardReason::Stale));
                     continue;
                 }
             }
--- a/massa-consensus-worker/src/state/prune.rs
+++ b/massa-consensus-worker/src/state/prune.rs
@@ -189,7 +189,7 @@ impl ConsensusState {
     }
 
     fn prune_waiting_for_dependencies(&mut self) -> Result<(), ConsensusError> {
-        let mut to_discard: PreHashMap<BlockId, Option<DiscardReason>> = PreHashMap::default();
+        let mut to_discard: PreHashMap<BlockId, DiscardReason> = PreHashMap::default();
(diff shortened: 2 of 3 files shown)


massa-consensus-worker/src/state/prune.rs
Leo-Besancon: In my understanding the `else` branch here becomes unreachable with this PR? If so, I think we can simplify this function a bit, and add the invariant that we should always have a reason? WDYT?

Leo-Besancon (commented):
Same as the other consensus-related PR, I'm not doing deep reviews. The implementation looks fine to me.

massa-consensus-worker/src/state/prune.rs
bilboquet: I second that.

bilboquet:
That looks good to me.
As Léo asked Damir to review, let's wait a bit. 
If Damir has to much to do, I think that we could approve this PR.

massa-consensus-worker/src/state/prune.rs
peterjah: i pushed i refacto commit about that

Leo-Besancon: Thanks, this refactor LGTM, resolving

Leo-Besancon (dismissed):
LGTM, waiting on other reviews (ping @damip) before merging

massa-consensus-worker/src/state/process_commands.rs
```diff
@@ -162,14 +162,19 @@ impl ConsensusState {
...
-                reason,
-                sequence_number,
-            })
-        });
+        self.blocks_state
+            .transition_map(block_id, |block_status, _| {
+                // If the block has already been pruned from consensus, there is nothing
+                // left to mark invalid. Avoid a None -> Discarded panic by treating it as
+                // a no-op; the attack attempt was already recorded above.
+                block_status.as_ref()?;
```
damip, line 158: Right fix. Two asks: a test (protocol marks invalid a block consensus never had, no panic), and a comment that says the real trigger, it is not a pruning race:

```rust
// The block may be unknown to consensus: a wishlisted parent is never inserted in
// `blocks_state` until its full block arrives, but protocol can decide it is invalid
// before that (committed ops exceed the max block size). `None -> Discarded` is a
// forbidden transition (panic), so do nothing: the attack attempt is already noted
// above, and the waiters depending on it will be pruned as stale.
```

massa-consensus-worker/src/state/prune.rs
```diff
@@ -128,7 +129,32 @@ impl ConsensusState {
         let len_slot_waiting = slot_waiting.len();
         (self.config.max_future_processing_blocks..len_slot_waiting).for_each(|idx| {
             let (_slot, block_id) = &slot_waiting[idx];
-            self.blocks_state.transition_map(block_id, |_, _| None);
+            let sequence_number = self.blocks_state.sequence_counter();
```
damip: Revert to `transition_map(block_id, |_, _| None)`. Evicted entries here are almost only far-future garbage (see #4988). Recording `Discarded(Stale)` for garbage pushes the real `Invalid` verdicts out of the `max_discarded_blocks = 100` set and counts attacker garbage in `stale_block_stats`. And for the rare legit header evicted here, `Discarded(Stale)` means its whole thread is discarded when it arrives for real, while `None` means it is refetched one slot later. Please add:

```rust
// Eviction for size is NOT a verdict on the block: forget it (None), do not mark it
// Discarded. Discarded is terminal (never reprocessed) and inherited by every child
// in check_header. A forgotten block is refetched through the wishlist as soon as a
// child references it, which is the recovery path we want.
```

massa-consensus-worker/src/state/prune.rs
```diff
@@ -267,7 +294,8 @@ impl ConsensusState {
                     .min();
                 if let Some((_seq_num, _slot, hash)) = remove_elt {
                     to_keep.remove(&hash);
-                    to_discard.insert(hash, None);
+                    // Preserve the over-limit entry as discarded instead of silently deleting it.
+                    to_discard.insert(hash, DiscardReason::Stale);
```
damip: Same thing: revert to `None` and keep `Option<DiscardReason>`. This eviction is exactly the case where we know nothing bad about the block, we just have too many. With `Stale` here, the cascade above ("has dependencies that will be discarded") now stamps `Stale` on every waiter built on it too, and every block arriving afterwards that references any of them is discarded on arrival. With `None` they are forgotten and come back through the wishlist. Same comment as in `prune_slot_waiting` please. In practice this path only triggers when the node is already beyond consensus catch-up and headed to `NeedSync`, so no drama, but there is no reason to add a poison mechanism for nothing.

damip (dismissed):
Two unrelated things in here, they get different answers.

1. `mark_invalid_block` no-op when consensus has no entry: correct, keep it, and it matters more than the issue says. This is not a race on a stale wishlist entry, it is the normal missing-parent flow. Consensus wishlists a parent `P` it has never seen (`P -> None` in `get_block_wishlist`, nothing in `blocks_state`). Protocol fetches `P`'s header and keeps it to itself (wishlisted headers are not registered to consensus), then the op list, then the ops. If the committed ops exceed the max block size, `gather_missing_block_ops` calls `mark_block_as_invalid(P)` -> `mark_invalid_block(P)` -> `transition_map` on `None` -> `panic!("Invalid transition from None to Discarded")`. Anyone can sign `P` (no draw check on a zero-endorsement header) and put whatever they want in its op list. A staker needs one slot to send a header with `P` as parent to each node, and every node that got it dies. Please say that in the PR description and add a test: mark invalid a block consensus never had, no panic.

2. Turning size evictions into `Discarded(Stale)`: please revert, both in `prune_slot_waiting` and in `prune_waiting_for_dependencies`, and put back `Option<DiscardReason>`. Details inline. Short version: `Discarded` is a verdict, it is terminal and it is inherited by every child (`check_header` discards a block whose parent is `Discarded`). A size eviction is not a verdict. The "blackhole" from the issue is a one-slot delay: the forgotten block is refetched through the wishlist as soon as a child references it. With `Discarded(Stale)` the child is discarded instead, so the change removes the recovery path it was supposed to protect. No upside.

To be clear on severity: in practice eviction only happens in states where the node is already lost (over 2048 waiters means 32+ periods of backlog, consensus catch-up is dead after `force_keep_final_periods` = 5 anyway, the node ends up in `NeedSync` either way). So this is not a "will desync nodes" objection, it is a "wrong direction, zero benefit, don't" objection.

peterjah:
@damip @Leo-Besancon @bilboquet 
Suggested comments and tests added

Leo-Besancon (approved):
LGTM

bilboquet (approved):
LGTM

PR #5262, opened 2026-09-07
[F25] fix(consensus-worker): defer denunciation and multistake side e…

### LINKED ISSUE #4988: [F25] Premature Future-Slot Header Processing Enables Multistake Poisoning and Denunciation Loss

peterjah:
…ffects for unverifiable slots

## What

A block header for a slot the selector cannot draw yet reaches `check_header`'s `WaitForSlot` path via `get_producer() -> Err` (`massa-consensus-worker/src/state/verifications.rs:217`), **before** the creator-turn check. Nothing is validated on that path except the signature — not even the producer. We nevertheless inserted the header into `nonfinal_active_blocks_per_slot` and forwarded a denunciation precursor for it.

## Attack

Anyone, with no stake, signs 2 junk headers for slot S with throwaway keys and sends them ~2 cycles ahead. Both are counted in the multistake index for S. When S arrives, the legitimately drawn producer's block is rejected as `more than 2 blocks for slot S` on every node that received the junk. Repeat per slot: the attacker suppresses blocks for free. The junk precursors also burn the pool's per-slot denunciation cache entry, so real equivocation at S goes undenounced.

## Fix

No side effects of any kind while a slot is unverifiable: no precursor, no multistake index entry, no attack note. `WaitingForSlot` only stores the header, already capped at 400 by `prune_slot_waiting`, which evicts the furthest slots first. Everything is redone when the slot is reprocessed and the producer can actually be checked.

Precursor forwarding and the multistake check are factored into a single `note_verifiable_header` helper so the header path (`convert_block_header`) and the full-block path (`process.rs`) cannot drift apart again. Order is precursor first, then multistake: a 3rd equivocating block is exactly what we want the pool to see, and the pool re-checks the PoS draw, expiry and `last_start_period` itself, so our local per-slot index must not gate what it receives.

## Why no MIP

Node-local only. Block validity rules, header/block serialization and hashing, the final-state hash and execution semantics are untouched. This changes when *this* node updates its own RAM indexes and what it feeds its own denunciation pool; blocks produced and validated are byte-identical, and old/new nodes agree on the same chain. No JSON-RPC or gRPC surface is affected.

## Bonus

Forwarding precursors at validation time rather than at reception closes a slashing-evasion window: the selector horizon is 128..256 periods while the pool's `denunciation_expire_periods` bound is 128, so equivocation submitted far enough ahead could previously be dropped by the pool before it was ever denounceable.

## Tests

| Test | Covers |
| --- | --- |
| `test_future_header_side_effects_deferred` | Precursor forwarded exactly once, when the slot becomes verifiable |
| `test_undrawable_slot_header_has_no_side_effects` | `get_producer -> Err`: header queued, precursor count stays 0 |
| `test_undrawable_slot_headers_do_not_poison_multistake` | 2 forged headers on an undrawable slot must not evict the legit block |
| `test_multistake_still_forwards_denunciation_precursor` | The 3rd equivocating block still reaches the pool before being dropped |

Each new test was verified to fail against the pre-fix behavior. The poisoning regression fails with exactly the reported symptom:

```
legit block for the poisoned slot was not accepted (status: Discarded, forged: [Discarded, Discarded])
```

### DIFF
--- a/massa-consensus-worker/src/state/process.rs
+++ b/massa-consensus-worker/src/state/process.rs
@@ -15,6 +15,7 @@ use massa_models::{
     block_header::SecuredHeader,
     block_id::BlockId,
     clique::Clique,
+    denunciation::DenunciationPrecursor,
     prehash::{PreHashMap, PreHashSet},
     slot::Slot,
     timeslots,
@@ -171,7 +172,16 @@ impl ConsensusState {
                             HeaderCheckOutcome::Discard(reason) => {
                                 self.maybe_note_attack_attempt(reason, &block_id)
                             }
+                            HeaderCheckOutcome::WaitForSlot => {
+                                // The slot is not yet verifiable: defer denunciation and
+                                // multistake side effects until the block is reprocessed.
+                            }
                             _ => {
+                                // Slot is verifiable: forward denunciation precursor and apply
+                                // the multistake limit so only validated blocks compete.
+                                self.channels.pool_controller.add_denunciation_precursor(
+                                    DenunciationPrecursor::from(&stored_block.content.header),
+                                );
                                 // Extra equivocation blocks must leave Incoming so Storage is
                                 // not retained indefinitely (Incoming is not pruned / slot-ticked).
                                 if self.detect_multistake(&stored_block.content.header) {
--- a/massa-consensus-worker/src/state/process.rs
+++ b/massa-consensus-worker/src/state/process.rs
@@ -180,14 +180,11 @@ impl ConsensusState {
                                 // multistake side effects until the block is reprocessed.
                             }
                             _ => {
-                                // Slot is verifiable: forward denunciation precursor and apply
-                                // the multistake limit so only validated blocks compete.
-                                self.channels.pool_controller.add_denunciation_precursor(
-                                    DenunciationPrecursor::from(&stored_block.content.header),
-                                );
+                                // Slot is verifiable: forward the denunciation precursor and
+                                // apply the multistake limit so only validated blocks compete.
                                 // Extra equivocation blocks must leave Incoming so Storage is
                                 // not retained indefinitely (Incoming is not pruned / slot-ticked).
-                                if self.detect_multistake(&stored_block.content.header) {
+                                if self.note_verifiable_header(&stored_block.content.header) {
                                     res = HeaderCheckOutcome::Discard(DiscardReason::Invalid(
                                         format!(
                                             "more than 2 blocks for slot {}",
--- a/massa-consensus-worker/src/state/process.rs
+++ b/massa-consensus-worker/src/state/process.rs
@@ -176,8 +176,11 @@ impl ConsensusState {
                                 );
                             }
                             HeaderCheckOutcome::WaitForSlot => {
-                                // The slot is not yet verifiable: defer denunciation and
-                                // multistake side effects until the block is reprocessed.
+                                // The slot is not verifiable yet: either the slot is simply
+                                // not reached, or the selector has no draw for it and not even
+                                // the producer was checked. No side effects on this path, they
+                                // are all redone when the slot is reprocessed. See
+                                // `convert_block_header` for the full rationale.
                             }
                             _ => {
                                 // Slot is verifiable: forward the denunciation precursor and
--- a/massa-consensus-worker/src/state/process.rs
+++ b/massa-consensus-worker/src/state/process.rs
@@ -170,7 +170,10 @@ impl ConsensusState {
                         );
                         match &res {
                             HeaderCheckOutcome::Discard(reason) => {
-                                self.maybe_note_attack_attempt(reason, &block_id)
+                                self.maybe_note_attack_attempt(reason, &block_id);
+                                self.channels.pool_controller.add_denunciation_precursor(
+                                    DenunciationPrecursor::from(&stored_block.content.header),
+                                );
                             }
                             HeaderCheckOutcome::WaitForSlot => {
                                 // The slot is not yet verifiable: defer denunciation and
--- a/massa-consensus-worker/src/state/verifications.rs
+++ b/massa-consensus-worker/src/state/verifications.rs
@@ -124,8 +124,13 @@ impl ConsensusState {
                 })
             }
             HeaderCheckOutcome::WaitForSlot => {
-                // The slot is not yet verifiable: defer denunciation and
-                // multistake side effects until the block is reprocessed.
+                // WaitForSlot covers two cases:
+                // 1. producer checked, slot just not reached yet (clock skew)
+                // 2. selector has no draw for this slot yet (> 2 cycles ahead): NOTHING is
+                //    validated, not even the producer. Anyone can forge these for free.
+                // => No side effects of any kind on this path: no precursor, no multistake
+                //    index, no attack note. Only store, capped by prune_slot_waiting.
+                //    Everything is redone when the slot is reprocessed.
                 Some(BlockStatus::WaitingForSlot(HeaderOrBlock::Header(header)))
             }
             HeaderCheckOutcome::Discard(reason) => {
--- a/massa-consensus-worker/src/state/verifications.rs
+++ b/massa-consensus-worker/src/state/verifications.rs
@@ -69,6 +69,22 @@ impl ConsensusState {
         false
     }
 
+    /// Handle the side effects of a header whose slot is verifiable: forward the
+    /// denunciation precursor to the pool, then apply the multi-stake limit.
+    ///
+    /// The precursor is forwarded unconditionally because the pool runs its own
+    /// eligibility checks (PoS draw, expiry, last start period) and keeps at most
+    /// one cache entry per slot: gating it on our local per-slot index could drop
+    /// evidence the pool never received.
+    ///
+    /// Returns `true` if the header is an extra equivocation block for its slot.
+    pub(crate) fn note_verifiable_header(&mut self, header: &SecuredHeader) -> bool {
+        self.channels
+            .pool_controller
+            .add_denunciation_precursor(DenunciationPrecursor::from(header));
+        self.detect_multistake(header)
+    }
+
     /// Check if the header is valid and if it could be processed when we will receive the full block
     pub(crate) fn convert_block_header(
         &mut self,
@@ -79,12 +95,9 @@ impl ConsensusState {
         let header_outcome = self.check_header(&block_id, &header, current_slot);
         match header_outcome {
             HeaderCheckOutcome::Proceed { .. } => {
-                if self.detect_multistake(&header) {
+                if self.note_verifiable_header(&header) {
                     return None;
                 }
-                self.channels
-                    .pool_controller
-                    .add_denunciation_precursor(DenunciationPrecursor::from(&header));
                 // set as waiting dependencies
                 let mut dependencies = PreHashSet::<BlockId>::default();
                 dependencies.insert(block_id); // add self as unsatisfied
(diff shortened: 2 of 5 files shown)


massa-consensus-worker/src/state/verifications.rs
```diff
@@ -107,9 +114,8 @@ impl ConsensusState {
...
             }
             HeaderCheckOutcome::WaitForSlot => {
-                if self.detect_multistake(&header) {
-                    return None;
-                }
+                // The slot is not yet verifiable: defer denunciation and
+                // multistake side effects until the block is reprocessed.
                 Some(BlockStatus::WaitingForSlot(HeaderOrBlock::Header(header)))
             }
             HeaderCheckOutcome::Discard(reason) => {
```
Leo-Besancon, line 136: I think we also need `.add_denunciation_precursor` there

massa-consensus-worker/src/state/process.rs
```diff
@@ -171,7 +172,16 @@ impl ConsensusState {
                             HeaderCheckOutcome::Discard(reason) => {
                                 self.maybe_note_attack_attempt(reason, &block_id)
```
Leo-Besancon: I think we also need `.add_denunciation_precursor` there

Leo-Besancon (commented):
The changes in this PR look ok to me aside from a few comments below.

As it changes the timings a bit I would indeed like additional reviews if some subtle assumptions are broken in any ways with these changes (but I don't see any).

Additional note: in `massa-consensus-worker/src/mod.rs`, the doc comment for `nonfinal_active_blocks_per_slot` still says:
```rust
    /// Blocks indexed by slot (used for multi-stake limiting). Blocks
    /// should be saved in this map when we receive the header or the full block directly.
```

This does not reflect the timings after this PR.

peterjah:
@Leo-Besancon  Good catches. I updated the nonfinal_active_blocks_per_slot doc comment to match the new deferred-insertion timing, and restored denunciation precursor forwarding on the Discard path for both headers and full blocks. Only WaitForSlot now defers side effects.

Leo-Besancon (dismissed):
LGTM, waiting on other reviews (ping @damip) before merging

bilboquet:
Looks good to me.
That being said, on the formal aspect, I'm seeing 2 different schemes:

- process.rs (full blocks, _ branch):
```rs
add_denunciation_precursor(...);          // precursor first
if detect_multistake(...) { res = Discard(...) }
```
- verifications.rs (headers, Proceed / WaitForDependencies):
```rs
if detect_multistake(&header) { return None; }    // multistake first
add_denunciation_precursor(...);
```

So one path does detect_multistake -> add_precursor and the other does add_precursor -> detect_multistake. Maybe I'm missing something, but it looks like we could do detect_multistake -> add_precursor everywhere (i.e. don't forward a precursor when the block is an extra block being discarded). 

That would let us factorize this into a single helper and make the two paths homogeneous. WDYT (@peterjah @Leo-Besancon @damip ?

massa-consensus-worker/src/state/verifications.rs
```diff
@@ -107,12 +114,14 @@ impl ConsensusState {
                 })
             }
             HeaderCheckOutcome::WaitForSlot => {
-                if self.detect_multistake(&header) {
-                    return None;
-                }
+                // The slot is not yet verifiable: defer denunciation and
+                // multistake side effects until the block is reprocessed.
```
damip: The dangerous case is not really "slot in the future", it is `get_producer` returning `Err` in `check_header` (the `// TODO properly handle PoS errors` line). That is the path where NOTHING is validated. Please make it explicit here, something like:

```rust
// WaitForSlot covers two cases:
// 1. producer checked, slot just not reached yet (clock skew)
// 2. selector has no draw for this slot yet (> 2 cycles ahead): NOTHING is
//    validated, not even the producer. Anyone can forge these for free.
// => No side effects of any kind on this path: no precursor, no multistake
//    index, no attack note. Only store, capped by prune_slot_waiting.
//    Everything is redone when the slot is reprocessed.
```

massa-consensus-worker/src/state/process.rs
```diff
@@ -169,9 +170,21 @@ impl ConsensusState {
...
+                                );
+                            }
+                            HeaderCheckOutcome::WaitForSlot => {
+                                // The slot is not yet verifiable: defer denunciation and
+                                // multistake side effects until the block is reprocessed.
                             }
                             _ => {
+                                // Slot is verifiable: forward denunciation precursor and apply
+                                // the multistake limit so only validated blocks compete.
+                                self.channels.pool_controller.add_denunciation_precursor(
```
damip: Here it is precursor then multistake, in `convert_block_header` it is multistake then precursor. Pick one (precursor first is better: a 3rd equivocating block is exactly what we want the pool to see) and factor the two `add_denunciation_precursor` + `detect_multistake` sequences into one helper so the two paths can not drift again. This is what @bilboquet asked for too.

massa-consensus-worker/src/tests/scenarios.rs
```diff
@@ -530,3 +534,121 @@ fn test_slot_maintenance_not_starved_by_command_flood() {
...
+    foreign_controllers
+        .pool_controller
+        .expect_add_denunciation_precursor()
+        .returning(move |_| {
+            precursor_count_clone.fetch_add(1, Ordering::SeqCst);
+        });
+    foreign_controllers
+        .selector_controller
+        .expect_get_producer()
+        .returning(move |_| Ok(staking_address));
```
damip, line 583: This mocks the selector to always answer, so the test only covers "producer checked, slot not reached yet". The actual attack is the `get_producer` -> `Err` case. Please add:

1. `expect_get_producer` returns `Err(PosError::CycleUnavailable(..))` for the future slot: header must end up `WaitingForSlot` and the precursor count must stay 0.
2. Regression for the actual poisoning: 2 headers for slot S signed by a random key while the selector says `Err`, then the selector starts answering and a legit block for S from the drawn producer is registered at slot time. It must be accepted (check `get_block_statuses`). On `main` it is dropped as "more than 2 blocks for slot".

damip (commented):
Fix is right and this is the most urgent PR of the batch (details on #4988).

Simple version of the bug: a header for a slot beyond the selector's horizon is `WaitForSlot` with NOTHING validated, and we still counted it in the multistake index. So 2 garbage headers per slot, from anyone, sent 2 cycles ahead, kill the legit block of that slot on every node that received them. Fix = no side effects until the slot is actually verifiable. Right shape: no new queue, no blocking, just moving the side effects after validation.

Checked the consequences:
- `nonfinal_active_blocks_per_slot` has only `detect_multistake` and prune as readers, so populating it later changes nothing else.
- `WaitingForSlot` is still capped at 400 by `prune_slot_waiting`, which evicts the furthest slots first. A validated producer flooding headers for its own future slot only evicts its own entries. At slot time 2 survive and it gets denounced. Fine.
- Header then full block sends the precursor twice, the pool dedups. Same as before.
- Discard now forwards a precursor: the pool applies expiry / creator check. Negligible traffic.
- Bonus: precursors at validation time close a slashing evasion window (selector horizon is 128..256 periods, pool bound is 128). Worth a line in the description.

Not approving yet: 3 things inline, and the PR description is the empty template. Please write 5 lines on what the attack was and why it is node-local (no MIP). It matters for the release notes.

peterjah:
@bilboquet @Leo-Besancon @damip 

Included proposed refacto and all improvements suggested in review

bilboquet (approved):
LGTM

Leo-Besancon (approved):

PR #5264, opened 2026-09-07
[F21] docs: rework list_required_active_blocks doc comment

### LINKED ISSUE #4984: [F21] Missing Parent Dependencies in `list_required_active_blocks` Output

---

## Source
AI security-scan **Finding 21** (severity: **info**). Tracked in `SUBMISSIONS.md` as F21.

**Affected file:** `massa-consensus-worker/src/state/mod.rs`

## Summary
`list_required_active_blocks` in `massa-consensus-worker/src/state/mod.rs` does not guarantee that the returned `kept_blocks` set is transitively closed over parent relationships. Because the function ends each of its two expansion rounds with `add_active_blocks_after`, blocks introduced in the second pass can be returned without all of their older cross-thread parents being revisited and added. The original claim that this necessarily causes bootstrap synchronization to fail or stall is not supported by the validated code path.

## Root cause
The confirmed issue is an algorithmic ordering flaw in `list_required_active_blocks`. After initializing `kept_blocks`, the function performs exactly two rounds where it first adds direct parents of the current set, then computes `earliest_blocks` with `list_earliest_blocks_of`, and finally calls `add_active_blocks_after`.

If the second and final call to `add_active_blocks_after` inserts an older active block whose parent lies in another thread and falls before that thread's computed lower bound, that parent is not added in the same round. Because there is no subsequent parent-expansion step after the final `add_active_blocks_after`, the returned set can contain a block while omitting one of its transitive parent dependencies.

The ruled-out portion of the original theory is the assumption that this incomplete parent closure directly breaks bootstrap processing. In the actual bootstrap path, `get_bootstrap_part` exports only `is_final` blocks, missing old parent references are tolerated during initialization, and later header checks do not require exhaustive traversal of such old cross-thread ancestry.

## Affected code
```rust
pub fn list_required_active_blocks(
        &self,
        end_slot: Option<Slot>,
    ) -> Result<PreHashSet<BlockId>, ConsensusError> {
        // if an end_slot is provided compute the latest final block for that given slot
        // if not use the latest_final_blocks_periods
        let effective_latest_finals: Vec<(BlockId, u64)> = if let Some(slot) = end_slot {
            self.list_latest_final_blocks_at(slot)?
        } else {
            self.latest_final_blocks_periods.clone()
        };

        // init kept_blocks using effective_latest_finals
        let mut kept_blocks: PreHashSet<BlockId> = effective_latest_finals
            .iter()
            .map(|(id, _period)| *id)
            .collect();

        // add all the active blocks that are after the effective_latest_finals of their thread
        self.add_active_blocks_after(&mut kept_blocks, &effective_latest_finals, end_slot);

        // do the following 2 times
        for _ in 0..2 {
            // extend kept_blocks with the parents of the current kept_blocks
            let mut cumulated_parents: PreHashSet<BlockId> = PreHashSet::default();
            for id in kept_blocks.iter() {
                let parents = self
                    .try_get_full_active_block(id)?
                    .0
                    .parents
                    .iter()
                    .map(|(id, _period)| *id);
                cumulated_parents.extend(parents);
            }
            kept_blocks.extend(cumulated_parents);
            // add all the active blocks whose slot is after the earliest kept_blocks of their thread
            let earliest_blocks = self.list_earliest_blocks_of(&kept_blocks, end_slot)?;
            self.add_active_blocks_after(&mut kept_blocks, &earliest_blocks, end_slot);
        }

        // check that we have the full blocks for every id we are about to return
        for id in kept_blocks.iter() {
            self.try_get_full_active_block(id)?;
        }

// … (truncated)
```

## Recommendation
Ensure that `list_required_active_blocks` returns a set that is transitively closed over parent relationships. In particular, avoid ending the procedure immediately after `add_active_blocks_after`, as blocks introduced during the final expansion can still have older cross-thread parents that were never revisited.

Restructure the algorithm so that parent expansion is performed until a fixpoint is reached, or add a final closure step that recursively includes all missing ancestors of every retained block before returning. If bounded execution is required, document and enforce the intended completeness guarantees explicitly, and ensure callers do not rely on the output being fully parent-complete unless that property is guaranteed.

## Why P0 / breaking

peterjah:
Note: read the discussion below. The only fix kept in this PR is a doc comment rewrite.

### DIFF
--- a/massa-consensus-worker/src/state/mod.rs
+++ b/massa-consensus-worker/src/state/mod.rs
@@ -315,38 +315,26 @@ impl ConsensusState {
         }
     }
 
-    /// Lists active block IDs that must be kept to preserve graph consistency
-    /// (latest finals, their descendants up to `end_slot`, and parents needed to fill holes).
+    /// Lists the active block IDs that must be kept: the latest finals of each thread, all active
+    /// blocks after them, and exactly 2 rounds of "add parents, then fill holes down to the
+    /// earliest kept block of each thread".
     ///
-    /// # Arguments
-    /// * `end_slot`: optional inclusive upper bound on block slots considered.
-    ///   - `None`: use `self.latest_final_blocks_periods` as the per-thread finals baseline
-    ///     (e.g. graph pruning).
-    ///   - `Some(slot)`: compute the latest final block in each thread with `block.slot <= slot`
-    ///     (e.g. bootstrap, aligned with the final-state cursor).
-    ///
-    /// # Invariants for `Some(end_slot)`
-    /// Slot ordering is `(period, thread)`. Genesis blocks live at
-    /// `Slot(last_start_period, thread)` for every thread, so a global upper bound only admits a
-    /// latest final in *every* thread when:
-    /// `end_slot >= Slot(last_start_period, thread_count - 1)` (the last genesis slot).
-    /// Callers must reject earlier values; otherwise `list_latest_final_blocks_at` cannot find a
-    /// final in higher threads (e.g. `Some(Slot(0, 0))` misses genesis at `(0, thread > 0)`).
-    ///
-    /// # Algorithm
-    /// if end_slot is None:
-    ///      set effective_latest_finals to be the IDs of the self.latest_final_blocks
-    /// else
-    ///      set effective_latest_finals to be the IDs of the Active Final blocks that have the highest period in each thread but are before end_slot (included)
-    ///
-    /// create a kept_blocks list of block IDs to keep
-    /// initialize it with effective_latest_finals as well as all the active blocks that are after the effective_latest_finals of their thread (included) (but before end_slot (included) if it is Some)
-    ///
-    /// do the following 2 times:
-    ///      extend kept_blocks with the parents of the current kept_blocks
-    ///      fill holes by adding to kept_blocks all the active block IDs whose slot is after the earliest kept_blocks of their thread (included) (but before end_slot (included) if it is Some)
-    ///
-    /// return kept_blocks
+    /// This is deliberately NOT a transitive closure over parents and must not become one:
+    /// - a fixpoint would walk back one ancestor per round until it hits a block already pruned
+    ///   (`Discarded { reason: Final }`), at which point `try_get_full_active_block` errors,
+    ///   `prune()` panics and `get_bootstrap_part` fails. Tests do not catch this because test
+    ///   graphs are never pruned.
+    /// - no consumer needs a closure:
+    ///   - `prune_active` uses this as a floor and additionally keeps
+    ///     `force_keep_final_periods_without_ops` periods of finals per thread. The set is
+    ///     self-sustaining (what it kept last time is still active next time), and a block whose
+    ///     parent was pruned can never become Active (the parent comes back Stale and the child
+    ///     is discarded), so no dangling parent can appear.
+    ///   - `get_bootstrap_part` exports only the final blocks of this set. A valid block compatible
+    ///     with final F_t can only reference F_t's own parent in thread t (grandpa rule), or one
+    ///     level deeper in thread-gap corner cases. Round 1 adds the parents of the finals, round 2
+    ///     covers the deeper case. Blocks validated after bootstrap are all after `end_slot` and
+    ///     are not in this set anyway, so a closure over blocks up to `end_slot` would not help.
     pub fn list_required_active_blocks(
         &self,
         end_slot: Option<Slot>,


Leo-Besancon (commented):
I've commented the issue directly (https://github.com/massalabs/massa/issues/4984#issuecomment-5566017991), I'm not sure if looping until we have a fixed-point is the right direction or not. If it is, I think we'd need additionnal proofs/estimates of performance impact.

I'm not sure versioning would be needed, why do you think it's necessary? No consumers in the codebase query the parents where the fix would matter if I understood correctly.

That's about as deep as I can review on my side, I'll let others (ping @damip) add details and evaluate if we want this or not :)

bilboquet:
Thinking a lot about @Leo-Besancon comment, I tend to think that this issue is a false positive, i.e. it can't have any impact. So Léo a) sounds good to me.

list_required_active_blocks has exactly two consumers:
- `prune_active()`: here a larger set is counter-productive
- `get_bootstrap_part()`: I don't find any path where a BS could succeed with some missing final block (hence creating a fork).

massa-consensus-worker/src/state/mod.rs
```diff
@@ -342,7 +369,7 @@ impl ConsensusState {
     /// create a kept_blocks list of block IDs to keep
     /// initialize it with effective_latest_finals as well as all the active blocks that are after the effective_latest_finals of their thread (included) (but before end_slot (included) if it is Some)
     ///
-    /// do the following 2 times:
+    /// do the following 2 times (or until no new blocks are added once MIP-0002 is active):
```
damip: Keep only a doc fix. Proposed replacement for the whole doc block (drop the "graph consistency" claim, state the exact guarantee, and say why it must not become a closure):

```rust
/// Lists the active block IDs that must be kept: the latest finals of each thread, all active
/// blocks after them, and exactly 2 rounds of "add parents, then fill holes down to the
/// earliest kept block of each thread".
///
/// This is deliberately NOT a transitive closure over parents and must not become one:
/// - a fixpoint would walk back one ancestor per round until it hits a block already pruned
///   (`Discarded { reason: Final }`), at which point `try_get_full_active_block` errors,
///   `prune()` panics and `get_bootstrap_part` fails. Tests do not catch this because test
///   graphs are never pruned.
/// - no consumer needs a closure:
///   - `prune_active` uses this as a floor and additionally keeps
///     `force_keep_final_periods_without_ops` periods of finals per thread. The set is
///     self-sustaining (what it kept last time is still active next time), and a block whose
///     parent was pruned can never become Active (the parent comes back Stale and the child
///     is discarded), so no dangling parent can appear.
///   - `get_bootstrap_part` exports only the final blocks of this set. A valid block compatible
///     with final F_t can only reference F_t's own parent in thread t (grandpa rule), or one
///     level deeper in thread-gap corner cases. Round 1 adds the parents of the finals, round 2
///     covers the deeper case. Blocks validated after bootstrap are all after `end_slot` and
///     are not in this set anyway, so a closure over blocks up to `end_slot` would not help.
```

massa-consensus-worker/src/state/mod.rs
```diff
@@ -368,8 +395,17 @@ impl ConsensusState {
         // add all the active blocks that are after the effective_latest_finals of their thread
         self.add_active_blocks_after(&mut kept_blocks, &effective_latest_finals, end_slot);
 
-        // do the following 2 times
-        for _ in 0..2 {
+        // MIP-0002 gates the fixpoint version of the algorithm.
+        let fixpoint_active = self.is_required_active_blocks_fixpoint_active(end_slot)?;
```
damip: No MIP for this. `list_required_active_blocks` only decides what this node prunes and what it exports to a bootstrap client, and a client accepts any set. There is nothing to coordinate between nodes. This plumbing (`MipStore` in the consensus worker, wall clock in a pruning routine, new component in MIP-0002) should all go away, and the PR should target `main`.

massa-consensus-worker/src/state/mod.rs
```diff
@@ -385,6 +421,10 @@ impl ConsensusState {
             // add all the active blocks whose slot is after the earliest kept_blocks of their thread
             let earliest_blocks = self.list_earliest_blocks_of(&kept_blocks, end_slot)?;
             self.add_active_blocks_after(&mut kept_blocks, &earliest_blocks, end_slot);
+
+            if fixpoint_active && kept_blocks.len() == before_len {
```
damip: This never converges on a real node. Each round adds the own-thread parent of the earliest kept block of every thread, and that parent is by construction not in the set yet, so `len()` grows every round. The walk stops only at genesis, or when it reaches a block that `prune_active` already turned into `Discarded { reason: Final }`. On any node that has run more than ~32 periods the second case comes first: `list_earliest_blocks_of` calls `try_get_full_active_block` on that id, gets `ContainerInconsistency`, and:

- `prune()` is `.expect("Error while pruning")` in `main_loop.rs`: consensus worker panics. Every node, first prune after MIP activation.
- `get_bootstrap_part` errors: nobody can serve bootstrap anymore.

Even without the error, walking back to genesis means never pruning anything.

CI is green because the test `MipStore` has MIP-0002 `Active` but test graphs are a few blocks from genesis and never prune finals, so the walk reaches genesis and stops. Nothing tests a pruned graph.

Please remove the fixpoint entirely (see my doc comment draft above), not just gate it.

damip (dismissed):
Please don't merge this. Not a P0, not a bug, and the fix would take the network down. Details on #4984, short version here.

1. The 2-round algorithm is deliberate and sufficient. `prune_active` has the 32-period buffer on top and the set is self-sustaining. `get_bootstrap_part` only needs the parents of the finals (round 1) and the thread-gap case (round 2). Nothing consumes a transitive closure. Leo and bilboquet are right.

2. The fixpoint crashes nodes. See inline. It does not converge before hitting pruned blocks, and a pruned block in the set is a `ContainerInconsistency` error, which is a panic in `prune()`. Tests pass only because they never prune.

3. No MIP for this. It is node-local, any peer accepts either output. The `MipStore` plumbing into the consensus worker, `MassaTime::now()` in a pruning routine and the edit of MIP-0002's components all go away.

What I want instead: keep only the doc comment (draft inline), retarget to `main`, close #4984 on merge.

Leo-Besancon:
@bilboquet @modship @peterjah I've kept the inline comment above and rebased on top of main.

bilboquet (approved):

PR #5265, opened 2026-09-11
feat(api): expose max_datastore_keys_query in get_status

### LINKED ISSUE #5228: Expose max_datastore_keys_query in get_status so clients can size datastore key queries
The node caps how many datastore keys a single query may return via max_datastore_keys_query, configured independently for each API surface in massa-node/base_config/config.toml (line 46 for JSON-RPC, line 127 for gRPC, both defaulting to 500).

The value is read at massa-node/src/settings.rs:158 / :405, wired into APIConfig.max_datastore_keys_queries (massa-api-exports/src/config.rs:93) and GrpcConfig.max_datastore_keys_queries (massa-grpc/src/config.rs:144) at massa-node/src/main.rs:988 / :1296, and consumed at massa-api/src/public.rs:1029 and massa-grpc/src/public.rs:1035.

That is the entire lifecycle — the limit is never published to clients.

Problem

Clients that enumerate an address's datastore keys must paginate, and to paginate they need a page size. With the limit unpublished they have to hardcode a guess. massa-web3 does exactly this: MAX_DATASTORE_KEYS_QUERY = 500 in src/provider/constants.ts:2, used by both the JSON-RPC pagination loop (src/provider/jsonRpcProvider/jsonRpcPublicProvider.ts:131) and getMultipleAddressesDatastoreKeys (src/client/storage.ts:23).

Against a node whose operator lowered the limit, that guess exceeds the node's max and cleanup_datastore_key_range_query returns an error — so get_addresses_datastore_keys fails outright rather than returning smaller pages. Pagination doesn't degrade, it breaks.

Neither status endpoint currently carries the value: NodeStatus (massa-api-exports/src/node.rs:14) exposes config: CompactConfig, and CompactConfig (massa-models/src/config/compact_config.rs:9) holds only protocol parameters (thread_count, t0, delta_f0, roll_price, max_block_size, …). The gRPC PublicStatus mirrors the same 10 fields.

Proposal

Add the effective limit to NodeStatus, sourced from api_settings:

// massa-api-exports/src/node.rs
/// max number of datastore keys returned by a single key query
pub max_datastore_keys_query: Option<u32>,

populated in get_status from self.0.api_settings.max_datastore_keys_queries.

On NodeStatus, not CompactConfig. CompactConfig is the shared protocol-parameter struct mirrored by the gRPC PublicStatus; adding a node-local API knob there would either force a massa-proto-rs change or make CompactConfig mean different things per surface. NodeStatus already carries operator-configured, non-protocol values from api_settings — chain_id and minimal_fees (massa-api/src/public.rs:615-616) — so this follows existing precedent.

Tasks

- [ ] Add the field to NodeStatus and populate it in get_status
- [ ] Update the NodeStatus Display impl
- [ ] Update the NodeStatus schema in massa-node/base_config/openrpc.json
- [ ] Update massa-client status output
- [ ] Test asserting the reported value matches the configured one

Backward compatibility

Additive and non-breaking: the field is optional, and serde ignores unknown fields, so existing clients are unaffected. No consensus impact — this is API surface only, so no MIP is required.

peterjah:


Leo-Besancon (commented):
The fix looks good to me. @peterjah @modship even if we'd need to update the protos (only adding the field), do we want the same response in the GRPC API? (to keep the two get_status responses aligned)

peterjah:
> The fix looks good to me. @peterjah @modship even if we'd need to update the protos only adding the field), do we want the same response in the GRPC API? (to keep the two get_status responses aligned)

My bad! i forgot GRPC :) i will add it

peterjah:
PR on massa-proto
https://github.com/massalabs/massa-proto/pull/82

peterjah:
Pr on massa-proto-rs
https://github.com/massalabs/massa-proto-rs/pull/48

massa-models/src/mapping_grpc.rs
```diff
@@ -210,7 +210,7 @@ impl From<OperationType> for grpc_model::OperationType {
                         .collect(),
                 };
                 grpc_operation_type.r#type =
-                    Some(grpc_model::operation_type::Type::ExecutSc(execute_sc));
+                    Some(grpc_model::operation_type::Type::ExecuteSc(execute_sc));
```
Leo-Besancon, line 213: @modship this change is related to a proto fix you've made I think? Do we want it here? (it's breakingI think?)

Leo-Besancon: (see here for the PR: https://github.com/massalabs/massa/pull/4928)

Leo-Besancon (commented):
LGTM apart from the inline comment

Leo-Besancon (approved):

PR #5266, opened 2026-09-11
[F40] bound query_state batch work: shared event budget, atomic reads

### LINKED ISSUE #5255: [Followup] Include Events and Datastore keys to query_state cumulative budget
PR https://github.com/massalabs/massa/pull/5236 (Finding F41) introduced a cumulative budge in `query_state` for Bytecode and datastore value, in order to reject exceeding the response size early. We should apply the same cumulative budget to `Events` (128 items x `max_event_per_query` is still 900k events under one lock) and to datastore key queries.

_Originally posted by @damip in https://github.com/massalabs/massa/pull/5236#pullrequestreview-5093042367_

### LINKED ISSUE #5268: [F40] Enforce single-lock batch reads in execution query getters
Extract from [#5057](https://github.com/massalabs/massa/issues/5057#issuecomment-5513735951) _Originally posted by @damip_

1. Comment on `query_state` and the other batch getters: responses stamped with cursors/fingerprint must be computed under a single `execution_state` lock, never release it between items, cost is bounded via budgets not by splitting the lock. Otherwise the next generated PR will try this again.

bilboquet:
## Attack

`query_state` holds `execution_state.read()` for the whole batch (up to 128 items via
gRPC `max_query_items_per_request`). `Events` items had no batch-wide bound: 7000 events
per item on the final-cache path plus a completely uncapped active-history path
(`flat_map().collect()`), i.e. up to ~896k events deserialized, collected and returned
under a single read lock from one remote batch. `parking_lot::RwLock` is fair, so the
execution writer (final/candidate slots) stalls behind the heaviest batch in flight.
Remote-triggerable node slowdown / stall with attacker-controlled gas upstream.

## Fix (2 commits)

**1. Atomic batch reads — Closes #5268 (F40 point 1, stronger than asked).**
The spec asked for a comment; the PR enforces it by typing instead: batch helpers
(`eval_query_item`, `*_under`) take `&ExecutionState` borrowed from the guard, so
releasing the lock mid-batch is a compile error (E0505, verified), not a silent
corruption. Public trait signatures unchanged. Pure move, no behavior change.

**2. Shared event budget — Closes #5255.**
`ExecutionQueryRequest.max_event_count`, fed by new `ApiConfig`/`GrpcConfig`
`max_events_per_query` settings (default 7000): each `Events` item takes
`min(remaining, per-item cap)` and decrements what it returns; past zero it gets a
per-item `TooLargeResponse` error (exists since #5236, no proto change) instead of
fetching. The limit is applied to the final-cache and active-history paths
(take-after-fetch: cost proportional to the returned plus one truncated fetch).
Event bytes (`data.len()` + estimated 128 B overhead, marked for later measurement)
and datastore-key bytes (sum of key lengths) now count against `max_response_size`.
Single-shot event queries are unchanged.

Left intentionally out: early-stop iterators (documented optimization), per-request
key-count forwarding (still #5189), `get_addresses_infos` cap and slot deadline
(phases 3-4 of #5057).

## Why no MIP

Node-local API behavior only. Block validity rules, serialization, hashing, execution
semantics and PoS draws are untouched. Observable change: batches that used to return
up to ~896k events now get per-item errors past the budget — recorded on the F42
breaking-changes list for builders (same list as #5189).

## Tests

- `test_query_state_events_budget_shared`: two identical `Events` items, budget of 1 —
  first takes it, second errors.
- `test_scan_datastore_count_none_is_unbounded`: tripwire documenting the unbounded
  `count=None` per-item scan left while #5189 is unmerged (breaks on purpose when it lands).
- Existing `datastore_manipulations` (has an `Events` item) green unchanged.
- `cargo check --workspace`, `clippy` clean.

Closes #5255
Closes #5268
Relates #5057


bilboquet:
so far only point 1 and 2 (#5255) of the issue.

bilboquet:
fix #5255 
fix #5268

massa-execution-worker/src/controller.rs
```diff
@@ -428,10 +185,13 @@ impl ExecutionController for ExecutionControllerImpl {
...
     /// * original caller address
     /// * operation id
+    // ponytail: single-shot path uncapped (usize::MAX), one full active-history scan per call;
+    // thread a handler-provided limit like query_state does if this is abused
+    // (breaking-changes list, F42).
     fn get_filtered_sc_output_event(&self, filter: EventFilter) -> Vec<SCOutputEvent> {
         self.execution_state
             .read()
-            .get_filtered_sc_output_event(filter)
+            .get_filtered_sc_output_event(filter, usize::MAX)
```
Leo-Besancon, line 196: I know that would make a lot more changes to the interfaces, but I think it would be cleaner to have an option instead, and setting it explicitely to None in all the callers going through this path?

Leo-Besancon (commented):
LGTM! Just have 1 comment inline

massa-execution-worker/src/controller.rs
bilboquet: done

Leo-Besancon: Unless I'm mistaken there is still `usize::MAX` limit here instead of None?

bilboquet: Sorry. Should be ok now. (remaining usize::MAX should be legit)

Leo-Besancon (approved):
LGTM

PR #5267, opened 2026-09-14
[F3] fix(api): treat pool in_pool lookup failures as non-fatal in get…

### LINKED ISSUE #5081: [F3] Public JSON-RPC get_operations turns pool metadata failures into total query failure

---

## Source
AI security-scan **Finding 3** (severity: **minor**). Tracked in `SUBMISSIONS.md` as F3.

**Affected file:** `massa-api/src/public.rs`

## Summary
`API<Public>::get_operations` in `massa-api/src/public.rs` retrieves operation data from storage, but the JSON-RPC path still performs `pool_command_sender.contains_operations(&ops, timeout)` to populate `in_pool` metadata and converts errors from that call into an `InternalServerError`. This creates a real robustness issue: a failure in the pool metadata lookup can cause the entire JSON-RPC request to fail even though the underlying operations were already available from storage. The original finding's broader claim that this creates a clear attacker-driven denial of service or fully hides data relative to gRPC is not supported as stated.

## Recommendation
Decouple storage-backed operation retrieval from the auxiliary pool lookup used to populate `in_pool`. If `pool_command_sender.contains_operations(&ops, timeout)` fails, the endpoint should still return the operation data already retrieved from storage and treat `in_pool` as unavailable, unknown, or omitted rather than converting the entire request into an `InternalServerError`.

Handle pool metadata lookup failures as non-fatal, log them for observability, and document the degraded-response behavior so clients can distinguish between missing operation data and missing `in_pool` metadata. Also align this behavior across API surfaces where appropriate to reduce inconsistent failure modes between JSON-RPC and gRPC.

## Why P0 / breaking
The fix changes **API / gRPC behavior** (Public JSON-RPC `get_operations` behavior). Per maintainer guidance, API/gRPC interface or observable-behavior changes are treated as **breaking** (they can break external tooling), so this is filed as an issue for design/versioning discussion rather than a direct PR. Note: the gRPC proto types are auto-generated and must not be hand-edited. See `SUBMISSIONS.md` Part D.

peterjah:
…_operations and get_endorsements

  * [ ] make tests compile
  * [ ] make tests pass 


massa-api/src/public.rs
```diff
@@ -713,11 +714,19 @@ impl MassaRpcServer for API<Public> {
         let ops: Vec<OperationId> = storage_info.iter().map(|(op, _)| op.id).collect();
 
         // ask pool whether it carries the operations
+        // this is auxiliary metadata: on pool failure (e.g. lock timeout) we still return the
+        // operations found in storage and report `in_pool: false`, like the gRPC endpoints do
```
Leo-Besancon: I think we should tweak this comment, as GRPC's get_operations does not have this field. I'd suggest something like: `like we do for get_endorsements for both JSON-RPC and gRPC endpoints`

Leo-Besancon (dismissed):
Small comment inline, otherwise LGTM!

massa-api/src/public.rs
peterjah: done

Leo-Besancon (approved):
LGTM

PR #5270, opened 2026-09-15
fix(grpc): report the matching id limit in too-many-ids error message…

### LINKED ISSUE #4658: Invalid max value used
in massa-grpc public.rs:
* max_operation_id_per_request instead of max_block_id_per_request used in get_blocks

peterjah:
…s for operation and endorsement filters


Leo-Besancon (approved):
LGTM

PR #5271, opened 2026-09-15
fix(bootstrap): reject more than one AskBootstrapPeers per bootstrap …

### LINKED ISSUE #4520: Bootstrap client can ask for bootstrap peers even if no bootstrap state were done
See scenario_ask_peers_only

peterjah:
…session


Leo-Besancon (approved):
LGTM, the tests are not that useful compared to the bloat IMO so we could drop them, but keeping them is fine too.

PR #5277, opened 2026-09-16
[F40] fix(execution): cap datastore keys in get_addresses_infos (#5057 point 3)

### LINKED ISSUE #5057: [F40] Unbounded `query_state` Batch Processing Can Cause Execution Lock Contention and Node DoS

bilboquet:
## Problem (#5057 item 3)

`get_addresses_infos` (behind `get_addresses`) lists EVERY datastore key of every
address with `count = None` hardcoded, under a single execution lock, with no length
check on the handler. Neither #5189 nor #5236 touches it. Unbounded RocksDB range
iteration + active-history merge per address, remotely triggerable.

## Fix (option 1 of the issue text)

Cap the per-address keys with `max_datastore_keys_query`, threaded as
`max_keys: Option<u32>` from the API handler through the trait into
`scan_datastore`. No new knob (existing setting).

Unconditional cap, no version gate (per review: an API response-only change needs
no MIP gate). The handler passes its configured setting directly; `None` keeps its
pre-existing meaning (no cap configured), same as `scan_datastore`.

Truncation (first N keys, no error) always applies once over the cap. The number N
of addresses per call stays unbounded (transport body limit only) — noted residual
for review; key-list removal (option 2) rejected: unknown external consumers,
default is maintain-and-evolve.

## Behavior change

First visible key truncation past the cap, on JSON-RPC `get_addresses` only
(no gRPC equivalent). Whether this needs an F42 breaking-changes list entry is to
confirm.

## Tests

- `cargo check` green on `massa-execution-exports` + `massa-execution-worker`;
  `cargo fmt --check` clean; zero errors in this change's files.
- Pre-existing on `main`, unrelated to this change: `massa_api` lib fails solo
  (`jsonrpsee::core::client` needs the `full` feature from dev-deps) and the
  `massa_api` test target fails on `GasCosts::default()` in `tests/mock.rs`
  (pinned sc-runtime rev).
- Existing `datastore_manipulations` caller updated (`None` on a test-only path).

Relates #5057


Leo-Besancon (changes requested):
The fix itself LGTM, but as it's only a breaking change in the sense of API response change, I would remove the MIP-gate from this PR.

Leo-Besancon (approved):
LGTM!

PR #5281, opened 2026-09-17
feat(massa-client): add --url and support full JSON-RPC URLs

### LINKED ISSUE #4915: massa-client should support mainnet (or buildnet) url instead of ip

Using the massa-client with one of the ip behing mainnet.massa.net is blocked. It could be useful if your node is down (server upgrading, massa node version upgrading) and you still want to transfer some coins.

So would be nice to be able to specify directly the url (https://mainnet.massa.net/api/v2). So at the end, we could run the massa-client like:

```
massa-client -u https://mainnet.massa.net/api/v2 ...
```

N/A

**Additional context**

N/A

peterjah:


Leo-Besancon (approved):
LGTM!

PR #5282, opened 2026-09-17
Remove dead aggregate (de)serializers left over from pre-DB bootstrap

### LINKED ISSUE #5254: [Followup] Remove test-only (de)serializers
Longer term we should probably just delete the dead aggregate (de)serializers for F35 (`DeferredCallRegistryChanges`) and F98 (`DeferredCreditsDeserializer`) in a cleanup PR, a parser that does not exist cannot be malleable.

_Originally posted by @damip in https://github.com/massalabs/massa/issues/5218#issuecomment-5508373537_

bilboquet:
Closes #5254.

Pure deletion per @damip's scoping: leftover aggregate (de)serializers from the pre-DB bootstrap that streamed StateChanges — a parser that does not exist cannot be malleable. Bootstrap streams DB batches and has zero mentions of them.

Deleted (aggregates + their round-trip tests, zero production callers):

    StateChanges*, PoSChanges*, AsyncPoolChanges*, DeferredCredits*, CycleHistory*
    DeferredRegistryChanges* + DeferredRegistrySlotChanges* (bulk impls, validator, tests)
    LedgerChanges* (ledger-exports and models), ExecutedOps*, ExecutedDenunciations*

Kept (in-memory types + the real DB sub-codecs): all *Changes/CycleInfo/Credits types, per-key codecs. PoSFinalState and DeferredCallRegistry now hold the sub-codecs directly (credits, cycle_info, slot, effective_total_gas) with identical bounds — no behavior change.

Two deviations from the scoping comment, both within its criterion:

    Point 3 says the registry fields were "never touched" — they were (4 prod sites use the sub-fields), so literal "drop the fields" was impossible; the sub-codecs were hoisted instead.
    LedgerChange* (models) and check_bounds weren't listed but are dead by the same "no production caller" rule, so they went too.

Verification: cargo check --workspace --all-targets, clippy --all-targets, cargo test --release


massa-executed-ops/src/ops_changes.rs
```diff
@@ -1,163 +1,6 @@
...
-    error::{context, ContextError, ParseError},
-    multi::length_count,
-    sequence::tuple,
-    IResult, Parser,
-};
-use std::ops::Bound::{Excluded, Included};
+use massa_models::{operation::OperationId, prehash::PreHashMap, slot::Slot};
 
 /// Changes for ExecutedOps (was_successful, op_expiry_slot)
 pub type ExecutedOpsChanges = PreHashMap<OperationId, (bool, Slot)>;
```
Leo-Besancon: nit: this is the only remaining line in this file, move it somewhere else?

massa-pos-exports/src/pos_final_state.rs
```diff
@@ -142,14 +146,14 @@ pub struct PoSFinalState {
     pub initial_rolls: BTreeMap<Address, u64>,
     /// initial seeds, used for negative cycle look back (cycles -2, -1 in that order)
     pub initial_seeds: Vec<Hash>,
-    /// deferred credits serializer
-    pub deferred_credits_serializer: DeferredCreditsSerializer,
-    /// deferred credits deserializer
-    pub deferred_credits_deserializer: DeferredCreditsDeserializer,
+    /// credits serializer
+    pub credits_serializer: CreditsSerializer,
```
Leo-Besancon: nit: with this rename, its less clear that it's the inner serializer for deferred credits. Should we rename these `DeferredCreditSerializer` and `DeferredCreditDeserializer`? (singular)

Leo-Besancon (dismissed):
Just two tiny inline comments, but it looks good to me!

Leo-Besancon (approved):
Thanks for the update! LGTM

Issue #5241, opened 2026-08-28
[F42] Residual unbounded datastore key enumeration: SC path, scan internals, and work-based pricing


peterjah:
#5059 (F42) reported that unbounded datastore key enumeration enables underpriced DoS. PR #5189  addressed only the API surface: it propagates the configured max_datastore_keys_query cap from the JSON-RPC and gRPC parsing layers into the scan, so get_addresses_datastore_keys and the gRPC QueryState datastore-key items can no longer trigger an unbounded enumeration.

That PR was linked to auto-close #5059. Three of the four problem areas in the original finding remain unfixed. This issue tracks them so they don't disappear when GitHub closes the original.

The API fix does not reduce the worst case reachable from a smart contract, which is the cheaper attack path — no RPC rate limiting applies there.

---
(a) The smart-contract path is still uncapped

massa-execution-worker/src/context.rs:660:

// TODO when updating the ABI, make sure to set this value to a maximum defined as a CONSTANT for determinism
// The API will use a different, user-configurable max value
let max_datastore_query = None;

Context::get_keys hardcodes None, so cleanup_datastore_key_range_query returns count = None and the SC reaches scan_datastore with no limit. A contract calling get_keys on a large datastore still walks the whole matching range.

The None is deliberate and correct as things stand: the API cap is node-operator configurable, and feeding a configurable value into execution would make block validity depend on local config — an instant fork. The fix must therefore use a protocol constant, not max_datastore_keys_query.

Consensus-observable: yes. Capping SC-visible results changes what contracts observe and can change execution outcomes. Requires a MipInfo in massa-versioning with MipComponent::Execution (or VM), the branch selected via the active component version, and tests covering both pre- and post-activation behaviour. Old behaviour must stay reachable for pre-activation slots.

Open question for the design discussion: what should a contract see when its query is truncated? Silent truncation is the failure mode we just documented as hostile on the API side, and it is worse here — an SC cannot detect it and may compute on a partial key set. Options include a hard constant cap with an explicit error, or an ABI change that returns a truncation flag.

---
(c) Pricing is not work-based

massa-node/base_config/gas_costs/abi_gas_costs.json:103:

"abi_get_ds_keys": 1825,

A flat constant. A call that scans and returns one key costs the same as one that walks  This is the underpricing at the heart of F42 — capping the count limits the worst caseper call but does not make cost proportional to work, so an attacker can still buy far more scanning than they pay for by issuing many capped calls.

[...]

PR #5286, opened 2026-09-22
perf(execution): bound scan_datastore internals to the requested count

### LINKED ISSUE #5283: [F42] (b) Bound scan_datastore internals without changing outputs
Follow-up of #5241 (residual F42 findings), part (b) of the suggested sequencing.

## Scope

Output-preserving optimisation of `scan_datastore` (`massa-execution-worker/src/datastore_scan.rs`). No observable behavior change, so no MIP gating. Targets `main`. Massa repo only.

Parts (a) (SC-path protocol cap) and (c) (work-based gas pricing) are tracked separately and target `dev_breaking` behind MIP-0002.

## Problem

Two spots materialise / fetch more keys than needed:

1. `Set` branch and absent-final branch: `key_updates` is fully materialised over the whole key range across all history layers, and `.take(count)` is only applied at the end. Intermediate collections should never hold more than `count` keys.
2. Merge replenish: the refetch passes `count` instead of what is still missing (`count - already_collected`), over-fetching final keys on every round.

## Acceptance criteria

- [ ] `scan_datastore` never materialises more than `count` keys in the Set and absent-final branches
- [ ] Merge replenish fetches only what is still needed
- [ ] Outputs identical: existing `tests_scan_datastore` suite green, no behavior change
- [ ] Benchmark showing the worst case bounded / reduced

bilboquet:
Stream the absolute datastore in the Set branch through an ordered merge instead of copying the full range, and replenish the final-state queue with the missing remainder only. Results are unchanged.

Also extracts the Set-branch merge into a `merge_set_case_keys` helper so `scan_datastore` stays readable.

Tests: 3 new deterministic regression tests (count bound on a large datastore, Set + newer Delete counterexample, merge delete-shadowing). `cargo test -p massa_execution_worker --lib`: 68 passed. `cargo fmt` + `cargo clippy --all-targets`: clean.

Closes #5283 (part of #5241).


massa-execution-worker/src/datastore_scan.rs
```diff
@@ -11,13 +12,91 @@ use massa_final_state::FinalStateController;
...
+                        }
+                        (None, None) => break,
+                    }
+                }
+            }
+        }
+        None => {
+            // reset after a speculative delete: updates are the only source
+            for (k, is_set) in key_updates.iter() {
+                if *is_set {
```
peterjah, line 91: With count == Some(0) this returns 1 key; the old code (filter_it.take(0)) returned 0, and every other branch still returns 0 (the Some(abs) branch checks at loop top, the no-reset paths use take(cnt) / len() >= cnt at loop top). count reaches here unvalidated — cleanup_datastore_key_range_query (massa-models/src/datastore.rs:256-265) only rejects count > max, never 0 — via the gRPC/JSON-RPC AddressDatastoreKeysCandidate query, so it's client-reachable. API-only, non-consensus, but it is an unintended behavior change introduced by this commit.

Fix is to hoist the check, mirroring the other branch:

```
for (k, is_set) in key_updates.iter() {
    if let Some(cnt) = count {
        if out.len() >= cnt as usize { break; }
    }
    if *is_set { out.insert(k.clone()); }
}
```

peterjah, line 91: With count == Some(0) this returns 1 key; the old code (filter_it.take(0)) returned 0, and every other branch still returns 0 (the Some(abs) branch checks at loop top, the no-reset paths use take(cnt) / len() >= cnt at loop top). count reaches here unvalidated — cleanup_datastore_key_range_query (massa-models/src/datastore.rs:256-265) only rejects count > max, never 0 — via the gRPC/JSON-RPC AddressDatastoreKeysCandidate query, so it's client-reachable. API-only, non-consensus, but it is an unintended behavior change introduced by this commit.

Fix is to hoist the check, mirroring the other branch:

```
for (k, is_set) in key_updates.iter() {
    if let Some(cnt) = count {
        if out.len() >= cnt as usize { break; }
    }
    if *is_set { out.insert(k.clone()); }
}
```

bilboquet: should ok now (with added test)

bilboquet: same as above

peterjah (approved):

PR #5287, opened 2026-09-22
fix(bootstrap): keep the client→server length prefix 4 bytes wide and…

peterjah:
… relax received element counts for MAIN.5.0 compatibility

# fix(bootstrap): restore `MAIN.5.0` wire compatibility

Closes the release blocker found by the simulated-network compatibility harness: a HEAD node
cannot bootstrap from a `MAIN.5.0` node — it connects and then hangs silently, with no error
and no timeout.

Consensus, block production and the P2P protocol were verified unaffected (a HEAD node started
at genesis stayed in consensus with the 5.0 nodes across 660 compared slots); the break is
confined to bootstrap.

## 1. The client→server frame header had shrunk from 36 to 35 bytes

`047913b2ea` (#4973) split the bootstrap message size limit per direction and pointed the
client→server length prefix at the new, much smaller `MAX_BOOTSTRAP_MESSAGE_FROM_CLIENT_SIZE`
(1 MiB). That constant is not just a policy cap: `SerializeMinBEInt::to_be_bytes_min` derives
the **width** of the prefix from the bound it is handed
(`skip_bytes = max_value.leading_zeros() / 8`), so lowering the bound narrowed the prefix from
4 bytes to 3.

| | `MAIN.5.0` | before this PR | after |
|---|---|---|---|
| bound used to encode the client→server prefix | `MAX_BOOTSTRAP_MESSAGE_SIZE` (300 MB) | `MAX_BOOTSTRAP_MESSAGE_FROM_CLIENT_SIZE` (1 MiB) | `BOOTSTRAP_MESSAGE_LEN_PREFIX_MAX` (290 MB) |
| prefix width | 4 | **3** | 4 |
| frame header (`HASH_SIZE_BYTES` + prefix) | 36 | **35** | 36 |

A 5.0 server does a `read_exact` for 36 bytes and blocks forever on a 36th byte a HEAD client
never sends — exactly the silent hang the harness observed. Server→client was unaffected (still
4 bytes), which is why a HEAD client could still read 5.0's throttle error; the framing break is
one-directional but fatal to the session either way.

Note that no version check catches this. HEAD still declares `MAIN.5.0`
(`massa-models/src/config/constants.rs`), so the two builds announce the *same* version and the
handshake passes trivially; even once bumped, `Version::is_compatible` matches on instance and
major only, so `MAIN.5.0` ↔ `MAIN.5.1` would connect and hang just the same.

### Fix

Separate the bound the prefix is *encoded against* — which is part of the wire format and shared
with every peer — from the per-direction *value cap*, which is ours to tighten freely:

- new `BOOTSTRAP_MESSAGE_LEN_PREFIX_MAX` / `BOOTSTRAP_MESSAGE_LEN_PREFIX_SIZE_BYTES`, pinned to
  the server→client bound, i.e. the encoding that has been on the wire since before `MAIN.5.0`;
- `BootstrapClientBinder::send_timeout` checks `msg_len` against
  `MAX_BOOTSTRAP_MESSAGE_FROM_CLIENT_SIZE` explicitly, then encodes against the prefix bound;
- `BootstrapServerBinder` reads `HASH_SIZE_BYTES + BOOTSTRAP_MESSAGE_LEN_PREFIX_SIZE_BYTES`,
  decodes against the prefix bound, then rejects a decoded length over the client cap.

#4973's actual goal — a client cannot announce a 300 MB frame — is preserved in full: the
oversized length is now refused by an explicit value check instead of by the encoding width, one
step later and still before any allocation. The now-dead fixed-width padding in the client is
removed, and `MAX_BOOTSTRAP_MESSAGE_FROM_CLIENT_SIZE_BYTES` (which only ever described the
mistaken width, and never shipped in a release) is gone.

Two guards so this cannot regress silently:

- a `const _: () = assert!(BOOTSTRAP_MESSAGE_LEN_PREFIX_SIZE_BYTES == 4, ...)`, which fails the
  build if any constant it derives from moves the width;
- `test_bootstrap_frame_header_widths_are_stable`, asserting the header widths as literal
  numbers (36 and 69) rather than recomputing them from the same constants the code uses — the
  existing binder tests build their frames from those constants, so they could never have caught
  this.

## 2. Second break, masked by the first: received `new_elements` count cap

Not reachable by the harness, because the framing hang happens before the first
`AskBootstrapPart`, but it would have blocked the release just as hard.

[F9] (#5198) added `MAX_BOOTSTRAP_FINAL_STATE_ELEMENTS_COUNT = 100_000` and enforces it while
*parsing* a received state part (`BootstrapServerMessageDeserializer::deserialize_kv_map`), as
well as when producing one (`massa-db-worker/src/massa_db.rs`). `MAIN.5.0` has no count notion
at all — it bounds a batch by **size only**, up to `MAX_BOOTSTRAP_FINAL_STATE_PARTS_SIZE`
(100 MB). Real state entries are on the order of tens of bytes, so a conforming 5.0 server packs
far more than 100 000 of them into a single part, and a HEAD client would reject it with
`too many new_elements entries`.

### Fix

The count we *send* stays at 100 000. The count we *accept* is now derived from the size bound
the section already carries (`MAX_RECEIVED_BOOTSTRAP_*_ELEMENTS_COUNT`): still a hard cap on how
many entries we will allocate, but unreachable by any conforming peer, 5.0 included. A `TODO`
marks it for tightening back in the first release that no longer has to bootstrap from a 5.0
server.

**This does weaken [F9] on the receiving side for 5.x, and that should be a conscious call.**
It is worth being precise about what is and is not lost:

- Version-gating the strict cap (enforce 100 000 only against peers announcing > 5.0) would
  *not* help. The threat [F9] addresses is a hostile bootstrap server, and a hostile server
  chooses the version it announces — it would simply claim 5.0 and take the relaxed path. The
  gate would add plumbing and no security.
- The remaining bound is not nothing: the `new_elements` section is still wrapped in a
  `length_value` capped at `max_final_state_elements_size`, so a part is bounded at 100 MB on
  the wire regardless. What is no longer bounded is the *allocator amplification* of many tiny
  entries into `BTreeMap` nodes, which is the specific OOM [F9] closed.
- The only way to keep the tight cap is to accept that 5.1 nodes cannot bootstrap from 5.0
  peers, i.e. ship a breaking major.


Leo-Besancon:
Q: For either 2. or both 1. and 2., would it be better to gate the behaviour behind the MIP?
So current behaviour is unchanged, and 5.1 (future updated nodes) will be able to bootstrap fine with 5.0 nodes, but then they switch to the good behaviour later on?

The goal of my proposal is to avoid weakening the [F9] fix.

Edit:
> Version-gating the strict cap (enforce 100 000 only against peers announcing > 5.0) would
not help. The threat [F9] addresses is a hostile bootstrap server, and a hostile server
chooses the version it announces — it would simply claim 5.0 and take the relaxed path. The
gate would add plumbing and no security.

Ah I did not understand this correctly. Indeed for the period of time before the MIP is activated, it does not change anything.
It would just need to be updated once we do the MAIN.6.0 cleanup release after the MIP activated I think.

bilboquet:
I think Léo raised a good point here. The F9 protection is effectively relaxed on the client receive path (50M/5M caps bound wire bytes, not allocator amplification), and we probably don't want to ship that as-is. As Léo suggested, we should reintroduce a mitigation — ideally a cap calibrated on real 5.0 batches rather than the theoretical 50M worst case — and then reintroduce the strong F9 limit once the MIP is activated / 5.0 no longer needs to be supported (6.0 cleanup). Note version-gating alone wouldn't help against a hostile server, since it chooses the version it announces.

peterjah:
Agreed on the substance — shipping the 50M/5M cap as the mitigation was not good enough, and
you're both right about why: it bounds **wire bytes**, and the thing [F9] closed was **allocator
amplification**. A 100 MB section of two-byte entries is within the wire bound and still costs
gigabytes as a map. Pushed a different mitigation rather than a different number.

### What changed

The receive-side bound is no longer a count. Each parsed `(key, value)` pair is charged the bytes
it occupied on the wire **plus** `BOOTSTRAP_BATCH_ENTRY_OVERHEAD` (96 B: the `BTreeMap` node slot,
two `Vec` headers, allocator rounding), and parsing fails when the section's budget is spent. That
bounds amplification directly instead of bounding the wire and hoping. Pairs are charged as
*parsed*, not as *inserted*, so the duplicate-key flood from #5198 is still charged per pair.

The budget is derived, not magic:

```rust
pub const fn bootstrap_batch_allocation_budget(section_wire_size: usize) -> usize {
    section_wire_size + (section_wire_size / MIN_EXPECTED_BOOTSTRAP_ELEMENT_WIRE_SIZE)
        * BOOTSTRAP_BATCH_ENTRY_OVERHEAD
}
```

so there is now exactly one knob to calibrate — `MIN_EXPECTED_BOOTSTRAP_ELEMENT_WIRE_SIZE`, the
smallest wire size we expect a *real* entry to have. It is set to **8**, which is deliberately
pessimistic: every state key carries a CF prefix (`ledger/`, `versioning/`, …) plus its own
identifying bytes, so real entries run to several tens of bytes.

### Calibration against a real 5.0 batch

A full 100 MB state part of ledger-shaped entries (7 B prefix + version byte + 33 B address + ident
= 42 B key, small value, 2 length varints ≈ 44 B/entry) is ~2.27M entries, so its real footprint is
**~318 MB**. Against that:

| `MIN_EXPECTED` | state budget (100 MB section) | versioning (10 MB) | updates (290 MB) |
|---|---|---|---|
| **8 (shipped)** | 1.30 GB | 130 MB | 3.77 GB |
| 16 | 0.70 GB | 70 MB | 2.03 GB |
| 24 | 0.50 GB | 50 MB | 1.45 GB |
| 32 | 0.40 GB | 40 MB | 1.16 GB |

So the ceiling for a hostile server goes from ~5 GB (50M tiny entries) to ~1.3 GB, with ~4x
headroom over an honest full part. Happy to raise `MIN_EXPECTED` to 16 or 24 if you want a tighter
ceiling — the constraint that kept me at 8 is the *versioning* section rather than the state one: a
10 MB section of short `versioning/`-prefixed entries would breach a 70 MB budget, and I would
rather not re-break bootstrap for a section whose real entry sizes I have not measured. If one of
you can measure a real mainnet part, that number is the one to set this from.

bilboquet:
looks good to me.

for the open question: I ran a full mainnet bootstrap with an instrumented deserialize_kv_map (see [bootstrap-measure-instrumentation.patch](https://github.com/user-attachments/files/32531605/bootstrap-measure-instrumentation.patch)
). 

Over 33× 100 MB state parts: up to ~3.57M entries/part, smallest entry 22 B wire (~443 MB footprint); versioning ≤1 entry (~1 KB) per part; updates negligible.

So **MIN_EXPECTED = 16** (700 MB budget, ~1.6x headroom, hostile ceiling halved to 700 MB).

bilboquet (changes requested):
see ma comment, only const value to agree on.
@Leo-Besancon

Leo-Besancon (commented):
Haven't check everything, so it's a pretty light review on my side, but I'm not sure I understand one point.
If we've updated the receiving cap behaviour to avoid breaking bootstrap on previous version, should we also update the sending cap behaviour (in `get_batch_to_stream`)?

peterjah:
> looks good to me.
> 
> for the open question: I ran a full mainnet bootstrap with an instrumented deserialize_kv_map (see [bootstrap-measure-instrumentation.patch](https://github.com/user-attachments/files/32531605/bootstrap-measure-instrumentation.patch) ).
> 
> Over 33× 100 MB state parts: up to ~3.57M entries/part, smallest entry 22 B wire (~443 MB footprint); versioning ≤1 entry (~1 KB) per part; updates negligible.
> 
> So **MIN_EXPECTED = 16** (700 MB budget, ~1.6x headroom, hostile ceiling halved to 700 MB).

Thx for the measures ! i will put 16

peterjah:
> Haven't check everything, so it's a pretty light review on my side, but I'm not sure I understand one point. If we've updated the receiving cap behaviour to avoid breaking bootstrap on previous version, should we also update the sending cap behaviour (in `get_batch_to_stream`)?

Good question — the answer splits in two, and the second half turned out to matter.

**For compatibility: no, the send side never needed a change.** The asymmetry is deliberate. Sending *fewer* entries per part is safe against every peer in both directions: a MAIN.5.0 client has no count notion at all, it bounds an incoming part by size only, so a part carrying 100 000 entries instead of 3.5M is just a smaller part and one more round trip. Nothing on any version's receive path can reject it. That is why the compat fix only had to touch what we *accept* — strict in what we send, liberal in what we accept.

**For throughput: yes, and I've now done it in this PR.** The count half of the condition was the binding constraint by a wide margin:

```rust
if new_elements_size <= self.config.max_final_state_elements_size
    && new_elements.len() < self.config.max_final_state_elements_count
```

With bilboquet's mainnet numbers, a part filled to the 100 MB size limit holds ~3.57M entries (~28 B/entry). Capped at 100 000 entries a part carries ~2.8 MB — about 3% of the size it is allowed. Full-state streaming would go from ~33 parts to ~1200: ~35x the round trips, each re-seeking the state iterator and recomputing `updates_on_previous_elements` over the change history. That cost is unreleased — the count cap came in with [F9] (42cc8ad, #5198) and HEAD still declares MAIN.5.0 — so this is about not shipping it, not about a regression already out there.

Rather than just dropping the cap, the send side now does the same accounting the receive side does:

- `MAX_BOOTSTRAP_FINAL_STATE_ELEMENTS_COUNT` / `MAX_BOOTSTRAP_VERSIONING_ELEMENTS_COUNT` and the two `MassaDBConfig` count fields are gone;
- `get_batch_to_stream` / `get_versioning_batch_to_stream` charge each entry `key + value + length varints + BOOTSTRAP_BATCH_ENTRY_OVERHEAD` and stop at `bootstrap_batch_allocation_budget(max_*_elements_size)` — the same formula, on the same size bound, as `deserialize_kv_map`.

That buys an invariant we did not have before: **a batch we are willing to build is a batch a peer of this version is willing to parse**, by construction, because both sides bound the same quantity with the same formula derived from the same constant. Server memory per streamed part stays explicitly bounded, and for real entries (≥ 22 B wire, above the 16 B floor) the size limit is the binding one again, so parts fill to 100 MB exactly as in 5.0.

One wrinkle worth flagging: the first entry of a batch is now always taken, even if it alone exceeds the budget. Without that, a batch admitting zero entries would stall the stream outright, since the client advances its cursor by what it receives — worth having explicitly rather than as an accident of the arithmetic.

bilboquet (approved):
ok for me

Leo-Besancon (approved):
LGTM if you've been able to test that this solve the bootstrap issue!

